PR #2327, opened 2020-10-28
FIX-#2326: move s3fs import in _read function

### LINKED ISSUE #2326: Move import s3fs in parquet_reader.py in function scope
### System information
- **Python version**:

### Describe the problem

`s3fs` - optional dependence, so we shouldn't import in global scope.

anmyachev:

## What do these changes do?

- [ ] passes `flake8 modin`
- [x] Resolves #2326 
- [ ] tests added and passing

### DIFF
--- a/modin/engines/base/io/column_stores/parquet_reader.py
+++ b/modin/engines/base/io/column_stores/parquet_reader.py
@@ -12,7 +12,6 @@
 # governing permissions and limitations under the License.
 
 import os
-import s3fs
 
 from modin.engines.base.io.column_stores.column_store_reader import ColumnStoreReader
 from modin.error_message import ErrorMessage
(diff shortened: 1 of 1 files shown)


prutskov:
@anmyachev Does it make sense to move `import s3fs` in `modin/engines/base/io/file_reader.py` to global scope of this file (this is used in two places in file).

anmyachev:
> @anmyachev Does it make sense to move `import s3fs` in `modin/engines/base/io/file_reader.py` to global scope of this file (this is used in two places in file).

In our case, this shouldn't be done, because we are claiming that `s3fs` is optional. That is, we have to import it only when its functionality is needed (otherwise, the `import modin.pandas` will break for a user who actually does not need this dependency)

prutskov:
@anmyachev ok, LGTM, ping when checks will pass.

anmyachev:
> @anmyachev ok, LGTM, ping me, please, when checks will pass.

ping @prutskov

prutskov (approved):

PR #2359, opened 2020-11-02
FEAT-#9999: install modin via pip for scaleout feature

anmyachev:

## What do these changes do?

- [ ] passes `flake8 modin`
- [ ] Resolves #? 
- [ ] tests added and passing


vnlitvinov (changes requested):
Mixing two different ways of installing packages could lead to a package conflict.
I would rather we have fixed our conda packages and continue using the current approach - we need conda packages anyway for OmniSci, so I don't see value of bringing pip into equation.

PR #2381, opened 2020-11-09
FIX-#2380: don't ignore lengths parameter for dask engine

### LINKED ISSUE #2380: Dask engine ignore lengths parameter

anmyachev:

## What do these changes do?

* Add `lengths` parameter handling for dask engine.
* Make `split_result_of_axis_func_pandas` always return a list value and removes a bug in which the wrong number of partitions is returned (in the case when `num_splits==1` and `lengths != 1`)

- [x] passes `flake8 modin`
- [x] Resolves #2380 
- [ ] tests added and passing

### DIFF
--- a/modin/data_management/utils.py
+++ b/modin/data_management/utils.py
@@ -85,7 +85,7 @@ def split_result_of_axis_func_pandas(axis, num_splits, result, length_list=None)
     if length_list is not None:
         length_list.insert(0, 0)
         sums = np.cumsum(length_list)
-        if axis == 0:
+        if axis == 0 or isinstance(result, pandas.Series):
             return [result.iloc[sums[i] : sums[i + 1]] for i in range(len(sums) - 1)]
         else:
             return [result.iloc[:, sums[i] : sums[i + 1]] for i in range(len(sums) - 1)]
@@ -94,7 +94,7 @@ def split_result_of_axis_func_pandas(axis, num_splits, result, length_list=None)
         return [result]
     # We do this to restore block partitioning
     chunksize = compute_chunksize(result, num_splits, axis=axis)
-    if axis == 0:
+    if axis == 0 or isinstance(result, pandas.Series):
         return [
             result.iloc[chunksize * i : chunksize * (i + 1)] for i in range(num_splits)
         ]
--- a/modin/data_management/utils.py
+++ b/modin/data_management/utils.py
@@ -82,8 +82,6 @@ def split_result_of_axis_func_pandas(axis, num_splits, result, length_list=None)
     list
         A list of Pandas DataFrames.
     """
-    if num_splits == 1:
-        return result
     if length_list is not None:
         length_list.insert(0, 0)
         sums = np.cumsum(length_list)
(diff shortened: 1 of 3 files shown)


modin/engines/dask/pandas_on_dask/frame/axis_partition.py
```diff
@@ -68,8 +70,6 @@ def deploy_func_between_two_axis_partitions(
             *partitions,
             pure=False,
         )
-        if num_splits == 1:
```
YarShev, line 71: Why is this removed? We have the code that handles returned not list values (see [here](https://github.com/modin-project/modin/blob/master/modin/engines/base/frame/axis_partition.py#L94)). Also, why is not `lengths` considered here?

modin/engines/dask/pandas_on_dask/frame/axis_partition.py
```diff
@@ -43,13 +43,15 @@ def deploy_axis_func(
...
```
YarShev, line 46: The same comment as for `deploy_func_between_two_axis_partitions`.

modin/data_management/utils.py
```diff
@@ -82,15 +82,16 @@ def split_result_of_axis_func_pandas(axis, num_splits, result, length_list=None)
...
     if length_list is not None:
         length_list.insert(0, 0)
         sums = np.cumsum(length_list)
         if axis == 0:
             return [result.iloc[sums[i] : sums[i + 1]] for i in range(len(sums) - 1)]
         else:
             return [result.iloc[:, sums[i] : sums[i + 1]] for i in range(len(sums) - 1)]
+
+    if num_splits == 1:
+        return [result]
```
YarShev, line 94: Why do we wrap to a list?

YarShev (commented):
Also, I think we need to move [this logic](https://github.com/modin-project/modin/blob/master/modin/engines/base/frame/axis_partition.py#L219) into `split_result_of_axis_func_pandas` with corresponding handling.

modin/data_management/utils.py
anmyachev: This is a simplification - now `split_result_of_axis_func_pandas` function returns `list` everywhere (as it is written in the docstring), which in turn localizes this edge case in this function, and not in several.

modin/engines/dask/pandas_on_dask/frame/axis_partition.py
anmyachev: Now there is no case when the type is different from the `list`.(Simplification)

modin/engines/dask/pandas_on_dask/frame/axis_partition.py
anmyachev: Answered above.

modin/engines/dask/pandas_on_dask/frame/axis_partition.py
anmyachev: The tests did not fall - so I did not pay attention.

I will see how to properly process `lengths`.

anmyachev: I think there is no reason to add length handling to `deploy_func_between_two_axis_partitions` because two frames will be combined into one by some binary operation.

anmyachev:
@YarShev ready for review

modin/engines/dask/pandas_on_dask/frame/axis_partition.py
YarShev: I see the point. There is no reason.

YarShev:
@anmyachev , apparently, now we don't have a case when return value is different from `list`. So, I think we need to remove this [check](https://github.com/modin-project/modin/blob/master/modin/engines/base/frame/axis_partition.py#L94) and keep returning `list`.

YarShev:
Also, can you rebase on current master to fix failing of Omnisci tests?

YarShev (approved):
@anmyachev , thanks, LGTM!

PR #2385, opened 2020-11-10
FIX-2200: Enable Calcite by default in OmniSci backend

### LINKED ISSUE #2200: Enable Calcite by default in OmniSci backend

amyskov:

## What do these changes do?

- [x] passes `flake8 modin`
- [x] Resolves #2200  
- [ ] tests added and passing

### DIFF
--- a/modin/config/envvars.py
+++ b/modin/config/envvars.py
@@ -196,6 +196,8 @@ class DoUseCalcite(EnvironmentVariable, type=bool):
     """
 
     varname = "MODIN_USE_CALCITE"
+    default = True
+    choices = ("True", "False")
 
 
 class TestDatasetSize(EnvironmentVariable, type=str):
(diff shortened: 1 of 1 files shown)


fexolm (dismissed):

modin/config/envvars.py
```diff
@@ -196,6 +196,8 @@ class DoUseCalcite(EnvironmentVariable, type=bool):
     """
 
     varname = "MODIN_USE_CALCITE"
+    default = True
+    choices = ("True", "False")
```
anmyachev, line 200: I think we don't need to define `choices` here, by default the allowed set of values is larger (more flexible solution)

amyskov: Ok, corrected.

anmyachev (approved):

PR #2460, opened 2020-11-18
FIX-#2459: Updated TeamCity tests image to use Ray as base image

### LINKED ISSUE #2459: Update ray base image for TeamCity tests
Ray has upgraded their [docker image](https://hub.docker.com/r/rayproject/ray) from Ubuntu 16.04 to Ubuntu 20.04. It is best if we build our teamcity images using ray image directly without creating an intermediate base image that has to be updated manually.

gshimansky:

## What do these changes do?

- [x] passes `flake8 modin`
- [x] Resolves #2459 
- [x] tests added and passing

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -32,5 +32,6 @@ dependencies:
   - rpyc==4.1.5
   - cloudpickle==1.4.1
   - boto3
+  - github
   - pip:
     - ray>=1.0.0
--- a/environment.yml
+++ b/environment.yml
@@ -32,7 +32,5 @@ dependencies:
   - rpyc==4.1.5
   - cloudpickle==1.4.1
   - boto3
-  - curl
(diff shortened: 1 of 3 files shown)


environment.yml
```diff
@@ -32,5 +32,7 @@ dependencies:
   - rpyc==4.1.5
   - cloudpickle==1.4.1
   - boto3
+  - curl
   - pip:
     - ray>=1.0.0
+    - PyGithub
```
anmyachev, line 38: `curl` and `PyGithub` are package specific for internal testing, let's install them explicitly in dockerfile.

gshimansky: Ok I moved `curl` and `PyGithub` to docker file.

anmyachev (approved):

PR #2483, opened 2020-11-27
DOCS-#2436: Explicit local / single node backend

### LINKED ISSUE #2436: Explicit backend use in case of LOCAL run
When we run MODIN in local whitout providing any dask or .. cluster does it run with the [multiprocessing ](https://docs.python.org/fr/3/library/multiprocessing.html) package of the lib python?  This could be explicited in the documentation and README.md

At first I was looking how and why would modin run  a "fake" dask cluster in local.

What do you think about it ? Thank you

raphaelauv:
Explain how in local ( single node ) modin work and use a simulated dark or ray cluster

Closes #2436

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,6 +131,8 @@ import numpy as np
 frame_data = np.random.randint(0, 100, size=(2**10, 2**8))
 df = pd.DataFrame(frame_data)
 ```
+**In local (without a cluster) modin will create and manage a local (dask or ray) cluster for the execution**
+
 
 To use Modin, you do not need to know how many cores your system has and you do not need
 to  specify how to distribute the data. In fact, you can continue using your previous
--- a/docs/using_modin.rst
+++ b/docs/using_modin.rst
@@ -21,6 +21,8 @@ functional parity with pandas.
 Using Modin on a Single Node
 ----------------------------
 
+**In local (without a cluster) modin will create and manage a local (dask or ray) cluster for the execution**
+
 In order to use the most up-to-date version of Modin, please follow the instructions on
 the `installation page`_
 


README.md
```diff
@@ -131,6 +131,8 @@ import numpy as np
 frame_data = np.random.randint(0, 100, size=(2**10, 2**8))
 df = pd.DataFrame(frame_data)
 ```
+**In local (without a cluster) modin will create and manage a simulated (dask or ray) cluster for the execution**
```
devin-petersohn, line 134: Can we change the word `simulated` to `local`? Simulated is not a term that will be familiar to the readers.

docs/using_modin.rst
```diff
@@ -21,6 +21,8 @@ functional parity with pandas.
 Using Modin on a Single Node
 ----------------------------
 
+**In local (without a cluster) modin will create and manage a simulated (dask or ray) cluster for the execution**
```
devin-petersohn, line 24: Same here for `simulated` -> `local`

devin-petersohn (changes requested):
Thanks @raphaelauv, great addition!

One wording change for using terminology more common to users.

devin-petersohn:
@raphaelauv Thanks for the patience with the review, it was a holiday last week in the US, so many were not able to review it yet.

README.md
raphaelauv: Okay

raphaelauv:
@devin-petersohn thank for your review , I updated the PR

anmyachev (approved):
Thanks @raphaelauv!

devin-petersohn (approved):

PR #2495, opened 2020-12-01
TEST-#2295: Cover by tests Quoting, Compression, and File Format parameters of read_csv

### LINKED ISSUE #2295: Cover by tests Quoting, Compression, and File Format parameters of read_csv
According to source files (https://github.com/pandas-dev/pandas/blob/v1.1.3/pandas/io/parsers.py#L540-L597) `pandas.read_csv` parameters roughly divided into several groups, this issue aims to cover `Quoting, Compression, and File Format` parameters.

amyskov:

## What do these changes do?

- [x] passes `flake8 modin`
- [x] Resolves #2295  
- [x] tests passing

### DIFF
--- a/modin/pandas/test/test_io.py
+++ b/modin/pandas/test/test_io.py
@@ -75,6 +75,16 @@ DATASET_SIZE_DICT = {
 
 # Number of rows in the test file
 NROWS = DATASET_SIZE_DICT.get(TestDatasetSize.get(), DATASET_SIZE_DICT["Small"])
+comp_to_ext = {"infer": "", "gzip": "gz", "bz2": "bz2", "xz": "xz", "zip": "zip"}
+
+test_csv_dialect_params = {
+    "delimiter": "_",
+    "doublequote": False,
+    "escapechar": "d",
+    "quotechar": "d",
+    "quoting": csv.QUOTE_ALL,
+}
+csv.register_dialect("test_csv_dialect", **test_csv_dialect_params)
 
 if not os.path.exists(IO_OPS_DATA_DIR):
     os.mkdir(IO_OPS_DATA_DIR)
@@ -235,11 +245,12 @@ def _make_csv_file(filenames):
                 df["col6"] = df["col6"].apply(
                     lambda x: f"{x:,f}".replace(",", thousands_separator)
                 )
+            filename = (
+                f"{filename}.{comp_to_ext[compression]}"
+                if compression != "infer"
+                else filename
+            )
 
-            if compression == "gzip":
-                filename = "{}.gz".format(filename)
-            elif compression == "zip" or compression == "xz" or compression == "bz2":
-                filename = "{fname}.{comp}".format(fname=filename, comp=compression)
             df.to_csv(
                 filename,
                 sep=delimiter,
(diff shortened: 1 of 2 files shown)


modin/pandas/test/test_io.py
```diff
@@ -860,6 +861,135 @@ def test_read_csv_iteration(self, make_csv_file, iterator):
...
+        unique_filename = get_unique_filename()
+        if dialect:
+            test_csv_dialect_params = {
+                "delimiter": "_",
+                "doublequote": False,
+                "escapechar": "d",
+                "quotechar": "d",
+                "quoting": csv.QUOTE_ALL,
+            }
+            csv.register_dialect(dialect, **test_csv_dialect_params)
```
anmyachev, line 926: Why this line?

modin/pandas/test/test_io.py
```diff
@@ -860,6 +861,135 @@ def test_read_csv_iteration(self, make_csv_file, iterator):
...
         df_equals(modin_df, pd_df)
 
+    # Quoting, Compression, and File Format parameters tests
+    @pytest.mark.parametrize("compression", ["infer", "gzip", "bz2", "xz", "zip"])
+    @pytest.mark.parametrize(
+        "encoding",
+        [None, "latin8", "ISO-8859-1", "latin1", "iso-8859-1", "cp1252", "utf8"],
+    )
+    @pytest.mark.parametrize("engine", [None, "python", "c"])
+    def test_read_csv_compression(self, make_csv_file, compression, encoding, engine):
```
anmyachev, line 871: Why are encoding and compression tests combined?

modin/pandas/test/utils.py
```diff
@@ -431,6 +431,9 @@
 # the type of this exceptions are the same
 io_ops_bad_exc = [TypeError, FileNotFoundError]
 
+# Files compression to extension mapping
+comp_to_ext = {"gzip": "gz", "bz2": "bz2", "xz": "xz", "zip": "zip"}
+
```
anmyachev, line 436: Move into `_csv_file_maker`. The context is too big for this variable considering how it is used.

anmyachev (changes requested):

modin/pandas/test/test_io.py
devin-petersohn: @anmyachev The combination of these two parameters is important to test so we ensure that when the file is compressed in a non UTF-8 format it can be properly read by Modin.

modin/pandas/test/test_io.py
amyskov: This line is needed because `dialect` should the instance of `csv.Dialect` class (`csv.get_dialect()` to get that instance is also added).

modin/pandas/test/utils.py
amyskov: `comp_to_ext` is also needed in `test_read_csv_compression` function, so it was moved to the `test_io.py`.

modin/pandas/test/test_io.py
anmyachev: got it, thanks @devin-petersohn

modin/pandas/test/test_io.py
```diff
@@ -76,6 +76,9 @@
 # Number of rows in the test file
 NROWS = DATASET_SIZE_DICT.get(TestDatasetSize.get(), DATASET_SIZE_DICT["Small"])
 
+# Files compression to extension mapping
+comp_to_ext = {"gzip": "gz", "bz2": "bz2", "xz": "xz", "zip": "zip"}
```
anmyachev, line 80: Variables in the global scope must be in upper case.
```suggestion
COMP_TO_EXT = {"gzip": "gz", "bz2": "bz2", "xz": "xz", "zip": "zip"}
```

anmyachev (approved):
LGTM

PR #2514, opened 2020-12-05
Ensure excel reader closes file if it is passed as path

### LINKED ISSUE #2465: read_excel prevent from further save of the Excel file while python session is running
### System information
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: Windows 10 20H2
- **Modin version** (`modin.__version__`): 0.8.2
- **Python version**: Python 3.8.3

An really simple example file to reproduce:
[test.xlsx](https://github.com/modin-project/modin/files/5570144/test.xlsx)

```python
import modin.pandas as pd
df = pd.read_excel('test.xlsx')
```
Then open the Excel file (if not already open), and try to save it.

### Describe the problem

Contrary to behavior in Pandas, using pd.read_excel with modin, prevent any further saving of the Excel files afterwards, until the python session is closed.
While trying to do it in Excel, get multiple error pop-up talking about "sharing issue", given some random number as Excel file then.
For my use case, it's quite handy to get the Excel file and the python session open in the same time, loading the data, working on it with pandas, and potentially (especially during developpement) making small change in Excel and reloading them right away in python. Also, that doesn't seem like a safe behavior and might hide something bigger ? don't know.

### Source code / logs

image
image
image
(sorry it's in french)

vnlitvinov:
## What do these changes do?

`openpyxl` has either a long-standing bug or a design decision that, when opening a file in read-only mode, it does not close the file handle (maybe it does so for caching purposes, no idea).
They have introduced a new `.close()` method, but people say it does not fully close all the handles.

So what I instead did was open the file if passed in a string and close it explicitly.
Also I've optimized the code a little bit by effectively inlining `load_workbook()` function which allowed to eliminate creation of _two_ `ExcelReader()` objects for a single read.

I've also decorated all the excel-reading tests with the checker ~~from Pandas test utils which they introduced in https://github.com/pandas-dev/pandas/pull/30096 when they were fixing this exact issue :)~~ inspired by Pandas but with a little more resilience (it only pays attention to file names and descriptors and ignores the order of open files).

- [x] passes `flake8 modin`
- [x] Resolves #2465  
- [x] tests added and passing

### DIFF
--- a/modin/config/envvars.py
+++ b/modin/config/envvars.py
@@ -12,6 +12,7 @@
 # governing permissions and limitations under the License.
 
 import os
+import sys
 from textwrap import dedent
 import warnings
 from packaging import version
@@ -206,6 +207,18 @@ class TestDatasetSize(EnvironmentVariable, type=str):
 
     varname = "MODIN_TEST_DATASET_SIZE"
     choices = ("Small", "Normal", "Big")
+
+
+class TrackFileLeaks(EnvironmentVariable, type=bool):
+    """
+    Whether to track for open file handles leakage during testing
+    """
+
+    varname = "MODIN_TEST_TRACK_FILE_LEAKS"
+    # Turn off tracking on Windows by default because
+    # psutil's open_files() can be extremely slow on Windows (up to adding a few hours).
+    # see https://github.com/giampaolo/psutil/pull/597
+    default = sys.platform != "win32"
 
 
 def _check_vars():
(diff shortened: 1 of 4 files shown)


vnlitvinov:
**Note**: if you take only first commit (where test decorators were added but no fix implemented), the test that fails for me is `test_io.py::test_from_excel_sheet_name` (it passes again after the fix).

anmyachev (approved):
Nice fix!

amyskov (approved):
LGTM

PR #2528, opened 2020-12-10
FIX-#2527: Use random name for hdf file test, clean file after testing

### LINKED ISSUE #2527: test_HDFStore uses hardcoded absolute filename which breaks parallel test sessions
### System information
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: any Linux
- **Modin version** (`modin.__version__`): current master (25928495ed5bf69ae54067952e41a05306fc596b)
- **Python version**: any
- **Code we can use to reproduce**: create a read-only file `/tmp/test_read_hdf.hdf5` and run `test_io.py::test_HDFStore`

### Describe the problem

Pandas up to (including) 1.1.5 use some "funny" logic when opening a HDFStore - if opening in write mode fails it silently (well, it prints to stdout which itself is a strange decision for a library) opens the store in read-only mode.
So if the file exists and, for example, was created by another user, all other test sessions on that box would fail.

This was fixed in Pandas via https://github.com/pandas-dev/pandas/pull/36859, but that isn't released yet.

### Source code / logs

Current proposal: do not use hardcoded absolute path for the store.

cc @YarShev

vnlitvinov:

## What do these changes do?
Ensure each run of `test_HDFStore` uses different test file, also clean it up after test is over.

- [x] passes `flake8 modin`
- [x] Resolves #2527 
- [x] tests added and passing

### DIFF
--- a/modin/pandas/test/test_io.py
+++ b/modin/pandas/test/test_io.py
@@ -26,6 +26,7 @@ import os
 import shutil
 import sqlalchemy as sa
 import csv
+import tempfile
 
 from .utils import (
     check_file_leaks,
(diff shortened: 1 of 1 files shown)


YarShev (approved):
@vnlitvinov , LTGM!

PR #2532, opened 2020-12-11
FIX-#2498: Fix possible number of partitions for Dask engine

### LINKED ISSUE #2498: Dask engine can only have 4 partitions - Pull Requests #2492
### System information
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: CentOS release 6.5 (Final)
- **Modin version** (`modin.__version__`): 0.8.2 installed from conda-forge channel
- **Python version**: 3.7.3
```python
client = Client(n_workers=10)
import modin.pandas as pd
df = pd.read_csv(filename)
```

### Describe the problem

Hi there, I recently got a really huge CSV file to read and found that no matter how I declare the number of the worker by using
```python
client = Client(n_workers=10)
import modin.pandas as pd
```
The partition of the Dataframe object is still 4 causing the efficiency really low.

So I reviewed some of the code briefly (actually it takes me a whole afternoon to track the problem) and found `pd.DEFAULT_NPARTITIONS` is 4 defaulted (in file `modin/modin/pandas/__init__.py`) and the init function (`_update_engine`) didn't change the value of it when using Dask engine.

Hence, I change `pd.DEFAULT_NPARTITIONS` after importing `pd` by 
```python
pd.DEFAULT_NPARTITIONS = 10
```
and it works well!

So I decided to add a line of code to solve this problem (See Pull Requests #2492 ). However, it failed in `test_read_csv_datetime`. Really confused about that.

YarShev:

## What do these changes do?

- [x] passes `flake8 modin`
- [x] Resolves #2498  
- [x] tests passing

### DIFF
--- a/modin/pandas/__init__.py
+++ b/modin/pandas/__init__.py
@@ -135,6 +135,8 @@ def _update_engine(publisher: Parameter):
 
                 dask_client = Client(n_workers=CpuCount.get())
 
+            num_cpus = len(dask_client.ncores())
+
     elif publisher.get() == "Cloudray":
         from modin.experimental.cloud import get_connection
 
--- a/modin/pandas/test/test_io.py
+++ b/modin/pandas/test/test_io.py
@@ -587,7 +587,7 @@ class TestCsv:
             None,
             {
                 "col1": lambda x: np.int64(x) * 10,
-                "col2": pd.to_datetime,
+                "col2": pandas.to_datetime,
                 "col4": lambda x: x.replace(":", ";"),
             },
         ],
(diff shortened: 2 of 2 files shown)


vnlitvinov (commented):
CI is failing for dask part of testing... is it a coincidence?

YarShev:
It looks like the same failure which we saw earlier. Tests are passing locally. I will rerun CI.

modin/pandas/__init__.py
```diff
@@ -134,6 +134,7 @@ def _update_engine(publisher: Parameter):
                 from distributed import Client
 
                 dask_client = Client(n_workers=CpuCount.get())
+                num_cpus = len(dask_client.cluster.plan)
```
vnlitvinov, line 137: shouldn't this be done outside of `except`? inside `except` there is the logic where we override the cpu cound (and we can get cpu count from `CpuCount.get()`)

YarShev: Definitely, thanks!

YarShev:
As I expected, `cluster` member is not always equal to not `None`. I am changing the fix.

vnlitvinov (approved):
A question for change in testing, otherwise it's fine.

modin/pandas/test/test_io.py
```diff
@@ -587,7 +587,7 @@ def test_read_csv_col_handling(
             None,
             {
                 "col1": lambda x: np.int64(x) * 10,
-                "col2": pd.to_datetime,
+                "col2": pandas.to_datetime,
```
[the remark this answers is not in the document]
YarShev: The reason is the following ([link](https://github.com/modin-project/modin/issues/2498#issuecomment-745808956)).

PR #2515, opened 2020-12-07
FEAT-#2342: Add axis partitions API

YarShev:
## What do these changes do?

- [x] commit message follows format outlined [here](https://modin.readthedocs.io/en/latest/CONTRIBUTING.html)
- [x] passes `flake8 modin`
- [x] passes `black --check modin`
- [x] signed commit with `git commit -s` 
- [x] Resolves #2342  
- [x] tests passing

PR #2545, opened 2020-12-16
FEAT-#2373: Add distributed xgboost on Modin with Ray

### LINKED ISSUE #2373: Prototype of xgboost + Modin predictor
### Describe the problem

Make prototype of xgboost scalable predictor with Modin.

prutskov:
## What do these changes do?
This new feature gives possibility to run distributed xgboost on Ray cluster using Modin DataFrame as input data. Currently supported classic xgboost interfaces `xgboost.train` and `xgboost.predict`.

`modin.experimental.xgboost` supports two ways of execution which can be switched by parameter `evenly_data_distribution` in `train/predict` functions:

1. `evenly_data_distribution==True`: in this case input data of `train/predict` functions will be distributed evenly between nodes in cluster to ensure evenly utilization of nodes (default behavior).
2. `evenly_data_distribution==False`: in this case partitions of input data of `train/predict` functions will not transfer between nodes in cluster. This method provides minimal data transfers between nodes but don't guarantee effective utilization of nodes. Most effective in case when all cluster nodes are occupied by data.

This PR contains temporary changes from #2515.

- [ ] passes `flake8 modin`
- [ ] Resolves #2373.  
- [ ] tests added and passing


prutskov:
> Is this somehow related to [ray-project/xgboost_ray#31](https://github.com/ray-project/xgboost_ray/pull/31) ?

@vnlitvinov This work is separate from `ray-project/xgboost_ray`, but discussion with [@richardliaw](https://github.com/richardliaw) about joining of work has been started.

YarShev:
Since this PR needs to be constantly updated because of changes in #2515, I would leave here only those changes which are related to this PR in order to do not confuse other reviewers.

prutskov:
> Since this PR needs to be constantly updated because of changes in #2515, I would leave here only those changes which are related to this PR in order to do not confuse other reviewers.

@YarShev, @devin-petersohn Changes from #2515 is required for local testing, but I don't see the need for constantly updating changes in files from #2515. Changes will be updated in case interfaces/behavior changing in #2515. 

**NOTE**: This PR consists only from `modin/experimental/xgboost/*` files. Other files are on review in #2515 and will be deleted from this PR after #2515 is merged into master branch.

modin/experimental/xgboost/xgboost.py
```diff
@@ -0,0 +1,499 @@
...
+    def __enter__(self):
+        xgb.rabit.init(self.args)
+        LOGGER.info("-------------- rabit started ------------------")
+
+    def __exit__(self, *args):
+        xgb.rabit.finalize()
+        LOGGER.info("-------------- rabit finished ------------------")
+
+
+@ray.remote
```
vnlitvinov, line 73: shouldn't this be in separate module, like `xgboost_ray.py`?

modin/experimental/xgboost/xgboost.py
```diff
@@ -0,0 +1,499 @@
...
+                else slice(start_idx, len(row_partitions))
+            )
+            row_partitions_by_actors.append(row_partitions[idx_slice])
+            start_idx += row_parts_last_idx
+    else:
+        actors_ips = [ray.get(actor.get_actor_ip.remote()) for actor in actors]
+        partitions_ips = [ray.get(row[0]) for row in row_partitions]
+
+        # If one of actors isn't contain partitions we make evenly
+        # distribution of data by nodes.
```
vnlitvinov, line 254: why?.. if I have 1000 actors, and 999 have evenly spaced data, it doesn't make much sense to redistribute it all IMHO... this should probably be checking that actor IPs match with `data_for_aligning` instead of this

modin/experimental/xgboost/xgboost.py
```diff
@@ -0,0 +1,499 @@
...
+        ):
+            return _get_evenly_distr_of_partitions_by_ips(
+                actors, actors_ips, row_partitions
+            )
+
+        if data_for_aligning is None:
+            for actor_ip in actors_ips:
+                actor_row_partitions = []
+                order_of_indexes = []
+                for i, row in enumerate(row_partitions):
```
vnlitvinov, line 267: this looks like `O(N^2)` algo where using a dict (or an OrderedDict if order matters) should be `O(N*log N)`... dunno if it matters as I don't know how big are actors' and partitions' amounts

modin/experimental/xgboost/xgboost.py
```diff
@@ -0,0 +1,499 @@
...
+        first = start_idx
+        last = (
+            (start_idx + row_parts_last_idx)
+            if (start_idx + row_parts_last_idx < len(row_partitions))
+            else len(row_partitions)
+        )
+
+        actor_row_partitions = []
+        order_of_indexes = []
+        while first < last:
```
vnlitvinov, line 307: umm, `for first in range(start_idx, last)`? :)

modin/experimental/xgboost/xgboost.py
```diff
@@ -0,0 +1,499 @@
...
+):
+    """
+    Train XGBoost model.
+
+    Parameters
+    ----------
+    params : dict
+        Booster params.
+    dtrain : ModinDMatrix
+        Data to be trained.
```
vnlitvinov, line 335:
```suggestion
        Data to be trained against.
```
as you cannot train the data itself to do anythin, but only a model against the data ;)

modin/experimental/xgboost/xgboost.py
```diff
@@ -0,0 +1,499 @@
...
+    nthread: int
+        Number of threads for using in each node. By default it is equal to
+        number of threads on master node.
+    evenly_data_distribution: boolean, default True
+        Whether make evenly distribution of partitions between nodes or not.
+        In case `False` minimal datatransfer between nodes will be provided
+        but the data may not be evenly distributed.
+    \\*\\*kwargs:
+        Other parameters are the same as `xgboost.train` except for
+        `evals_result`, which is returned as part of function return value
```
vnlitvinov, line 345: why diverge from xgboost API?

modin/experimental/xgboost/xgboost.py
```diff
@@ -0,0 +1,499 @@
...
+    assert len(X) == len(y)
+    X_row_parts = unwrap_row_partitions(X, bind_ip=not evenly_data_distribution)
+    y_row_parts = unwrap_row_partitions(y, bind_ip=not evenly_data_distribution)
+    assert len(X_row_parts) == len(y_row_parts), "Unaligned train data"
+
+    # Create remote actors
+    actors = create_actors(nthread=nthread)
+
+    add_as_eval_method = None
+    if len(evals):
```
vnlitvinov, line 374:
```suggestion
    if evals:
```

modin/experimental/xgboost/xgboost.py
```diff
@@ -0,0 +1,499 @@
...
+        for (eval_data, method) in evals:
+            if id(eval_data) == id(dtrain):
```
vnlitvinov, line 376:
```suggestion
            if eval_data is dtrain:
```

modin/experimental/xgboost/xgboost.py
```diff
@@ -0,0 +1,499 @@
...
+                add_as_eval_method = method
+                evals.remove((eval_data, method))
```
vnlitvinov, line 378: you're modifying a list which you're iterating over... this is unsafe and could lead to subtle bugs, you should iterate over a copy of the list then, like `for (eval_data, method) in evals[:]:`

modin/experimental/xgboost/xgboost.py
```diff
@@ -0,0 +1,499 @@
...
+        evals_unwrapped = [
+            (
+                (
+                    unwrap_row_partitions(eval_X, bind_ip=not evenly_data_distribution),
+                    unwrap_row_partitions(eval_y, bind_ip=not evenly_data_distribution),
+                    eval_method,
+                )
+            )
+            for ((eval_X, eval_y), eval_method) in evals
+        ]
```
vnlitvinov, line 389: why do this here instead of explicitly calling `unwrap_row_partitions(eval_X)` for arguments of `_split_data_across_actors`? seems like two loops where one is enough

modin/experimental/xgboost/xgboost.py
```diff
@@ -0,0 +1,499 @@
...
+from typing import Dict, Optional
+from multiprocessing import cpu_count
+
+import xgboost as xgb
+import numpy as np
+import pandas
+
+from modin.developer import unwrap_row_partitions
+
+if os.environ.get("MODIN_ENGINE", "Ray").title() == "Ray":
```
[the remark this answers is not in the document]
prutskov: fixed

modin/experimental/xgboost/xgboost.py
prutskov: fixed

modin/experimental/xgboost/xgboost.py
prutskov: fixed

modin/experimental/xgboost/xgboost.py
prutskov: fixed

modin/experimental/xgboost/xgboost.py
prutskov: fixed

modin/experimental/xgboost/xgboost.py
prutskov: Good point, it's fixed.

modin/experimental/xgboost/xgboost.py
prutskov: Thanks, explicitly calling was added for argoments of `_split_data_across_actors`.

modin/experimental/xgboost/xgboost.py
prutskov: I think it will make sense in case supporting another distributed engines by modin.xgboost.

modin/experimental/xgboost/xgboost.py
prutskov: This is made accordance with [xgboost.dask](https://github.com/dmlc/xgboost/blob/master/python-package/xgboost/dask.py#L871) API for aligning with distributed API.

modin/experimental/xgboost/xgboost.py
```diff
@@ -0,0 +1,482 @@
...
+
+    def set_predict_data(
+        self,
+        *data_X,
+    ):
+        for X in data_X:
+            self._dpredict.append(xgb.DMatrix(X, None))
+
+    def add_eval_data(self, *data_X_y, eval_method):
+        self._evals.append((self._get_dmatrix(*data_X_y), eval_method))
```
YarShev, line 117: Why as list?

modin/experimental/xgboost/xgboost.py
```diff
@@ -0,0 +1,482 @@
...
+        for (eval_data, method) in evals[:]:
+            if eval_data is dtrain:
+                add_as_eval_method = method
+                evals.remove((eval_data, method))
+
+        for ((eval_X, eval_y), eval_method) in evals:
+            # Split data across workers
+            _split_data_across_actors(
+                actors,
+                lambda actor, *Xy: actor.add_eval_data.remote(
```
YarShev, line 382: `Xy` -> `X_y`.

modin/experimental/xgboost/xgboost.py
```diff
@@ -0,0 +1,482 @@
...
+                    *Xy, eval_method=eval_method
```
YarShev, line 383: `Xy` -> `X_y`.

modin/experimental/xgboost/xgboost.py
```diff
@@ -0,0 +1,482 @@
...
+            len(X) == len(y) and len(X) > 0
+        ), "X and y should have the equal length more than 0"
+
+        X = pandas.concat(X, axis=0)
+        y = pandas.concat(y, axis=0)
+        LOGGER.info(f"Concat time: {time.time() - s} s")
+
+        return xgb.DMatrix(X, y)
+
+    def set_train_data(self, *data_X_y, add_as_eval_method=None):
```
YarShev, line 103: `data_X_y` -> `X_y`.

modin/experimental/xgboost/xgboost.py
```diff
@@ -0,0 +1,482 @@
...
+        self._dtrain = self._get_dmatrix(*data_X_y)
+
+        if add_as_eval_method is not None:
+            self._evals.append((self._dtrain, add_as_eval_method))
+
+    def set_predict_data(
+        self,
+        *data_X,
```
YarShev, line 111: `data_X` -> `X`.

modin/experimental/xgboost/xgboost.py
```diff
@@ -0,0 +1,482 @@
...
+    ):
+        for X in data_X:
```
YarShev, line 113: `for X in data_X:` -> `for x in X:`.

modin/experimental/xgboost/xgboost.py
```diff
@@ -0,0 +1,482 @@
...
+            self._dpredict.append(xgb.DMatrix(X, None))
```
YarShev, line 114: `X` -> `x`.

modin/experimental/xgboost/xgboost.py
```diff
@@ -0,0 +1,482 @@
...
+
+    def add_eval_data(self, *data_X_y, eval_method):
```
YarShev, line 116: `data_X_y` -> `X_y`.

modin/experimental/xgboost/xgboost.py
```diff
@@ -0,0 +1,482 @@
...
+        else:
+            y_parts = (
+                y_parts_by_actors[i]
+                if evenly_data_distribution
+                else y_parts_by_actors[i][2]
+            )
+            set_func(actor, *(X_parts + y_parts))
+
+
+def _set_row_partitions_by_actors(
```
YarShev, line 228: `_set_row_partitions_by_actors` -> `_set_row_partitions_to_actors`.

modin/experimental/xgboost/xgboost.py
```diff
@@ -0,0 +1,482 @@
...
+    actors, row_partitions, data_for_aligning=None, evenly_data_distribution=True
+):
+
+    row_partitions_by_actors = []
```
YarShev, line 232: `row_partitions_by_actors` -> `row_partitions_to_actors`

modin/experimental/xgboost/xgboost.py
```diff
@@ -0,0 +1,482 @@
...
+                if start_idx + row_parts_last_idx < len(row_partitions)
+                else slice(start_idx, len(row_partitions))
+            )
+            row_partitions_by_actors.append(row_partitions[idx_slice])
+            start_idx += row_parts_last_idx
+    else:
+        actors_ips = [ray.get(actor.get_actor_ip.remote()) for actor in actors]
+        partitions_ips = [ray.get(row[0]) for row in row_partitions]
+
+        # If one of actors isn't contain partitions we make evenly
```
YarShev, line 253: `If one of actors isn't contain partitions we make evenly` -> `If one of the actors doesn't contain partitions we make evenly`

modin/experimental/xgboost/xgboost.py
```diff
@@ -0,0 +1,482 @@
...
+
+
+def _get_evenly_distr_of_partitions_by_ips(actors, actors_ips, row_partitions):
+    num_actors = len(actors)
+    row_parts_last_idx = (
+        len(row_partitions) // num_actors
+        if len(row_partitions) % num_actors == 0
+        else len(row_partitions) // num_actors + 1
+    )
+    row_partitions_by_actors = []
```
YarShev, line 295: `row_partitions_by_actors` -> `row_partitions_to_actors`

modin/experimental/xgboost/xgboost.py
```diff
@@ -0,0 +1,482 @@
...
+    nthread: Optional[int] = cpu_count(),
+    evenly_data_distribution: Optional[bool] = True,
+    **kwargs,
+):
+    """
+    Run prediction with a trained booster.
+
+    Parameters
+    ----------
+    model: A Booster or a dictionary returned by `modin.experimental.xgboost.train`.
```
YarShev, line 430: `model:` -> `model :` and below in such way

modin/experimental/xgboost/xgboost.py
YarShev: Number of partitions is more than number of actors. As a rule, partitions' amount is measured from decades to thousands and actors' amount is measured from some to decades/hundreds (in depend of a cluster used).

modin/experimental/xgboost/xgboost.py
YarShev: It's a good point.

YarShev (changes requested):

modin/experimental/xgboost/xgboost.py
prutskov: fixed

modin/experimental/xgboost/xgboost.py
prutskov: fixed

modin/experimental/xgboost/xgboost.py
prutskov: fixed

modin/experimental/xgboost/xgboost.py
```diff
@@ -0,0 +1,499 @@
...
+        LOGGER.info("-------------- rabit started ------------------")
+
+    def __exit__(self, *args):
+        xgb.rabit.finalize()
+        LOGGER.info("-------------- rabit finished ------------------")
+
+
+@ray.remote
+class ModinXGBoostNodeActor:
+    def __init__(self, node, nthread=cpu_count()):
```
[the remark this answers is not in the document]
prutskov: fixed

modin/experimental/xgboost/xgboost.py
```diff
@@ -0,0 +1,499 @@
...
+        self._evals = []
+        self._dpredict = []
+        self._node = node
+        self._nthreads = nthread
+
+        LOGGER.info(
+            f"Actor <{self._node}>, nthread = {self._nthreads} was initialized."
```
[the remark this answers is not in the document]
prutskov: fixed

modin/experimental/xgboost/xgboost.py
```diff
@@ -0,0 +1,499 @@
...
+        xgb.rabit.finalize()
+        LOGGER.info("-------------- rabit finished ------------------")
+
+
+@ray.remote
+class ModinXGBoostNodeActor:
+    def __init__(self, node, nthread=cpu_count()):
+        self._evals = []
+        self._dpredict = []
+        self._node = node
```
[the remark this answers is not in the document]
prutskov: fixed

modin/experimental/xgboost/xgboost.py
```diff
@@ -0,0 +1,499 @@
...
+        self._nthreads = nthread
+
+        LOGGER.info(
+            f"Actor <{self._node}>, nthread = {self._nthreads} was initialized."
+        )
+
+    def get_actor_ip(self):
+        return self._node
```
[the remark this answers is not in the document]
prutskov: fixed

modin/experimental/xgboost/xgboost.py
```diff
@@ -0,0 +1,499 @@
...
+        row_partitions_by_actors.append((ip, order_of_indexes, actor_row_partitions))
+
+    return row_partitions_by_actors
+
+
+def train(
+    params: Dict,
+    dtrain: ModinDMatrix,
+    *args,
+    evals=(),
```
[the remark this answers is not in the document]
prutskov: Dosctring was added

modin/experimental/xgboost/xgboost.py
```diff
@@ -0,0 +1,482 @@
...
+
+        s = time.time()
+        predictions = [booster.predict(X, *args, **kwargs) for X in local_dpredict]
+        LOGGER.info(f"Local prediction time: {time.time() - s} s")
+        return np.concatenate(predictions)
+
+
+class ModinDMatrix(xgb.DMatrix):
+    """
+    DMatrix holding on references to Modin.DataFrame.
```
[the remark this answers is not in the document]
prutskov: fixed

modin/experimental/xgboost/xgboost.py
```diff
@@ -0,0 +1,482 @@
...
+
+    Parameters
+    ----------
+    data : modin.DataFrame
```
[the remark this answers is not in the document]
prutskov: fixed

modin/experimental/xgboost/xgboost.py
```diff
@@ -0,0 +1,482 @@
...
+        data source of DMatrix.
+    label: modin.DataFrame
```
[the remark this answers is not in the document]
prutskov: fixed

modin/experimental/xgboost/xgboost.py
```diff
@@ -0,0 +1,482 @@
...
+        xgb.rabit.init(self.args)
+        LOGGER.info("-------------- rabit started ------------------")
+
+    def __exit__(self, *args):
+        xgb.rabit.finalize()
+        LOGGER.info("-------------- rabit finished ------------------")
+
+
+@ray.remote
+class ModinXGBoostNodeActor:
```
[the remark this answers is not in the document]
prutskov: fixed

modin/experimental/xgboost/xgboost.py
```diff
@@ -0,0 +1,482 @@
...
+        yield self.data
+        yield self.label
+
+
+def create_actors(num_cpus=1, nthread=cpu_count()):
+    num_nodes = len(ray.nodes())
+
+    # Create remote actors
+    actors = [
+        ModinXGBoostNodeActor.options(num_cpus=num_cpus, resources={node: 1.0}).remote(
```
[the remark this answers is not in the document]
prutskov: fixed

modin/experimental/xgboost/xgboost.py
```diff
@@ -0,0 +1,482 @@
...
+        self._nthreads = nthread
+
+        LOGGER.info(
+            f"Actor <{self._node}>, nthread = {self._nthreads} was initialized."
+        )
+
+    def get_actor_ip(self):
+        return self._node
+
+    def _get_dmatrix(self, *data_X_y):
```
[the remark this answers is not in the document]
prutskov: fixed

modin/experimental/xgboost/xgboost.py
prutskov: fixed

modin/experimental/xgboost/xgboost.py
prutskov: fixed

modin/experimental/xgboost/xgboost.py
prutskov: fixed

modin/experimental/xgboost/xgboost.py
prutskov: fixed

modin/experimental/xgboost/xgboost.py
prutskov: fixed

modin/experimental/xgboost/xgboost.py
prutskov: fixed

modin/experimental/xgboost/xgboost.py
prutskov: fixed

modin/experimental/xgboost/xgboost.py
prutskov: fixed

modin/experimental/xgboost/xgboost.py
prutskov: fixed

modin/experimental/xgboost/xgboost.py
```diff
@@ -0,0 +1,499 @@
...
+    rabit_tracker = xgb.RabitTracker(hostIP=host, nslave=num_workers)
+
+    # Get tracker Host + IP
+    env.update(rabit_tracker.slave_envs())
+    rabit_tracker.start(num_workers)
+
+    # Wait until context completion
+    thread = Thread(target=rabit_tracker.join)
+    thread.daemon = True
+    thread.start()
```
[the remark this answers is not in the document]
prutskov: This is need for correct lifetime of `rabit_tracker`.

modin/experimental/xgboost/xgboost.py
```diff
@@ -0,0 +1,499 @@
...
+
+from modin.developer import unwrap_row_partitions
+
+if os.environ.get("MODIN_ENGINE", "Ray").title() == "Ray":
+    import ray
+    from ray.services import get_node_ip_address
+else:
+    raise ValueError("Current version supports only Ray engine as MODIN_ENGINE.")
+
+LOGGER = logging.getLogger("[xgboost.modin]")
```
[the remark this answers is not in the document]
prutskov: fixed

modin/experimental/xgboost/xgboost.py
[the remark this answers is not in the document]
vnlitvinov: It would achieve almost the same effect by just adding it to a global variable.
Also, given `RabitTracker` uses a thread internally which target has its contained as a closure variable these 3 lines seem pointless to me.

vnlitvinov (commented):
Please also refrain from rebasing all the time as it makes re-reviewing much harder (reader has to read the whole file over and over again instead of just checking what has changed).

modin/experimental/xgboost/xgboost.py
prutskov: It was fixed. Empty actors won't used in computations but users will notified using warning message about this.

modin/experimental/xgboost/xgboost.py
prutskov: Thank you! Now `dict` is used for set of `actors`, part of loops was changed for decreasing complexity.

modin/experimental/xgboost/xgboost.py
prutskov: Ray-specific functionality was moved to `xgboost_ray.py` file.

modin/experimental/xgboost/xgboost.py
```diff
@@ -12,140 +12,22 @@
...
-
-    # Wait until context completion
-    thread = Thread(target=rabit_tracker.join)
-    thread.daemon = True
-    thread.start()
-
-    return env
-
-
-class RabitContext:
```
vnlitvinov: why this class was moved?

modin/experimental/xgboost/xgboost_ray.py
```diff
@@ -0,0 +1,339 @@
...
+        s = time.time()
+        X = X_y[: len(X_y) // 2]
+        y = X_y[len(X_y) // 2 :]
+
+        assert (
+            len(X) == len(y) and len(X) > 0
+        ), "X and y should have the equal length more than 0"
+
+        X = pandas.concat(X, axis=0)
+        y = pandas.concat(y, axis=0)
```
vnlitvinov, line 91: that's a lot of memory copying (because `X_y` is a tuple due to being an asterisk argument), can this be reduced somehow?

modin/experimental/xgboost/xgboost_ray.py
```diff
@@ -0,0 +1,339 @@
...
+
+
+def _split_data_across_actors(
+    actors: Dict, set_func, X_parts, y_parts=None, evenly_data_distribution=True
+):
+    if not evenly_data_distribution:
+        actors_ips = list(actors.keys())
+        partitions_ips = [ray.get(row_part[0]) for row_part in X_parts]
+        if len(np.unique(actors_ips)) != len(np.unique(partitions_ips)):
+            empty_nodes_ips = list(set(actors_ips) - set(partitions_ips))
```
vnlitvinov, line 170: we might want to introduce some threshold here, as it seems both wasteful to redistribute data when 999 of 1000 have data _and_ to not redistribute if only 1 actor out of 1000 has data... maybe we could redistribute if e.g. less than `90%` or smth actors have data?

modin/experimental/xgboost/xgboost_ray.py
```diff
@@ -0,0 +1,339 @@
...
+    return actors
+
+
+def _split_data_across_actors(
+    actors: Dict, set_func, X_parts, y_parts=None, evenly_data_distribution=True
+):
+    if not evenly_data_distribution:
+        actors_ips = list(actors.keys())
+        partitions_ips = [ray.get(row_part[0]) for row_part in X_parts]
+        if len(np.unique(actors_ips)) != len(np.unique(partitions_ips)):
```
vnlitvinov, line 169: This check feels strange... take `actor_ips = [1]` and `partition_ips = [2]` - they have same amount of unique items but are actually different.
I think you should compute `empty_nodes_ips` here and check for it being non-empty.

modin/experimental/xgboost/xgboost.py
prutskov: This class isn't used `xgboost.py` and applicable for implementation only

modin/experimental/xgboost/xgboost_ray.py
prutskov: Do you mean something like that?
```python
def _get_dmatrix(self, X_y):
    X = X_y[: len(X_y) // 2]
    y = X_y[len(X_y) // 2 :]
    ...
    
    
def set_train_data(self, *X_y, add_as_eval_method=None):
    data = list(X_y)
    self._dtrain = self._get_dmatrix(X_y)
```

modin/experimental/xgboost/xgboost.py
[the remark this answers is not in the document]
prutskov: Function `stop_rabit_tracker` was added for stopping tracker after usage.

modin/experimental/xgboost/xgboost.py
prutskov: Fixed by next way: data will be redistributed in case when 10% or more number of nodes are empty.

modin/experimental/xgboost/xgboost.py
prutskov: This class and `start/stop_rabit_tracker` functions was moved into separate file `utils.py` because this functions can be used in implementation with other distribute engines.

modin/experimental/xgboost/xgboost_ray.py
prutskov: `*X_y`(tuple) was changed to `X_y`(list) by a way from comment above.

modin/experimental/xgboost/xgboost_ray.py
prutskov: Fixed by next way: data will be redistributed in case when 10% or more number of nodes are empty.

modin/experimental/xgboost/xgboost_ray.py
prutskov: IPs of partitions is subset of set of actors IPs because the same ray cluster is used. Therefore, in your example `actor_ips = [1,2]` and check will work correctly. But it was changed using list of empty nodes.

modin/experimental/xgboost/xgboost_ray.py
vnlitvinov: I don't think it changes much, might be even changing it to being slower... some profiling and considerations of what is the input are needed. If `X_y` would be a numpy array then selecting half of it would be just making a view, which is almost zero-cost.

modin/experimental/xgboost/utils.py
```diff
@@ -0,0 +1,52 @@
...
+# ANY KIND, either express or implied. See the License for the specific language
+# governing permissions and limitations under the License.
+
+import logging
+import xgboost as xgb
+
+LOGGER = logging.getLogger("[modin.xgboost]")
+
+
+def start_rabit_tracker(num_workers: int, host_ip):
```
vnlitvinov, line 20: why it is not a context manager for the tracker?.. its `__enter__()` could return the tracker, not the context manager, making stopping a tracker automated...

modin/experimental/xgboost/xgboost_ray.py
```diff
@@ -78,7 +42,7 @@ def __init__(self, ip, nthread=cpu_count()):
     def get_actor_ip(self):
         return self._ip
 
-    def _get_dmatrix(self, *X_y):
+    def _get_dmatrix(self, X_y):
```
vnlitvinov: this is certainly better, as passing arguments by asterisk takes extra time

modin/experimental/xgboost/xgboost_ray.py
```diff
@@ -94,7 +58,7 @@ def _get_dmatrix(self, *X_y):
         return xgb.DMatrix(X, y)
 
     def set_train_data(self, *X_y, add_as_eval_method=None):
-        self._dtrain = self._get_dmatrix(*X_y)
+        self._dtrain = self._get_dmatrix(list(X_y))
```
vnlitvinov: this is probably not needed... I was actually wondering if we can pass `X_y` from outside as non-crumpled data

modin/experimental/xgboost/xgboost_ray.py
```diff
@@ -107,7 +71,7 @@ def set_predict_data(
             self._dpredict.append(xgb.DMatrix(x, None))
 
     def add_eval_data(self, *X_y, eval_method):
-        self._evals.append((self._get_dmatrix(*X_y), eval_method))
+        self._evals.append((self._get_dmatrix(list(X_y)), eval_method))
```
vnlitvinov: same

modin/experimental/xgboost/xgboost_ray.py
```diff
@@ -202,28 +153,43 @@ def _set_row_partitions_to_actors(
...
         if data_for_aligning is None:
-            for i, row_part in enumerate(row_partitions):
-                row_partitions_by_actors[partitions_ips[i]][0].append(row_part[1])
-                row_partitions_by_actors[partitions_ips[i]][1].append(i)
+            actors_ips = list(actors.keys())
+            partitions_ips = [ray.get(row_part[0]) for row_part in row_partitions]
+
+            empty_actor_ips = []
+            for ip in actors_ips:
+                if ip not in set(partitions_ips):
```
vnlitvinov, line 169: this `set()` should be computed once, as now it's computed for every actor, which is wasteful

modin/experimental/xgboost/xgboost_ray.py
```diff
@@ -202,28 +153,43 @@ def _set_row_partitions_to_actors(
...
+                    empty_actor_ips.append(ip)
+
+            # In case portion of nodes without data is more than 10%, data
+            # redistribution between nodes will be applied.
```
vnlitvinov, line 173:
```suggestion
            # In case portion of nodes without data is less than 10%,
            # no data redistribution between nodes will be performed.
```
so that comment applies to `then` branch instead of `else`

modin/experimental/xgboost/xgboost_ray.py
```diff
@@ -202,28 +153,43 @@ def _set_row_partitions_to_actors(
...
+            if len(empty_actor_ips) / len(actors_ips) < 0.1:
+                import warnings
+
+                for ip in empty_actor_ips:
+                    actors.pop(ip)
+                    row_partitions_by_actors.pop(ip)
+                    warnings.warn(
+                        f"Node {ip} isn't used due to it doesn't contain any data."
```
vnlitvinov, line 181:
```suggestion
                        f"Node {ip} isn't used as it doesn't contain any data."
```

vnlitvinov (commented):
Getting there... :)

modin/experimental/xgboost/xgboost_ray.py
prutskov: fixed

modin/experimental/xgboost/xgboost_ray.py
prutskov: fixed

modin/experimental/xgboost/xgboost_ray.py
prutskov: fixed

modin/experimental/xgboost/utils.py
prutskov: Context manager was added instead of `start/stop..` separate functions.

modin/experimental/xgboost/utils.py
```diff
@@ -17,23 +17,23 @@
...
+        self.rabit_tracker = xgb.RabitTracker(hostIP=host_ip, nslave=num_workers)
 
-    # Get tracker Host + IP
-    env.update(rabit_tracker.slave_envs())
-    rabit_tracker.start(num_workers)
-
-    return rabit_tracker, env
+        # Get tracker Host + IP
+        self.env.update(self.rabit_tracker.slave_envs())
+        self.rabit_tracker.start(num_workers)
```
vnlitvinov, line 30: I think this should go to `__enter__()` method (logically), not that it would really change the execution flow :)

vnlitvinov (approved):
Code is mostly fine by me (one change noted is optional), please fix the CI.

modin/experimental/xgboost/xgboost_ray.py
prutskov: fixed

modin/experimental/xgboost/xgboost_ray.py
prutskov: fixed

modin/experimental/xgboost/utils.py
prutskov: I agree with you, it was fixed.

modin/experimental/xgboost/xgboost_ray.py
prutskov: Time measurements for approaches 'with asterisk/without asterisk' is the next:
```
Time of asterisk = 0.00313
Time of tuple is = 0.00162
```

<details><summary>Script for measurements</summary>

```python
import time
import numpy as np
from numpy.random import randn
import pandas

data = tuple([pandas.DataFrame(randn(10000)) for i in range(0, 100000)])

def func_asterisk_tuple(dat):
    def func(*tup):
        l1 = tup[: len(tup) // 2]
        l2 = tup[len(tup) // 2 :]

    func(*dat)


def func_tuple(dat):
    def func(tup):
        l1 = tup[: len(tup) // 2]
        l2 = tup[len(tup) // 2 :]

    func(dat)


t_asterisk = 0
t_tuple = 0
N = 10
for it in range(0, N):
    s = time.time()
    func_asterisk_tuple(data)
    t_asterisk += time.time() - s

    s = time.time()
    func_tuple(data)
    t_tuple += time.time() - s


print(f"Time of asterisk is {t_asterisk / N}")
print(f"Time of tuple is {t_tuple / N}")

```
</details>

modin/experimental/xgboost/xgboost_ray.py
```diff
@@ -0,0 +1,338 @@
...
+class ModinXGBoostActor:
+    def __init__(self, ip, nthread=cpu_count()):
+        self._evals = []
+        self._dpredict = []
+        self._ip = ip
+        self._nthreads = nthread
+
+        LOGGER.info(f"Actor <{self._ip}>, nthread = {self._nthreads} was initialized.")
+
+    def get_actor_ip(self):
```
YarShev, line 42: This can be removed because no used.

modin/experimental/xgboost/xgboost_ray.py
```diff
@@ -0,0 +1,338 @@
...
+    }
+
+    assert num_nodes == len(
+        actors
+    ), f"Number of nodes {num_nodes} is not equal to number of actors {len(actors)}."
+
+    return actors
+
+
+def _split_data_across_actors(
```
YarShev, line 127: Please, add docstring for this.

modin/experimental/xgboost/xgboost_ray.py
```diff
@@ -0,0 +1,338 @@
...
+    for ip, actor in actors.items():
+        X_parts = X_parts_by_actors[ip][0]
+        if y_parts is None:
+            set_func(actor, *X_parts)
+        else:
+            y_parts = y_parts_by_actors[ip][0]
+            set_func(actor, *(X_parts + y_parts))
+
+
+def _set_row_partitions_to_actors(
```
YarShev, line 151: Please, add docstring for this.

modin/experimental/xgboost/xgboost_ray.py
```diff
@@ -0,0 +1,338 @@
...
+        else:
+            for ip, (_, order_of_indexes) in data_for_aligning.items():
+                row_partitions_by_actors[ip][1].extend(order_of_indexes)
+                for row_idx in order_of_indexes:
+                    row_partitions_by_actors[ip][0].append(row_partitions[row_idx][1])
+
+    return row_partitions_by_actors
+
+
+def _distribute_partitions_evenly(
```
YarShev, line 202: Please, add docstring for this.

YarShev (changes requested):
LGTM! Left some minor comments. Please, also update the description of the PR to comply the changes.

modin/experimental/xgboost/xgboost_ray.py
prutskov: Docstring was added.

modin/experimental/xgboost/xgboost_ray.py
prutskov: Docstring was added.

modin/experimental/xgboost/xgboost_ray.py
prutskov: Docstring was added.

modin/experimental/xgboost/xgboost_ray.py
prutskov: Thanks, it was removed

vnlitvinov (approved):
LGTM given CI is green.

modin/experimental/xgboost/xgboost_ray.py
vnlitvinov: Cool, so "tuple" is approx. two times faster than asterisk.

If the data passed is a tuple of partitions then I think the overhead would be negligible, so we should go with the easiest implementation (I was worried the data would be raw items - this would mean tuples of hundreds of MB or more, here I assume it would be a few MB at most - no one would notice).

YarShev:
@prutskov , thanks for the changes! @vnlitvinov , thanks for your review! 😄

YarShev (approved):

PR #2562, opened 2020-12-18
Switch to Ray from conda-forge

### LINKED ISSUE #2561: Move to Ray from conda-forge in the environment to speed things up
This should speed up the testing a little bit by speeding up environment set up.

vnlitvinov:

## What do these changes do?

- [x] passes `flake8 modin`
- [x] Resolves #2561
- [ ] tests added and passing

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -67,47 +67,13 @@ jobs:
       - run: pip install flake8 flake8-print
       - run: flake8 --enable=T modin
 
-  prepare-cache:
-    runs-on: ${{ matrix.os }}
-    strategy:
-      matrix:
-        os: ["ubuntu-latest",  "windows-latest"]
-        python-version: ["3.6", "3.7", "3.8"]
-    steps:
-      - uses: actions/checkout@v2
-        with:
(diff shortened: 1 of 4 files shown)


amyskov:
Looks good, but what about  `.github/workflows/push.yml` file?

vnlitvinov:
Good point, I forgot about that. Will fix now.

amyskov:
Also docker files for benchmark running contains explicit ray installation (https://github.com/modin-project/modin/blob/master/examples/docker/census-on-omnisci/census-omnisci.dockerfile#L51), can it be removed now?

vnlitvinov:
> Also docker files for benchmark running contains explicit ray installation (https://github.com/modin-project/modin/blob/master/examples/docker/census-on-omnisci/census-omnisci.dockerfile#L51), can it be removed now?

This is a completely separate issue, and it can not be removed right now, because we saw cases when just installing `modin` without asking for `ray` brings us conda-forge modin without Omnisci.

This file would have to be fixed after our modin moves to conda-forge ray, too (cc @anmyachev), and we push a release.

amyskov (approved):

PR #2579, opened 2021-01-05
FEAT-#2091: add distributed dataframe compare

### LINKED ISSUE #2091: `compare` function implementation
So far, we are just defaulting to pandas.

kvu35:
## What do these changes do?
The current implementation of compare defaults to pandas. This PR uses `broadcast_apply_full_axis` to parallelize the operation and add additional optional checks to `pandas.base.validate_other` for inter DF operations. These additional checks are necessary because of the compare api only works on DFs with equal index objects.

- [x] passes `flake8 modin`
- [x] Resolves #2091 
- [x] tests added and passing


anmyachev:
@kvu35 in general, everything is good.

Please make the following changes:
* add a mention in the `docs\supported_apis\dataframe_supported.rst` about new implementation;
* make rebase from master.

kvu35:
@anmyachev Finished. Please review again when you have the time.

anmyachev:
@kvu35 please also add test and implementation note for series.

kvu35:
@anmyachev All done. Please check when you have time.

anmyachev (approved):
LGTM

anmyachev:
thanks @kvu35!

modin/backends/pandas/query_compiler.py
```diff
@@ -3023,3 +3023,12 @@ def cat_codes(self):
         return self.default_to_pandas(lambda df: df[df.columns[0]].cat.codes)
 
     # END Cat operations
+
+    def compare(self, other, **kwargs):
```
YarShev, line 3027: I'm sorry for the late question but still did you check the performance for moving `other` object to pandas and calling `_apply_full_axis`? Because for some cases this works faster.

kvu35: No problem. Can you please specify what you mean by moving other to pandas? If we convert other to pandas right away using other.to_pandas() then we would get a single pandas dataframe object. We want to perform the compare across likewise partitions between self and other.

devin-petersohn: Broadcast join is probably what @YarShev means. There's no need to implement only for `compare`, we need some cost analysis to do this for all of the `BinaryFunction` operations.

devin-petersohn: Probably it belongs in a separate issue.

YarShev: Yes, it is that what I mean. That makes sense to consider in a separate issue then.

PR #2581, opened 2021-01-05
REFACTOR-#2580: Move automatic engine init to after data ingestion

### LINKED ISSUE #2580: Change engine initialization to happen on first operation
The current approach will start the engine on import, however this presents a number of issues.

devin-petersohn:
* Resovles #2580

Instead of automatically starting the engine when Modin is imported,
we start it after the first time the user reads or creates a dataframe.
This is intended to help downstream libraries not need the engine to
check for typing, as well as clear up some transient errors that can
occur with certain engines on large machines.

I have also added a warning message that informs the user how to clear
the message. We will likely need a way to suppress these errors, because
many users will not care about them and potentially want to suppress.
We will probably also want to add a benchmarking page on best practices
for benchmarking because this change can give the impression of a
performance degradation on data ingestion even though nothing is
changing from that perspective.

## What do these changes do?

- [x] passes `flake8 modin`
- [x] Resolves #2580 
- [ ] tests ~added and~ passing

### DIFF
--- a/modin/engines/ray/utils.py
+++ b/modin/engines/ray/utils.py
@@ -158,21 +158,6 @@ def initialize_ray(
                 _memory=object_store_memory,
                 _lru_evict=True,
             )
-
-            global_node = ray.worker._global_node
-            # Check only for head node
-            if global_node.head:
-                import psutil
-                from modin.error_message import ErrorMessage
-
-                ray_session_dir = os.path.dirname(global_node._session_dir)
-                ray_free_space_GB = psutil.disk_usage(ray_session_dir).free // 10 ** 9
-                ErrorMessage.single_warning(
-                    f"Modin Ray engine was started with {ray_free_space_GB} GB free space avaliable, "
-                    "if it is not enough for your application, please set environment variable "
-                    "MODIN_ON_RAY_PLASMA_DIR=/directory/without/space/limiting"
-                )
-
         _move_stdlib_ahead_of_site_packages()
         ray.worker.global_worker.run_function_on_all_workers(
             _move_stdlib_ahead_of_site_packages
--- a/modin/engines/ray/utils.py
+++ b/modin/engines/ray/utils.py
@@ -12,7 +12,6 @@
 # governing permissions and limitations under the License.
 
 import builtins
-import threading
 import os
 import sys
 
@@ -94,7 +93,7 @@ def initialize_ray(
     """
     import ray
 
-    if threading.current_thread().name == "MainThread" or override_is_cluster:
+    if not ray.is_initialized() or override_is_cluster:
         import secrets
 
         cluster = override_is_cluster or IsRayCluster.get()
@@ -111,6 +110,17 @@ def initialize_ray(
                 logging_level=100,
             )
         else:
+            from modin.error_message import ErrorMessage
+
+            # This string is intentionally formatted this way. We want it indented in
+            # the warning message.
+            ErrorMessage.not_initialized(
+                "Ray",
+                """
+    import ray
+    ray.init()
+""",
+            )
             object_store_memory = Memory.get()
             plasma_directory = RayPlasmaDir.get()
             if IsOutOfCore.get():
--- a/modin/pandas/dataframe.py
+++ b/modin/pandas/dataframe.py
@@ -45,11 +45,12 @@ import warnings
 
 from modin.error_message import ErrorMessage
 from modin.utils import _inherit_docstrings, to_pandas, hashable
-from modin.config import IsExperimental
+from modin.config import Engine, IsExperimental
 from .utils import (
     from_pandas,
     from_non_pandas,
 )
+from . import _update_engine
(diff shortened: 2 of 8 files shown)


modin/pandas/dataframe.py
```diff
@@ -87,6 +88,7 @@ def __init__(
         query_compiler: query_compiler
             A query compiler object to manage distributed computation.
         """
+        Engine.subscribe(_update_engine)
```
williamma12, line 91: By updating the engine at every call to dataframe, we have undefined (or at least undocumented) behavior when a user decides to change the engine on-the-fly. For example,
```
import modin.pandas as pd

# Build dataframe with ray engine.
import ray
ray.init()

df = pd.DataFrame([[1,2,3],[4,5,6]])

# Change to dask engine.
os.environ["MODIN_ENGINE"] = "dask"

df.sum()
```

Right now, it runs the `df.sum()` and returns a result. I am assuming that it is running with the ray engine. Do we want this to be the intended behavior? If so, we should move this line to only be triggered when there is `query_compiler` is None (i.e., when we are creating the dataframe). The same should be done for series.

modin/engines/ray/utils.py
```diff
@@ -111,6 +110,17 @@ def initialize_ray(
                 logging_level=100,
             )
         else:
+            from modin.error_message import ErrorMessage
+
+            # This string is intentionally formatted this way. We want it indented in
+            # the warning message.
+            ErrorMessage.not_initialized(
+                "Ray",
+                """
```
williamma12, line 119: Very minor, but would this look better if the quotes weren't indented?

williamma12 (commented):
Looks good but I have some concerns about changing engines on-the-fly and the possibility of using multiple engines (possibly one person using multiple different clusters for whatever reason).

modin/pandas/dataframe.py
williamma12: Just noticed this, but doesn't all of the IO functions eventually hit this line as well?

devin-petersohn: The execution of each dataframe is tied to it when it is created. All child dataframes inherit the execution context, so the behavior is defined but undocumented. I actually don't want users to do this, because you won't be able to join the dataframes.

devin-petersohn: Which IO functions are not covered? They all should be.

modin/engines/ray/utils.py
devin-petersohn: Black did this. The first quotes are indented but after that the string is interpreted literally, so the last quotes don't have an indent otherwise the indent would be reflected in the message.

modin/pandas/dataframe.py
williamma12: > The execution of each dataframe is tied to it when it is created. All child dataframes inherit the execution context, so the behavior is defined but undocumented. I actually don't want users to do this, because you won't be able to join the dataframes.
Ah ok that makes sense. I mistakenly believed that the `__init__` would be called for every function.

> Which IO functions are not covered? They all should be.
I believe they all are. I wanted to find a way to move the `Engine.subscribe(_update_engine)` to one place as there may be other things we want to do when initializing new dataframes. However, we can push that off until later when we know we have something else that needs to be done.

modin/engines/ray/utils.py
williamma12: Ah ok i guess there is no choice in the matter then.

williamma12 (approved):

modin/pandas/dataframe.py
aregm: I don't think changing the execution engine on the fly is a good idea. I would prefer to prescribe and document the engine set at the beginning of the session. I cannot think of a good example when you need to change the engine in the middle of the processing.

aregm (approved):

modin/pandas/dataframe.py
devin-petersohn: The engine is subscribed to exactly once. It is the same as importing modin multiple times in the current iteration. Even if the flag is changed, the engine will not be updated for existing or new dataframes, it is only set the first time the user does any operation.

This example shows that changing the `MODIN_ENGINE` does not change anything while computing, only when the first engine is subscribed to does any update internally happen. After that there is no update

```python
import modin.pandas as pd
import os

os.environ["MODIN_ENGINE"] = "dask"
df = pd.DataFrame([1,2,3])
df = pd.DataFrame([3,4,5])  # still dask
```

We document that you should not change the `"MODIN_ENGINE"` in the middle of computation, but it is still safe to do so in that we will just ignore it.

YarShev: > Which IO functions are not covered? They all should be.

Why `read_clipboard` and `read_feather` are not covered?

devin-petersohn: @YarShev they should be. I am adding them

williamma12 (approved):

PR #2600, opened 2021-01-13
FIX-#2596: Update pandas version to 1.2.1

### LINKED ISSUE #2596: Update pandas version to 1.2.1

YarShev:

## What do these changes do?

- [x] passes `flake8 modin`
- [x] Resolves #2596  
- [x] tests passing


.github/workflows/ci.yml
```diff
@@ -106,7 +106,7 @@ jobs:
         with:
           activate-environment: modin
           environment-file: environment.yml
-          python-version: 3.6
+          python-version: 3.8
```
devin-petersohn, line 109: Please change tests in a different PR

.github/workflows/ci.yml
```diff
@@ -283,7 +283,7 @@ jobs:
         run: pytest modin/experimental/engines/omnisci_on_ray/test/test_dataframe.py
       - shell: bash -l {0}
         run: bash <(curl -s https://codecov.io/bash)
-  
+
```
devin-petersohn, line 286: Please remove extra newline

modin/pandas/test/test_api.py
```diff
@@ -154,7 +154,17 @@ def test_top_level_api_equality():
...
                 pass
 
-    assert not len(difference), "Extra params found in API: {}".format(difference)
+    try:
+        assert not len(difference), "Extra params found in API: {}".format(difference)
+    except AssertionError:
+        if (
+            len(difference) == 1
+            and difference[0][0] == "read_table"
+            and list(difference[0][1].keys())[0] == "storage_options"
```
devin-petersohn, line 163: Why is this needed?

.github/workflows/ci.yml
YarShev: Because now pandas 1.2.0 supports Python 3.7.1 and higher there was dependencies conflict in `test-headers` job. I changed python version for testing purposes to see that helps. Okay, I will change that in a different PR.

.github/workflows/ci.yml
YarShev: Okay, will be removed.

modin/pandas/test/test_api.py
YarShev: pandas added `storage_options` parameter to `read_csv`, but not to `read_table`. Because we are implementing `read_csv` and `read_table` using one function `_make_parser_func` there was inconsistency in signature of the functions (test-api job failed). I asked a question related to adding the parameter to `read_table` in pandas repo and currently waiting for a response.

.github/workflows/ci.yml
devin-petersohn: I see, yes it would be good to change the tests separately to have more fine-grained commit docs.

modin/pandas/test/test_api.py
YarShev: pandas guys said it looks like an omission of one of their commits. They took the question on consideration.

.github/workflows/ci.yml
YarShev: Actually, we have everywhere blank line between jobs. I think, here blank should be added too.

modin/pandas/test/test_api.py
devin-petersohn: Can you link the pandas issue?

YarShev: Yes, sure. [39167](https://github.com/pandas-dev/pandas/issues/39167)

YarShev:
`modin/pandas/test/data/excel_sheetname_title.xlsx` was automatically changed by excel because wasn't able to open it. See pandas issue [39250](https://github.com/pandas-dev/pandas/issues/39250) for more info.

devin-petersohn:
@YarShev That sheet should not have been changed in this PR. It is intentionally that way because of an issue with older excel files.

YarShev:
> @YarShev That sheet should not have been changed in this PR. It is intentionally that way because of an issue with older excel files.

@devin-petersohn , should we leave it as it was?

devin-petersohn:
>@devin-petersohn , should we leave it as it was?

Yes, it should be reverted. We do not want excel to try to repair it because it came from an older version of excel.

YarShev:
Okay, I am reverting it.

YarShev:
rabased on master

devin-petersohn:
@YarShev What else is needed for this?

YarShev:
@devin-petersohn , Currently I am fixing tests for Series and DataFrame, and @dchigarev is fixing tests for groupby. Can you take these tests `modin/pandas/test/test_rolling.py`?

dchigarev:
@YarShev `.rolling` failings may be related with groupby, so we should fix groupby first. I'll look at `test_rolling.py` after I fix groupby tests

modin/pandas/test/dataframe/test_join_sort.py
```diff
@@ -316,32 +316,38 @@ def test_merge(test_data, test_data2):
 )
 @pytest.mark.parametrize("na_position", ["first", "last"], ids=["first", "last"])
 def test_sort_index(axis, ascending, na_position):
-    data = test_data["float_nan_data"]
-    modin_df, pandas_df = pd.DataFrame(data), pandas.DataFrame(data)
-
-    # Change index value so sorting will actually make a difference
-    if axis == 0:
-        length = len(modin_df.index)
+    # pandas raises the exception when `ascending=None` (pandas issue 39434).
```
devin-petersohn, line 319: For this one, I would rather just raise the error that pandas does temporarily until the issue is fixed. Changing the test is probably not what we want.

modin/pandas/test/test_api.py
devin-petersohn: Now that we have them split in the API, we can just add the parameter to `read_csv`.

modin/pandas/test/test_series.py
```diff
@@ -2914,31 +2908,38 @@ def test_slice_shift(data, index, periods):
...
-    modin_series, pandas_series = create_test_series(data)
-    df_equals(
-        modin_series.sort_index(
-            ascending=ascending, sort_remaining=sort_remaining, na_position=na_position
-        ),
-        pandas_series.sort_index(
-            ascending=ascending, sort_remaining=sort_remaining, na_position=na_position
-        ),
-    )
+    # pandas raises the exception when `ascending=None` (pandas issue 39434).
```
devin-petersohn, line 2911: See about about `ascending=None`

devin-petersohn (changes requested):
Lots of questions so far.

modin/data_management/functions/mapreducefunction.py
```diff
@@ -23,8 +23,8 @@ def caller(query_compiler, *args, **kwargs):
             return query_compiler.__constructor__(
                 query_compiler._modin_frame._map_reduce(
                     cls.validate_axis(axis),
-                    lambda x: map_function(x, *args, **kwargs),
-                    lambda y: reduce_function(y, *args, **kwargs),
+                    lambda x, *args, **kwargs: map_function(x, *args, **kwargs),
```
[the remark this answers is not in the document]
YarShev: Yes, this was left by mistake.

modin/engines/base/frame/data.py
```diff
@@ -1628,9 +1633,9 @@ def broadcast_apply_full_axis(
             None,
             dtypes,
         )
-        if new_index is not None:
+        if new_axes[0] is not None:
```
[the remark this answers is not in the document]
YarShev: Since we compute new variable `new_axis` above based on `new_index` and `new_columns` I would prefer using `new_axis`.

modin/pandas/dataframe.py
```diff
@@ -2089,7 +2099,6 @@ def __delitem__(self, key):
     __mod__ = mod
     __imod__ = mod  # pragma: no cover
     __rmod__ = rmod
-    __div__ = div
```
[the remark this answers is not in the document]
YarShev: pandas DataFrame doesn't support this attribute in new versions.

modin/pandas/test/test_api.py
YarShev: Yes, I agree.

modin/pandas/test/dataframe/test_join_sort.py
YarShev: Okay, fixed.

modin/pandas/test/test_series.py
YarShev: Okay, fixed.

modin/data_management/functions/binary_function.py
```diff
@@ -24,6 +24,7 @@ def caller(query_compiler, other, *args, **kwargs):
             axis = kwargs.get("axis", 0)
             broadcast = kwargs.pop("broadcast", False)
             join_type = call_kwds.get("join_type", "outer")
+            squeeze_self = kwargs.pop("squeeze_self", None)
```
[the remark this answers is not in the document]
YarShev: Until now we could pass `df` and `axis` parameter to binary functions (`__add__`, `__and__`, etc.) and it worked successfully. Now pandas doesn't pass `axis` parameter to these functions and throws the exception:
```python
import pandas
import numpy as np
df = pandas.DataFrame([1, 2])
a = np.array([1, 2])
df.__and__(a)
ValueError: Unable to coerce to Series, length must be 1: given 2
```

That's why we need `squeeze_self`.

modin/engines/base/frame/data.py
```diff
@@ -1187,6 +1187,7 @@ def _apply_full_axis(
         new_index=None,
         new_columns=None,
         dtypes=None,
+        build_mapreduce_func=True,
```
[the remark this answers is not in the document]
YarShev: This is needed to handle a result for binary operations in different way (see [here](https://github.com/modin-project/modin/pull/2600/files#diff-342ff10433091e4da2ecea2ae0656fbb1a6fe9cd2f4f9ffdfb1d5a994d0d24ccR83)).

modin/data_management/functions/binary_function.py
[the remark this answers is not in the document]
devin-petersohn: Can we just squeeze in the API layer? Convert the Series to a dataframe then do the `add` then convert back with `squeeze`?

modin/engines/base/frame/data.py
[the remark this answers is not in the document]
devin-petersohn: If this is needed because of the `squeeze`, then please see above about how we can handle the squeeze differently.

modin/engines/base/frame/data.py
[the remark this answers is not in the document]
devin-petersohn: This change will impact performance. `new_index` indicates that there will be a mismatch, so external metadata should be propagated into internal. In your changes, the external is always broadcast to the partitions because `new_axes[0]` will always be `not None`. I think it should be reverted and if we want to change this discuss in a separate issue or new PR.

devin-petersohn (changes requested):

modin/data_management/functions/binary_function.py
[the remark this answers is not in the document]
YarShev: We can't do it in such way. We can perform `__and__` between one column `pandas.DataFrame` and `np.array` correctly only by squeezing `pandas.DataFrame` to `pandas.Series` in a backend.

modin/engines/base/frame/data.py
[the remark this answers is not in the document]
YarShev: The answer is in the same place.

modin/engines/base/frame/data.py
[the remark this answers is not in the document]
YarShev: Okay, I see. This should be reverted.

modin/data_management/functions/binary_function.py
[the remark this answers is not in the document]
devin-petersohn: The numpy array can be converted to a `Series`, then the behavior should be correct.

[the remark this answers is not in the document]
dchigarev: The problem that causes us to pass `squeeze_self` and `squeeze_other`/`broadcast` to the backend, is that on the query compiler level we can't determine whether this is a Series or a single-column frame. On a backend, we have common binary functions that deals with DataFrames and Series at the same time, but its behavior depends on an object type to which we apply a binary function (DataFrame or Series)


```python
>>> df1
   col1
0     0
1     1
2     2
3     3
>>> arr
array([10, 20, 30, 40])
>>> df1 + arr  # can't do '__add__' with DataFrame and 1D numpy array
  File "C:\Program Files (x86)\Microsoft Visual Studio\Shared\Python37_64\lib\site-packages\pandas\core\ops\__init__.py", line 651, in f
    self, other = _align_method_FRAME(self, other, axis, flex=True, level=level)
  File "C:\Program Files (x86)\Microsoft Visual Studio\Shared\Python37_64\lib\site-packages\pandas\core\ops\__init__.py", line 474, in _align_method_FRAME
    right = to_series(right)
  File "C:\Program Files (x86)\Microsoft Visual Studio\Shared\Python37_64\lib\site-packages\pandas\core\ops\__init__.py", line 466, in to_series
    msg.format(req_len=len(left.columns), given_len=len(right))
ValueError: Unable to coerce to Series, length must be 1: given 4
>>> df1.squeeze() + arr
0    10
1    21
2    32
3    43
Name: col1, dtype: int64
```


</details>

So the workaround for this problem is passing parameters to a backend to indicate what we're dealing with (Frame/Series). 

I don't like either that there is too much responsibility for a backend to handle minor differences between high-level abstractions, but currently, I can't see a quick solution for this, I think it is a general issue about API design. 

@devin-petersohn I know you've been working on a design documentation for Modin, anything you can share with us about this case?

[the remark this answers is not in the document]
dchigarev: And answering to your suggestion: previously we were just broadcasting a numpy array to do a binary op, converting it to Series would add a decent overhead
>The numpy array can be converted to a Series, then the behavior should be correct.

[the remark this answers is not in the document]
YarShev: pandas issue [#39519](https://github.com/pandas-dev/pandas/issues/39519).

YarShev:
Rebased on master.

modin/data_management/functions/binary_function.py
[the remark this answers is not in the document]
devin-petersohn: I prefer to disable the operation in the front end than to add this to the backend. Let's just default to pandas in this case and find a fix later. Ideally we will update the pandas version as soon as possible.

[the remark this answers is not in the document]
YarShev: I don't mind. Reverted the changes.

YarShev:
Rebased on master.

modin/pandas/test/test_groupby.py
```diff
@@ -1367,16 +1365,14 @@ def test_agg_exceptions(operation):
     "kwargs",
     [
         {
-            "Max": ("cnt", np.max),
-            "Sum": ("cnt", np.sum),
-            "Num": ("c", pd.Series.nunique),
```
devin-petersohn, line 1372: Why was this removed?

modin/data_management/functions/mapreducefunction.py
```diff
@@ -20,11 +20,20 @@ def call(cls, map_function, reduce_function, **call_kwds):
         def caller(query_compiler, *args, **kwargs):
             preserve_index = call_kwds.pop("preserve_index", True)
             axis = call_kwds.get("axis", kwargs.get("axis"))
+            if kwargs.get("numeric_only", "is_not_exist") == "is_not_exist":
```
devin-petersohn, line 23: Why is this change necessary? `numeric_only` should be handled at a higher level than this.

modin/pandas/test/test_groupby.py
dchigarev: test frame doesn't have any column named `"cnt"`, so previously this test was useless. It was just testing that modin and pandas will raise a `KeyError`, now this test works properly.

I think that I made these changes when I was debugging the new behavior of `groupby`, so it seems this is not correlated with updating the pandas version. Should we fix it in a different PR, or it's acceptable to keep it here?

devin-petersohn: We may want to revert and discuss in a different PR. It is useful to check for the `KeyError`.

devin-petersohn: What I mean is that we may want both, with the `KeyError` and with these changes. A new issue/PR is probably best for a discussion like this.

modin/data_management/functions/mapreducefunction.py
YarShev:
```python
import pandas
import numpy as np
# plain df
df = pandas.DataFrame(
    {
        "int_col": [-5, 2, 7, 16],
        "float_col": [np.NaN, -9.4, 10.1, np.NaN],
        "str_col": ["a", np.NaN, "c", "d"],
        "bool_col": [False, True, True, False],
    }
)
df
   int_col  float_col  bool_col
0       -5        NaN     False
1        2       -9.4      True
2        7       10.1      True
3       16        NaN     False
# it is that what is happening during our `map` phase
# pandas 1.1.5 behavior:
df.max(axis=0, numeric_only=True)
int_col      16.0
float_col    10.1
bool_col      1.0
dtype: float64
# pandas 1.2.1 behavior:
df.max(axis=0, numeric_only=True)
int_col        16
float_col    10.1
bool_col     True
dtype: object

# the difference is in last two lines
bool_col      1.0
dtype: float64
# and
bool_col     True
dtype: object
```
Since we perform `max` and `min` through 2 stages (`map` and `reduce`) we need to pass `not numeric_only` to `reduce` phase.
I am not sure that we can handle this at a higher level. Some changes are required here in any case.

YarShev: pandas issue [39607](https://github.com/pandas-dev/pandas/issues/39607).

modin/pandas/test/test_groupby.py
dchigarev: I've reverted changes at `test_groupby` and also disabled this test. It revealed, that because of another bug, the dictionary for aggregate is empty in this test, which behavior was changed in 1.2.0, it's unclear was it intended or not, so I've asked a [question](https://github.com/pandas-dev/pandas/issues/39609) at pandas repo.

I've also created an issue about our groupby tests to add reverted changes later (#2698)

modin/data_management/functions/mapreducefunction.py
YarShev: One more of the solutions might be passing parameter `_numeric_only` to `map_reduce_function` and doing similar logic that is now.

devin-petersohn: If we need to convert types, thatThere can be done with `as_type`. There shouldn't be any addition to this file for changing the type implicitly in this way. It will also cause `dtypes` to be an issue, so it's best handled at the higher layers.

modin/pandas/test/test_rolling.py
```diff
@@ -78,6 +77,11 @@ def test_dataframe(data, window, min_periods, win_type):
             pandas_rolled.aggregate([np.sum, np.mean]),
         )
         df_equals(modin_rolled.quantile(0.1), pandas_rolled.quantile(0.1))
+        # `Rolling.counts` has a buggy side-effect on other rolling functions described in:
+        # https://github.com/pandas-dev/pandas/issues/39554
+        # So the testing of `.count` should always be the last until this bug
+        # will be fixed in pandas, to avoid this side-effect
+        df_equals(modin_rolled.count(), pandas_rolled.count())
```
dchigarev, line 84: This bug is already fixed (pandas-dev/pandas#39554), so we need to revert this workaround when updating pandas to 1.2.2

YarShev: Definitely.

modin/data_management/functions/mapreducefunction.py
YarShev: `astype` can't be used at the API layer because we need to handle the case at the time point between `map` and `reduce` phases. Even if we could use `astype` it could add some overhead. So, moved the handling to the query compiler.

devin-petersohn: If we convert the types beforehand, it will be correct in the output. There should be no need to handle anything in between map and reduce phases if the types are handled first.

YarShev: I am not sure when we can convert the types beforehand. If we convert them at the API level as `data.astype(object)`, then we have empty dataframe after map phase that is incorrect.

YarShev: And if we set `numeric_only=False/None`, then we have other failures.

devin-petersohn: This will change only bool to float: `self.astype({c: "float" for c in self.columns if self.dtypes[c] == np.bool})`

YarShev: Okay, I see, but then we will have a behavior that match pandas 1.1.5 behavior mentioned above. Also, some overhead will be added because of use of `astype`. Are we okay with this?

YarShev: Okay, I see, but then we will have a behavior that matches pandas 1.1.5 behavior mentioned above. Also, some overhead will be added because of use of astype. Are we okay with this?

devin-petersohn: What is the current Modin behavior in case of bool for `numeric_only=True` if the extra handling of `numeric_only` is removed from this?

YarShev: If we perform `self.astype({c: "float" for c in self.columns if self.dtypes[c] == np.bool})` and leave `max = MapReduceFunction.register(pandas.DataFrame.max)`, we will have the behavior matching pandas 1.1.5 for both `numeric_only=True` and `numeric_only=False`.

devin-petersohn: We need behavior to match pandas 1.2, what if we do not do the `astype`? What behavior in Modin was changed here? Is the dtype of the result different depending on `numeric_only`?

YarShev: If we do not do the `astype` and leave the changes as it is in the PR, we match pandas 1.2.1 behavior. But if we do the `astype`, we match 1.1.5 behavior regardless the processing of `numeric_only` (True/False/None). So, I would leave the changes as they are now.

modin/data_management/functions/groupby_function.py
```diff
@@ -78,7 +78,7 @@ def reduce(
         **kwargs,
     ):
         by_part = list(df.index.names)
-        if drop and len(df.columns.intersection(by_part)) > 0:
+        if drop and len(df.columns.intersection(pandas.Index(by_part))) > 0:
```
devin-petersohn, line 81: This change should not be needed, I have tested it locally.

modin/pandas/base.py
```diff
@@ -2094,11 +2105,33 @@ def set_axis(self, labels, axis=0, inplace=False):
             obj.set_axis(labels, axis=axis, inplace=True)
             return obj
 
-    def shift(self, periods=1, freq=None, axis=0, fill_value=None):
+    def set_flags(
+        self, *, copy: bool = False, allows_duplicate_labels: Optional[bool] = None
+    ):
+        return self._default_to_pandas(
```
devin-petersohn, line 2111: Can we add a test for this?

modin/pandas/base.py
```diff
@@ -2094,11 +2105,33 @@ def set_axis(self, labels, axis=0, inplace=False):
...
+            pandas.DataFrame.set_flags,
+            copy=copy,
+            allows_duplicate_labels=allows_duplicate_labels,
+        )
+
+    @property
+    def flags(self):
+        return self._default_to_pandas(pandas.DataFrame.flags)
```
devin-petersohn, line 2119: Can we add a test for this?

modin/pandas/groupby.py
```diff
@@ -904,16 +894,18 @@ def _wrap_aggregation(
...
+    def _apply_agg_function(self, f, overwrite_groupby_kwargs=None, *args, **kwargs):
         """
         Perform aggregation and combine stages based on a given function.
 
         TODO: add types.
 
         Parameters
         ----------
-        f:
+        f: callable,
```
devin-petersohn, line 905:
```suggestion
        f: callable
```

modin/pandas/groupby.py
```diff
@@ -634,7 +622,9 @@ def corr(self, **kwargs):
         return self._default_to_pandas(lambda df: df.corr(**kwargs))
 
     def fillna(self, **kwargs):
-        result = self._apply_agg_function(lambda df: df.fillna(**kwargs))
+        result = self._apply_agg_function(
+            lambda df: df.fillna(**kwargs), overwrite_groupby_kwargs={"as_index": True}
```
devin-petersohn, line 626: Why not just do the `kwargs.update({"as_index": True}) in the function instead of adding a new argument to `_apply_agg_function`?

modin/pandas/groupby.py
```diff
@@ -904,16 +894,18 @@ def _wrap_aggregation(
             return result.squeeze()
         return result
 
-    def _apply_agg_function(self, f, drop=True, *args, **kwargs):
+    def _apply_agg_function(self, f, overwrite_groupby_kwargs=None, *args, **kwargs):
```
devin-petersohn, line 897: I think this can be reverted, see above.

devin-petersohn (changes requested):
Thanks @YarShev, a couple more changes for maintainability.

modin/data_management/functions/mapreducefunction.py
devin-petersohn: I see, I had not seen that the changes to `MapReduceFunction` were reverted, so the current way of handling it is fine.

modin/pandas/groupby.py
dchigarev: Because `kwargs` in this context is an argument for `fillna` function, not for groupby.

If you were speaking about `self._kwargs` then we should also revert this change after getting the result since we don't want to modify GroupBy object. Is it what you've proposed?

devin-petersohn: Yes, that is what I mean, I see we should update the `self._kwargs` afterward to reset it. This looks like a regression in pandas, was this reported there?

modin/pandas/base.py
YarShev: Added to the tests of ops that are default to pandas.

modin/pandas/base.py
YarShev: Added to the tests of ops that are default to pandas.

modin/data_management/functions/groupby_function.py
dchigarev: The behavior for `MultiIndex.intersection` was changed for some reason, so now it's raising a TypeError when trying to find an intersection between multi-index and a list.
```python
import pandas

multi_index = pandas.MultiIndex.from_arrays([["a", "b"], ["c", "d"]])
try:
    result = multi_index.intersection(["some_col"])
except Exception as e:
    result = e
print(f"Intersection with a list: {repr(result)}")

result = multi_index.intersection(pandas.Index(["some_col"]))
print(f"Intersection with an Index: {result}")
```

<details><summary>Output for pandas 1.1.5</summary>

```
Intersection with a list: MultiIndex([], )
Intersection with an Index: MultiIndex([], )
```

</details>

<details><summary>Output for pandas 1.2.1</summary>

```
Intersection with a list: TypeError('other must be a MultiIndex or a list of tuples')
Intersection with an Index: MultiIndex([], )
```

</details>

devin-petersohn: I see, let's leave a comment above this then and report the regression to pandas. Leave the issue number in the comment

modin/pandas/groupby.py
dchigarev: This is not regression, 'fillna` does not consider `as_index` parameter at all, so to bypass our `as_index=False` logic that incorrectly processes `fillna` index, it's better set `as_index=True`.

Why we haven't seen this problem before? At 1.2.0 version pandas fixed a bug, where `dropna=True` parameter doesn't drop `nan` groups.
```python
import pandas
import numpy as np

data = {"a": [np.nan, 2, np.nan, 2], "b": [1, 2, 3, np.nan]}
df = pandas.DataFrame(data)
res1 = df.groupby("a", as_index=False).fillna(method="ffill")
res2 = df.groupby("a", as_index=True).fillna(method="ffill")
assert res1.equals(res2)
print(res1)

# pandas 1.1.5 output:
#      b
# 0  NaN
# 1  2.0
# 2  NaN
# 3  2.0

# pandas 1.2.0+ output
#      b
# 1  2.0
# 3  2.0
```
The latest pandas version processes the` dropna=True` parameter correctly and drops `nan` groups, so the index is not monotonic range anymore (as our logic considers it when `as_index=False`).

dchigarev: The problem with modifying and restoring class attribute is that if an exception occurs somewhere between modifying and restoring the value, then we will get a modified object:
<details><summary>Code example</summary>

```python
class Example:
    def __init__(self):
        self.attr = "initial value"
    
    def method(self):
        previous_value = self.attr
        self.attr = "modified value"
        self._private_method()
        self.attr = previous_value
    
    def _private_method(self):
        raise Exception

obj = Example()
print(obj.attr)
try:
    obj.method()
except:
    pass
print(obj.attr)

# Output:
# initial value
# modified value
```

</details>

modin/data_management/functions/groupby_function.py
dchigarev: done it

modin/pandas/groupby.py
dchigarev: There is another way how we can modify groupby parameter for a single operation - we can create a new groupby object like we do in [`groupby.size()`](https://github.com/modin-project/modin/blob/8a50c4a147de94b10bf098b5f540d22c43c76268/modin/pandas/groupby.py#L492). But IMHO this makes code pretty heavy, and also triggers [unnecessary computations](https://github.com/modin-project/modin/blob/8a50c4a147de94b10bf098b5f540d22c43c76268/modin/pandas/groupby.py#L75-L92) in the GroupBy constructor.

devin-petersohn: I think it is probably fine to create a new GroupBy object, the computation should be relatively small (`by` should not be very long).

dchigarev: done

devin-petersohn (approved):
LGTM, thanks everyone involved.

PR #2617, opened 2021-01-15
TEST-#2616: remove DEFAULT_NPARTITIONS values assigning

### LINKED ISSUE #2616: Manual partitions number setting in tests is broken
### System information
- **Modin version**: d6637304a24f986b5abb12e2b525bc3af97d6cac
- **Python version**:  3.8.6
```
import os

import modin.pandas as pd
pd.DEFAULT_NPARTITIONS = 1

data = [x for x in range(10000)]

df = pd.DataFrame(data)
print(len(df._query_compiler._modin_frame._partitions)) # <available number of cores>
```

### Describe the problem

Occurred after 9a6695d3b36cb6c21e835b10d62f26a474583f15 commit.

amyskov:

## What do these changes do?

- [ ] passes `flake8 modin`
- [ ] Resolves #2616 
- [ ] tests passing

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -159,6 +159,8 @@ jobs:
   test-internals:
     needs: [lint-commit, lint-flake8, lint-black, test-api, test-headers]
     runs-on: ubuntu-latest
+    env:
+      MODIN_CPUS: 4
     name: test-internals
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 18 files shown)


devin-petersohn:
I will fix this issue. I think we should set up a config to be able to change it at runtime as well instead of setting once.

amyskov:
> I will fix this issue. I think we should set up a config to be able to change it at runtime as well instead of setting once.

Great, config is better way.

devin-petersohn:
Fix is in #2622

devin-petersohn:
Superceded by #2622

PR #2625, opened 2021-01-16
FEAT-#2624: Allow read_csv,json,fwf to use file handles without falling back to Pandas 

### LINKED ISSUE #2624: `read_csv` significantly slower than pandas when passing in a file handle
Hello! I've done a bunch of research around "modin" and "file handle" and couldn't find anything, so I'm opening this issue.

I'm using the `DM_ALUNO.csv` (about 2.9G in size with 12m rows) in an effort to benchmark `modin` and `pandas` from this URL `http://download.inep.gov.br/microdados/microdados_educacao_superior_2018.zip`.

I'm running on a 13" Macbook Pro 2019 with 16gb RAM and a 2.8 GHz Quad-Core Intel Core i7 CPU. `modin` (with the dask backend) performs faster than Pandas consistently by about 25% when doing:

```
import modin.pandas as pd

pd.read_csv("DM_ALUNO.CSV")
```

on my computer, the above takes about 17s on `modin` and 25s on `pandas`. However, if I pass in a file handle instead:

```
import modin.pandas as pd

with open("DM_ALUNO.CSV") as f:
  pd.read_csv(f)
```

then it performs _much_ worse. It takes about 60s on `modin` and continues to take 25s on `pandas`. Does anybody know why this is happening? I had a browse through the `read_csv` code on `modin` and couldn't fathom why. Does this have something to do with `dask` and it being unable to pickle file handlers?

Note, I also get the following when doing this:

```
>>> with open("data/01_raw/DM_ALUNO.CSV") as f:
...   import modin.pandas as pd; start_time=time.time();pd.read_csv(f);end_time=time.time()
...
UserWarning: Parameters provided defaulting to pandas implementation.
distributed.worker - WARNING - Memory use is high but worker has no data to store to disk.  Perhaps some other process is leaking memory?  Process memory: 1.52 GB -- Worker memory limit: 2.15 GB
distributed.worker - WARNING - Memory use is high but worker has no data to store to disk.  Perhaps some other process is leaking memory?  Process memory: 1.53 GB -- Worker memory limit: 2.15 GB
```

mzjp2:
Resolves #2624 

## What do these changes do?

See some of the context in #2624 -- what this PR does is adds in some code so that when a file handle is passed into `modin.pandas.read_*` - we extract the filepath from it using the `.name` attribute of the buffer, check if it exists, convert it to an absolute path and then let the rest of the `read_*` code move ahead as though that string was passed in initially.

The reason that a file handle has to fall back to Pandas in a single worker is because file handles can't be pickled and sent across processes, since they aren't serialisable. This PR should hopefully do nothing but make the `read_*` method more performant where possible, and continue to fallback to single worker Pandas where not. So hopefully a win-win.

Certainly on my laptop (Macbook Pro 13" 2019 with 16G Ram, quad-core CPU) it turned `modin.pandas.read_csv` for a `2.9G` csv file with `12m` rows from 60s to 17s and got rid of the Dask warnings saying worker memory limit was nearing its limit when using file handles (which are unavoidable with the data framework I'm using). 

This is my first contribution to the repo, so please feel free to let me know if I've done anything wrong!

- [ ] passes `flake8 modin`
- [x] Resolves #? 
- [ ] tests added and passing

### DIFF
--- a/modin/pandas/test/test_io.py
+++ b/modin/pandas/test/test_io.py
@@ -1279,6 +1279,20 @@ class TestCsv:
 
         df_equals(modin_df, pandas_df)
 
+    @pytest.mark.parametrize("read_mode", ["r", "rb"])
+    def test_read_csv_file_handle(self, request, read_mode, make_csv_file):
+        if request.config.getoption("--simulate-cloud").lower() != "off":
+            pytest.skip("Cannot pickle file handles. See comments in PR #2625")
+
+        unique_filename = get_unique_filename()
+        make_csv_file(filename=unique_filename)
+
+        with open(unique_filename, mode=read_mode) as buffer:
+            df_pandas = pandas.read_csv(buffer)
+            buffer.seek(0)
+            df_modin = pd.read_csv(buffer)
+            df_equals(df_modin, df_pandas)
+
 
 class TestTable:
     def test_read_table(self, make_csv_file):
@@ -1466,6 +1480,21 @@ class TestJson:
         eval_to_file(
             modin_obj=modin_df, pandas_obj=pandas_df, fn="to_json", extension="json"
         )
+
+    @pytest.mark.parametrize("read_mode", ["r", "rb"])
+    def test_read_json_file_handle(self, request, read_mode):
+        if request.config.getoption("--simulate-cloud").lower() != "off":
+            pytest.skip("Cannot pickle file handles. See comments in PR #2625")
+        unique_filename = get_unique_filename(extension="json")
+        try:
+            setup_json_file(filename=unique_filename)
+            with open(unique_filename, mode=read_mode) as buf:
+                df_pandas = pandas.read_json(buf)
+                buf.seek(0)
+                df_modin = pd.read_json(buf)
+                df_equals(df_pandas, df_modin)
+        finally:
+            teardown_test_files([unique_filename])
 
 
 class TestExcel:
@@ -1965,6 +1994,19 @@ class TestFwf:
             )
         finally:
             teardown_test_files([unique_filename])
+
+    @pytest.mark.parametrize("read_mode", ["r", "rb"])
+    def test_read_fwf_file_handle(self, request, read_mode):
+        if request.config.getoption("--simulate-cloud").lower() != "off":
+            pytest.skip("Cannot pickle file handles. See comments in PR #2625")
+        unique_filename = get_unique_filename("txt")
+        setup_fwf_file(filename=unique_filename)
+
+        with open(unique_filename, mode=read_mode) as buffer:
+            df_pandas = pandas.read_fwf(buffer)
+            buffer.seek(0)
+            df_modin = pd.read_fwf(buffer)
+            df_equals(df_modin, df_pandas)
 
 
 class TestGbq:
--- a/modin/engines/base/io/text/csv_dispatcher.py
+++ b/modin/engines/base/io/text/csv_dispatcher.py
@@ -22,6 +22,7 @@ import sys
 class CSVDispatcher(TextFileDispatcher):
     @classmethod
     def _read(cls, filepath_or_buffer, **kwargs):
+        filepath_or_buffer = cls.get_path_or_buffer(filepath_or_buffer)
         if isinstance(filepath_or_buffer, str):
             if not cls.file_exists(filepath_or_buffer):
                 return cls.single_worker_read(filepath_or_buffer, **kwargs)
--- a/modin/engines/base/io/text/fwf_dispatcher.py
+++ b/modin/engines/base/io/text/fwf_dispatcher.py
@@ -22,6 +22,7 @@ import sys
 class FWFDispatcher(TextFileDispatcher):
     @classmethod
     def read(cls, filepath_or_buffer, **kwargs):
+        filepath_or_buffer = cls.get_path_or_buffer(filepath_or_buffer)
         if isinstance(filepath_or_buffer, str):
             if not cls.file_exists(filepath_or_buffer):
                 return cls.single_worker_read(filepath_or_buffer, **kwargs)
--- a/modin/engines/base/io/text/json_dispatcher.py
+++ b/modin/engines/base/io/text/json_dispatcher.py
@@ -22,6 +22,7 @@ from csv import QUOTE_NONE
 class JSONDispatcher(TextFileDispatcher):
     @classmethod
     def _read(cls, path_or_buf, **kwargs):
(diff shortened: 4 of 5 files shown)


mzjp2:
I'm not sure why the tests are failing on CI - they pass locally for me?

devin-petersohn:
@mzjp2 The Dask tests appear to be working, but it looks like there are failures for Ray and Python (debug mode). This is what I see (you should also be able to see the logs by clicking on failed tests). Does this help?

```
_____________________ TestCsv.test_read_csv_file_handle[r] _____________________
  File "/modin/modin/pandas/test/test_io.py", line 1289, in test_read_csv_file_handle
    eval_io(fn_name="read_csv", filepath_or_buffer=buffer)
  File "/modin/modin/pandas/test/utils.py", line 751, in eval_io
    call_eval_general()
  File "/modin/modin/pandas/test/utils.py", line 736, in call_eval_general
    eval_general(
  File "/modin/modin/pandas/test/utils.py", line 689, in eval_general
    values = execute_callable(
  File "/modin/modin/pandas/test/utils.py", line 672, in execute_callable
    md_result = fn(modin_df, **md_kwargs)
  File "/modin/modin/pandas/test/utils.py", line 730, in applyier
    result = getattr(module, fn_name)(*args, **kwargs)
  File "/modin/modin/pandas/io.py", line 116, in parser_func
    return _read(**kwargs)
  File "/modin/modin/pandas/io.py", line 135, in _read
    pd_obj = EngineDispatcher.read_csv(**kwargs)
  File "/modin/modin/data_management/factories/dispatcher.py", line 104, in read_csv
    return cls.__engine._read_csv(**kwargs)
  File "/modin/modin/data_management/factories/factories.py", line 87, in _read_csv
  File "/modin/modin/engines/base/io/io.py", line 161, in read_csv
    return cls._read(**kwargs)
  File "/modin/modin/engines/base/io/io.py", line 172, in _read
    pd_obj = pandas.read_csv(**kwargs)
  File "/root/anaconda3/envs/modin/lib/python3.8/site-packages/pandas/io/parsers.py", line 688, in read_csv
    return _read(filepath_or_buffer, kwds)
  File "/root/anaconda3/envs/modin/lib/python3.8/site-packages/pandas/io/parsers.py", line 454, in _read
    parser = TextFileReader(fp_or_buf, **kwds)
  File "/root/anaconda3/envs/modin/lib/python3.8/site-packages/pandas/io/parsers.py", line 948, in __init__
    self._make_engine(self.engine)
  File "/root/anaconda3/envs/modin/lib/python3.8/site-packages/pandas/io/parsers.py", line 1180, in _make_engine
    self._engine = CParserWrapper(self.f, **self.options)
  File "/root/anaconda3/envs/modin/lib/python3.8/site-packages/pandas/io/parsers.py", line 2010, in __init__
    self._reader = parsers.TextReader(src, **kwds)
  File "pandas/_libs/parsers.pyx", line 540, in pandas._libs.parsers.TextReader.__cinit__
pandas.errors.EmptyDataError: No columns to parse from file
```

mzjp2:
> @mzjp2 The Dask tests appear to be working, but it looks like there are failures for Ray and Python (debug mode). This is what I see (you should also be able to see the logs by clicking on failed tests). Does this help?

Thanks for replying! I can see the logs, but can't seem to reproduce them locally when running the tests :/

devin-petersohn:
There must be a race condition with the tests. Nothing obvious pops out at me, I will try to block of some time today to look.

modin/pandas/test/test_io.py
```diff
@@ -1279,6 +1279,15 @@ def wrapped_read_csv(file, method):
...
         df_equals(modin_df, pandas_df)
 
+    @pytest.mark.parametrize("read_mode", ["r", "rb"])
+    def test_read_csv_file_handle(self, read_mode, make_csv_file):
+        unique_filename = get_unique_filename()
+
+        make_csv_file(filename=unique_filename)
+
+        with open(unique_filename, mode=read_mode) as buffer:
+            eval_io(fn_name="read_csv", filepath_or_buffer=buffer)
```
amyskov, line 1289:
```suggestion
            df_pandas = pandas.read_csv(buffer)
            buffer.seek(0)
            df_modin = pd.read_csv(buffer)
            df_equals(df_modin, df_pandas)
```
I think here and in `test_read_json_file_handle` can be used direct pandas and modin results comparison, because after the first usage of file handle by `pandas.read_csv` in `eval_io`, file handle pointer is shifted and the second call of the same file handle in `pd.read_csv` causes `pandas.errors.EmptyDataError: No columns to parse from file` error

mzjp2: Many thanks, that’s addressed the TeamCity tests. I bizzarely have the cloud test failing on one parametrised case of one test now... 🤔

amyskov: You are welcome! looks like `test cloud` test fails because file handle can't be serialized in the cloud mode and it should be converted to the path object as it was done in the `csv_dispatcher.py`. @anmyachev can you take a look and suggest which place this conversion should be done?

anmyachev: We can do the same replacement here:
https://github.com/modin-project/modin/blob/5f03eb8b1e887f9f3b6990fb023576070d0e6cee/modin/data_management/factories/factories.py#L285

To do this, just check `a[0]`. 

Several suggestions:
 * It would be nice to make a warning that a filename will be used instead of a buffer;
 * try to move the actions of replacing the buffer to a filename into a separate function (remove duplicated code).

mzjp2: Thanks very much guys, super helpful! I actually ended up leaving the cloud stuff for now and have added a skip to the pytest in when running with `--simulate-cloud=normal` because I think the issue there of handling non-picklable objects is a wider one that is coincidentally picked up by my tests, but perhaps not in scope for this PR but also because I couldn't immediately see a nice way to re-use the code I've defined in `text_file_dispatcher.py` in that factory file.

@anmyachev I've taken your feedback onboard and removed duplicate code by moving the functionality into the base class and then using `cls.get_path_from_buffer(...)` in the `csv,json,fwf` classes. Thanks for this, it does make it nicer and also made it easier to implement the warning that the filename was used instead of the buffer, since I could put it in that one place.

amyskov:
LGTM, @anmyachev @devin-petersohn do you have any comments?

modin/pandas/test/test_io.py
anmyachev: @mzjp2 skipping cloud tests is ok for now. Good job!

anmyachev (approved):
Thanks @mzjp2! LGTM

@amyskov merging is on you.

PR #2626, opened 2021-01-16
experimental ray implementation of read_pickle, to_pickle functions

anmyachev:
## What do these changes do?

- [ ] passes `flake8 modin`
- [ ] Resolves #? 
- [ ] tests added and passing


modin/engines/base/frame/data.py
```diff
@@ -1310,6 +1310,7 @@ def _apply_full_axis(
         new_index=None,
         new_columns=None,
         dtypes=None,
+        enumerate_partitions=False,
```
devin-petersohn, line 1313: I am generally against adding more arguments to the internal APIs. Is there no other way?

anmyachev: I'll look for another way.

anmyachev:
Current plan:
 * write tests;
 * improve error handling;
 * remove duplication;
 * add performance numbers;
 * replace `_apply_full_axis` -> `broadsact_apply_full_axes`;
 * in order for the api to cease to be experimental, it is necessary to add changes to the functions of the pandas (for compatibility; for example, processing a list of files);

devin-petersohn (changes requested):
Can we add these tests to TeamCity? That's where the code coverage gets calculated I believe.

Is it possible to add a `PickleFileDispatcher` and move the logic here into that similar to other file dispatchers? 

Why does this need to be in experimental for regular pickle files? Can we allow regular pickle files to be done with non-experimental API?

anmyachev:
> Can we add these tests to TeamCity? That's where the code coverage gets calculated I believe.
> 
> Is it possible to add a `PickleFileDispatcher` and move the logic here into that similar to other file dispatchers?
> 
> Why does this need to be in experimental for regular pickle files? Can we allow regular pickle files to be done with non-experimental API?

Appropriate pickle parser and dispatcher were be created for read_pickle function.

For regular files, the pandas implementation is used, since for the case when one file is transferred, the experimental implementation does not give any speedup.

anmyachev:
> Can we add these tests to TeamCity? That's where the code coverage gets calculated I believe.
> 

The question of the correctness of the coverage is now being decided by @vnlitvinov (if I remember correctly), so let's not run experimental tests on TeamCity for now.

Also, I think we can leave the implementation of `to_picke` as is.

@devin-petersohn PR is ready for review.

modin/engines/base/io/text/text_file_dispatcher.py
```diff
@@ -38,22 +37,6 @@ def get_path_or_buffer(cls, filepath_or_buffer):
                 return cls.get_path(buffer_filepath)
         return filepath_or_buffer
 
-    @classmethod
```
devin-petersohn, line 41: I prefer these refactoring changes to go in a new PR.

modin/experimental/engines/pandas_on_ray/io_exp.py
```diff
@@ -172,6 +181,43 @@ def read_sql(
...
         )
 
+    @classmethod
+    def to_pickle(cls, qc, **kwargs):
+        """
+        In experimental mode, all partitions are written to their own separate file.
+
+        The filenames is determined as follows:
+        - if `*` in the filename then it will be replaced by the increasing sequence 0, 1, 2, …
+        - if ``*` is not the filename, then by default it is assumed that it is between
```
devin-petersohn, line 191:
```suggestion
        - if `*` is not the filename, then by default it is assumed that it is between
```

modin/experimental/engines/pandas_on_ray/io_exp.py
```diff
@@ -172,6 +181,43 @@ def read_sql(
             cls.frame_cls(np.array(partition_ids), new_index, cols_names)
         )
 
+    @classmethod
+    def to_pickle(cls, qc, **kwargs):
+        """
+        In experimental mode, all partitions are written to their own separate file.
```
devin-petersohn, line 187: Is this useful? How would someone reuse this group of pickle files somewhere else?

devin-petersohn (commented):
I understand the need to read a list of pickle files, but why are we writing a list of pickle files for `to_pickle`? Was this requested specifically?

anmyachev:
> I understand the need to read a list of pickle files, but why are we writing a list of pickle files for `to_pickle`? Was this requested specifically?

In order to take advantage of the read function, the write function has been modified accordingly. So far this is only useful for modin users, requires additional work outside.

devin-petersohn:
>In order to take advantage of the read function, the write function has been modified accordingly. So far this is only useful for modin users, requires additional work outside.

In that case, we should rename `to_pickle` to reflect this, since users might uknowingly think it is the same as pandas. `to_pickle_partitions` or something like that

anmyachev:
> > In order to take advantage of the read function, the write function has been modified accordingly. So far this is only useful for modin users, requires additional work outside.
> 
> In that case, we should rename `to_pickle` to reflect this, since users might uknowingly think it is the same as pandas. `to_pickle_partitions` or something like that

I slightly changed the behavior of `to_pickle`, now if there is no `*` symbol in the filename, then the default implementation will be used (dataframe will be pickled in a one file)

anmyachev:

anmyachev:
@devin-petersohn friendly reminder :)

devin-petersohn (commented):
Here is what I understand from this PR, please correct me if I am wrong:

* Parallel reading of multiple pickle files with wildcard in path
* Parallel writing of multiple partitions back into multiple files
* Single file reading defaults to pandas
* Single file writing defaults to pandas

Is this correct?

anmyachev:
> Here is what I understand from this PR, please correct me if I am wrong:
> 
> * Parallel reading of multiple pickle files with wildcard in path
> * Parallel writing of multiple partitions back into multiple files
> * Single file reading defaults to pandas
> * Single file writing defaults to pandas
> 
> Is this correct?

* Parallel reading if list of filename is passed.
* Parallel writing if filename contains wildcard.

The rest is correct

devin-petersohn:
If that is the case, we are deviating from pandas very heavily and should separate the behaviors into different top level APIs. Similar to how `read_csv_glob` was implemented.

anmyachev:

devin-petersohn (commented):
Let's mark this for the next release.

anmyachev:
> Let's mark this for the next release.

ok

anmyachev:
@devin-petersohn can we continue working on the PR?

devin-petersohn (commented):
Looks good overall. I think all that is needed is to add docs to this page https://modin.readthedocs.io/en/latest/flow/modin/experimental/pandas.html

devin-petersohn:
@anmyachev Feel free to ping me when this is ready for review.

anmyachev:
@devin-petersohn main changes since the last review:
- `read_pickle_distributed` also accepts a filename with the globstring now;
- added testcase with defaulting to pandas;
- added experimental features to pandas.rst doc;
- added fixes for problems that appeared due to previous changes.

anmyachev:

devin-petersohn (commented):
Overall, looks great @anmyachev. Can you rebase on current master, then I believe it can be merged. Thanks!

anmyachev:
> Overall, looks great @anmyachev. Can you rebase on current master, then I believe it can be merged. Thanks!

Rebased :)

Issue #2342, opened 2020-10-30
Row Partition API


richardliaw:
**Is your feature request related to a problem? Please describe.**
A clear and concise description of what the problem is. What kind of performance improvements would you like to see with this new API?

I'd like to be able to obtain a list of row partitions from a distributed Modin dataframe.

This would allow me to easily do distributed data-parallel training by feeding in these dataframe shards. This would be useful for both XGBoost and Horovod.

cc @tgaddair

devin-petersohn:
Thanks @richardliaw for the request!

We should consider a developer API for cases like this that enable access to Modin internal partitions without needing to dig into them. I have been thinking about this some.

Design can look something like this:

```python
from modin.developer.partitioning import extract_row_partitions, extract_col_partitions
import ray

row_oid_list = extract_row_partitions(df, assert_type=ray.ObjectRef)
col_oid_list =  extract_col_partitions(df, assert_type=ray.ObjectRef)
```

Thoughts @richardliaw?

cc @aregm

fschlimb:
I ran into a similar issue.
In addition to being able to extract the partitions it would be good to know the locality of each partition. Knowing where a partition (currently) resides can be used to minimize data movement when passing partitions to ray actors. Minimal data movement requires pairing actors and partitions which are 'pinned' to the same node.

richardliaw:
@devin-petersohn sounds good!

devin-petersohn:
I have a prototype ready to try. Install via `pip install git+https://github.com/devin-petersohn/modin@feat/2342`

Usage is simple:
```python
from modin.developer import unwrap_row_partitions

unwrap_row_partitions(df)
```

You will get a list of `ray.ObjectRef` each contain a row partition.

fschlimb:
Thanks, works great for me!
What about getting sufficient information to be able to map data to local actors?

fschlimb:
How would this work with modin on dask?

YarShev:
> Thanks, works great for me!
> What about getting sufficient information to be able to map data to local actors?

I think, we

YarShev:
> Thanks, works great for me!
> What about getting sufficient information to be able to map data to local actors?

Maybe it would be good to have some kind of `RowPartitionsActor` and `ColPartitionsActor` that will keep row partitions and column partitions, respectively, assigned to them. Also, these actors should keep some info on node ip address (it is supposed an actor per node).

YarShev:
> How would this work with modin on dask?

[...]

PR #2643, opened 2021-01-25
REFACTOR-#2642: Refactor partition API

### LINKED ISSUE #2642: Refactor partition API
### Describe the problem

For Modin on Ray engine with pandas backend IPs of the remote partitions may not match actual locations if the partitions are lower than 100 kB. Ray saves such objects (<= 100 kB, by default) in in-process store of the calling process. We can't get IPs for such objects with saving good performance. So, we need to add a doc page for this and perform some refactoring.

YarShev:

## What do these changes do?

- [x] passes `flake8 modin`
- [x] Resolves #2642  
- [x] tests passing


modin/config/envvars.py
```diff
@@ -155,6 +155,26 @@ def _get_default(cls):
...
+class EnablePartitionIPs(EnvironmentVariable, type=bool):
+    """
+    Whether to enable use of IPs for remote partitions of not
+    """
+
+    varname = "MODIN_ENABLE_PARTITIONS_IPS"
+
+    @classmethod
+    def _get_default(cls):
+        return False
```
vnlitvinov, line 167:
```suggestion
    default = False
```

`_get_default()` method is needed when default value is not a code-time constant (like CPU core count which differs based on machine code is run), for fully static defaults there is a special field

modin/engines/base/frame/axis_partition.py
```diff
@@ -136,12 +137,22 @@ def unwrap(self, squeeze=False, bind_ip=False):
         """
         if squeeze and len(self.list_of_blocks) == 1:
             if bind_ip:
-                return self.list_of_ips[0], self.list_of_blocks[0]
+                if EnablePartitionIPs.get():
+                    return self.list_of_ips[0], self.list_of_blocks[0]
+                else:
+                    raise ValueError(
+                        "Passed `bind_ip=True` but `MODIN_ENABLE_PARTITIONS_API` env var was not exported."
```
vnlitvinov, line 144: same note about error message, won't re-type it here :)

modin/engines/dask/pandas_on_dask/frame/axis_partition.py
```diff
@@ -27,7 +28,12 @@ def __init__(self, list_of_blocks, bind_ip=False):
             obj.drain_call_queue()
         self.list_of_blocks = [obj.future for obj in list_of_blocks]
         if bind_ip:
-            self.list_of_ips = [obj.ip for obj in list_of_blocks]
+            if EnablePartitionIPs.get():
+                self.list_of_ips = [obj.ip for obj in list_of_blocks]
+            else:
+                raise ValueError(
+                    "Passed `bind_ip=True` but `MODIN_ENABLE_PARTITIONS_API` env var was not exported."
```
vnlitvinov, line 35: same about error message

modin/engines/ray/pandas_on_ray/frame/axis_partition.py
```diff
@@ -13,11 +13,15 @@
...
+from modin.config import EnablePartitionIPs
 from modin.engines.base.frame.axis_partition import PandasFrameAxisPartition
 from .partition import PandasOnRayFramePartition
 
 import ray
-from ray.services import get_node_ip_address
+
+
+if EnablePartitionIPs.get():
+    from ray.services import get_node_ip_address
```
vnlitvinov, line 24:
```suggestion
from ray.services import get_node_ip_address
```
This is not entirely corrent, the config variable could be changed in runtime.
I suggest removing the check here.

modin/engines/ray/pandas_on_ray/frame/axis_partition.py
```diff
@@ -27,7 +31,12 @@ def __init__(self, list_of_blocks, bind_ip=False):
             obj.drain_call_queue()
         self.list_of_blocks = [obj.oid for obj in list_of_blocks]
         if bind_ip:
-            self.list_of_ips = [obj.ip for obj in list_of_blocks]
+            if EnablePartitionIPs.get():
+                self.list_of_ips = [obj.ip for obj in list_of_blocks]
+            else:
+                raise ValueError(
+                    "Passed `bind_ip=True` but `MODIN_ENABLE_PARTITIONS_API` env var was not exported."
```
vnlitvinov, line 38: same note about not using env var name in error message

modin/engines/ray/pandas_on_ray/frame/partition.py
```diff
@@ -13,15 +13,19 @@
...
+from modin.engines.base.frame.partition import BaseFramePartition
 from modin.engines.ray.utils import handle_ray_task_error
 
 import ray
-from ray.services import get_node_ip_address
 from ray.worker import RayTaskError
 
 
+if EnablePartitionIPs.get():
+    from ray.services import get_node_ip_address
```
vnlitvinov, line 26:
```suggestion
from ray.services import get_node_ip_address
```
same note about config variable being changed at runtime

modin/engines/ray/pandas_on_ray/frame/partition.py
```diff
@@ -207,7 +222,7 @@ def get_index_and_columns(df):
     return len(df.index), len(df.columns)
 
 
-@ray.remote(num_returns=4)
+@ray.remote(num_returns=4 if EnablePartitionIPs.get() else 3)
```
vnlitvinov, line 225: Ideally this should be changed so that `deploy_ray_func()` is regenerated each time `EnablePartitionsIPs` is changed.
I suggest rewriting it like this (untested):
```python
def _deploy_ray_func(call_queue, parition):
    # actual code of the func
    return result

deploy_ray_func = None

def _update_deploy(enabled):
    global deploy_ray_func
    deploy_ray_func = ray.remote(num_returns=4 if enabled else 3)(_deploy_ray_func)
EnablePartitionIPs.subscribe(_update_deploy)
```

docs/partition_ips.rst
```diff
@@ -0,0 +1,64 @@
...
+when unwrapping of the remote partitions with their IPs. Several options are provided to handle the case in
+``How to handle objects that are lower 100 kB`` section. Wherein, there is no such issue for Modin on ``Dask`` engine
+with ``pandas`` backend because ``Dask`` saves any objects in the worker process that processes a function.
+Please let us know what you think!
+
+Install Modin Partition IPs
+----------------------------
+
+Modin now comes with all the dependencies for partitions IPs functionality by default! See
+the `installation page`_ for more information on installing Modin.
```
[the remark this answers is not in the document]
YarShev: for cross-reference

modin/engines/base/frame/axis_partition.py
YarShev: fixed

modin/engines/dask/pandas_on_dask/frame/axis_partition.py
YarShev: fixed

modin/engines/ray/pandas_on_ray/frame/axis_partition.py
YarShev: fixed

modin/engines/ray/pandas_on_ray/frame/partition.py
YarShev: I am not sure it can be fixed by following your suggestion. Fixed by another way.

vnlitvinov: I don't see it fixed in current state of PR. Also my proposal should be valid :)

vnlitvinov (approved):

YarShev:
I would merge #2606 firstly, and then this PR. @devin-petersohn , @vnlitvinov , what do you think?

vnlitvinov:
They seem to overlap, so we should merge in most logical order. I personally am fine with any one you as author thinks reasonable :)

YarShev:
So, let's consider #2606 to merge firstly.

gshimansky:
Are you planning to run CI tests in two modes now, one with disabled API and one with enabled? I see that currently blocks of code with API enabled are not tested at all.

YarShev:
Yes, I am going to do it after we merge #2606 .

devin-petersohn:
@YarShev Will you ping me when #2606 is merged?

vnlitvinov:
@devin-petersohn I've merged the PR closing #2606

YarShev:
@devin-petersohn , @vnlitvinov , I will ping you after I make necessary changes for this PR.

YarShev:
@devin-petersohn , @vnlitvinov , just a friendly reminder to review.

docs/partition_ips.rst
```diff
@@ -0,0 +1,62 @@
...
+------------------------------------------
+
+Partition IPs is detected from an environment variable.
+
+.. code-block:: bash
+
+   export MODIN_ENABLE_PARTITION_API=True
+
+How to handle objects that are lower than 100 kB
+-------------------------------------------
```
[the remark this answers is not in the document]
YarShev: Fixed.

.github/workflows/ci.yml
```diff
@@ -405,6 +405,67 @@ jobs:
...
+    needs: [lint-commit, lint-flake8, lint-black, test-api, test-headers]
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: ["3.7", "3.8"]
+        engine: ["ray", "dask"]
+    env:
+      MODIN_ENGINE: ${{matrix.engine}}
+      MODIN_MEMORY: 1000000000
+      MODIN_ENABLE_PARTITION_IPS: "True"
```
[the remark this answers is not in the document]
YarShev: Applied your suggestion. But now we have dummy jobs ([1st](https://github.com/modin-project/modin/pull/2643/checks?check_run_id=1829564447), [2nd](https://github.com/modin-project/modin/pull/2643/checks?check_run_id=1829564582)) that do nothing. @vnlitvinov , @devin-petersohn , any thoughs?

.github/workflows/push.yml
```diff
@@ -191,6 +191,66 @@ jobs:
...
+  test-all-with-partition-ips-enabled:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: ["3.7", "3.8"]
+        engine: ["ray", "dask"]
+    env:
+      MODIN_ENGINE: ${{matrix.engine}}
+      MODIN_MEMORY: 1000000000
+      MODIN_ENABLE_PARTITION_IPS: "True"
```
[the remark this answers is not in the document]
YarShev: See the comment in discussion for `co.yml`.

modin/engines/dask/pandas_on_dask/frame/partition.py
```diff
@@ -26,7 +27,10 @@ def apply_list_of_funcs(funcs, df):
         if isinstance(func, bytes):
             func = pkl.loads(func)
         df = func(df, **kwargs)
-    return df, get_ip()
+    if EnablePartitionIPs.get():
```
devin-petersohn, line 30: What is the use case for getting the IP of the partition each time? I prefer not adding this to the existing `deploy_func` rather add a new function specific to this. Having multiple return types from this function can only invite more bugs.

devin-petersohn (changes requested):
I'm not sure about the design of mixing getting the IP and the result of the data.

I think I prefer it always being a part of the metadata, but hidden from the user or developer unless they want it.

modin/engines/dask/pandas_on_dask/frame/partition.py
YarShev: The use case is that an user can call `unwrap_partitions(df, axis=None, bind_ip=True) at any time point during his workflow and IPs of the partitions have to be valid.

YarShev: I am not sure on what bugs you are saying about but we can add an additional function to return multiple values.

YarShev:
> I'm not sure about the design of mixing getting the IP and the result of the data.
> 
> I think I prefer it always being a part of the metadata, but hidden from the user or developer unless they want it.

Your preference is that partitions should always have IPs, isn't it? We can add a note into the doc page in such case. The note would be something like [this](https://github.com/modin-project/modin/pull/2643/files#diff-364a73f30337967a29d44fa72a1e0f584073b42bb6fc3f17be924e484edf399dR31).

docs/partition_ips.rst
```diff
@@ -0,0 +1,62 @@
+Partition IPs in Modin (experimental)
+=====================================
+
+If you are working with Modin DataFrame and would like to unwrap its remote partitions
+for your needs  (pass them to another function that will be processed on a concrete node of the cluster,
+for instance), you can use IPs of the remote partitions. In that case you can pass the partitions
+having needed IPs to your function. It can help with minimazing of data movement between nodes. However,
+it is worth noting that for Modin on ``Ray`` engine with ``pandas`` backend IPs of the remote partitions may not match
+actual locations if the partitions are lower than 100 kB. Ray saves such objects (<= 100 kB, by default) in in-process store
```
prutskov, line 9: I think we should insert link on Ray docpage with info about in-process store.

docs/partition_ips.rst
```diff
@@ -0,0 +1,62 @@
...
+with ``pandas`` backend because ``Dask`` saves any objects in the worker process that processes a function.
+Please let us know what you think!
+
+Install Modin Partition IPs
+---------------------------
+
+Modin now comes with all the dependencies for partitions IPs functionality by default! See
+the :doc:`installation page </installation>` for more information on installing Modin.
+
+Starting Modin with Partition IPs enabled
```
prutskov, line 22: Do you plan to add method level documentation for `unwrap_partitions`/`from_partitions` functions in this docpage? I think it will be useful for users who want fast start work with this feature.

modin/distributed/dataframe/pandas/partitions.py
```diff
@@ -47,6 +48,9 @@ def unwrap_partitions(api_layer_object, axis=None, bind_ip=False):
             f"Only API Layer objects may be passed in here, got {type(api_layer_object)} instead."
         )
 
+    if bind_ip and not EnablePartitionIPs.get():
+        raise ValueError("Passed `bind_ip=True` but partition IPs API was not enabled.")
```
prutskov, line 52: As I understand, this exception will raised now when user will work with Modin XGBoost, because `bind_ip=True` already used in XGB. Could you add mentioning of partition IPs enabling in Modin XGBoost docpage?

docs/partition_ips.rst
```diff
@@ -0,0 +1,62 @@
...
+-------------------------------------
+
+Before you run this, please make sure you follow the instructions listed above.
+
+.. code-block:: python
+
+  import modin.pandas as pd
+  from modin.distributed.dataframe.pandas import unwrap_partitions, from_partitions
+  df = pd.read_csv("/path/to/your/file")
+  partitions = unwrap_partitions(df, axis=0, bind_ip=True)
```
prutskov, line 58: Do we want to show the user a small tutorial how to use partitions and IPs which are placed in `partitions `?

prutskov (dismissed):

modin/engines/dask/pandas_on_dask/frame/partition.py
devin-petersohn: We should just always return ip as a part of the metadata in that case instead of optionally. The user can optionally see it, but it will always be there.

If a user changes `EnablePartitionIPs` in the middle of a workflow, I believe it can cause some problems if we conditionally return different numbers of objects. That is the bug I was talking about.

YarShev: Yes, that is true on the bug you were saying about. So we could add a note on `EnablePartitionIPs` changing during a workflow. However, it looks not user-friendly. Let's always return ip as a part of the metadata then.

docs/partition_ips.rst
YarShev: We probably want a user to refer to the entire ray documentation.

docs/partition_ips.rst
YarShev: Yes, it will be useful.

modin/distributed/dataframe/pandas/partitions.py
YarShev: not actual.

docs/partition_ips.rst
YarShev: We don't want to show a tutorial to a user. Small running example in the doc page will be enough.

YarShev:
@devin-petersohn , some more changes are required to add. I will ping you after that.

modin/distributed/dataframe/pandas/partitions.py
```diff
@@ -13,34 +13,34 @@
...
     list
-        A list of Ray.ObjectRef/Dask.Future to partitions of the `api_layer_object`
+        A list of Ray.ObjectRef/Dask.Future to partitions of the ``api_layer_object``
         if Ray/Dask is used as an engine.
 
     Notes
     -----
-    In case bind_ip=True, a list containing tuples of Ray.ObjectRef/Dask.Future to node ip addresses
-    and partitions of the `api_layer_object`, respectively, is returned if Ray/Dask is used as an engine.
+    In case ``get_ip=True``, a list containing tuples of node ip addresses and Ray.ObjectRef/Dask.Future to
```
devin-petersohn, line 42:
```suggestion
    If ``get_ip=True``, a list containing tuples of node ip addresses and Ray.ObjectRef/Dask.Future to
```

docs/developer/pandas/partition_api.rst
```diff
@@ -0,0 +1,70 @@
+Pandas Partition API in Modin
+=============================
+
+If you are working with Modin DataFrame, you can unwrap its remote partitions for your needs
```
devin-petersohn, line 4:
```suggestion
If you are working with a Modin Dataframe, you can unwrap its remote partitions
to get the raw future object compatible with the execution engine (e.g. ``ray.ObjectRef`` for Ray).
```

docs/developer/pandas/partition_api.rst
```diff
@@ -0,0 +1,70 @@
...
+(pass them to another function that will be processed on a concrete node of the cluster,
```
devin-petersohn, line 5:
```suggestion
```

docs/developer/pandas/partition_api.rst
```diff
@@ -0,0 +1,70 @@
...
+for instance) and get IPs of the remote partitions as well. In that case you can pass the partitions
```
devin-petersohn, line 6:
```suggestion
You can use this API to get the IPs of the nodes that hold these objects as well. In that case you can pass the partitions
```

docs/developer/pandas/partition_api.rst
```diff
@@ -0,0 +1,70 @@
...
+having needed IPs to your function. It can help with minimazing of data movement between nodes. However,
+it is worth noting that for Modin on ``Ray`` engine with ``pandas`` backend IPs of the remote partitions may not match
```
devin-petersohn, line 8: Separate discussion on Ray into separate section

devin-petersohn (changes requested):

docs/developer/pandas/partition_api.rst
YarShev: The PR is going to be divided into two. The first is for refactor, the second is for documentation. This comment will be applied in the second PR.

modin/engines/dask/pandas_on_dask/frame/partition_manager.py
```diff
@@ -62,7 +36,6 @@ class DaskFrameManager(BaseFrameManager):
     def get_indices(cls, axis, partitions, index_func):
         """
         This gets the internal indices stored in the partitions.
-
```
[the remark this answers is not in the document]
YarShev: That was done accidentally.

modin/distributed/dataframe/pandas/partitions.py
```diff
@@ -13,34 +13,34 @@
...
-        A list of Ray.ObjectRef/Dask.Future to partitions of the `api_layer_object`
+        A list of Ray.ObjectRef/Dask.Future to partitions of the ``api_layer_object``
         if Ray/Dask is used as an engine.
 
     Notes
     -----
-    In case bind_ip=True, a list containing tuples of Ray.ObjectRef/Dask.Future to node ip addresses
-    and partitions of the `api_layer_object`, respectively, is returned if Ray/Dask is used as an engine.
+    If ``get_ip=True``, a list containing tuples of node ip addresses and Ray.ObjectRef/Dask.Future to
+    partitions of the ``api_layer_object``, respectively, is returned if Ray/Dask is used as an engine.
```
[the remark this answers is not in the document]
YarShev: Fixed.

modin/distributed/dataframe/pandas/partitions.py
```diff
@@ -90,15 +90,16 @@ def from_partitions(partitions, axis):
     Parameters
     ----------
     partitions : list
-        List of Ray.ObjectRef/Dask.Future referencing to partitions in depend of the engine used.
-        Or list containing tuples of Ray.ObjectRef/Dask.Future referencing to ip addresses of partitions
-        and partitions itself in depend of the engine used.
+        List of Ray.ObjectRef/Dask.Future to partitions depending on the engine used.
+        Or list of tuples of node ip addresses and Ray.ObjectRef/Dask.Future to partitions
+        depending on the engine used.
```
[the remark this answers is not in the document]
YarShev: Fixed.

modin/config/envvars.py
```diff
@@ -155,6 +155,14 @@ def _get_default(cls):
         return CpuCount.get()
 
 
+class RayPartitionThreshold(EnvironmentVariable, type=int):
```
devin-petersohn, line 158: This is quite specific, do users actually need such low level control over the storage semantics of this specific engine?

modin/engines/base/frame/axis_partition.py
```diff
@@ -95,33 +95,33 @@ def shuffle(self, func, lengths, **kwargs):
     def _wrap_partitions(self, partitions):
         return [self.partition_type(obj) for obj in partitions]
 
-    def coalesce(self, bind_ip=False):
+    def coalesce(self, get_ip=False):
```
devin-petersohn, line 98: Do you think `force_materialize` is a better name for this? `coalesce` generally happens anyway.

devin-petersohn (commented):
After reading this, would it be possible to have a separate API for the IP that returns a dictionary IP -> list of `ray.ObjectRef`/`Future` objects? That seems more useful and explicit than returning a tuple.

Generally the use case for getting the IP is to use for scheduling, so as a developer I would prefer to have a dictionary instead of a list of tuples. It would simplify the logic here too, I believe.

prutskov:
> After reading this, would it be possible to have a separate API for the IP that returns a dictionary IP -> list of `ray.ObjectRef`/`Future` objects? That seems more useful and explicit than returning a tuple.
> 
> Generally the use case for getting the IP is to use for scheduling, so as a developer I would prefer to have a dictionary instead of a list of tuples. It would simplify the logic here too, I believe.

It's good point, but in this case it will be good to know position of each partition from original list of partitions. This will be useful for user operations and `from_partitions` function in those interfaces.

In [Modin XGBoost](https://github.com/modin-project/modin/blob/master/modin/experimental/xgboost/xgboost_ray.py#L169) is used dictionary style `{'ip':[list of partitions], [positions of partitions in original list]}`, for example.

modin/config/envvars.py
YarShev: I think, this may be useful for those who do not want to add additional lines with `import ray` and `ray init(...)` to their workflow/scrpipts. The whole contol is left to Modin, but not to the developer.

modin/engines/base/frame/axis_partition.py
YarShev: I was thinking on this. I would prefer something of these: `use_ip`, `materialize_ip`, or `get_ip` at last :) because ObjectRefs to data will not be materilized yet.

YarShev:
@devin-petersohn , I am not sure that I see what you mean by the separate API. Something like that `unwrap_partitions_with_ips`?

modin/engines/base/frame/axis_partition.py
devin-petersohn: In the case of `axis_partition`, we return a single object which will have to be materialized. `force_materialize` provides a future, so I think it still makes sense instead of `coalesce`.

devin-petersohn:
Maybe something like `map_ip_addresses_to_partitions`. The API name will need to be worked out.

modin/engines/base/frame/axis_partition.py
YarShev: Do you want to rename `coalesce` to `force_materialization`? If so, I don't follow your thought. Why do we want to materialize `axis_partition` after its coaliscing? We need to return ObjectRef to the `axis_partitions` with saving asynchrony when `get_ip=False`.

YarShev:
`map_partitions_to_ips`?

YarShev:
Actually, I think, API for `unwrap_partitions` and `from_partitions` should be similar. I mean a container that is returned from `unwrap_partitions` and that is passed in `from_partitions` should be identical. The list of tuples, which is returned and that is passed in currently, looks more appropriate. On additional function `map_partitions_to_ips` that returns dict ip -> list of partitions, it doesn't look as huge effort for developer to perform such action, but to have such operation in Modin, I don't mind.

YarShev:
@devin-petersohn , just a friendly reminder.

modin/engines/base/frame/axis_partition.py
devin-petersohn: My thinking was: the assumption from the developers is that they can treat the axis partition as if it were already coalesced, so a `coalesce` api might cause confusion. `force_materialization` is a scarier name that will not encourage misuse. The fact that it is asynchronous does not change that the whole axis partition is materialized.

modin/config/envvars.py
devin-petersohn: Let's postpone this specific update to a new issue and PR. I'm not sure it is directly related to the partition API. It has to do with initialization of Ray.

devin-petersohn (commented):
@YarShev I am saying that I don't think it's useful to have the (partition, ip) tuple form. The only time it would be useful to get the IP addresses is if you want to know which objects are on which machines. A dictionary is a much more natural object for that mapping than a list of tuples.

modin/config/envvars.py
YarShev: Okay, it makes sense.

modin/engines/base/frame/axis_partition.py
YarShev: Okay, I see. We probably need to rename it to `force_materilization`.

YarShev:
@devin-petersohn , `map_partitions_to_ips` function was added to return dict ip -> list of partitions.

YarShev:
@devin-petersohn , I changed the way of ip getting for a partition. Now the whole logic is encapsulated in `ip` property of a partition. Also, now we don't have the issue for `small` objects of Ray that are stored in in-process store.

modin/distributed/dataframe/pandas/partitions.py
```diff
@@ -13,34 +13,35 @@
...
     Parameters
     ----------
     api_layer_object : DataFrame or Series
         The API layer object.
     axis : None, 0 or 1. Default is None
         The axis to unwrap partitions for (0 - row partitions, 1 - column partitions).
-        If axis is None, all the partitions of the API layer object are unwrapped.
-    bind_ip : boolean. Default is False
-        Whether to bind node ip address to each partition or not.
+        If ``axis=None``, all the partitions of the API layer object are unwrapped.
```
devin-petersohn, line 30:
```suggestion
        If ``axis is None``, the partitions are unwrapped as they are currently stored.
```

modin/engines/ray/pandas_on_ray/frame/partition.py
```diff
@@ -189,6 +192,25 @@ def width(self):
                 handle_ray_task_error(e)
         return self._width_cache
 
+    def ip(self):
+        if self._ip_cache is None:
+            if len(self.call_queue):
+                self.drain_call_queue()
+            # Since `ray.wait` returns a tuple of completed and not completed object refs as lists
+            # we need to retrive only completed object ref via `[0][0]` indexing.
+            oid = ray.wait([self.oid])[0][0]
```
devin-petersohn, line 201: I don't think we want to block on each IP address, it will force everything to block before it can move to the next partition. This combined with the `drain_call_queue` means that we will be operating single-threaded.

devin-petersohn (changes requested):

modin/engines/ray/pandas_on_ray/frame/partition.py
YarShev: I see these two possible cases:
1) We can notify a user as follows, for example, if you are going to use IPs, please note, that getting ip for each of the partitions is a blocking call.
2) We can return back to getting node ip addresses from remote functions taking into account the issue related to in-process store. `unwrap_partitions` would return a list of tuples of `ray.ObjectRefs` to ips and partitions, respectively. And `map_partitions_to_ips` would return a dictionary of ip -> list of partitions like `{str: [ObjectRef, ...], ...}` where we would materialize all of ips. For this `ip` property looks unnecessary.

Your thoughts?

prutskov:
@YarShev could you please attach performance comparison three approaches:
1. Approach with returning `[(str, ObjectRef),..]` using way with `ray.wait`+`ray.objects()`
2. Approach with returning `[(str, ObjectRef),..]` using way with internal getting of str IP's from ObjectRef IP's
3. Current approach from master with returning `[(IP_ObjectRef, Part_ObjectRef)]` + getting of str IP's from ObjectRef IP's on user-side.

YarShev:
@prutskov , of course, the 3rd item you mentioned is faster than other items because we get all IPs at once.

```python
import ray
import modin.pandas as pd
from modin.distributed.dataframe.pandas import unwrap_partitions
df = pd.read_csv("/my/file")
print('df.shape =', df.shape)
df.shape = (2260668, 145)
# the 1st item you mentioned
%time partitions = unwrap_partitions(df, axis=0, get_ip=True)
CPU times: user 980 ms, sys: 309 ms, total: 1.29 s
Wall time: 13.9 s
# the 2nd item you mentioned
%time partitions = unwrap_partitions(df, axis=0, get_ip=True)
CPU times: user 974 ms, sys: 301 ms, total: 1.27 s
Wall time: 12.8 s
# the 3rd item you mentioned
%time partitions = unwrap_partitions(df, axis=0, get_ip=True)
CPU times: user 335 ms, sys: 39.1 ms, total: 374 ms
Wall time: 303 ms
%time ips = ray.get([part[0] for part in partitions]) # @devin-petersohn , pay attention, this has about the same time as `%time ips = [ray.get(part[0]) for part in partitions]`
CPU times: user 522 ms, sys: 210 ms, total: 732 ms
Wall time: 3.12 s
```

prutskov:
> @prutskov , of course, the 3rd item you mentioned is faster than other items because we get all IPs at once.
> 
> ```python
> import ray
> import modin.pandas as pd
> from modin.distributed.dataframe.pandas import unwrap_partitions
> df = pd.read_csv("/my/file")
> print('df.shape =', df.shape)
> df.shape = (2260668, 145)
> # the 1st item you mentioned
> %time partitions = unwrap_partitions(df, axis=0, get_ip=True)
> CPU times: user 980 ms, sys: 309 ms, total: 1.29 s
> Wall time: 13.9 s
> # the 2nd item you mentioned
> %time partitions = unwrap_partitions(df, axis=0, get_ip=True)
> CPU times: user 974 ms, sys: 301 ms, total: 1.27 s
> Wall time: 12.8 s
> # the 3rd item you mentioned
> %time partitions = unwrap_partitions(df, axis=0, get_ip=True)
> CPU times: user 335 ms, sys: 39.1 ms, total: 374 ms
> Wall time: 303 ms
> %time ips = ray.get([part[0] for part in partitions]) # @devin-petersohn , pay attention, this has about the same time as `%time ips = [ray.get(part[0]) for part in partitions]`
> CPU times: user 522 ms, sys: 210 ms, total: 732 ms
> Wall time: 3.12 s
> ```

Thanks, @YarShev!
The general idea of this feature, in my opinion, is providing easy interface of getting IP's for partitions to make possible running user-commands around partitions in-place. This interface provides possibilities to performance improvements of user applications and this interface should work fast, otherwise the meaning is lost.

Can we add `ips = [ray.get(part[0]) for part in partitions]` in our implementation to save the performance?

YarShev:
@prutskov , once again, we have fully asynchronous getting ips in master and `unwrap_partitions` returns a list of tuples of futures objects to node ip addresses and partitions themselves. We can add materializing of all IPs and return a list of tuples of strings and futures to partitions from `unwrap_partitions`. However, we have the issue related to in-process store for ray in that case and we need to decide what is better. [These two](https://github.com/modin-project/modin/pull/2643#discussion_r578959486) are my thoughts. We need to choose from them or think of another way.

devin-petersohn:
I still prefer the dictionary `{ip_string: [ObjectRef]}` approach.

YarShev:
@devin-petersohn , that is what `map_partitions_to_ips` will return. We need to define one of [these ways](https://github.com/modin-project/modin/pull/2643#discussion_r578959486). I suggest going by the 2nd way and update the doc page adding the issue related to in-process store for Ray. Your thoughts? We do not need to put this off on the back burner.

devin-petersohn:
What about something like this?
```python
import itertools
import operator

ips = ray.get([part[0] for part in partitions])
ip_tuple = zip(ips, (part[1] for part in partitions))
ip_dict = {k: list(v) for k, v in dict(itertools.groupby(ip_tuple, operator.itemgetter(0))).items}
```

YarShev:
So, I guess, you are okay with the [2nd clause](https://github.com/modin-project/modin/pull/2643#discussion_r578959486) (return partition and ip from each remote call). I am returning back the changes. What about your suggestions above, this one
```python
ip_tuple = zip(ips, (part[1] for part in partitions))
```
we will return it from `unwrap_partitions`. It is necessary to keep initial order of partitions. And this one
```python
ip_dict = {k: list(v) for k, v in dict(itertools.groupby(ip_tuple, operator.itemgetter(0))).items}
```
we will return it from `map_partitions_to_ips`. It is for those users who don't care about order of partitions.

devin-petersohn:
I see, perhaps it would be better to provide an `OrderedDict` object in the dictionary case. The order is important likely for all implementations, it is a part of the assumptions for a dataframe.

YarShev:
That doesn't make sense because several partitions can fit to one ip but have different places in a dataframe.

YarShev:
@devin-petersohn , the changes were reverted. Please, look at them.

modin/distributed/dataframe/pandas/partitions.py
YarShev: Applied.

modin/engines/ray/pandas_on_ray/frame/partition.py
YarShev: Agreed with the 2nd clause.

prutskov (commented):
Need to add changes to Modin XGBoost related using `unwrap_partiitons` function. Current implementation uses `[(ip_ObjectRef, part_ObjectRef),..]` output style of this function. Also, could you please give performance numbers of getting IP this implementation VS previous implementation with materialization of IP's on user-side.

williamma12:
This pr looks really good! It was both a good catch on finding that subtle ray issue that originally led to this pr and everyone has wonderful ideas on clear, user-friendly, performant API for partitions!

However, I think this PR has slowly grown beyond the scope of its initial and intended purpose of fixing the bug described in #2642. For the sake of being able to make progress on the PR and make it easier to review, it would be helpful if we could split this into multiple PRs? 

It looks like that there is a bug fix for #2642, refactoring of the existing partitions API, and addition of a new partitions API. I'd be happy to clear out time in my schedule to help push the PR in, but we might be able to get through with these issues faster and move on to the bigger problems like the ones discussed [here](https://github.com/modin-project/modin/issues/2342#issuecomment-727870801).

YarShev:
@williamma12 , thanks for your attention to this PR! We would like to merge it as soon as possible too. I agree the PR has grown beyond the scope of its initial and intended purpose described in #2642. The goal was to add a doc page containing a note regarding ray in-process store issue (it is that what was added in #2721) and partition API declaring and also make some refactor. I suggest removing a new API (`map_partitions_to_ips`) from the PR, leave the changes related to refactor and add the doc page, which is in #2721, to here as well. Regarding performance measurements, I got the following numbers:

|time|n_parts = 16|n_parts = 32|n_parts = 64|n_parts = 112|
|------|------|------|------|------|
|unwrap_partitions_master|0.0321981377071804|0.0597010718451605|0.0962678591410319|0.2718964682685|
|getting_ip_master|0.00045331319173177|0.000844796498616536|0.00155636999342176|0.00207601653205023|
|unwrap_partitions_pr2643|1.51400627030266|1.12596117125617|1.12180537647671|1.0700610478719|
|getting_ip_pr2643|2.72856818305121E-06|4.42398918999566E-06|8.00026787651909E-06|0.0000138547685411241|

<details>

```python
from time import time
import ray
import modin.pandas as pd
from modin.distributed.dataframe.pandas import unwrap_partitions


def time_unwrap_partitions(df):
    partitions = df._query_compiler._modin_frame._partitions
    partitions = [part.oid for row in partitions for part in row]
    while partitions:
        _, partitions = ray.wait(partitions)

    common_time = 0.0
    n = 10
    for i in range(n):
        if i == 0:
            partitions = unwrap_partitions(df, axis=0, get_ip=True)
        else:
            t0 = time()
            partitions = unwrap_partitions(df, axis=0, get_ip=True)
            common_time += (time() - t0)
    common_time /= (n - 1)
    print('time_unwrap_partitions =', common_time)
    _partitions = [part[1] for part in partitions]
    while _partitions:
        _, _partitions = ray.wait(_partitions)
    return partitions


def time_getting_ip_master(partitions):
    common_time = 0.0
    n = 10
    for i in range(n):
        if i == 0:
            ips = ray.get([part[0] for part in partitions])
            # ips = [ray.get(part[0]) for part in partitions]
        else:
            t0 = time()
            ips = ray.get([part[0] for part in partitions])
            # ips = [ray.get(part[0]) for part in partitions]
            common_time += (time() - t0)
    common_time /= (n - 1)
    print('time_getting_ip_master =', common_time)
    return ips

def time_getting_ip_pr2643(partitions):
    common_time = 0.0
    n = 10
    for i in range(n):
        if i == 0:
            ips = [part[0] for part in partitions]
        else:
            t0 = time()
            ips = [part[0] for part in partitions]
            common_time += (time() - t0)
    common_time /= (n - 1)
    print('time_getting_ip_pr2643 =', common_time)
    return ips

df = pd.read_csv("/my/file")
print('df.shape =', df.shape)
df.shape = (2260668, 145)
partitions = time_unwrap_partitions(df)
# ips = time_getting_ip_master(partitions)
ips = time_getting_ip_pr2643(partitions)
```
</details>

As we can see, `unwrap_partitions_master` and `getting_ip_master` shows better performance than `unwrap_partitions_pr2643` and `getting_ip_pr2643`. So, I also suggest keeping a list of tuples of ObjRefs as a return value from `unwrap_partitions`. Then a user can materialize IPs himself and create more appropriate container for IPs and partitions for his goals on his side. @williamma12 , please, let me know what you think so I can start to make the changes.

modin/engines/dask/pandas_on_dask/frame/axis_partition.py
```diff
@@ -21,13 +21,13 @@
...
-    def __init__(self, list_of_blocks, bind_ip=False):
+    def __init__(self, list_of_blocks, get_ip=False):
         # Unwrap from BaseFramePartition object for ease of use
         for obj in list_of_blocks:
             obj.drain_call_queue()
         self.list_of_blocks = [obj.future for obj in list_of_blocks]
-        if bind_ip:
-            self.list_of_ips = [obj.ip for obj in list_of_blocks]
+        if get_ip:
+            self.list_of_ips = [obj._ip_cache for obj in list_of_blocks]
```
williamma12, line 30: Is it ok if the cache is empty?

modin/engines/ray/pandas_on_ray/frame/partition_manager.py
```diff
@@ -151,35 +128,27 @@ def get_indices(cls, axis, partitions, index_func=None):
 
     @classmethod
     def broadcast_apply(cls, axis, apply_func, left, right, other_name="r"):
-        def mapper(df, others):
-            other = pandas.concat(others, axis=axis ^ 1)
+        def map_func(df, others):
+            other = pandas.concat(ray.get(others), axis=axis ^ 1)
```
williamma12, line 132: Why are we blocking on others?

williamma12 (commented):
@YarShev That sounds like a great plan! 

This is a great refactor! It is a lot more readable! I just have some questions about a few things (I think they are staying in this PR)

YarShev:
> @YarShev That sounds like a great plan! I can see arguments for having us or the user to materialize since this is a developer API. If we decide on only having one, we can always add in the other later if people want it.

@williamma12 , I fully agree with you. So, we continue using a list of tuple of ObjRefs as a return value from `unwrap_partitions` as it is now in master. I am starting making the changes and moving the doc page, which is in #2721, to here.

modin/engines/dask/pandas_on_dask/frame/axis_partition.py
YarShev: The cache will be valid value after any remote call (any operation on a Modin DataFrame).

modin/engines/ray/pandas_on_ray/frame/partition_manager.py
YarShev: We are blocking on `others` because `others` is a list of `ray.ObjectRefs`. Ray doesn't materialize `ray.ObjectRefs` if they are in a list. We have to materialize them ourselves.

williamma12: Ah ok, thanks for letting me know about this weird ray behavior!

YarShev:
@williamma12 , just a friendly reminder to review.

modin/engines/ray/pandas_on_ray/frame/partition_manager.py
devin-petersohn: If we can pass `others` as `*others` to `apply` it should avoid the need to `ray.get`. I think this way it doesn't give Ray enough information for scheduling and in distributed mode can have some heavy performance implications for scheduling and placement.

YarShev: Currently we can't pass `others` as `*others` to `apply` because `apply` has different interface, i.e. `apply(self, func, **kwargs)`). We already have similar logic related to `ray.get` in `deploy_ray_func`. If we want to change the logic, i.e. get rid of explicit `ray.get`, we should perform that as a part of another issue because that may take a long time, I believe.

devin-petersohn: In that case I prefer to revert these changes since they are mostly refactoring anyway and make the right changes in a later PR.

YarShev: Okay, let's do it as a part of another PR. Reverted the changes.

devin-petersohn: Thanks, the main concern is negatively affecting current cluster users because we do not have sufficient scheduling hints to Ray yet. It is a good change to clean up the code but I do not want to make such performance critical changes so close to a release.

YarShev: I see, agree.

YarShev:
@devin-petersohn , @williamma12 , let's converge to merge the PR because it has been hanging for over a month.

williamma12 (approved):
This looks good!

However, I was wrong and I think for future PRs it might have been better if we did the refactor and docs separately. That way, the PRs would be smaller and more closely follow the one PR, one change rule of thumb, which will make things easier to review (aka faster to get PRs in). But this is already basically done so I don’t think its worth the extra effort of separating it into PRs after merging them into one.

Regardless, this is a great pr that improves performance, actually document the partitions API, and reduces technical debt!

devin-petersohn (approved):
Thanks @YarShev, glad we can converge!

Issue #2646, opened 2021-01-25
Add testing for Modin XGBoost


prutskov:
### Describe the problem

Need to add automated testing for Modin In XGBoost.

PR #2640, opened 2021-01-22
DOCS-2633: Add documentation for distributed XGBoost on Modin

prutskov:
Signed-off-by: Alexey Prutskov <alexey.prutskov@intel.com>

## What do these changes do?

This changes add documentation for distributed XGBoost on Modin

- [ ] commit message follows format outlined [here](https://modin.readthedocs.io/en/latest/CONTRIBUTING.html)
- [ ] passes `flake8 modin`
- [ ] passes `black --check modin`
- [ ] signed commit with `git commit -s` 
- [ ] Resolves #2633 ? 
- [ ] tests added and passing

PR #2659, opened 2021-01-28
FIX-#2658: Move backend check in xgb to train/predict

### LINKED ISSUE #2658: Move backend check in Modin XGBoost from module to funcs
### Describe the problem

In current implementation check for the backend is in `modin.experimental.xgboost`  module. Need to move this check to `predict/train` function directly to make possible importing this module (for building docs, for examples).

prutskov:

## What do these changes do?
This changes move checking of backend to separate function to make importing module possible.

- [x] passes `flake8 modin`
- [x] Resolves #2658

### DIFF
--- a/modin/experimental/xgboost/xgboost.py
+++ b/modin/experimental/xgboost/xgboost.py
@@ -98,7 +98,7 @@ def train(
     if Engine.get() == "Ray":
         from .xgboost_ray import _train
     else:
-        raise ValueError("Current version supports only Ray engine as MODIN_ENGINE.")
+        raise ValueError("Current version supports only Ray engine.")
 
     result = _train(
         dtrain, nthread, evenly_data_distribution, params, *args, evals=evals, **kwargs
@@ -141,7 +141,7 @@ def predict(
     if Engine.get() == "Ray":
         from .xgboost_ray import _predict
     else:
-        raise ValueError("Current version supports only Ray engine as MODIN_ENGINE.")
+        raise ValueError("Current version supports only Ray engine.")
 
     if isinstance(model, xgb.Booster):
         booster = model
--- a/modin/experimental/xgboost/xgboost.py
+++ b/modin/experimental/xgboost/xgboost.py
@@ -20,11 +20,6 @@ import xgboost as xgb
 
 from modin.config import Engine
 
-if Engine.get() == "Ray":
-    from .xgboost_ray import _train, _predict
-else:
-    raise ValueError("Current version supports only Ray engine as MODIN_ENGINE.")
-
 LOGGER = logging.getLogger("[modin.xgboost]")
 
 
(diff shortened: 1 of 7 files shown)


devin-petersohn (changes requested):
This looks good, is it possible to add tests to verify that the correct error gets thrown if the engine is not Ray?

modin/experimental/xgboost/xgboost.py
```diff
@@ -99,6 +94,12 @@ def train(
                          'eval': {'logloss': ['0.480385', '0.357756']}}}
     """
     LOGGER.info("Training started")
+
+    if Engine.get() == "Ray":
+        from .xgboost_ray import _train
+    else:
+        raise ValueError("Current version supports only Ray engine as MODIN_ENGINE.")
```
YarShev, line 101: https://github.com/modin-project/modin/pull/2643#discussion_r563659173

prutskov: Thanks, it was fixed

prutskov:
> This looks good, is it possible to add tests to verify that the correct error gets thrown if the engine is not Ray?

@devin-petersohn test was added, complex testing will be added in #2646

prutskov:
@devin-petersohn, @YarShev `xgboost` package was added as dependency for Modin. It will be needed for testing in #2646, for example. Also, we will not force the user to install `xgboost` self. What do you think? In case you agree I'll remove user note for installing `xgboost` in #2640.

devin-petersohn:
@prutskov The `requirements.txt` should really be renamed to `test_requires.txt`. It will not be installed for users when they install modin. I think this is the correct option.

prutskov:
@devin-petersohn Are you talking about on renaming `requirements.txt` -> `test_requires.txt` (and possible `environment.yml`->`test_enironment.yml` for conda) or about adding new files `test_requires.txt` and `test_enironment.yml`?

Also, it's found that conda package `py-xgboost` doesn't support `python=3.8`. We use python 3.8 in a part of testing.

devin-petersohn:
Conda forge package should just be xgboost: https://anaconda.org/conda-forge/xgboost

Yes, maybe a rename would be more appropriate. It is not a pressing matter.

prutskov:
@devin-petersohn Conda forge package `xgboost` doesn't support Windows platform. Instead of that I propose to use `xgboost` package [from pip](https://github.com/modin-project/modin/pull/2659/files#diff-9efd195f4e9bfb79ccd456a1d8370fafcc4bcb0b00ea3799222667d2ae818533R37-R38) in conda environment.

Renaming of files for environment will affect several files, I think it should be fixed as separate issue(#2667).

devin-petersohn (dismissed):
Thanks @prutskov LGTM!

devin-petersohn (approved):

PR #2660, opened 2021-01-29
FIX-#1806 Resolved error when reverting to Pandas for Multiindex

todd-yu:
## What do these changes do?

Added testing for multi-index functionality when reverting to Pandas.

- [x] passes `flake8 modin`
- [x] Resolves #? 
- [x] tests added and passing


modin/pandas/base.py
```diff
@@ -1687,30 +1687,37 @@ def reindex(
...
-            level is not None
-            or (
-                (columns is not None or axis == 1)
-                and self._query_compiler.has_multiindex(axis=1)
-            )
-            or (
-                (index is not None or axis == 0)
-                and self._query_compiler.has_multiindex()
-            )
+        if (columns is not None and self._query_compiler.has_multiindex(axis=1)) or (
```
devin-petersohn, line 1690: What is the problem if we just move the `axis = self._get_axis_number(axis)` statement to below this?

devin-petersohn (changes requested):
I think it's best to put the tests near the existing tests for `reset_index` rather than create a new file.

modin/pandas/base.py
todd-yu: If the user specifies an axis that's 1 or 0 during reindex with multi-index, we run into the original issue of providing both an axis and index/column to the Pandas API when defaulting.  

Here's some minimal reproducible code:

`index = np.array(["AUD","BRL","CAD","EUR","INR"])
data = np.random.randint(1, 20, (5,5))
A = pd.DataFrame(data=data, index=index, columns=index)
np.random.seed(42)
midx1 = pd.MultiIndex.from_product([["Bank_1", "Bank_2"], ["AUD","CAD","EUR"]], names=["Bank", "Curency"])
B = pd.DataFrame(np.random.randint(10,25,6), midx1)
B.columns = ["Notional"]
B.reindex(pd.MultiIndex.from_product([B.index.levels[0], A.index], names=["Bank", "Curency"]), fill_value=0, axis=0)`

devin-petersohn (approved):
Thanks @todd-yu!

PR #2662, opened 2021-01-30
FEAT-#2451: Read multiple csv files simultaneously via glob paths

### LINKED ISSUE #2451: Read from multiple csv files
It would be interesting to add the functionality to read from multiple files, much in the way that [dask does it](https://docs.dask.org/en/latest/dataframe-api.html#dask.dataframe.read_csv) (not sure about ray). 

Assume I have several csv files named 'data-001.csv', 'data-002.csv', ... , in dask I would load them all in one single dataframe using:

```python
import dask.dataframe as dd
ddf = dd.read_csv(
    'data-*.csv'
)
df = ddf.compute()
```

I understand that this is specific to dask and is not in the general pandas API, but perhaps could be interesting to add this functionality?

Currently I am doing the following to rapidly load the dataset and still work with modin: 

```python
import dask.dataframe as dd
import modin.pandas as pd
ddf = dd.read_csv(
    'data-*.csv'
)
df = ddf.compute()
df = pd.DataFrame(df)
```

---

This can be extremely useful when working, for instance, with Google Cloud Platform. I realized while working with a BigQuery table, which I exported to a bucket as multiple CSVs (this is the way Google allows me to do it via their GUI). Once downloaded, it'd be nice to load them all at once and avoid doing `pd.concat`, as it requires lot of memory.

williamma12:
## What do these changes do?
Add support for reading multiple csv files simultaneously via glob paths like dask.

To make this possible, I added a `CSVGlobDispatcher` to separate the code path from the regular `CSVDispatcher`. In doing so, glob support is only available via the ray experimental engine. 

`CSVGlobDispatcher` is the same as `CSVDispatcher` but assumes that the filepath is a glob path. s3 files are also supported via glob paths. To support glob paths, we search for the file via `glob.glob()` (or `s3fs.glob()`). Then, we calculate the partitions via `CSVGlobDispatcher.partitioned_multiple_files()`, which is the same as `TextFileDispatcher.partitioned_file()` but supports multiple files. Then, we pass it to `PandasCSVGlobParser`, which parses the file but supports the output from `CSVGlobDispatcher.partitioned_multiple_files()`.

- [x] passes `flake8 modin`
- [x] Resolves #2451 
- [x] tests added and passing

### DIFF
--- a/modin/data_management/factories/factories.py
+++ b/modin/data_management/factories/factories.py
@@ -232,7 +232,7 @@ class ExperimentalPandasOnRayFactory(ExperimentalBaseFactory, PandasOnRayFactory
         cls.io_cls = ExperimentalPandasOnRayIO
 
     @classmethod
-    def _read_csv(cls, **kwargs):
+    def _read_csv_glob(cls, **kwargs):
         return cls.io_cls.read_csv_glob(**kwargs)
 
 
--- a/modin/data_management/factories/factories.py
+++ b/modin/data_management/factories/factories.py
@@ -218,10 +218,6 @@ class ExperimentalBaseFactory(BaseFactory):
                 del kwargs["max_sessions"]
         return cls.io_cls.read_sql(**kwargs)
 
-    @classmethod
-    def _read_csv(cls, **kwargs):
-        return cls.io_cls.read_csv_glob(**kwargs)
-
 
 class ExperimentalPandasOnRayFactory(ExperimentalBaseFactory, PandasOnRayFactory):
     @classmethod
@@ -234,6 +230,10 @@ class ExperimentalPandasOnRayFactory(ExperimentalBaseFactory, PandasOnRayFactory
         )
 
         cls.io_cls = ExperimentalPandasOnRayIO
+
+    @classmethod
+    def _read_csv(cls, **kwargs):
+        return cls.io_cls.read_csv_glob(**kwargs)
 
 
 class ExperimentalPandasOnPythonFactory(ExperimentalBaseFactory, PandasOnPythonFactory):
--- a/modin/data_management/factories/factories.py
+++ b/modin/data_management/factories/factories.py
@@ -218,6 +218,10 @@ class ExperimentalBaseFactory(BaseFactory):
                 del kwargs["max_sessions"]
         return cls.io_cls.read_sql(**kwargs)
 
+    @classmethod
+    def _read_csv(cls, **kwargs):
+        return cls.io_cls.read_csv_glob(**kwargs)
+
 
 class ExperimentalPandasOnRayFactory(ExperimentalBaseFactory, PandasOnRayFactory):
     @classmethod
--- a/modin/engines/base/io/file_dispatcher.py
+++ b/modin/engines/base/io/file_dispatcher.py
@@ -13,8 +13,6 @@
 
 import os
 import re
-from typing import List
-
 from modin.config import Backend
 
 S3_ADDRESS_REGEX = re.compile("[sS]3://(.*?)/(.*)")
@@ -53,22 +51,9 @@ class FileDispatcher:
         raise NotImplementedError(NOT_IMPLEMENTED_MESSAGE)
 
     @classmethod
-    def get_path(cls, file_path: str) -> str:
-        """
-        Returns the path of the file(s).
-
-        Parameters
-        ----------
-        file_path: str
-            String representing a path.
-
-        Returns
-        -------
-        str
-            String of strings of absolute file paths.
-        """
+    def get_path(cls, file_path):
         if S3_ADDRESS_REGEX.search(file_path):
-            return cls._s3_path(file_path, False)[0]
+            return file_path
         else:
             return os.path.abspath(file_path)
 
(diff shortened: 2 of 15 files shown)


modin/engines/base/io/file_dispatcher.py
```diff
@@ -117,12 +139,31 @@ def file_size(cls, f):
         return size
 
     @classmethod
-    def file_exists(cls, file_path):
+    def file_exists(cls, file_path: str, glob_path: bool = False) -> bool:
```
devin-petersohn, line 142: I think I would prefer a different utility specific to globs here instead of a flag.

modin/engines/base/io/file_dispatcher.py
```diff
@@ -51,9 +54,28 @@ def _read(cls, *args, **kwargs):
         raise NotImplementedError(NOT_IMPLEMENTED_MESSAGE)
 
     @classmethod
-    def get_path(cls, file_path):
+    def get_path(cls, file_path: str, glob_path: bool = False) -> Union[list, str]:
```
devin-petersohn, line 57: I think I prefer a different utility for globs here.

devin-petersohn (commented):
@williamma12 Great PR! I have a few questions on it:

- I am wondering if there should be a separate codepath for globs outside of the single file codepath. 
- Also, the semantics of the glob are that it creates a signle dataframe from all files passed into the glob? Does this mean it's possible to have tons of partitions if there are many (e.g. hundreds of) files?
- I think we need this support in the experimental API since pandas does not support it.

modin/engines/base/io/file_dispatcher.py
williamma12: fixed

modin/engines/base/io/file_dispatcher.py
williamma12: fixed.

williamma12:
> * Also, the semantics of the glob are that it creates a signle dataframe from all files passed into the glob? Does this mean it's possible to have tons of partitions if there are many (e.g. hundreds of) files?
No, `CSVGlobDispatcher.partitioned_multiple_files()` will calculate the splits such that it will try to pack the files into the regular number of partitions. It will try to keep a file together but will have multiple files in one partition if the files are either not big enough or not exactly the partition size. This is why I added the `PandasCSVGlobParser`, which is now able to read in multiple files if necessary.

modin/engines/base/io/text/csv_dispatcher.py
```diff
@@ -11,26 +11,30 @@
...
+
 from modin.config import NPartitions
+from modin.data_management.utils import compute_chunksize
+from modin.engines.base.io.text.text_file_dispatcher import TextFileDispatcher
 
 
 class CSVDispatcher(TextFileDispatcher):
     @classmethod
-    def _read(cls, filepath_or_buffer, **kwargs):
+    def _read(cls, filepath_or_buffer: FilePathOrBuffer, **kwargs):
```
devin-petersohn, line 28: These are great changes, but can we split them in a different PR relating to code quality and documentation?

All CSVDispatcher changes can go in a different PR.

modin/engines/base/io/file_dispatcher.py
```diff
@@ -116,24 +131,67 @@ def file_size(cls, f):
...
+        try:
+            return get_file_path(s3fs)
+        except NoCredentialsError:
+            pass
+        s3fs = S3FS.S3FileSystem(anon=True)
+        return get_file_path(s3fs)
+
     @classmethod
-    def file_exists(cls, file_path):
+    def file_exists(cls, file_path: str) -> bool:
```
devin-petersohn, line 178: Instead of modifying the original, can we override it? This should simplify the code for the existing readers.

devin-petersohn (changes requested):
Left a couple of comments. Overall looks good.

My main question revolves around overriding the original `file_exists` and `open_file` instead of modifying the existing one. Do you think that would be better?

modin/engines/base/io/text/csv_dispatcher.py
williamma12: Makes sense! I have removed those changes

modin/engines/base/io/file_dispatcher.py
williamma12: Yes, I think I was thinking along these lines but didnt go as far. I have since made these changes

modin/experimental/engines/pandas_on_ray/io_exp.py
```diff
@@ -53,6 +57,14 @@ def _read_parquet_columns(path, columns, num_splits, kwargs):  # pragma: no cove
 
 
 class ExperimentalPandasOnRayIO(PandasOnRayIO):
+    build_args = dict(
+        frame_partition_cls=PandasOnRayFramePartition,
+        query_compiler_cls=PandasQueryCompiler,
+        frame_cls=PandasOnRayFrame,
+    )
+    read_csv = type(
```
devin-petersohn, line 65: Can we rename this to avoid confusion? Or do you disagree with this?
```suggestion
    read_csv_glob = type(
```

williamma12: Sure that sounds good. I have made the change

modin/experimental/pandas/io_exp.py
```diff
@@ -68,3 +72,113 @@ def read_sql(
...
+        reader = pd_obj.read
+        pd_obj.read = lambda *args, **kwargs: DataFrame(
+            query_compiler=reader(*args, **kwargs)
+        )
+        return pd_obj
+
+    return DataFrame(query_compiler=pd_obj)
+
+
+read_csv = _make_parser_func(sep=",")
```
devin-petersohn, line 184:
```suggestion
read_csv_glob = _make_parser_func(sep=",")
```

modin/data_management/factories/factories.py
```diff
@@ -231,6 +231,10 @@ def prepare(cls):
 
         cls.io_cls = ExperimentalPandasOnRayIO
 
+    @classmethod
+    def _read_csv(cls, **kwargs):
```
devin-petersohn, line 235:
```suggestion
    def _read_csv_glob(cls, **kwargs):
```

modin/experimental/pandas/io_exp.py
```diff
@@ -68,3 +72,113 @@ def read_sql(
...
+    ----------
+    filepath_or_buffer:
+        The filepath of the csv file.
+        We only support local files for now.
+    kwargs: Keyword arguments in pandas.read_csv
+    """
+    from modin.data_management.factories.dispatcher import EngineDispatcher
+
+    Engine.subscribe(_update_engine)
+    pd_obj = EngineDispatcher.read_csv(**kwargs)
```
devin-petersohn, line 171:
```suggestion
    pd_obj = EngineDispatcher.read_csv_glob(**kwargs)
```

devin-petersohn: Maybe want to wrap this one in `try..except`

devin-petersohn (commented):
@williamma12 I'm thinking we should change everything, even the top level API to `read_csv_glob`. That way it completely separates the behavior and is handled explicitly.

devin-petersohn (dismissed):
Thanks @williamma12, LGTM!

devin-petersohn (approved):
Thanks @williamma12

PR #2715, opened 2021-02-11
FEAT-#2215: implement to_csv in a distributed way

### LINKED ISSUE #2215: write dataframe to csv
I'm using Modin and it looks cool. However, when I try to write my dataframe to csv, it is defaulting to pandas. How far is the next update for this implementation?

anmyachev:

## What do these changes do?

```
Current algorithm:
	1) Combine all partitions in a pandas dataframe
	2) Call to_csv function: 
		a. Convert the dataframe to a string
		b. Write the string into the file

New algorithm:
	1) Convert each partition to a string (parallel stage)
             In order not to put the received strings into object storage,
             it is proposed to use the synchronization mechanism - `ray.util.queue.Queue`
        2) Write strings to the file in the desired order
```

- [x] passes `flake8 modin`
- [x] Resolves #2215 
- [x] tests added and passing

**Benchmark results**:

Dataframe size | Pandas | Modin - 4 cores | Modin - 8 cores
-|-|-|-
~ 800 MB | 39.2 sec | 11.5 sec | 6.5 sec |


modin/engines/ray/generic/io.py
```diff
@@ -44,3 +46,86 @@ def func(df):
...
+
+            with open(path_or_buf, **open_kwargs) as _f:
+                _f.write(content)
+
+            queue.put(get_value + 1)
+
+            return pandas.DataFrame()
+
+        queue.put(0)
+        result = qc._modin_frame.broadcast_apply_full_axis(
```
devin-petersohn, line 122: Do we really need to use `broadcast_apply`? I think the regular `map_axis_partitions` would be better here.

devin-petersohn (changes requested):
Just one question on the implementation, but overall looks great.

Can we add more inline documentation to the approach? For example, why return empty pandas.DataFrame() from the function? Please also try to remove several empty newline characters. Vertical whitespace should be use sparingly.

anmyachev:

devin-petersohn (approved):
Thanks @anmyachev, LGTM

PR #2724, opened 2021-02-11
TEST-#2722: add ASV read_csv skiprows benchmark

### LINKED ISSUE #2722: [ASV] add read_csv benchmark for skiprows parameter

amyskov:

## What do these changes do?
This PR added ASV benchmark for `read_csv` function with `skiprows` parameter. Also some common for IO and others benchmarks functions and configuration parameters were moved into `utils` module.

Note: subpart of https://github.com/modin-project/modin/pull/2607

- [x] passes `flake8 modin`
- [x] Resolves #2722  
- [x] tests passing

### DIFF
--- a/asv_bench/benchmarks/io/csv.py
+++ b/asv_bench/benchmarks/io/csv.py
@@ -1,0 +1,60 @@
+# Licensed to Modin Development Team under one or more contributor license agreements.
+# See the NOTICE file distributed with this work for additional information regarding
+# copyright ownership.  The Modin Development Team licenses this file to you under the
+# Apache License, Version 2.0 (the "License"); you may not use this file except in
+# compliance with the License.  You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software distributed under
+# the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific language
+# governing permissions and limitations under the License.
+
+import modin.pandas as pd
+import numpy as np
+
+from ..utils import (
+    generate_dataframe,
+    RAND_LOW,
+    RAND_HIGH,
+    ASV_USE_IMPL,
+    ASV_DATASET_SIZE,
+    UNARY_OP_DATA_SIZE,
+    IMPL,
+    execute,
+)
+
+# ray init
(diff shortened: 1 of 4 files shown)


asv_bench/benchmarks/io/csv.py
```diff
@@ -0,0 +1,60 @@
...
+        [
+            None,
+            lambda x: x % 2,
+            np.arange(1, UNARY_OP_DATA_SIZE[ASV_DATASET_SIZE][0][0] // 10),
+            np.arange(1, UNARY_OP_DATA_SIZE[ASV_DATASET_SIZE][0][0], 2),
+        ],
+    ]
+
+    def time_skiprows(self, test_filenames, shape, skiprows):
+        data_id = "_".join([str(size) for size in shape])
```
anmyachev, line 59: Can you move this line outside `time_skiprows` func? It affects result time.

amyskov: It is difficult to avoid using some file identification at all because we need to select needed test file in accordance to `shape` parameter and i hope we can leave this identification since it takes microseconds for execution that is much less than measurement error tolerance. Also special function for this case was added.

anmyachev: It's should be easy, just move data_id creation into `setup` function and save result in `self.data_id`.

amyskov: Yes, this is good option, right.

asv_bench/benchmarks/io/csv.py
```diff
@@ -0,0 +1,65 @@
...
+        for shape in UNARY_OP_DATA_SIZE[ASV_DATASET_SIZE]:
+            data_id = get_array_id(shape)
+            test_filenames[data_id] = f"{test_filename}_{data_id}.csv"
+            df = generate_dataframe("pandas", "str_int", *shape, RAND_LOW, RAND_HIGH)
+            df.to_csv(test_filenames[data_id], index=False)
+
+        return test_filenames
+
+    def setup(self, test_filenames, shape, *args, **kwargs):
+        self.data_id = get_array_id(shape)
```
anmyachev, line 47: Maybe `get_shape_id`?

anmyachev, line 47:
```suggestion
        self.shape_id = get_shape_id(shape)
```

amyskov: Ok, function was renamed.

anmyachev (approved):
LGTM!

PR #2740, opened 2021-02-16
REFACTOR-#2739: io tests refactoring

### LINKED ISSUE #2739: IO tests refactoring
IO tests common functions can be moved into `utils` module and fixtures can be moved into `modin/conftest.py` file where they can accessed from `modin/experimental` directories.

amyskov:

## What do these changes do?

- [x] passes `flake8 modin`
- [x] Resolves #2739  
- [x] tests added and passing

### DIFF
--- a/modin/pandas/test/utils.py
+++ b/modin/pandas/test/utils.py
@@ -39,8 +39,20 @@ DATASET_SIZE_DICT = {
     "Big": (2 ** 7, 2 ** 12),
 }
 
+# dataset sizes for IO operations
+DATASET_SIZE_DICT_IO = {
+    "Small": 64,
+    "Normal": 2000,
+    "Big": 20000,
+}
+
 # Size of test dataframes
 NCOLS, NROWS = DATASET_SIZE_DICT.get(TestDatasetSize.get(), DATASET_SIZE_DICT["Normal"])
+
+# Number of rows for IO test files
+NROWS_IO = DATASET_SIZE_DICT_IO.get(
+    TestDatasetSize.get(), DATASET_SIZE_DICT_IO["Small"]
+)
 
 # Range for values for test data
 RAND_LOW = 0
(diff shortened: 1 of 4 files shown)


modin/pandas/test/utils.py
```diff
@@ -1071,3 +1086,138 @@ def wrapped_function(self, *args, **kwargs):
...
+                    lambda x: f"{x:,f}".replace(",", thousands_separator)
+                )
+            filename = (
+                f"{filename}.{COMP_TO_EXT[compression]}"
+                if compression != "infer"
+                else filename
+            )
+            # create test data dir if it is not exists yet
+            if not os.path.exists(IO_OPS_DATA_DIR):
+                os.mkdir(IO_OPS_DATA_DIR)
```
anmyachev, line 1164: Why is this folder created for only `csv` files? What about other types, in instance `parquet`?

modin/pandas/test/utils.py
```diff
@@ -39,9 +39,21 @@
     "Big": (2 ** 7, 2 ** 12),
 }
 
+# dataset sizes for IO operations
+DATASET_SIZE_DICT_IO = {
+    "Small": 64,
+    "Normal": 2000,
+    "Big": 20000,
+}
+
```
anmyachev, line 48: You also have a similar variable in `modin/pandas/test/test_io.py` - `DATASET_SIZE_DICT`. Can we reuse it?

modin/pandas/test/utils.py
amyskov: Right, this directory should be created for all io tests, so directory creation was moved on the top of `conftest.py` file.

modin/pandas/test/utils.py
amyskov: Yes, i think we can use number of rows from `DATASET_SIZE_DICT`, default number of rows is larger, but tests execution time ingresses negligible.

anmyachev (approved):

PR #2744, opened 2021-02-17
REFACTOR-#2656: Refactor Modin to follow algebra laid out in paper.

### LINKED ISSUE #2656: Redesign Modin to align with DataFrame Algebra

Redesign Modin's architecture to implement the DataFrame Algebra proposed in the vision paper found [here](https://people.eecs.berkeley.edu/~totemtang/paper/Modin.pdf). Predecessor to #2204

RehanSD:
Signed-off-by: Rehan S. Durrani <rdurrani@berkeley.edu>

## What do these changes do?

Address Issue #2656 - redesign Modin to align with the Dataframe algebra discussed in the Modin paper.

- [x] passes `flake8 modin`
- [ ] Resolves #2656 ? 
- [ ] tests added and passing


modin/engines/modin_dataframe.py
```diff
@@ -0,0 +1,406 @@
...
+        pass
+
+    def filter_by_types(self, types):
+        """Allows the user to specify a type or set of types by which to filter the columns.
+
+        Parameters
+        ----------
+        types: hashable or list of hashables
+            The types to filter columns by.
+
```
RehanSD, line 59: Do we want to include a discussion of when exceptions are thrown here?

modin/engines/modin_dataframe.py
```diff
@@ -0,0 +1,406 @@
...
+        -------
+        ModinDataframe
+            A new ModinDataframe with the same columns as the previous, with only a single row.
+        """
+        pass
+
+    def infer_types(self, columns_list):
+        """Determines the compatible type shared by all values in the specified columns,
+        and converts all values to that type.
+
```
RehanSD, line 265: Do we want to include a discussion of when exceptions are thrown here?

modin/engines/modin_dataframe.py
```diff
@@ -0,0 +1,406 @@
...
+    def transpose(self):
+        """Swaps the row and column axes.
+
+        Notes
+        -----
+            Transposing a dataframe is expensive, and thus, while the axes are swapped
+                logically immediately, the physical swap does not occur until absolutely necessary,
+                which helps motivate the axis argument to the other operators in this algebra.
+
+            This operation explicitly manipulates dataframe metadata.
```
RehanSD, line 341: I added this note to any function that operated on metadata - do we want to keep it?

modin/engines/modin_dataframe.py
devin-petersohn: Yes, it should just be Python TypeErrors in this case. Any others?

modin/engines/modin_dataframe.py
```diff
@@ -0,0 +1,406 @@
...
+    This class is intended to specify the behaviours that a DataFrame must implement.
+
+    For more details about how these methods were chosen, please refer to this
+    (http://www.vldb.org/pvldb/vol13/p2033-petersohn.pdf) paper, which specifies
+    a DataFrame algebra that this class exposes.
+
+    """
+
+    def mask(
+        self, row_indices=None, row_positions=None, col_indices=None, col_positions=None
```
devin-petersohn, line 29: Can you add type hints?

modin/engines/modin_dataframe.py
```diff
@@ -0,0 +1,406 @@
...
+            The column indices to extract.
+
+        Returns
+        -------
+        ModinDataframe
+             A new ModinDataframe from the mask provided.
+        """
+        pass
+
+    def filter_by_types(self, types):
```
devin-petersohn, line 52: Let's make this always be a list, similar to `mask`

modin/engines/modin_dataframe.py
```diff
@@ -0,0 +1,406 @@
...
+            The types to filter columns by.
+
+        Returns
+        -------
+        ModinDataframe
+             A new ModinDataframe from the filter provided.
+        """
+        pass
+
+    def map(self, axis, function, result_schema=None):
```
devin-petersohn, line 67: Type hints here too. I think we can add them for every method here.

modin/engines/modin_dataframe.py
devin-petersohn: I don't think it's necessarily useful.

devin-petersohn (changes requested):

modin/engines/modin_dataframe.py
```diff
@@ -0,0 +1,406 @@
...
+        -----
+            This does not change the number of rows.
+
+            The user-defined function may increase the number of columns (rows if axis=1),
+                but it should not remove or drop columns and each invocation of the function
+                must generate the same number of new columns.
+
+        Parameters
+        ----------
+            axis: int
```
devin-petersohn, line 80: axis can also be `None` here.

devin-petersohn (changes requested):
One thing missed from last review.

modin/engines/modin_dataframe.py
RehanSD: This was meant to be a continuation of the discussion stemming from the section of the paper that stated that infer_types would throw an error - you mentioned that if the types couldn't be coerced, an error should be thrown; however, depending on the type system, no errors may be thrown?

devin-petersohn: I don't think we're adding rules for coercing types. The `TypeError` would probably not happen in this method.

devin-petersohn (commented):
At a high level looks good. We'll spend some time going over it in detail together and leave comments here.

devin-petersohn (changes requested):

RehanSD:
@devin-petersohn Do we want to change all uses of `{row|col}_indices` and `{row|col}_numeric_idx` to `{row|col}_labels` and `{row|col}_positions` (respectively)?

devin-petersohn:
@RehanSD Yes

modin/engines/modin_dataframe.py
```diff
@@ -0,0 +1,477 @@
...
+# Unless required by applicable law or agreed to in writing, software distributed under
+# the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific language
+# governing permissions and limitations under the License.
+
+from abc import ABC, abstractmethod
+from typing import List, Hashable, Optional, Callable, Union, Dict
+
+
+class ModinDataframe(ABC):
```
YarShev, line 18: I think we should change `ModinDataframe` to `ModinDataFrame`.

modin/engines/modin_dataframe.py
```diff
@@ -0,0 +1,477 @@
...
+    @abstractmethod
+    def map(
+        self,
+        function: Callable,
+        axis: Optional[int] = None,
+        result_schema: Optional[Dict[Hashable, type]] = None,
+    ) -> "ModinDataframe":
+        """Apply a user-defined function row- wise (or column-wise if axis=1).
+
+        Notes
```
YarShev, line 85: I would prefer `Notes` follows after `Returns` as it is done in other projects. Also, numpydoc guide describes `Notes` as one of the last sections in a docstring.

modin/engines/modin_dataframe.py
YarShev: By the way, there was a discussion on removing part `Data` from `Modin DataFrame` some time ago. What way do we go?
@devin-petersohn , @vnlitvinov

devin-petersohn: `DataFrame` is camel case for "data frame", but I think it should be one word. There are historical reasons for this.

`ModinFrame` is more concise, but I wouldn't call a Modin database `ModinBase`. Interested in other opinions too on this.

modin/engines/modin_dataframe.py
devin-petersohn: +1 let's follow numpydoc standard

modin/engines/modin_dataframe.py
vnlitvinov: I think calling something a `DataFrame` (or `Dataframe`) when we provide `modin.pandas.DataFrame` as `pandas.DataFrame` replacement looks ambiguous.

For that matter I think I would prefer seeing `BaseFrame` here, or even something without "Frame" word in it to remove even smallest possible confusion here, like `ModinBaseBody` maybe?

YarShev: > I think calling something a `DataFrame` (or `Dataframe`) when we provide `modin.pandas.DataFrame` as `pandas.DataFrame` replacement looks ambiguous.

Probably, yes. It looks ambiguous a bit.

> For that matter I think I would prefer seeing `BaseFrame` here, or even something without "Frame" word in it to remove even smallest possible confusion here, like `ModinBaseBody` maybe?

I wouldn't tie `Modin` part to any name of a class/method. I like name `BaseFrame`. Also, if we decide to support another distributed object at one day, it would be good to name base class like `BaseTensor`.

devin-petersohn: >I wouldn't tie Modin part to any name of a class/method. I like name BaseFrame.

That makes sense. `BaseDataframe` or `CoreDataframe`. I don't think shorthanding `Dataframe` to `Frame` is worth the extra characters saved. I'm necessary agree that there's ambiguity here with the API layer here because users will not see this for the most part. Developers will understand the abstractions (provided we document them well enough 😄).

YarShev: I think I see your idea about saving `DataFrame` instead of `Frame`. Let's keep `DataFrame` then :)

modin/engines/base/frame/dataframe.py
```diff
@@ -18,12 +18,49 @@
...
+    ):
+        raise TypeError(
+            "Resultant schema:\n{}\nDoes not match expected result schema:\n{}".format(
+                new_df.dtypes, new_dtypes
+            )
+        )
+    return new_df
+
+
+class BasePandasFrame(ModinDataframe):
```
YarShev, line 63: Do we want to fully abstract from pandas as base backend? We need to know this to form the right structure/architecture of Modin being discussed in #2957.

YarShev: I think we should distinguish base classes (frame, manager, partitions) and classes for pandas backend. We need to define general API for base classes that has to be overridden in child classes. Those things are currently mixed in base folder. @devin-petersohn , what do you think?

docs/comparisons/dask.rst
```diff
@@ -59,7 +59,7 @@ Dask DataFrame
 Dask DataFrame uses row-based partitioning, similar to Spark. This can be seen in their
 `documentation`_. They also have a custom index object for indexing into the object,
 which is not pandas compatible. Dask DataFrame seems to treat operations on the
-DataFrame as MapReduce operations, which is a good paradigm for the subset of the pandas
+DataFrame as TreeReduce operations, which is a good paradigm for the subset of the pandas
```
YarShev, line 62: Why was `MapReduce` renamed to `TreeReduce`?

modin/engines/base/frame/dataframe.py
devin-petersohn: I agree @YarShev.

docs/comparisons/dask.rst
devin-petersohn: `TreeReduce` is a more descriptive way to describe the pattern expected. `MapReduce` has certain connotations related to defining number of mappers/reducers, etc. I think we want to avoid any concerns there.

YarShev: I see, that makes sense, thanks!

YarShev:
@devin-petersohn , I would prefer the way I mentioned above ([link](https://github.com/modin-project/modin/pull/2744#issuecomment-944234928)). But if he really wants to work on this after #3412 and #3413 are merged, some documentation changes will be required as well.

RehanSD:
Obsoleted by #3661.

Issue #1611, opened 2020-06-16
Support different datetime extractions


Garra1980:
Extract_year already supported, we need to support other similar operations. 

Note that so far we use parameter shape_hint="column" during construction of Series object to prevent redundant frame materialization. This parameter so far is in omnisci-on-ray branch only

abykovsk:
Need to add more datetime functions in omniscidb.
Also OmniSci backend has different behavior from pandas in functions such as `dt.dayofweek`, `dt.microsecond` and `dt.nanosecond`.
In `dt.dayofweek`, we have different start for a week (Monday for pandas and Sunday for OmniSci).
In `dt.microsecond` and `dt.nanosecond` we have different results for pandas (number of micro or nanoseconds) and OmniSci (micro or nanoseconds + seconds)

abykovsk:
[PR](https://github.com/modin-project/modin/pull/2726) with adding mod operation for `dt.microsecond` and `dt.nanosecond` was merged.
[PR](https://github.com/intel-ai/omniscidb/pull/125) with adding option for `dt.dayofweek` was merged.
[PR](https://github.com/intel-ai/omniscidb/pull/126) with adding functions for `dt.dayofweek`, `dt.dayofyear`, `dt.daysinmonth`, `dt.is_leap_year`, `dt.is_month_end`, `dt.is_month_start`, `dt.is_quarter_end`, `dt.is_quarter_start`, `dt.is_year_end`, `dt.is_year_start` in omnisci backend was merged.
[PR](https://github.com/modin-project/modin/pull/2750) with adding functions for `dt.week`, `dt.day`, `dt.hour`, `dt.minute`, `dt.second`, `dt.microsecond`, `dt.nanosecond`, `dt.quarter`, `dt.dayofweek`, `dt.dayofyear`, `dt.daysinmonth`, `dt.is_leap_year`, `dt.is_month_end`, `dt.is_month_start`, `dt.is_quarter_end`, `dt.is_quarter_start`, `dt.is_year_end`, `dt.is_year_start` is ready for review.

Need to add functions for TimeDeltas

PR #2750, opened 2021-02-17
FEAT-#1611: Add some datetime extraction functions

abykovsk:
Signed-off-by: Alina <alina.bykovskaya@intel.com>

## What do these changes do?

- [x] passes `flake8 modin`
- [ ] Resolves #? 
- [ ] tests added and passing


modin/pandas/series_utils.py
```diff
@@ -484,11 +484,11 @@ def second(self):
 
     @property
     def microsecond(self):
-        return Series(query_compiler=self._query_compiler.dt_microsecond())
+        return Series(query_compiler=self._query_compiler.dt_microsecond()) % 1000000
```
ienkovich, line 487: Why do you need these changes?

modin/experimental/backends/omnisci/query_compiler.py
```diff
@@ -429,6 +429,196 @@ def dt_month(self):
             self._modin_frame.dt_extract("month"), self._shape_hint
         )
 
+    def dt_week(self):
+        """Extract week from Datetime info
```
ienkovich, line 433: I know you follow docs style from other methods in this class, but there was an agreement on it and we should use new formatting in the new code. You just need to use something like this:
```
Returns
-------
DFAlgQueryCompiler
    A new query compiler.
```

ienkovich (changes requested):
Did you check tests pass with updated OmniSci?

modin/pandas/series_utils.py
abykovsk: Omnisci returns **seconds * 1000000 + microseconds** instead of just **microseconds**. Same behavior for nanoseconds.

abykovsk:
> Did you check tests pass with updated OmniSci?

Yes, I checked it with related changes in OmniSci

devin-petersohn:
@Garra1980 should this be closed?

Garra1980:
Nope, we plan to return to this some time this year

YarShev:
@Garra1980, do we plan to continue this PR?

Garra1980:
Yes

anmyachev:
Continued in #6568

PR #2757, opened 2021-02-18
FIX-#2751: speed up dtypes

### LINKED ISSUE #2751: speed up _compute_dtypes function
The function works extremely inefficiently when the number of columns is approximately equal to the number of rows.

For example, if `df.shape = (5000, 5000)` and `MODIN_CPUS=44` then `count_partitions=44^2`.

anmyachev:

## What do these changes do?

- [ ] passes `flake8 modin`
- [x] Resolves #2751 
- [ ] tests added and passing

### DIFF
--- a/modin/engines/base/frame/data.py
+++ b/modin/engines/base/frame/data.py
@@ -141,11 +141,9 @@ class BasePandasFrame(object):
         def dtype_builder(df):
             return df.apply(lambda col: find_common_type(col.values), axis=0)
 
-        map_func = self._build_mapreduce_func(0, lambda df: df.dtypes)
-        reduce_func = self._build_mapreduce_func(0, dtype_builder)
         # For now we will use a pandas Series for the dtypes.
         if len(self.columns) > 0:
-            dtypes = self._map_reduce(0, map_func, reduce_func).to_pandas().iloc[0]
+            dtypes = self._map_reduce(0, lambda df: df.dtypes, dtype_builder)
         else:
             dtypes = pandas.Series([])
         # reset name to None because we use "__reduced__" internally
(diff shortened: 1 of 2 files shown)


modin/engines/base/frame/data.py
```diff
@@ -141,11 +141,9 @@ def _compute_dtypes(self):
         def dtype_builder(df):
             return df.apply(lambda col: find_common_type(col.values), axis=0)
 
-        map_func = self._build_mapreduce_func(0, lambda df: df.dtypes)
-        reduce_func = self._build_mapreduce_func(0, dtype_builder)
         # For now we will use a pandas Series for the dtypes.
         if len(self.columns) > 0:
-            dtypes = self._map_reduce(0, map_func, reduce_func).to_pandas().iloc[0]
+            dtypes = self._map_reduce(0, lambda df: df.dtypes, dtype_builder)
```
devin-petersohn, line 146: This is incorrect, it will always return `object` for every column because the dtype of a `Series` of dtypes is `object`. Is this the cause of the performance?

devin-petersohn (changes requested):

modin/engines/base/frame/data.py
anmyachev: I have an example of dtypes output:

```python

(Pdb) test.df.dtypes
col0       float64
col1       float64
col2       float64
col3       float64
col4       float64
            ...
col4995    float64
col4996    float64
col4997    float64
col4998    float64
col4999    float64
Length: 5000, dtype: object
(Pdb) type(test.df.dtypes)

```

anmyachev: Oh, I forgot to remove some of the code

devin-petersohn: Yes, sorry. I missed the third argument `dtype_builder`

anmyachev:
In case when number of partition close to used cpus, previous approach much faster.

PR #2794, opened 2021-03-02
FEAT-#0000: use ray wait to trigger computation

anmyachev:

## What do these changes do?

- [ ] passes `flake8 modin`
- [ ] Resolves #? 
- [ ] tests added and passing


devin-petersohn (commented):
@anmyachev Maybe we should add a "synchronous mode" to Modin. What do you think?

anmyachev:
> @anmyachev Maybe we should add a "synchronous mode" to Modin. What do you think?

I don't see reasons why the user would want this behavior. But on the other hand, it will be convenient for us to measure the time of different functions inside a script or notebook.

Perhaps we should try adding such a mode and see how it complicates the code.

devin-petersohn:
I think it will mostly be useful for benchmarking, not for users.

anmyachev:
"Synchronous mode" added in #2800

PR #2801, opened 2021-03-03
FEAT-#2800: add wait_computations, wait apis

### LINKED ISSUE #2800: Add wait function for _modin_frame

This is the most lightweight way to wait for the end of parallel computations on partitions.

anmyachev:

## What do these changes do?

- [x] passes `flake8 modin`
- [x] Resolves #2800 
- [x] tests added and passing

### DIFF
--- a/asv_bench/benchmarks/utils.py
+++ b/asv_bench/benchmarks/utils.py
@@ -221,15 +221,20 @@ def execute(df):
     "Make sure the calculations are done."
     if ASV_USE_IMPL == "modin":
         partitions = df._query_compiler._modin_frame._partitions
-        map(lambda partition: partition.drain_call_queue(), partitions)
+        all(
+            map(
+                lambda partition: partition.drain_call_queue() or True,
+                partitions.flatten(),
+            )
+        )
         if ASV_USE_ENGINE == "Ray":
             from ray import wait
 
-            map(lambda partition: wait(partition.oid), partitions)
+            all(map(lambda partition: wait([partition.oid]), partitions.flatten()))
         elif ASV_USE_ENGINE == "Dask":
             from dask.distributed import wait
 
-            map(lambda partition: wait(partition.future), partitions)
+            all(map(lambda partition: wait(partition.future), partitions.flatten()))
         elif ASV_USE_ENGINE == "Python":
             pass
 
--- a/modin/config/envvars.py
+++ b/modin/config/envvars.py
@@ -261,6 +261,15 @@ class ProgressBar(EnvironmentVariable, type=bool):
         cls.put(False)
 
 
+class BenchmarkMode(EnvironmentVariable, type=bool):
+    """
+    Whether or not to perform computations syncronous.
+    """
+
+    varname = "MODIN_BENCHMARK_MODE"
+    default = False
+
+
 def _check_vars():
     """
     Look out for any environment variables that start with "MODIN_" prefix
--- a/modin/config/envvars.py
+++ b/modin/config/envvars.py
@@ -260,6 +260,12 @@ class ProgressBar(EnvironmentVariable, type=bool):
     def disable(cls):
         cls.put(False)
 
+    @classmethod
+    def put(cls, value):
+        if value and BenchmarkMode.get():
+            raise ValueError("ProgressBar isn't compatible with BenchmarkMode")
+        super().put(value)
+
 
 class BenchmarkMode(EnvironmentVariable, type=bool):
     """
@@ -268,6 +274,12 @@ class BenchmarkMode(EnvironmentVariable, type=bool):
 
     varname = "MODIN_BENCHMARK_MODE"
     default = False
+
+    @classmethod
+    def put(cls, value):
+        if value and ProgressBar.get():
+            raise ValueError("BenchmarkMode isn't compatible with ProgressBar")
+        super().put(value)
 
 
 def _check_vars():
--- a/modin/engines/base/frame/data.py
+++ b/modin/engines/base/frame/data.py
@@ -2102,6 +2102,10 @@ class BasePandasFrame(object):
 
         return df
 
+    def wait_computations(self):
+        """Wait for computation results."""
+        self._frame_mgr_cls.wait_computations(self._partitions)
+
     def to_numpy(self, **kwargs):
         """
         Convert a Modin DataFrame to a 2D NumPy array.
--- a/modin/engines/base/frame/data.py
+++ b/modin/engines/base/frame/data.py
@@ -21,7 +21,6 @@ from typing import List, Hashable
 from modin.backends.pandas.query_compiler import PandasQueryCompiler
 from modin.error_message import ErrorMessage
 from modin.backends.pandas.parsers import find_common_type_cat as find_common_type
-from modin.config import BenchmarkMode
 
 
 class BasePandasFrame(object):
@@ -81,9 +80,6 @@ class BasePandasFrame(object):
         self._column_widths_cache = column_widths
         self._dtypes = dtypes
         self._filter_empties()
-
-        if BenchmarkMode.get():
-            self.wait_computations()
 
     @property
     def _row_lengths(self):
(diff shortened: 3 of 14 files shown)


modin/engines/base/frame/partition_manager.py
```diff
@@ -555,6 +555,13 @@ def to_pandas(cls, partitions):
         else:
             return cls.concatenate(df_rows)
 
+    @classmethod
+    def wait_computations(cls, partitions):
+        """Wait for computation results."""
+        for row in partitions:
+            for partition in row:
+                partition.wait()
```
anmyachev, line 563: Maybe `flatten`?

Garra1980 (dismissed):

modin/engines/base/frame/data.py
```diff
@@ -2102,6 +2102,10 @@ def to_pandas(self):
 
         return df
 
+    def wait_computations(self):
```
devin-petersohn, line 2105: What would you think if we made this a decorator instead? Then decorate all of the methods that trigger computation. 

If it is implemented this way, we could add a config for `modin.config.BenmarkMode` and check within the decorator if that flag is set. What do you think?

devin-petersohn (commented):
Looks great! One question on your thoughts for structure.

modin/engines/base/frame/data.py
anmyachev: I suggest leaving both the decorator and the function, since in different situations it is convenient to use either one or the other.

anmyachev: It seems that we should cast map object to list

asv_bench/benchmarks/utils.py
```diff
@@ -221,15 +221,20 @@ def execute(df):
...
+            map(
+                lambda partition: partition.drain_call_queue() or True,
+                partitions.flatten(),
+            )
+        )
         if ASV_USE_ENGINE == "Ray":
             from ray import wait
 
-            map(lambda partition: wait(partition.oid), partitions)
+            all(map(lambda partition: wait([partition.oid]), partitions.flatten()))
```
anmyachev, line 233: I forgot that `map` changed behavior in python3.

Python3 - type(map return value) like iterator
Python2 - type(map return value) - list

anmyachev:

modin/engines/base/frame/data.py
devin-petersohn: I don't think we need both. In fact, I prefer this functionality not be exposed here. We can move the decorator into the frame manager.

I prefer these be deleted here because we don't want others to be accidentally calling them. Benchmark Mode is for internal and developer use only.

modin/engines/base/frame/partition_manager.py
```diff
@@ -555,6 +555,14 @@ def to_pandas(cls, partitions):
         else:
             return cls.concatenate(df_rows)
 
+    @classmethod
+    def wait_computations(cls, partitions):
```
devin-petersohn, line 559: Along with changes above, I would move this outside the class as a decorator and decorate all functions, first checking if `BenchmarkMode.get()` then executing the line below. This way, it's not an API that is explicitly available for users.

devin-petersohn (changes requested):
I think there's a less invasive design explained below. Please let me know if it makes sense!

modin/engines/base/frame/partition_manager.py
anmyachev: @devin-petersohn thanks for review! It's make sense. 

`wait_computations` was made as decorator function and other wait ways were removed. Please review again :)

devin-petersohn (commented):
The approach looks great! I tested locally but I can't seem to get the synchronous mode to work. Here is how I tested:

```python
import modin.pandas as pd
import numpy as np

BenchmarkMode.put(True)
df = pd.DataFrame(np.random.randint(0, 100, size=(2**20, 2**8))).add_prefix("col")
a = df.applymap(lambda x: x)  # returns immediately
```

Do I need to do something else?

anmyachev:
> The approach looks great! I tested locally but I can't seem to get the synchronous mode to work. Here is how I tested:
> 
> ```python
> import modin.pandas as pd
> import numpy as np
> from modin.config import BenchmarkMode
> 
> BenchmarkMode.put(True)
> df = pd.DataFrame(np.random.randint(0, 100, size=(2**20, 2**8))).add_prefix("col")
> a = df.applymap(lambda x: x)  # returns immediately
> ```
> 
> Do I need to do something else?

No, it's enough.

I tested your script with `time` calls. I got next times: `3.08` sec in sync mode and 0.19 sec in async mode.
Environment: `MODIN_CPUS=44, MODIN_ENGINE=ray`

```python
import modin.pandas as pd
import numpy as np
from time import time

BenchmarkMode.put(True)
df = pd.DataFrame(np.random.randint(0, 100, size=(2**20, 2**8))).add_prefix("col")
print(f"dataframe created")
start = time()
a = df.applymap(lambda x: x)
print(f"applymap: {time()-start}")

```

anmyachev:
@devin-petersohn are you using the latest commit from the PR? What about modin environment variables? Which ones have you tested with?

devin-petersohn:
@anmyachev This time I tried it, it is working. I do not know why it failed the first time, I was using the ProgressBar with it. I have noticed that the ProgressBar does not work properly if both are set to `True` because `BenchmarkMode` will block until the end and `ProgressBar` also blocks. Can you add a check for when `BenchmarkMode` and `ProgressBar` are enabled that throws an Exception for being incompatible?

anmyachev:
> @anmyachev This time I tried it, it is working. I do not know why it failed the first time, I was using the ProgressBar with it. I have noticed that the ProgressBar does not work properly if both are set to `True` because `BenchmarkMode` will block until the end and `ProgressBar` also blocks. Can you add a check for when `BenchmarkMode` and `ProgressBar` are enabled that throws an Exception for being incompatible?

@devin-petersohn thanks for review! The check was be added.

devin-petersohn (approved):
Thanks @anmyachev LGTM

devin-petersohn:
@modin-project/modin-omnisci waiting for your approval.

Garra1980 (approved):

modin/config/envvars.py
```diff
@@ -260,6 +260,27 @@ def enable(cls):
...
+    @classmethod
+    def put(cls, value):
+        if value and BenchmarkMode.get():
+            raise ValueError("ProgressBar isn't compatible with BenchmarkMode")
+        super().put(value)
+
+
+class BenchmarkMode(EnvironmentVariable, type=bool):
+    """
+    Whether or not to perform computations syncronous.
```
YarShev, line 272: syncronous -> synchronous

modin/engines/base/frame/partition_manager.py
```diff
@@ -18,12 +18,44 @@
...
+    Note
+    ----
+        `func` should return numpy array with partitions.
+    """
+    if BenchmarkMode.get():
+
+        def wait(*args, **kwargs):
+            """Wait for computation results."""
+            result = func(*args, **kwargs)
+            if isinstance(result, tuple):
```
YarShev, line 45: Why is this check necessary?

modin/experimental/engines/omnisci_on_ray/frame/partition.py
```diff
@@ -43,6 +43,9 @@ def get(self):
             return self.arrow_table
         return ray.get(self.oid)
 
+    def wait(self):
+        ray.wait([self.oid])
```
YarShev, line 47: Why don't we handle ray operations here as we do for pandas on ray? (see [here](https://github.com/modin-project/modin/pull/2801/files#diff-2df6c73d4a63d1c1345522920436cf6b0aec9abd1105fee23ce74a524981a35aR90))

modin/engines/base/frame/partition_manager.py
anmyachev: To distinguish situations where one or more values are returned.

modin/experimental/engines/omnisci_on_ray/frame/partition.py
anmyachev: I did in the same style as `get` function in `OmnisciOnRayFramePartition` class.

modin/engines/base/frame/partition_manager.py
YarShev: Which situations do you mean? Please, give an example.

modin/experimental/engines/omnisci_on_ray/frame/partition.py
YarShev: We probably need to do the same handling for `OmnisciOnRayFramePartition`'s operations that include ray calls.

modin/engines/base/frame/partition_manager.py
anmyachev: In instance, `from_pandas` and `map_partitions`

YarShev: I see, it makes sense for `from_pandas` but not for `map_partitions`. The latter returns np.array.

modin/experimental/engines/omnisci_on_ray/frame/partition.py
anmyachev: https://github.com/modin-project/modin/issues/2837

PR #2828, opened 2021-03-05
FIX-#2798: Fix number of partitions for dataframe on a cluster

### LINKED ISSUE #2798: Wrong number of partitions of a Modin DataFrame on a cluster
### Describe the problem

Details are in https://discuss.modin.org/t/dividing-a-modin-dataframe-into-partitions-on-a-cluster/183

YarShev:

## What do these changes do?

- [x] passes `flake8 modin`
- [x] Resolves #2798  
- [x] tests passing

### DIFF
--- a/modin/config/envvars.py
+++ b/modin/config/envvars.py
@@ -123,7 +123,7 @@ class RayRedisAddress(EnvironmentVariable, type=ExactStr):
 
 class CpuCount(EnvironmentVariable, type=int):
     """
-    How may CPU cores to utilize across the whole distribution
+    How many CPU cores to use when initialization of the Modin engine.
     """
 
     varname = "MODIN_CPUS"
--- a/modin/config/envvars.py
+++ b/modin/config/envvars.py
@@ -145,10 +145,20 @@ class Memory(EnvironmentVariable, type=int):
 
 class NPartitions(EnvironmentVariable, type=int):
     """
-    How many partitions to use by default
+    How many partitions to use for a Modin DataFrame (along each axis)
     """
 
     varname = "MODIN_NPARTITIONS"
+    _is_put = False
+
+    @classmethod
+    def put(cls, value):
+        cls._is_put = True
+        return super().put(value)
+
+    @classmethod
+    def is_put(cls):
+        return cls._is_put
 
     @classmethod
     def _get_default(cls):
--- a/modin/config/envvars.py
+++ b/modin/config/envvars.py
@@ -149,16 +149,17 @@ class NPartitions(EnvironmentVariable, type=int):
     """
 
     varname = "MODIN_NPARTITIONS"
-    _is_put = False
+    # This flag is used to detect whether NPartitions is default value or not
+    _is_default = True
 
     @classmethod
     def put(cls, value):
-        cls._is_put = True
+        cls._is_default = False
         return super().put(value)
 
     @classmethod
-    def is_put(cls):
-        return cls._is_put
+    def is_default(cls):
+        return cls._is_default
 
     @classmethod
     def _get_default(cls):
--- a/modin/config/envvars.py
+++ b/modin/config/envvars.py
@@ -150,7 +150,7 @@ class NPartitions(EnvironmentVariable, type=int):
 
     varname = "MODIN_NPARTITIONS"
     # This flag is used to detect whether NPartitions is default value or not
-    _is_default = True
+    _is_default = False
 
     @classmethod
     def put(cls, value):
@@ -158,11 +158,21 @@ class NPartitions(EnvironmentVariable, type=int):
         return super().put(value)
 
     @classmethod
-    def is_default(cls):
-        return cls._is_default
+    def put_if_default(cls, value):
+        """
+        Put specific value if NPartitions wasn't set by a user yet
+
+        Notes
+        -----
+        This method is used to set NPartitions from cluster resources internally
+        and should not be called by a user.
+        """
+        if cls._is_default:
+            cls.put(value)
 
     @classmethod
     def _get_default(cls):
+        cls._is_default = True
         return CpuCount.get()
 
 
--- a/modin/engines/dask/utils.py
+++ b/modin/engines/dask/utils.py
@@ -34,5 +34,6 @@ def initialize_dask():
         )
         client = Client(n_workers=CpuCount.get())
 
-    num_cpus = len(client.ncores())
-    NPartitions.put(num_cpus)
+    if not NPartitions.is_put():
+        num_cpus = len(client.ncores())
+        NPartitions.put(num_cpus)
--- a/modin/engines/dask/utils.py
+++ b/modin/engines/dask/utils.py
@@ -34,6 +34,6 @@ def initialize_dask():
         )
         client = Client(n_workers=CpuCount.get())
 
-    if not NPartitions.is_put():
+    if NPartitions.is_default():
         num_cpus = len(client.ncores())
         NPartitions.put(num_cpus)
--- a/modin/engines/dask/utils.py
+++ b/modin/engines/dask/utils.py
@@ -34,6 +34,5 @@ def initialize_dask():
         )
         client = Client(n_workers=CpuCount.get())
 
-    if NPartitions.is_default():
-        num_cpus = len(client.ncores())
-        NPartitions.put(num_cpus)
+    num_cpus = len(client.ncores())
+    NPartitions.put_if_default(num_cpus)
--- a/modin/engines/dask/utils.py
+++ b/modin/engines/dask/utils.py
@@ -11,7 +11,7 @@
 # ANY KIND, either express or implied. See the License for the specific language
 # governing permissions and limitations under the License.
 
-from modin.config import CpuCount
+from modin.config import CpuCount, NPartitions
 from modin.error_message import ErrorMessage
 
 
@@ -19,7 +19,7 @@ def initialize_dask():
     from distributed.client import get_client
 
     try:
-        get_client()
+        client = get_client()
     except ValueError:
         from distributed import Client
 
@@ -32,4 +32,7 @@ def initialize_dask():
     client = Client()
 """,
         )
-        Client(n_workers=CpuCount.get())
+        client = Client(n_workers=CpuCount.get())
+
+    num_cpus = len(client.ncores())
+    NPartitions.put(num_cpus)
--- a/modin/engines/ray/utils.py
+++ b/modin/engines/ray/utils.py
@@ -166,6 +166,5 @@ def initialize_ray(
 
         ray.worker.global_worker.run_function_on_all_workers(_import_pandas)
 
-    if NPartitions.is_default():
-        num_cpus = int(ray.cluster_resources()["CPU"])
-        NPartitions.put(num_cpus)
+    num_cpus = int(ray.cluster_resources()["CPU"])
+    NPartitions.put_if_default(num_cpus)
--- a/modin/engines/ray/utils.py
+++ b/modin/engines/ray/utils.py
@@ -166,5 +166,6 @@ def initialize_ray(
 
         ray.worker.global_worker.run_function_on_all_workers(_import_pandas)
 
-    num_cpus = int(ray.cluster_resources()["CPU"])
-    NPartitions.put(num_cpus)
+    if not NPartitions.is_put():
+        num_cpus = int(ray.cluster_resources()["CPU"])
+        NPartitions.put(num_cpus)
--- a/modin/engines/ray/utils.py
+++ b/modin/engines/ray/utils.py
@@ -22,6 +22,7 @@ from modin.config import (
     Memory,
     RayPlasmaDir,
     IsOutOfCore,
+    NPartitions,
 )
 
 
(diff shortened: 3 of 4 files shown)


devin-petersohn (dismissed):
Thanks @YarShev, LGTM!

devin-petersohn:
@YarShev Do you know what tests failed? They seem unrelated to the changes.

YarShev:
@devin-petersohn , the changes shouldn't affect the tests but it seems they did somehow. So far, I am not sure of why tests failed.

modin/engines/ray/utils.py
```diff
@@ -164,3 +165,6 @@ def initialize_ray(
         )
 
         ray.worker.global_worker.run_function_on_all_workers(_import_pandas)
+
+    num_cpus = int(ray.cluster_resources()["CPU"])
+    NPartitions.put(num_cpus)
```
anmyachev, line 170: What if `MODIN_CPUS` was be defined?

anmyachev: In cluster case `CpuCount` isn't equal to actual used cores, is it?

YarShev: Yes, it is. In cluster case `CpuCount` would be equal to number of cpus on a head node.

anmyachev: This is contrary to her definition - `How may CPU cores to utilize across the whole distribution`

YarShev: We could update `CpuCount` to actual number of cpus after engine initialization. Then, a user will be able to use `CpuCount.get()` to get the number.

YarShev: However, it is confusing a little because a user can set `MODIN_CPUS` at first working on a cluster and then call `CpuCount.get()` after engine initialization. The numbers will be different.

YarShev: I think, correcting the definition of `CpuCount` will be enough. Then, a user will be able to compute the number of cpus myself via `CpuCount.get() * num_nodes`.

YarShev:
@devin-petersohn , I found out that some of our io tests are failing even without my changes if we change `NPartitions` [here](https://github.com/modin-project/modin/blob/master/modin/pandas/test/test_io.py#L55) to value that is equal or more than 43 (NPartitions >= 43).

YarShev:
@devin-petersohn , I agree, we can mark failing tests as xfail and fix them as a part of separate issue.

vnlitvinov (dismissed):

YarShev:
Marked failing tests as xfail.

modin/config/envvars.py
```diff
@@ -123,7 +123,7 @@ class RayRedisAddress(EnvironmentVariable, type=ExactStr):
 
 class CpuCount(EnvironmentVariable, type=int):
     """
-    How may CPU cores to utilize across the whole distribution
+    How many CPU cores to use when initialization of the Modin engine.
```
amyskov, line 126: `How many CPU cores to use during initialization of the Modin engine.` sounds better.

amyskov (dismissed):

modin/config/envvars.py
amyskov: Ok, lets leave your version.

anmyachev: CPU cores are used not only at the time of initialization, but also in the process of all modin calculations.

It would also be nice to add a note on how to understand this variable in cluster computing.

devin-petersohn (commented):
@YarShev How will this work when users specify their own `NPartitions` values? It seems like we will be overwriting that value.

```python
import modin.pandas as pd
from modin.confix import NPartitions

NPartitions.put(16)
df = pd.read_csv("some.csv")  # will overwrite user-preferred value
```

It seems that we should only overwrite `NPartitions` in cases where the user already has a running Ray or Dask cluster. Does that make sense?

YarShev:
@devin-petersohn , great note! We certainly need to provide a possibility for a user to set the value and not override it. If `NPartitions` is not set, we should initialize it from cluster resources. @vnlitvinov's suggestion looks pretty good so the issue is fixed by following his idea.

vnlitvinov (dismissed):
`NPartitions.is_put` ideally should be a read-only `property` so it's impossible to reset from outside... though defining a property on a class type (instead of class instance) is a funny deal.

Maybe convert it to
```python
_is_put = False
@classmethod
def is_put(cls):
    return cls._is_put
```
to highlight that it's read-only?

YarShev:
Yes, it looks more reliable (albeit bloated a little, I think). Fixed.

modin/engines/dask/utils.py
```diff
@@ -32,4 +32,8 @@ def initialize_dask():
     client = Client()
 """,
         )
-        Client(n_workers=CpuCount.get())
+        client = Client(n_workers=CpuCount.get())
+
+    if not NPartitions.is_put():
+        num_cpus = len(client.ncores())
+        NPartitions.put(num_cpus)
```
vnlitvinov, line 39:
```suggestion
        NPartitions.put(len(client.ncores()))
```

modin/engines/ray/utils.py
```diff
@@ -164,3 +165,7 @@ def initialize_ray(
         )
 
         ray.worker.global_worker.run_function_on_all_workers(_import_pandas)
+
+    if not NPartitions.is_put():
+        num_cpus = int(ray.cluster_resources()["CPU"])
+        NPartitions.put(num_cpus)
```
vnlitvinov, line 171:
```suggestion
        NPartitions.put(int(ray.cluster_resources()["CPU"]))
```

vnlitvinov (dismissed):
We could save two LoC if we want 😆

YarShev:
I don't think it is critical issue. I suggest leaving as it is.

devin-petersohn:
@vnlitvinov For testing and debugging purposes I often set and reset the number of partitions. Part of the appeal of Modin's architecture is the flexibility in partitioning so making it read-only would be counter to this.

I am imagining at some point in the future that the system can change the global number of partitions on the fly, like when new compute nodes become available or nodes die/go away. Would we still have this flexibility with your approach?

YarShev:
@devin-petersohn , we can set and reset the number of partitions as it was before. The PR is adding a separate class method `is_put`, which is read-only, to handle initialization of the number of partitions correctly.

YarShev:
@devin-petersohn , yes, you are right, though. It won't work for the following case:
```python
import modin.pandas as pd
import numpy as np
NPartitions.put(10)
data = np.random.randint(0, 100, size=(2 ** 10, 2 ** 10))
df1 = pd.DataFrame(data)
NPartitions.put(11)
df2 = pd.DataFrame(data) # the number of partitions will be 10
```

vnlitvinov:
> @vnlitvinov For testing and debugging purposes I often set and reset the number of partitions. Part of the appeal of Modin's architecture is the flexibility in partitioning so making it read-only would be counter to this.

I never suggested to make the _variable_ read-only, only the field `is_put` to highlight that it isn't expected to be modified outside of `NPartitions` class.

YarShev:
> I am imagining at some point in the future that the system can change the global number of partitions on the fly, like when new compute nodes become available or nodes die/go away. Would we still have this flexibility with your approach?

However, if a user doesn't touch `NPartitions` in any way and new compute nodes become available or nodes die/go away in the middle of a workflow, the number of partitions will be equal to its initial value. An example:

```python
# Let's say that we have 3 nodes with 16 cores on each one
import modin.pandas as pd
df = pd.DataFrame(frame_data) # NPartitions is equal to 48
...
# one node dies unexpectedly
df2 = pd.DataFrame(frame_data2) # NPartitions is equal to 48, not 32. A user should set NPartitions himself
```

modin/config/envvars.py
```diff
@@ -145,10 +145,20 @@ class Memory(EnvironmentVariable, type=int):
...
     varname = "MODIN_NPARTITIONS"
+    _is_put = False
+
+    @classmethod
+    def put(cls, value):
+        cls._is_put = True
+        return super().put(value)
+
+    @classmethod
+    def is_put(cls):
```
devin-petersohn, line 160:
```suggestion
    def is_default(cls):
```

modin/config/envvars.py
```diff
@@ -145,10 +145,20 @@ class Memory(EnvironmentVariable, type=int):
...
-    How many partitions to use by default
+    How many partitions to use for a Modin DataFrame (along each axis)
     """
 
     varname = "MODIN_NPARTITIONS"
+    _is_put = False
+
+    @classmethod
+    def put(cls, value):
+        cls._is_put = True
```
devin-petersohn, line 156:
```suggestion
        cls._is_default = False
```

modin/config/envvars.py
```diff
@@ -145,10 +145,20 @@ class Memory(EnvironmentVariable, type=int):
 
 class NPartitions(EnvironmentVariable, type=int):
     """
-    How many partitions to use by default
+    How many partitions to use for a Modin DataFrame (along each axis)
     """
 
     varname = "MODIN_NPARTITIONS"
+    _is_put = False
```
devin-petersohn, line 152:
```suggestion
    _is_default = True
```

devin-petersohn: Can we add a single line comment above this too for clarifying its purpose?

devin-petersohn (commented):
@YarShev @vnlitvinov I understand now. Rewording the variable will clear up the confusion on it for me (and future me 😄 ).

modin/config/envvars.py
YarShev: Why did you change the value to `False`?

YarShev: Oh, I see.

modin/config/envvars.py
YarShev: Added.

modin/config/envvars.py
```diff
@@ -145,10 +145,21 @@ class Memory(EnvironmentVariable, type=int):
...
+    _is_default = True
+
+    @classmethod
+    def put(cls, value):
+        cls._is_default = False
+        return super().put(value)
+
+    @classmethod
+    def is_default(cls):
+        return cls._is_default
```
vnlitvinov, line 162: it came to my mind that there's probably even more elegant solution:
```suggestion
    @classmethod
    def put_if_default(cls, value):
        """Put specific value if NPartitions wasn't set by user yet"""
        if cls._is_default:
            cls.put(value)
```

and then you do stuff like `NPartitions.put_if_default(len(client.cores()))`

modin/config/envvars.py
```diff
@@ -145,10 +145,21 @@ class Memory(EnvironmentVariable, type=int):
 
 class NPartitions(EnvironmentVariable, type=int):
     """
-    How many partitions to use by default
+    How many partitions to use for a Modin DataFrame (along each axis)
     """
 
     varname = "MODIN_NPARTITIONS"
+    # This flag is used to detect whether NPartitions is default value or not
+    _is_default = True
```
vnlitvinov, line 153: I would say this should have initial value of `False`:
```suggestion
    _is_default = False
```
and be reset to `True` only in `_get_default()` method, otherwise it would be impossible for a user to override the cluster-defined value by setting an environment variable.

vnlitvinov (changes requested):

modin/config/envvars.py
YarShev: Yes, you are right but in order to get all cases working we need to reset `_is_default` to `False` in `put` because the following case might be possible:
```python
import modin.pandas as pd
df = pd.DataFrame(frame_data) # NPartitions will be caught from cluster resources 
NPartitions.put(some_value)
df2 = pd.DataFrame(frame_data2) # NPartitions will still be caught from cluster resources 
```
Thanks for notice!

modin/config/envvars.py
YarShev: Fixed.

vnlitvinov (approved):

modin/pandas/test/test_io.py
```diff
@@ -1836,6 +1845,10 @@ def test_to_stata(self):
 
 
 class TestFeather:
+    @pytest.mark.skipif(
```
devin-petersohn, line 1848:
```suggestion
    @pytest.mark.xfail(
```

devin-petersohn (changes requested):
Looks great, one minor change for `xfail` instead of `skip`. `xfail` can take a condition I believe.

modin/pandas/test/test_io.py
YarShev: It seems `xfail` couldn't take a condition in former times. Applied.

devin-petersohn: Yes, it's a good addition. We should change all of our `skipif` to `xfail` in the test suite.

YarShev: That is what you want to resolve as a part of #2846 ?

devin-petersohn: We can do it there, yes.

YarShev:
Something went wrong with TeamCity tests.

YarShev:
Rebased on the latest master to restart the tests.

devin-petersohn (approved):
LGTM, thanks @YarShev!

PR #2835, opened 2021-03-06
Fix Pickle support for DataFrame and Series

### LINKED ISSUE #672: Modin objects are not able to be pickled
```
File ".../site-packages/cache_memoize/__init__.py", line 116, in inner
    result = cache.get(cache_key, MARKER)
  File ".../site-packages/django_redis/cache.py", line 32, in _decorator
    return method(self, *args, **kwargs)
  File ".../site-packages/django_redis/cache.py", line 81, in get
    client=client)
  File ".../site-packages/django_redis/client/default.py", line 210, in get
    return self.decode(value)
  File ".../site-packages/django_redis/client/default.py", line 318, in decode
    value = self._serializer.loads(value)
  File ".../site-packages/django_redis/serializers/pickle.py", line 35, in loads
    return pickle.loads(value)
  File ".../site-packages/modin/pandas/base.py", line 3337, in __getattribute__
    if self.empty and is_callable:
  File ".../site-packages/modin/pandas/base.py", line 3343, in __getattribute__
    return object.__getattribute__(self, item)
  File ".../site-packages/modin/pandas/dataframe.py", line 211, in empty
    return len(self.columns) == 0 or len(self.index) == 0

  File ".../site-packages/modin/pandas/dataframe.py", line 131, in _get_columns
    return self._query_compiler.columns
  File ".../site-packages/modin/pandas/dataframe.py", line 1898, in __getattr__
    if key in self.columns:
  File ".../site-packages/modin/pandas/base.py", line 3343, in __getattribute__
    return object.__getattribute__(self, item)

#  ... So many times!

  File ".../site-packages/modin/pandas/dataframe.py", line 131, in _get_columns
    return self._query_compiler.columns
  File ".../site-packages/modin/pandas/base.py", line 3343, in __getattribute__
    return object.__getattribute__(self, item)
RecursionError: maximum recursion depth exceeded while calling a Python object
```
 I use django-cache to cache the result of modin df, When django want to fetch the cache,It hit this issue.

vnlitvinov:
## What do these changes do?
* Fix Pickle support for Modin `DataFrame` and `Series` by implementing `__reduce__` protocol (ref: [Python docs](https://docs.python.org/3/library/pickle.html#object.__reduce__))
* Add new protected method `_materialize` to Modin Frame which makes sure all deferred functions for each partition are executed
* Add configuration variable `PersistentPickle` that controls how objects are pickled (backed by `MODIN_PERSISTENT_PICKLE` environment variable): in light mode we store all internals (like partitions), which should be faster for e.g. huge distributed DataFrames (we'll store just arrays of `Ray.ObjectIDs` instead of full data), in full mode it is possible to save pickled objects to disk and restore them in other Modin session but it is slower because we pull in all the data to the driver process
* Extract the construction of the set of default attributes to a global variable - slightly speeds up `__getattribute__` _(unrelated to the PR but small enough so I hope to make it through)_

- [x] passes `flake8 modin`
- [x] Resolves #672 (and potentially some more)
- [x] tests added and passing

### DIFF
--- a/modin/backends/pandas/query_compiler.py
+++ b/modin/backends/pandas/query_compiler.py
@@ -201,6 +201,9 @@ class PandasQueryCompiler(BaseQueryCompiler):
         else:
             return result
 
+    def finalize(self):
+        self._modin_frame.finalize()
+
     def to_pandas(self):
         return self._modin_frame.to_pandas()
 
--- a/modin/pandas/dataframe.py
+++ b/modin/pandas/dataframe.py
@@ -2425,6 +2425,7 @@ class DataFrame(BasePandasDataset):
         return cls(query_compiler=query_compiler)
 
     def __reduce__(self):
+        self._query_compiler._modin_frame._materialize()
         return self._inflate, (self._query_compiler,)
 
 
--- a/modin/pandas/dataframe.py
+++ b/modin/pandas/dataframe.py
@@ -2094,9 +2094,6 @@ class DataFrame(BasePandasDataset):
     def __round__(self, decimals=0):
         return self._default_to_pandas(pandas.DataFrame.__round__, decimals=decimals)
 
-    def __setstate__(self, state):
-        return self._default_to_pandas(pandas.DataFrame.__setstate__, state)
-
     def __delitem__(self, key):
         if key not in self:
             raise KeyError(key)
@@ -2421,16 +2418,14 @@ class DataFrame(BasePandasDataset):
     def _inflate_light(cls, query_compiler):
         """
         Re-creates the object from previously-serialized lightweight representation.
+
         The method is used for faster but not disk-storable persistence.
         """
         return cls(query_compiler=query_compiler)
 
     @classmethod
     def _inflate_full(cls, pandas_df):
-        """
-        Re-creates the object from previously-serialized disk-storable
-        representation.
-        """
+        """Re-creates the object from previously-serialized disk-storable representation."""
         return cls(data=from_pandas(pandas_df))
 
     def __reduce__(self):
--- a/modin/pandas/dataframe.py
+++ b/modin/pandas/dataframe.py
@@ -2429,7 +2429,7 @@ class DataFrame(BasePandasDataset):
         return cls(data=from_pandas(pandas_df))
 
     def __reduce__(self):
-        self._query_compiler._modin_frame._materialize()
+        self._query_compiler.finalize()
         if PersistentPickle.get():
             return self._inflate_full, (self._to_pandas(),)
         return self._inflate_light, (self._query_compiler,)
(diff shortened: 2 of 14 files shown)


modin/pandas/series.py
```diff
@@ -1734,6 +1734,20 @@ def _getitem(self, key):
...
+    @classmethod
+    def _inflate_light(cls, query_compiler, name):
+        return cls(query_compiler=query_compiler, name=name)
+
+    @classmethod
+    def _inflate_full(cls, pandas_series):
+        return cls(data=pandas_series)
+
+    def __reduce__(self):
+        self._query_compiler._modin_frame._materialize()
```
devin-petersohn, line 1746: I think this leaks too much into the API layer.

devin-petersohn (commented):
Can we avoid accessing deep internals from the API layer?

modin/pandas/series.py
vnlitvinov: I've renamed the method to `.finalize()` and added the same method to query compiler level, so API level does not directly access the frame level anymore.

modin/backends/pandas/query_compiler.py
```diff
@@ -201,6 +201,9 @@ def default_to_pandas(self, pandas_op, *args, **kwargs):
         else:
             return result
 
+    def finalize(self):
+        self._modin_frame.finalize()
```
anmyachev, line 205: What about other backends? (`pyarrow`, `omnisci`) Should we override for these?

vnlitvinov: We should override it for every backend which defers operations, so most likely yes.

modin/pandas/dataframe.py
```diff
@@ -2416,6 +2413,27 @@ def _getitem(self, key):
         else:
             return self._getitem_column(key)
 
+    # persistance support
```
anmyachev, line 2416: Maybe this comment should be above `def _inflate_full` function, not `def _inflate_light`?

vnlitvinov: this comment denotes the place where persistance support methods start... I should probably make it clearer.

vnlitvinov: I tried to make it clearer by rewording the comments.

modin/backends/pandas/query_compiler.py
vnlitvinov: Do note that for OmniSci it is already overridden (with `raise NotImplementedError()` stub, cc @ienkovich), and `pyarrow` inherits the one from Pandas so it doesn't seem to be necessary for now.

devin-petersohn (approved):
@vnlitvinov LGTM! Thanks

vnlitvinov:
@devin-petersohn CI is green but code coverage (which is weird at best as of late anyway).

devin-petersohn:
Awaiting approval from @modin-project/modin-omnisci since there were code changes there as well.

ienkovich (approved):
LGTM!

PR #2843, opened 2021-03-09
FEAT-#2842: add dockerfile for omnisci engine development

### LINKED ISSUE #2842: Create docker file for development omnisci engine
This engine is available only on Linux, to simplify development on other platforms, it is proposed to use docker.

anmyachev:

## What do these changes do?

- [ ] passes `flake8 modin`
- [x] Resolves #2842 
- [ ] tests added and passing

### DIFF
--- a/docker/omnisci-engine-dev.dockerfile
+++ b/docker/omnisci-engine-dev.dockerfile
@@ -1,0 +1,70 @@
+# Licensed to Modin Development Team under one or more contributor license agreements.
+# See the NOTICE file distributed with this work for additional information regarding
+# copyright ownership.  The Modin Development Team licenses this file to you under the
+# Apache License, Version 2.0 (the "License"); you may not use this file except in
+# compliance with the License.  You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software distributed under
+# the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific language
+# governing permissions and limitations under the License.
+
+FROM ubuntu:18.04
+ENV http_proxy ${http_proxy}
+ENV https_proxy ${https_proxy}
+
+RUN apt-get update --yes \
+    && apt-get install wget git --yes \
+    #
+    # cleanup
+    && apt-get autoremove -y \
+    && apt-get clean -y \
+    && rm -rf /var/lib/apt/lists/*
+
+ENV USER modin
+ENV UID 1000
+ENV HOME /home/$USER
+
+RUN adduser --disabled-password \
+    --gecos "Non-root user" \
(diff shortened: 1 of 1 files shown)


anmyachev:
Related to https://discuss.modin.org/t/running-omnisci-tests-locally/187

devin-petersohn:
@anmyachev Restarting CI, please let me know when it passes.

anmyachev:
> @anmyachev Restarting CI, please let me know when it passes.

Thanks @devin-petersohn. CI passed.

docker/omnisci-engine-dev.dockerfile
```diff
@@ -0,0 +1,70 @@
...
+ARG modin_dir="${HOME}/modin"
+
+# Clone modin repo
+RUN mkdir "$modin_dir" \
+    && git clone "https://github.com/$gh_username/modin.git" "$modin_dir" \
+    && cd "$modin_dir" \
+    && git remote add upstream "https://github.com/modin-project/modin.git"
+
+# install modin dependencies
+RUN conda env create -n modin -f "$modin_dir/requirements/env_omnisci.yml"
```
RehanSD, line 57: Do we want to specify a python version here? We can make it an `ARG` so people can specify it too. At the very least, we want to specify a version so that we can use Ray, since Ray does not support 3.9, and this will default to the latest version.

RehanSD (changes requested):
Thank you for the contribution! Looks good to me, besides the python arg. The error I flagged earlier was because my docker settings reset so I wasn't giving the container enough memory.

docker/omnisci-engine-dev.dockerfile
```diff
@@ -0,0 +1,70 @@
...
+
+ENV USER modin
+ENV UID 1000
+ENV HOME /home/$USER
+
+RUN adduser --disabled-password \
+    --gecos "Non-root user" \
+    --uid $UID \
+    --home $HOME \
+    $USER
```
RehanSD, line 34: Why do we create another user if we do not use it?

docker/omnisci-engine-dev.dockerfile
anmyachev: @RehanSD  thanks for review! At the moment, `omnisci engine` only supports `python 3.7`. In this case, I think this is not necessary.

docker/omnisci-engine-dev.dockerfile
anmyachev: Good catch! I think for development container, we can use default user.

docker/omnisci-engine-dev.dockerfile
RehanSD: Makes sense, thank you!

anmyachev:
@RehanSD can we merge the PR?

RehanSD (approved):
Looks good to me - the only concern is that I get a warning `mesg: ttyname failed: inappropriate ioctl for device` while building the container. We're holding off on merging PRs till after release 0.9.1, which should happen later today!

PR #2851, opened 2021-03-11
FIX-#2844: Check for both types of ray.ObjectRef

### LINKED ISSUE #2844: Fix defensive assertions for Ray Client users
Related to https://github.com/ray-project/ray/issues/14042

For now, we will check for both types.

devin-petersohn:
Resolves #2844

## What do these changes do?

- [x] passes `flake8 modin`
- [x] Resolves #2844 
- [ ] tests added and passing

### DIFF
--- a/modin/engines/ray/pandas_on_ray/frame/partition.py
+++ b/modin/engines/ray/pandas_on_ray/frame/partition.py
@@ -20,11 +20,12 @@ from modin.engines.ray.utils import handle_ray_task_error
 import ray
 from ray.worker import RayTaskError
 from ray.services import get_node_ip_address
+from ray.util.client.common import ClientObjectRef
 
 
 class PandasOnRayFramePartition(BaseFramePartition):
     def __init__(self, object_id, length=None, width=None, ip=None, call_queue=None):
-        assert type(object_id) is ray.ObjectRef
+        assert isinstance(object_id, (ray.ObjectRef, ClientObjectRef))
 
         self.oid = object_id
         if call_queue is None:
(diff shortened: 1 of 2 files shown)


modin/engines/ray/pandas_on_ray/frame/partition.py
```diff
@@ -20,11 +20,18 @@
 import ray
 from ray.worker import RayTaskError
 from ray.services import get_node_ip_address
+from packaging import version
+ObjectIDType = ray.ObjectRef
```
RehanSD, line 24: Is this a good name @devin-petersohn, or would you prefer something like `object_id_cls`?

modin/engines/ray/pandas_on_ray/frame/partition.py
```diff
@@ -206,7 +213,7 @@ def ip(self):
                 self.drain_call_queue()
             else:
                 self._ip_cache = self.apply(lambda df: df)._ip_cache
-        if isinstance(self._ip_cache, ray.ObjectID):
+        if isinstance(self._ip_cache, (ray.ObjectID, ClientObjectRef)):
```
RehanSD, line 216:
```suggestion
        if isinstance(self._ip_cache, ObjectIDType):
```

williamma12 (commented):
This fix by itself looks good to me. 

However, would it be cleaner if we made `ObjectIDType` a class variable instead? Every engine would have to know its own `ObjectIDType` so that can add some more consistency between the different children of `BaseFramePartition`

richardliaw:
hey @devin-petersohn is there anything blocking this PR?

devin-petersohn:
@richardliaw Sorry for the delay. I don't think there are any specific blockers here.

@williamma12 I think it would be fine to merge this and add that minor refactoring as a future step. This is still consistent and correct at least. We will need to add some testing for Ray client, which we don't have set up yet.

williamma12 (approved):
Sounds good! Let's plan to refactor this soon

PR #2860, opened 2021-03-11
FIX-#2859: Fix metadata calculation on reduce operations

### LINKED ISSUE #2859: dtype computation on reduce operations can be incorrect
This is an issue where dtypes are carried through in reductions, which is incorrect.

```python
# on any df
df.isna().count().dtypes  # should be `int64`, but is `bool` 
```

Caused by lines: https://github.com/modin-project/modin/blame/master/modin/engines/base/frame/data.py#L1093-L1098

The dtypes are being carried through, but shouldn't be. This is caused by the `preserve_index` logic, which I couldn't find being used internally.

devin-petersohn:
Resolves #2859

## What do these changes do?

- [x] passes `flake8 modin`
- [x] Resolves #2859 
- [ ] tests added and passing

### DIFF
--- a/modin/engines/base/frame/data.py
+++ b/modin/engines/base/frame/data.py
@@ -1056,12 +1056,17 @@ class BasePandasFrame(object):
                 # axis=1 does not have this requirement because the index already will
                 # line up with the index of the data based on how pandas creates a
                 # DataFrame from a Series.
-                return pandas.DataFrame(series_result).T
-            return pandas.DataFrame(series_result)
+                result = pandas.DataFrame(series_result).T
+                result.index = ["__reduced__"]
+            else:
+                result = pandas.DataFrame(series_result)
+                if isinstance(series_result, pandas.Series):
+                    result.columns = ["__reduced__"]
+            return result
 
         return _map_reduce_func
 
-    def _compute_map_reduce_metadata(self, axis, new_parts, preserve_index=True):
+    def _compute_map_reduce_metadata(self, axis, new_parts):
         """
         Compute the metadata for the result of reduce function.
 
@@ -1071,8 +1076,6 @@ class BasePandasFrame(object):
             The axis on which reduce function was applied
         new_parts: numpy 2D array
             Partitions with the result of applied function
-        preserve_index: boolean
-            The flag to preserve labels for the reduced axis.
 
         Returns
         -------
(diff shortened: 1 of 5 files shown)


modin/engines/base/frame/data.py
```diff
@@ -1082,32 +1085,21 @@ def _compute_map_reduce_metadata(self, axis, new_parts, preserve_index=True):
...
-        else:
-            new_dtypes = None
+        new_dtypes = None
         result = self.__constructor__(
             new_parts,
             *new_axes,
             *new_axes_lengths,
             new_dtypes,
         )
-        result._apply_index_objs(axis)
```
RehanSD, line 1107: Do we no longer need this because we don't need to explicitly preserve the index?

modin/engines/base/frame/data.py
```diff
@@ -1056,12 +1056,17 @@ def _map_reduce_func(df, *args, **kwargs):
                 # axis=1 does not have this requirement because the index already will
                 # line up with the index of the data based on how pandas creates a
                 # DataFrame from a Series.
-                return pandas.DataFrame(series_result).T
-            return pandas.DataFrame(series_result)
+                result = pandas.DataFrame(series_result).T
+                result.index = ["__reduced__"]
+            else:
+                result = pandas.DataFrame(series_result)
+                if isinstance(series_result, pandas.Series):
```
RehanSD, line 1063: Is this for the case where the map reduce doesn't fully reduce down to a single value per row? (since this is a column-wise reduction).

RehanSD (commented):
Looks good to me - just left some comments for my personal clarification. Based off of the changes, it seems that the default behavior is the `preserve_index` behavior. I just wanted to confirm that that's the case.

modin/engines/base/frame/data.py
devin-petersohn: Row/column Index is always preserved for these functionalities.

modin/engines/base/frame/data.py
devin-petersohn: Yes, the result does not have a name, so we give it a name that is shared externally and internally.

RehanSD: Is there a case where series_result is not a pandas.Series? What would it be otherwise?

devin-petersohn: It might not always be a `Series`, it could be a dataframe. We always convert it to a dataframe here in order to treat them correctly because `Series` will not always have a name.

RehanSD (approved):
Looks good to me!

PR #2862, opened 2021-03-12
FIX-#2857: Correctly handle identical index binary operations

### LINKED ISSUE #2857: modin fails to correctly deliver division results in dataframe rows
### System information
- **Modin version** (`modin.__version__`):  0.9.0
- **Python version**: 3.7.2
- **Code we can use to reproduce**: `new['effratio'] = new['efflen']/new['txtlen']`

### Describe the problem
If you directly use rows in a DataFrame to do division, the result is not correct. I discovered this when I want to calculate a ratio. Some ratio values are larger than 1. I ran the same code with original pandas and it delivered the correct result.

I have listed several items in this dataframe for you to see the problem. See the image.

`new` is a DataFrame object I read in.

issue

devin-petersohn:
Resolves #2857

## What do these changes do?

- [x] passes `flake8 modin`
- [x] Resolves #2857 
- [x] tests added and passing

### DIFF
--- a/modin/engines/base/frame/data.py
+++ b/modin/engines/base/frame/data.py
@@ -973,7 +973,7 @@ class BasePandasFrame(object):
                 return left_index.join(right_index, how=how, sort=sort)
 
         # define condition for joining indexes
-        all_indices_equal = all(indexes[0].equals(index) for index in [indexes[1:]])
+        all_indices_equal = all(indexes[0].equals(index) for index in indexes[1:])
         do_join_index = how is not None and not all_indices_equal
 
         # define condition for joining indexes with getting indexers
@@ -1783,8 +1783,6 @@ class BasePandasFrame(object):
             return self._partitions, [o._partitions for o in other], joined_index
 
         base_frame_idx = non_empty_frames_idx[0]
-        base_frame = frames[base_frame_idx]
-
         other_frames = frames[base_frame_idx + 1 :]
 
         # Picking first non-empty frame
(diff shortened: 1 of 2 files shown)


modin/engines/base/frame/data.py
```diff
@@ -1783,8 +1783,6 @@ def get_axis_lengths(partitions, axis):
             return self._partitions, [o._partitions for o in other], joined_index
 
         base_frame_idx = non_empty_frames_idx[0]
```
RehanSD, line 1785: Why do we define this variable if we only use it in line 1786? Why not use it in line 1789 as well?

modin/engines/base/frame/data.py
```diff
@@ -973,7 +973,7 @@ def merge(left_index, right_index):
                 return left_index.join(right_index, how=how, sort=sort)
 
         # define condition for joining indexes
-        all_indices_equal = all(indexes[0].equals(index) for index in [indexes[1:]])
+        all_indices_equal = all(indexes[0].equals(index) for index in indexes[1:])
```
RehanSD, line 976: I'm not sure I understand what these changes do - before were we making a 2D list with one row that contains indices, but now we're just looping through the list of indices?

RehanSD (commented):
Left some comments for clarification.

modin/engines/base/frame/data.py
devin-petersohn: Yes, it was a bug. Deleting these two characters fixes the case in the issue.

modin/engines/base/frame/data.py
devin-petersohn: The code quality in this helper method is in need of some TLC, I do not disagree. Let's put that later.

RehanSD (approved):
Looks good to me!

PR #2868, opened 2021-03-15
TEST-#2867: cast all env variables in lowercase for ASV benchmarking

### LINKED ISSUE #2867: Benchmarks aren't trigger computations in case when MODIN_ENGINE=ray
This appears in case when `from modin.config import TestDatasetSize, AsvImplementation, Engine` ends with `ImportError` and user directly define environment variable as `export MODIN_ENGINE=ray`. But the rest of the code relies on the form where the first character is in uppercase.

anmyachev:

## What do these changes do?

- [x] passes `flake8 modin`
- [x] Resolves #2867 
- [x] tests added and passing


devin-petersohn (approved):
LGTM

PR #2870, opened 2021-03-15
FIX-#2869: Fix setting NPartitions via put

### LINKED ISSUE #2869: Fix setting NPartitions via put
### Describe the problem
We need to reset `is_default` to `False` after base `put` call because we may meet `_get_default` call inside of base `put` (if `MODIN_NPARTITIONS` wasn't set) that can change `is_default` to `True`.

### Source code / logs
```python
import modin.pandas as pd
import numpy as np
NPartitions.put(1)
df = pd.DataFrame(np.arange(2 ** 10))
df._query_compiler._modin_frame._partitions.shape[0] == 1
False
```

YarShev:

## What do these changes do?

- [x] passes `flake8 modin`
- [x] Resolves #2869  
- [x] tests added and passing

### DIFF
--- a/modin/config/envvars.py
+++ b/modin/config/envvars.py
@@ -154,8 +154,8 @@ class NPartitions(EnvironmentVariable, type=int):
 
     @classmethod
     def put(cls, value):
+        super().put(value)
         cls._is_default = False
-        return super().put(value)
 
     @classmethod
     def put_if_default(cls, value):


devin-petersohn (approved):
Thanks @YarShev LGTM!

YarShev:
@devin-petersohn , I reran the tests because io tests hung for some reason.

devin-petersohn:
@YarShev Thank you, I could not get them to rerun.

Windows test can sometimes hang for 6 hours+. Do we have an open issue for that? It's hard to identify why it happens, sometimes it just does, but setting a lower timeout would help a lot.

YarShev:
Yes, I see Windows tests hanging not at first time as well. We don't have an open issue for that. Probably, yes, the timeout would be good solution.

devin-petersohn:
@YarShev tests are hanging again. I will restart once they are shut down.

YarShev:
I have restarted the tests 😄  Will see how it goes at this time.

YarShev:
@devin-petersohn , tests passed.

PR #2873, opened 2021-03-15
Experimental SQL support via dfsql

btseytlin:
## What do these changes do?

Adds the ability to query Modin DataFrames with SQL statements.

How it works:
1. Install modin with [dfsql](https://github.com/mindsdb/dfsql).
``pip install modin[sql]``
2. Query using the `sql_query` interface:
```
>>> import modin.pandas as pd
>>> import modin.experimental.sql as mdsql
>>> df = pd.DataFrame({'col1': [1, 2, 3]})
>>> df.head()
   col1
0     1
1     2
2     3
>>> mdsql.sql_query("SELECT col1 FROM df WHERE col1 = 1", from_tables={'df': df})
1
``` 
3. Alternatively, use the added `.sql` DataFrame accessor. It's registered when `modin.experimental.sql` is imported. In this case the `FROM` can be omitted:
```
>>> df.sql("SELECT col1 WHERE col1 = 1")
UserWarning: `from_dict` defaulting to pandas implementation.
1
```

- [x] passes `flake8 modin`
- [x] Resolves #? 
- [x] tests added and passing

### DIFF
--- a/modin/experimental/sql/__init__.py
+++ b/modin/experimental/sql/__init__.py
@@ -1,0 +1,26 @@
+# Licensed to Modin Development Team under one or more contributor license agreements.
+# See the NOTICE file distributed with this work for additional information regarding
+# copyright ownership.  The Modin Development Team licenses this file to you under the
+# Apache License, Version 2.0 (the "License"); you may not use this file except in
+# compliance with the License.  You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software distributed under
+# the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific language
+# governing permissions and limitations under the License.
+
+import warnings
+
+try:
+    from dfsql import sql_query as query
+    import dfsql.extensions  # noqa: F401
+except ImportError:
+    warnings.warn(
+        "Modin experimental sql interface requires dfsql to be installed."
+        " Run `pip install modin[sql]` to install it."
+    )
+    raise
+
+__all__ = ["query"]
--- a/setup.py
+++ b/setup.py
@@ -41,8 +41,12 @@ dask_deps = ["dask>=2.12.0,<=2.19.0", "distributed>=2.12.0,<=2.19.0"]
 ray_deps = ["ray>=1.0.0,<1.2.0", "pyarrow==1.0"]
 remote_deps = ["rpyc==4.1.5", "cloudpickle==1.4.1", "boto3==1.4.8"]
 spreadsheet_deps = ["modin-spreadsheet>=0.1.0"]
-
+sql_deps = ["dfsql>=0.2.6"]
 all_deps = dask_deps + ray_deps + remote_deps + spreadsheet_deps
+
+# dfsql does not support Windows yet
+if os.name != 'nt':
+    all_deps += sql_deps
 
 setup(
     name="modin",
@@ -63,6 +67,7 @@ setup(
         "ray": ray_deps,
         "remote": remote_deps,
         "spreadsheet": spreadsheet_deps,
+        "sql": sql_deps,
         "all": all_deps,
     },
     python_requires=">=3.7.1",
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -415,6 +415,8 @@ jobs:
       - shell: bash -l {0}
         run: python -m pytest modin/experimental/pandas/test/test_io_exp.py
       - shell: bash -l {0}
+        run: pip install dfsql && pytest modin/experimental/sql/test/test_sql.py
+      - shell: bash -l {0}
         run: bash <(curl -s https://codecov.io/bash)
 
   test-experimental:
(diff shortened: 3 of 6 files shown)


modin/experimental/sql/__init__.py
```diff
@@ -0,0 +1,27 @@
...
+# Unless required by applicable law or agreed to in writing, software distributed under
+# the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific language
+# governing permissions and limitations under the License.
+
+import warnings
+
+try:
+    import dfsql
+    from dfsql import sql_query
```
devin-petersohn, line 18:
```suggestion
    from dfsql import sql_query as query
```

modin/experimental/sql/__init__.py
```diff
@@ -0,0 +1,27 @@
...
+    import dfsql.extensions
+except ImportError:
+    warnings.warn(
+        "Modin experimental sql interface requires dfsql to be installed."
+        " Run `pip install modin[sql]` to install it."
+    )
+    raise
+
+__all__ = ["dfsql", "sql_query"]
```
devin-petersohn, line 27:
```suggestion
__all__ = ["query"]
```

modin/experimental/sql/__init__.py
```diff
@@ -0,0 +1,27 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software distributed under
+# the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific language
+# governing permissions and limitations under the License.
+
+import warnings
+
+try:
+    import dfsql
```
devin-petersohn, line 17: Will users use this import? 
```suggestion
```

modin/experimental/sql/__init__.py
```diff
@@ -0,0 +1,27 @@
...
+    from dfsql import sql_query
+    import dfsql.extensions
+except ImportError:
+    warnings.warn(
```
devin-petersohn, line 21: I prefer to throw our own error message and not rethrow the `ImportError`. I think it might give too many error messages if we add an extra warning.

setup.py
```diff
@@ -41,8 +41,8 @@ def is_pure(self):
 ray_deps = ["ray>=1.0.0,<1.2.0", "pyarrow==1.0"]
 remote_deps = ["rpyc==4.1.5", "cloudpickle==1.4.1", "boto3==1.4.8"]
 spreadsheet_deps = ["modin-spreadsheet>=0.1.0"]
-
-all_deps = dask_deps + ray_deps + remote_deps + spreadsheet_deps
+sql_deps = ["dfsql==0.2.2"]
```
devin-petersohn, line 44: How quickly will this evolve? Is pinning to 0.2.2 a good idea or should we have a `>=` so we don't need to bump every release?

devin-petersohn (changes requested):
I created an issue for you to link in your PR: https://github.com/modin-project/modin/issues/2874. The commit message will need to have it as well. Something like:

```
FEAT-#2874: Experimental SQL support via dfsql
```
We should also add some sanity tests.

I can help you finish this up if you prefer.

modin/experimental/sql/__init__.py
btseytlin: I would rather leave `dfsql` exposed in case users want to use the full functonality of dfsql, like persisting tables

setup.py
btseytlin: I think you are right, should be a good idea to use `>=`

modin/experimental/sql/__init__.py
devin-petersohn: Would you prefer `from modin.experimental.sql import dfsql` or `import dfsql`?

btseytlin: Alright, it makes sense

modin/experimental/sql/__init__.py
btseytlin: I haven't found a fitting custom exception class in Modin. What should I use to throw the error?

btseytlin:
> I created an issue for you to link in your PR: #2874. The commit message will need to have it as well. Something like:
> 
> ```
> FEAT-#2874: Experimental SQL support via dfsql
> ```
> 
> We should also add some sanity tests.
> 
> I can help you finish this up if you prefer.

Added test, squashed commits and set the commit message

modin/experimental/sql/__init__.py
devin-petersohn: You can just raise an `ImportError`.

devin-petersohn (commented):
Looks good. Left one comment, let me know if it makes sense!

modin/experimental/sql/__init__.py
btseytlin: Maybe remove the `try except` altogether then? Not very Pythonic to override an `ImportError` with custom text, if I understood your meaning correctly

devin-petersohn: Let's just leave it as you have it. I think it will be ok.

devin-petersohn (approved):
@btseytlin LGTM, thanks!

PR #2877, opened 2021-03-15
FIX-#2881: Move to new private Ray API

### LINKED ISSUE #2881: Ray APIs moved to private subpackage on latest master, breaking compatibility
### Describe the problem

Ray [recently moved](https://github.com/ray-project/ray/pull/14607) some of their internal APIs to an explicit `_private` subpackage. Modin calls one of these functions (`ray.utils.get_system_memory()`), leading to errors on tests that run on Ray's latest master.

### Source code / logs
[Example CI logs with failing tests](https://github.com/ray-project/xgboost_ray/pull/66/checks?check_run_id=2114689375)

```
                    object_store_memory = int(
>                       0.6 * ray.utils.get_system_memory() // 10 ** 9 * 10 ** 9
                    )
E                   AttributeError: module 'ray' has no attribute 'utils'

/opt/hostedtoolcache/Python/3.6.9/x64/lib/python3.6/site-packages/modin/engines/ray/utils.py:132: AttributeError
```

krfricke:
## What do these changes do?

Ray recently moved utility functions to a private subpackage on latest master. This PR addresses these API changes and makes sure to stay compatible with them. 

- [ ] passes `flake8 modin`
- [x] Resolves #2881
- [ ] tests added and passing

### DIFF
--- a/modin/engines/ray/utils.py
+++ b/modin/engines/ray/utils.py
@@ -133,15 +133,21 @@ def initialize_ray(
                 # want to overwrite that value if we have.
                 if object_store_memory is None:
                     # Round down to the nearest Gigabyte.
-                    mem_bytes = ray.utils.get_system_memory() // 10 ** 9 * 10 ** 9
+                    try:
+                        system_memory = ray._private.utils.get_system_memory()
+                    except AttributeError:  # Compatibility with Ray <= 1.2
+                        system_memory = ray.utils.get_system_memory()
+                    mem_bytes = system_memory // 10 ** 9 * 10 ** 9
                     # Default to 8x memory for out of core
                     object_store_memory = 8 * mem_bytes
             # In case anything failed above, we can still improve the memory for Modin.
             if object_store_memory is None:
                 # Round down to the nearest Gigabyte.
-                object_store_memory = int(
-                    0.6 * ray.utils.get_system_memory() // 10 ** 9 * 10 ** 9
-                )
+                try:
+                    system_memory = ray._private.utils.get_system_memory()
+                except AttributeError:  # Compatibility with Ray <= 1.2
+                    system_memory = ray.utils.get_system_memory()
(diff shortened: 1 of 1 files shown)


devin-petersohn (commented):
Thanks @krfricke! Is there an open issue for this?

The commitlint failure means the first line of your commit message needs to tag the issue. Something like:

```
FIX-#<ISSUE_NUM>: Move to new private Ray API
```

Let me know if this doesn't make sense. Thanks again!

krfricke:
There's no open issue for this, I noticed the failure on our latest master CI and decided to just write a quick patch. I can open an issue if this is required )or move to `REFACTOR` instead?) Let me know what you prefer.

devin-petersohn:
@krfricke Thanks for clarifying, is there a test run for Modin on the Ray CI?

Opening an issue would be best, ideally each PR has an associated issue.

krfricke:
The test is not on the upstream Ray CI, but we're integrating with Modin in `https://github.com/ray-project/xgboost_ray` and run tests for data sources there.

I opened an issue for this fix, can you re-trigger the commit check?

devin-petersohn:
@krfricke Thanks for opening the issue! The commit lint checks check the actual commit message for be98b38, not the PR name. To resolve it you need to rebase and reword be98b38, then force push to your `ray-api` branch. We have found that `git commit --amend` does not work, but rebasing does.

krfricke:
The `codecov/project` check fails though I can't see how this would be related to this PR.

devin-petersohn:
@krfricke Failing codecov is ok in this case. The difference is that we are not yet testing on Ray nightly wheels, so the clause for handling the new API is not yet tested by CI.

Codecov reports are also a bit flaky lately, we are planning to go in and fix it soon.

devin-petersohn (approved):

PR #2889, opened 2021-03-16
FIX-#2886: Match pandas error message on Series.kurt

### LINKED ISSUE #2886: Series `kurt` and `kurtosis` error messages do not match pandas
Discovered as a part of #2847

devin-petersohn:
Resolves #2886

## What do these changes do?

- [x] passes `flake8 modin`
- [x] Resolves #2886 
- [x] tests enabled~added~ and passing

### DIFF
--- a/modin/pandas/series.py
+++ b/modin/pandas/series.py
@@ -822,6 +822,14 @@ class Series(BasePandasDataset):
     def keys(self):
         return self.index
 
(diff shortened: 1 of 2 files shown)


RehanSD (approved):
Looks good to me - only comment is same as in #2890 - do we want `eval_general` to check not just that the type of error is the same, but also that the messages match?

PR #2890, opened 2021-03-16
FIX-#2887: Match pandas error message on `Series.sum`

### LINKED ISSUE #2887: `Series.sum` does not give the same error message as pandas on `numeric_only=True`
Discovered as a part of #2847

devin-petersohn:
Resolves #2887

## What do these changes do?

- [x] passes `flake8 modin`
- [x] Resolves #2887 
- [x] tests enabled~added~ and passing

### DIFF
--- a/modin/pandas/test/test_series.py
+++ b/modin/pandas/test/test_series.py
@@ -3059,19 +3059,7 @@ def test_subtract(data):
 @pytest.mark.parametrize(
     "skipna", bool_arg_values, ids=arg_keys("skipna", bool_arg_keys)
 )
-@pytest.mark.parametrize(
-    "numeric_only",
-    [
-        None,
-        False,
-        pytest.param(
-            True,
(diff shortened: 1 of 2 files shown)


modin/pandas/test/test_series.py
```diff
@@ -3059,19 +3059,7 @@ def test_subtract(data):
...
-        False,
-        pytest.param(
-            True,
-            marks=pytest.mark.xfail(
-                reason="numeric_only not implemented for pandas.Series"
-            ),
-        ),
-    ],
-)
+@pytest.mark.parametrize("numeric_only", [None, False, True])
```
RehanSD, line 3062: Do we want to make `eval_general` check not just that the type of error is the same, but also that the error messages are the same?

RehanSD (approved):
Looks good to me - left one comment, but other than that, I think this can be merged.

PR #2907, opened 2021-03-19
DOCS-#2914: add docstrings to data_management/functions component

### LINKED ISSUE #2914: add documentation to data_management/functions component

anmyachev:

## What do these changes do?

- [ ] passes `flake8 modin`
- [x] Resolves #2914 
- [ ] tests added and passing

### DIFF
--- a/modin/backends/pandas/query_compiler.py
+++ b/modin/backends/pandas/query_compiler.py
@@ -37,7 +37,7 @@ from modin.data_management.functions import (
     ReductionFunction,
     BinaryFunction,
     GroupbyReduceFunction,
-    groupby_reduce_functions,
+    GROUPBY_REDUCE_FUNCTIONS,
 )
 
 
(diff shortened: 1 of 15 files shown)


anmyachev:
While writing the documentation for various functions, I came across the fact that `call` method can be deleted without changing the logic.  @devin-petersohn do you have anything against deleting it? (in the future, if necessary, I will divide into two pull requests with refactoring and documentation)

devin-petersohn:
@anmyachev Ideally, call would be used in places where `register` is followed by some parenthesis: https://github.com/modin-project/modin/blob/07c2f86c890f31532ee49bf091cef4e8b036b0bd/modin/backends/pandas/query_compiler.py#L765-L767

I would prefer it be used in those places rather than deleted altogether.

anmyachev:
> @anmyachev Ideally, call would be used in places where `register` is followed by some parenthesis:
> 
> https://github.com/modin-project/modin/blob/07c2f86c890f31532ee49bf091cef4e8b036b0bd/modin/backends/pandas/query_compiler.py#L765-L767
> 
> I would prefer it be used in those places rather than deleted altogether.

If we are talking about the problem of code readability, then I see two options:
1. Just rewrite the calling code a little
```python
max = MapReduceFunction.register(map_func, reduce_func)
return max(self, axis=axis, **kwargs)
```
2. Make `register` return a functor, with implemented method `call`, which will call `__call__` inside itself.
```python
return MapReduceFunction.register(map_func, reduce_func).call(
    self, axis=axis, **kwargs
)
```
It seems to me that the slight improvement in readability is not worth the effort that should be made in the second case.

Can we choose the first option?

anmyachev:
Closed in favor of #2956.

PR #2910, opened 2021-03-22
Fix-#2798: Fix setting NPartitions

### LINKED ISSUE #2798: Wrong number of partitions of a Modin DataFrame on a cluster

YarShev:

## What do these changes do?

- [x] passes `flake8 modin`
- [x] Resolves #2798  
- [x] tests passing

### DIFF
--- a/modin/config/envvars.py
+++ b/modin/config/envvars.py
@@ -149,13 +149,6 @@ class NPartitions(EnvironmentVariable, type=int):
     """
 
     varname = "MODIN_NPARTITIONS"
-    # This flag is used to detect whether NPartitions is default value or not
-    _is_default = True
-
-    @classmethod
-    def put(cls, value):
-        super().put(value)
-        cls._is_default = False
 
     @classmethod
     def _put(cls, value):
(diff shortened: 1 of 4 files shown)


modin/config/envvars.py
```diff
@@ -167,14 +167,28 @@ def put_if_default(cls, value):
...
+        if cls._is_default and not cls._is_set_from_config():
             cls.put(value)
 
     @classmethod
     def _get_default(cls):
         cls._is_default = True
         return CpuCount.get()
 
+    @classmethod
+    def _is_set_from_config(cls):
```
vnlitvinov, line 179: It might be useful to be migrated to `Parameter` and made an internal flag for all configuration parameters... that way you can set it in `Parameter.get()` method (and it can be a tri-state - "set by config source", "set by default value", "set by user")

Right now you're duplicating the `.get()` efforts to an extent, which isn't necessary IMHO.

YarShev: In that case we will need to call `.get()` before `._put()` to initialize value of that flag that is not main goal of `.get()`. It is confusing a little, I think.

YarShev: On the other hand, we can leave a comment that it is a dummy call to initialize the flag. Thereby, we can get rid of code duplication that you mentioned.

YarShev: Changed way to set NPartitions from engines.

modin/config/pubsub.py
```diff
@@ -74,6 +74,9 @@ class Parameter(object):
     type = str
     default = None
     is_abstract = True
+    # This flag is used to detect a way of setting value for an entity.
+    # 0 - default value, 1 - value from config source, 2 - value set by a user.
```
vnlitvinov, line 78: I would rather we have these as constant values, not inside a comment

vnlitvinov (commented):
Looks better

modin/config/pubsub.py
YarShev: Yes, it makes sense. Fixed.

modin/config/pubsub.py
```diff
@@ -74,6 +74,9 @@ class Parameter(object):
     type = str
     default = None
     is_abstract = True
+    # This flag is used to detect a way of setting value for an entity.
+    way_of_set = None
+    WAYS_OF_SET = {"default": 0, "from_config": 1, "from_user": 2}
```
devin-petersohn, line 79: What is the difference between `from_user` and `from_config`? Can we document it?

YarShev: The difference is that `from_user` maps a value set by `put` and `from_config` maps a value set by env var. Probably, it makes sense to rename `from_user` to `put` and `from_config` to `envvar`. So, added comment on the dictionary and its content.

devin-petersohn (approved):
Thanks @YarShev, LGTM!

PR #2919, opened 2021-03-24
REFACTOR-#2918: Refactor internal functionality of partitions

### LINKED ISSUE #2918: Refactor internal functionality of partitions
We can do the following things:
1) Get rid of passing `ray.ObjectRef(s)` within a list to a partition's `apply` (and to a remote call then) to avoid explicit call `ray.get`.
2) Add preprocessing a function for dask engine as it is done for ray engine.

YarShev:

## What do these changes do?

- [x] passes `flake8 modin`
- [x] Resolves #2918  
- [x] tests passing

### DIFF
--- a/modin/engines/dask/pandas_on_dask/frame/partition.py
+++ b/modin/engines/dask/pandas_on_dask/frame/partition.py
@@ -91,7 +91,7 @@ class PandasOnDaskFramePartition(PandasFramePartition):
         The keyword arguments are sent as a dictionary.
         """
         client = default_client()
-        call_queue = self.call_queue + [(func, args, kwargs)]
+        call_queue = self.call_queue + [[func, args, kwargs]]
         if len(call_queue) > 1:
             future = client.submit(
                 apply_list_of_funcs, call_queue, self.future, pure=False
@@ -129,7 +129,7 @@ class PandasOnDaskFramePartition(PandasFramePartition):
         The keyword arguments are sent as a dictionary.
         """
         return PandasOnDaskFramePartition(
-            self.future, call_queue=self.call_queue + [(func, args, kwargs)]
+            self.future, call_queue=self.call_queue + [[func, args, kwargs]]
         )
 
     def drain_call_queue(self):


YarShev:
@modin-project/modin-core , just a friendly reminder to review.

modin/engines/ray/pandas_on_ray/frame/partition.py
```diff
@@ -239,30 +252,57 @@ def get_index_and_columns(df):
...
+    )
+
+
+@ray.remote(num_returns=4)
+def apply_list_of_funcs(partition, funcs):  # pragma: no cover
     def deserialize(obj):
         if isinstance(obj, ObjectIDType):
             return ray.get(obj)
-        return obj
+        elif isinstance(obj, tuple) and any(isinstance(o, ObjectIDType) for o in obj):
```
vnlitvinov, line 276:
```suggestion
        elif isinstance(obj, (tuple, list)) and any(isinstance(o, ObjectIDType) for o in obj):
```
maybe even generic `ABC.Sequence`?

modin/engines/ray/pandas_on_ray/frame/partition.py
```diff
@@ -239,30 +252,57 @@ def get_index_and_columns(df):
...
+            return ray.get(list(obj))
+        elif isinstance(obj, dict) and any(
+            isinstance(obj[o], ObjectIDType) for o in obj
```
vnlitvinov, line 279:
```suggestion
            isinstance(val, ObjectIDType) for val in obj.values()
```

modin/engines/ray/pandas_on_ray/frame/partition.py
```diff
@@ -239,30 +252,57 @@ def get_index_and_columns(df):
...
+        ):
+            return ray.get(list(obj.values()))
```
vnlitvinov, line 281:
```suggestion
            return dict(zip(obj.keys(), ray.get(list(obj.values()))))
```
not sure if this is needed... but I think it's better to retain the object type (`dict` in this case) upon deserialization, otherwise e.g. deserializing `**kwargs` would break

modin/engines/base/frame/partition.py
```diff
@@ -36,7 +36,7 @@ def get(self):
         """
         pass
 
-    def apply(self, func, **kwargs):
+    def apply(self, func, *args, **kwargs):
```
devin-petersohn, line 39: Are there any `*args` being passed here? Why do we need this change?

modin/engines/dask/pandas_on_dask/frame/partition.py
```diff
@@ -80,25 +71,42 @@ def apply(self, func, **kwargs):
...
+        futures = [
+            client.submit(lambda l, i: l[i], future, i, pure=False) for i in range(2)
+        ]
         return PandasOnDaskFramePartition(futures[0], ip=futures[1])
 
-    def add_to_apply_calls(self, func, **kwargs):
+    def add_to_apply_calls(self, func, *args, **kwargs):
         return PandasOnDaskFramePartition(
-            self.future, call_queue=self.call_queue + [[pkl.dumps(func), kwargs]]
+            self.future, call_queue=self.call_queue + [[func, args, kwargs]]
```
devin-petersohn, line 90: If this change works and is fast, it would be great. Originally, this was meant to speed up serialization of functions and arguments.

modin/engines/dask/pandas_on_dask/frame/partition.py
```diff
@@ -80,25 +71,42 @@ def apply(self, func, **kwargs):
...
+        else:
+            func, args, kwargs = call_queue[0]
+            future = client.submit(apply_func, self.future, func, *args, **kwargs)
+        futures = [
+            client.submit(lambda l, i: l[i], future, i, pure=False) for i in range(2)
+        ]
         return PandasOnDaskFramePartition(futures[0], ip=futures[1])
 
-    def add_to_apply_calls(self, func, **kwargs):
+    def add_to_apply_calls(self, func, *args, **kwargs):
```
devin-petersohn, line 88: I see a `*args` change here also, are we using it? In the past, I have only used `**kwargs` to aid in debugging.

modin/engines/dask/pandas_on_dask/frame/partition.py
```diff
@@ -176,7 +184,7 @@ def preprocess_func(cls, func):
         Returns:
             An object that can be accepted by `apply`.
         """
-        return func
+        return get_client().scatter(func, hash=False)
```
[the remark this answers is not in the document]
devin-petersohn: I agree, we need to be passing `broadcast=True` to `scatter` here, I believe.

https://distributed.dask.org/en/latest/api.html#distributed.Client.scatter

modin/engines/ray/pandas_on_ray/frame/partition.py
devin-petersohn: We can also use `is_list_like` from pandas.

devin-petersohn (commented):
Looks like a nice overall cleanup @YarShev, left some comments.

modin/engines/dask/pandas_on_dask/frame/partition.py
vnlitvinov: I think this is supposed work because we now preprocess the `func` for Dask, too.

modin/engines/base/frame/partition.py
vnlitvinov: I do not know the exact details but I would assume this is to support the generic way of calling arbitrary Python functions where you can pass arguments both as position and as named ones... it shouldn't be hurting much anyway.

vnlitvinov:
@YarShev you have a merge conflict now, plus some feedback that is not addressed :)

modin/engines/dask/pandas_on_dask/frame/partition.py
```diff
@@ -80,25 +71,42 @@ def apply(self, func, **kwargs):
...
-        futures = [get_client().submit(lambda l: l[i], future) for i in range(2)]
+        client = get_client()
+        call_queue = self.call_queue + [[func, args, kwargs]]
+        if len(call_queue) > 1:
+            future = client.submit(
+                apply_list_of_funcs, self.future, call_queue, pure=False
+            )
+        else:
+            func, args, kwargs = call_queue[0]
+            future = client.submit(apply_func, self.future, func, *args, **kwargs)
```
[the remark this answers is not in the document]
YarShev: We could have a function wrapper in which `if...else...` juggling occurs.

modin/engines/dask/pandas_on_dask/frame/partition.py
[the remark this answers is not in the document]
YarShev: I agree, it makes sense.

modin/engines/ray/pandas_on_ray/frame/partition.py
YarShev: We can't use `is_list_like` from pandas. Otherwise, we have to add a check `not isinstance(obj, (DataFrame, Series))` because `DataFrame` and `Series` are list-like.

modin/engines/ray/pandas_on_ray/frame/partition.py
YarShev: I agree, will be changed.

modin/engines/dask/pandas_on_dask/frame/partition.py
YarShev: @vnlitvinov , yes, you are correct.

modin/engines/base/frame/partition.py
YarShev: @devin-petersohn , yes, we have `*args` passed here for now ([link](https://github.com/modin-project/modin/pull/2919/files#diff-216a3c97cca7e07c5687d171206e7ebf2678c10bb2308f6b9fe1604468b6bed2R142)). Also, this change is necessary to pass `*args` directly to a remote call in order to a backend itself (ray/dask) is able to materialize data if the `len(call queue) == 1`. This speedups map operations. That's why I added remote `apply_func` function.

@vnlitvinov , your thoughts are on the right way as well.

modin/engines/dask/pandas_on_dask/frame/partition.py
YarShev: See the comment ([link](https://github.com/modin-project/modin/pull/2919#discussion_r639504163)).

modin/engines/dask/pandas_on_dask/frame/partition.py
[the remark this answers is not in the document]
vnlitvinov: I've forgotten the exact details as it was 28 days since previous review, but I think there should be a way to re-do it so it would be uniform regardless of single- or multi-call (without hiding the juggling - it's just sweeping stuff under the rug)

devin-petersohn (commented):
@YarShev shall we postpone this to the next release?

YarShev:
@devin-petersohn , yes, we should postpone this to the next release.

modin/engines/dask/pandas_on_dask/frame/partition.py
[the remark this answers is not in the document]
YarShev: Fixed.

modin/engines/ray/pandas_on_ray/frame/partition.py
YarShev: Fixed.

modin/engines/ray/pandas_on_ray/frame/partition.py
YarShev: Fixed.

modin/engines/dask/pandas_on_dask/frame/partition.py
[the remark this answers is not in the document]
YarShev: Left as it is. See the commit ([link](https://github.com/modin-project/modin/pull/2919#discussion_r639504163)) why I had done this way. If you have any suggestions how to improve it, I would glad to hear them.

modin/engines/dask/pandas_on_dask/frame/partition.py
```diff
@@ -351,3 +346,52 @@ def empty(cls):
             A new ``PandasOnDaskFramePartition`` object.
         """
         return cls(pandas.DataFrame(), 0, 0)
+
+
+def apply_func(partition, func, *args, **kwargs):
```
devin-petersohn, line 351: Can we merge this with https://github.com/modin-project/modin/blob/ccb5a537d5b776c0c5015f1d3dcf13886d21a250/modin/engines/dask/task_wrapper.py#L23-L46?

If we do the same for Ray, we can reduce the size of these files because it will be standardized.

modin/engines/ray/pandas_on_ray/frame/partition.py
```diff
@@ -131,12 +139,21 @@ def drain_call_queue(self):
             return
         oid = self.oid
         call_queue = self.call_queue
-        (
-            self.oid,
-            self._length_cache,
-            self._width_cache,
-            self._ip_cache,
-        ) = deploy_ray_func.remote(call_queue, oid)
+        if len(call_queue) > 1:
```
devin-petersohn, line 142: I agree with @vnlitvinov where he mentions above about this issue, it does seem a little weird. What is the benefit of special casing the call queue having length 1?

modin/engines/dask/pandas_on_dask/frame/partition.py
YarShev: I don't think we can merge this exactly with RayTask.deploy because the latter is applicable for deploying io function. If we really want to merge such things, I would suggest to make the following structure:
```
.task_wrapper/
   .io_task.py # RayIOTask.deploy with own parameters for deploying function
   .partition_task.py # RayPartitionTask.deploy with own parameters for deploying function
   .axis_partition_task.py # RayAxisPartitionTask.deploy with own parameters for deploying function
   . # and probably we will need something like two_axis_partition_task.py containing RayTwoAxisPartitionTask.deploy with own parameters for deploying function

# or we can do the following
# task_wrapper.py
class RayTask:
    def deploy_io_task(...):
        ...
    def deploy_partition_task(...):
        ...
    def deploy_axis_partition_task(...):
        ...
    def deploy_two_axis_partition_task(...):
        ...
```
And then add a class member of RayTask to Partition, AxisPartition and IO classes to call deploy_... functions.
But I would suggest doing this as part of #2957 or after we define a stable Modin structure. What do you think?

modin/engines/ray/pandas_on_ray/frame/partition.py
YarShev: This speeds up the operations that are map-like or include map phase.

**pandas_on_ray**

*BenchmarkMode.put(True)*

| df.shape = (8192, 1024) | df.round() | df.abs() | df.max() | df.prod() |
| - | - | - | - | - |
| master                           | 2.3758      | 2.0388  | 3.1617    | 4.0645    |
| PR                                 | 0.8098      | 0.7027  | 1.2626    | 1.6392    |

*BenchmarkMode.put(False)*

|df.shape = (8192, 1024)|df.round()|df.abs()|df.max()|df.prod()|
| - | - | - | - | - |
| master                           | 0.6013      | 0.5945  | 2.4051    | 3.2238    |
| PR                                 | 0.5044      | 0.4410  | 1.2138    | 1.2128    |

devin-petersohn: Very interesting. Can we add a comment in the code to mention that it is written this way because it dramatically improves performance?

modin/engines/dask/pandas_on_dask/frame/partition.py
devin-petersohn: @YarShev Yes, that makes sense. We can do it later.

modin/engines/ray/pandas_on_ray/frame/partition.py
YarShev: Sure, will add.

modin/engines/dask/pandas_on_dask/frame/partition.py
YarShev: Cool!

modin/engines/ray/pandas_on_ray/frame/partition.py
YarShev: Added.

modin/engines/dask/pandas_on_dask/frame/partition.py
```diff
@@ -114,24 +90,34 @@ def apply(self, func, **kwargs):
...
         The keyword arguments are sent as a dictionary.
         """
-        func = pkl.dumps(func)
-        call_queue = self.call_queue + [[func, kwargs]]
-        future = default_client().submit(
-            apply_list_of_funcs, call_queue, self.future, pure=False
-        )
-        futures = [default_client().submit(lambda l: l[i], future) for i in range(2)]
+        client = default_client()
+        call_queue = self.call_queue + [[func, args, kwargs]]
```
vnlitvinov, line 94:
```suggestion
        call_queue = self.call_queue + [(func, args, kwargs)]
```
Let's use tuples where possible - they take less memory and could be slightly faster in certain operations.

modin/engines/dask/pandas_on_dask/frame/partition.py
```diff
@@ -114,24 +90,34 @@ def apply(self, func, **kwargs):
...
+                apply_list_of_funcs, self.future, call_queue, pure=False
+            )
+        else:
+            # We handle `len(call_queue) == 1` in a different way because
+            # this improves performance a bit.
+            func, args, kwargs = call_queue[0]
+            future = client.submit(apply_func, self.future, func, *args, **kwargs)
+        futures = [
+            client.submit(lambda l, i: l[i], future, i, pure=False) for i in range(2)
+        ]
```
vnlitvinov, line 106: I wonder if there's more elegant way of unpacking a "future tuple"...

modin/engines/dask/pandas_on_dask/frame/partition.py
```diff
@@ -143,16 +129,29 @@ def add_to_apply_calls(self, func, **kwargs):
         The keyword arguments are sent as a dictionary.
         """
         return PandasOnDaskFramePartition(
-            self.future, call_queue=self.call_queue + [[pkl.dumps(func), kwargs]]
+            self.future, call_queue=self.call_queue + [[func, args, kwargs]]
```
vnlitvinov, line 132:
```suggestion
            self.future, call_queue=self.call_queue + [(func, args, kwargs)]
```

modin/engines/dask/pandas_on_dask/frame/partition.py
```diff
@@ -143,16 +129,29 @@ def add_to_apply_calls(self, func, **kwargs):
...
+                apply_list_of_funcs, self.future, call_queue, pure=False
+            )
+        else:
+            # We handle `len(call_queue) == 1` in a different way because
+            # this improves performance a bit.
+            func, args, kwargs = call_queue[0]
+            future = client.submit(apply_func, self.future, func, *args, **kwargs)
+        futures = [
+            client.submit(lambda l, i: l[i], future, i, pure=False) for i in range(2)
+        ]
```
vnlitvinov, line 152: can we extract this in a common function? I think this piece of code is exactly identical in `apply()` and here...

modin/engines/dask/pandas_on_dask/frame/partition_manager.py
```diff
@@ -151,30 +104,20 @@ def broadcast_apply(cls, axis, apply_func, left, right, other_name="r"):
...
-                                if axis
-                                else [obj.future for obj in right[row_idx]]
-                            ),
-                            pure=False,
-                        )
+                    part.apply(
+                        map_func,
+                        *rt_axis_parts[col_idx].list_of_blocks
+                        if axis
+                        else rt_axis_parts[row_idx].list_of_blocks,
```
vnlitvinov, line 120: if the star ("unpacking a tuple") is applied to the whole `a if foo else b` statement I suggest wrapping it in braces for readability... if it's applied to only "true" branch then why such asymmetry?

modin/engines/ray/pandas_on_ray/frame/partition.py
vnlitvinov: @YarShev numbers look interesting... but can we see those comparing this PR against "this PR but without special handling of 1-long queue" to be sure we're not measuring some other effect?

modin/engines/ray/pandas_on_ray/frame/partition.py
```diff
@@ -395,26 +458,38 @@ def deploy_ray_func(call_queue, partition):  # pragma: no cover
...
+        elif isinstance(obj, dict) and any(
+            isinstance(val, ObjectIDType) for val in obj.values()
+        ):
+            return dict(zip(obj.keys(), ray.get(list(obj.values()))))
+        else:
+            return obj
 
-    if len(call_queue) > 1:
-        for func, kwargs in call_queue[:-1]:
+    if len(funcs) > 1:
```
vnlitvinov, line 472: why do we need this `if` and why do we special-case the last entry in the list?..

modin/engines/dask/pandas_on_dask/frame/partition.py
YarShev: So far I have not seen it.

modin/engines/dask/pandas_on_dask/frame/partition.py
YarShev: I think this relates to this conversation ([link](https://github.com/modin-project/modin/pull/2919#discussion_r652781272)) and it makes sense to be fixed later.

modin/engines/dask/pandas_on_dask/frame/partition_manager.py
YarShev: Yes, the star should be applied to the whole statement. Will be fixed.

modin/engines/ray/pandas_on_ray/frame/partition.py
YarShev: The numbers look even more interesting for the case without special handling of 1-long queue.

**pandas_on_ray**
*BenchmarkMode.put(True)*

| df.shape = (8192, 1024) | df.round() | df.prod() |
| - | - | - |
| master                           | 2.3758      | 4.0645    |
| PR (with the special casing)    | 0.8098      | 1.6392    |
| PR (with **no** the special casing)    | 3.2039      | 5.9272    |

*BenchmarkMode.put(False)*

| df.shape = (8192, 1024) | df.round() | df.prod() |
| - | - | - |
| master                           | 0.6013      | 3.2238    |
| PR (with the special casing)    | 0.5044      | 1.2128   |
| PR (with **no** the special casing)    | 0.8461      | 4.8586    |

It turns out if we pass `call_queue/funcs` in `apply_list_of_func` as second parameter, we will see that slowdown. And if we pass `call_queue/funcs` in `apply_list_of_func` as the first parameter, we will see numbers that are about equal to master. I am not sure of what is the reason but it will be fixed.

modin/engines/ray/pandas_on_ray/frame/partition.py
YarShev: I left this code as it was but yes, we can rid of special-casing the last entry in the list.

YarShev: Fixed

modin/engines/dask/pandas_on_dask/frame/partition_manager.py
YarShev: Fixed

modin/engines/ray/pandas_on_ray/frame/partition.py
YarShev: Fixed

modin/engines/dask/pandas_on_dask/frame/partition.py
vnlitvinov: Makes sense, let's fix all at once in a later PR.

vnlitvinov (commented):
@YarShev changes themselves look good to me, but you now have failing tests in CI...

YarShev:
@vnlitvinov , oh, I have remembered back that we can't use tuples in call_queue for Dask. There is an issue with handling tuples in Dask but I couldn't reproduce the issue some time ago without Modin code to submit the reproducer to Dask repo.

vnlitvinov (approved):
LGTM assuming tests pass.

YarShev:
@devin-petersohn , friendly reminder!

devin-petersohn (approved):
Thanks @YarShev!

PR #2921, opened 2021-03-24
FEAT-#2920: add support of write into cloud storage like S3 and GCS

### LINKED ISSUE #2920: Add support of write into cloud storage like S3 or GCS for to_csv function

anmyachev:

## What do these changes do?

The key idea is to use the pandas functionality to create file object from strings.
This way we can support URLS like `s3://` or `gs://`.

Meanwhile, the writing remains sequential and can be a bottleneck in the case of low internet speed.

- [x] passes `flake8 modin`
- [x] Resolves #2920 
- [x] tests added and passing

### DIFF
--- a/modin/engines/ray/generic/io.py
+++ b/modin/engines/ray/generic/io.py
@@ -99,6 +99,7 @@ class RayIO(BaseIO):
                 kwargs["path_or_buf"] = io.StringIO()
             df.to_csv(**kwargs)
             content = kwargs["path_or_buf"].getvalue()
+            kwargs["path_or_buf"].close()
 
             # each process waits for its turn to write to a file;
             # in case of violation of the order of receiving messages from the queue,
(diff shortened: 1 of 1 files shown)


modin/engines/ray/generic/io.py
```diff
@@ -99,6 +99,7 @@ def func(df, **kw):
                 kwargs["path_or_buf"] = io.StringIO()
             df.to_csv(**kwargs)
             content = kwargs["path_or_buf"].getvalue()
+            kwargs["path_or_buf"].close()
```
anmyachev, line 102: Ability to free memory earlier

devin-petersohn (commented):
How can this be tested?

anmyachev:
> How can this be tested?

I have tested manually.

To do this, I created a bucket in S3. (You need to do the same)

Import paths are configured to run the script in `asv_bench/benchmarks` folder.

```python
from time import time
import modin.pandas as pd
from utils import generate_dataframe


def measure_csv():
    filename = [SOME S3 URL]

    df = generate_dataframe("modin", "int", 4_000_000, 30, 0, 100)
    start = time()
    df.to_csv(filename)
    print(f"to_csv time: {time()-start}")

    start = time()
    read_df = pd.read_csv(filename, index_col=0)
    print(f"read_csv time: {time()-start}")

    print(f"equals: {df._query_compiler.to_pandas().equals(read_df._query_compiler.to_pandas())}")

measure_csv()
```

I have not found a way to test automatically (in CI) and for free.

devin-petersohn (approved):
@anmyachev LGTM, thanks for the testing code!

PR #2923, opened 2021-03-24
TEST-#2922: add read_csv true/false parameters for asv tests

### LINKED ISSUE #2922: [ASV] add read_csv benchmark with true/false_values parameters for testing with OmniSci backend

amyskov:

## What do these changes do?

- [x] passes `flake8 modin`
- [x] Resolves #2922

### DIFF
--- a/asv_bench/benchmarks/io/csv.py
+++ b/asv_bench/benchmarks/io/csv.py
@@ -24,19 +24,14 @@ from ..utils import (
     IMPL,
     execute,
     get_shape_id,
+    prepare_io_data,
 )
 
 
 class BaseReadCsv:
     # test data file can de created only once
     def setup_cache(self, test_filename="io_test_file"):
-        test_filenames = {}
-        for shape in UNARY_OP_DATA_SIZE[ASV_DATASET_SIZE]:
-            shape_id = get_shape_id(shape)
-            test_filenames[shape_id] = f"{test_filename}_{shape_id}.csv"
-            df = generate_dataframe("pandas", "str_int", *shape, RAND_LOW, RAND_HIGH)
-            df.to_csv(test_filenames[shape_id], index=False)
-
+        test_filenames = prepare_io_data(test_filename, "str_int")
         return test_filenames
 
     def setup(self, test_filenames, shape, *args, **kwargs):
@@ -72,6 +67,27 @@ class TimeReadCsvSkiprows(BaseReadCsv):
         execute(
             IMPL[ASV_USE_IMPL].read_csv(
                 test_filenames[self.shape_id], skiprows=self.skiprows
+            )
+        )
+
+
+class TimeReadCsvTrueFalseValues(BaseReadCsv):
+    param_names = ["shape"]
+    params = [
+        UNARY_OP_DATA_SIZE[ASV_DATASET_SIZE],
+    ]
+
+    # test data file can de created only once
+    def setup_cache(self, test_filename="io_test_file"):
+        test_filenames = prepare_io_data(test_filename, "true_false_int")
+        return test_filenames
+
+    def time_true_false_values(self, test_filenames, shape):
+        execute(
+            IMPL[ASV_USE_IMPL].read_csv(
+                test_filenames[self.shape_id],
+                true_values=["Yes", "true"],
+                false_values=["No", "false"],
             )
         )
 
--- a/asv_bench/benchmarks/io/csv.py
+++ b/asv_bench/benchmarks/io/csv.py
@@ -20,6 +20,7 @@ from ..utils import (
     RAND_HIGH,
     ASV_USE_IMPL,
     ASV_DATASET_SIZE,
+    ASV_USE_BACKEND,
     UNARY_OP_DATA_SIZE,
     IMPL,
     execute,
@@ -69,8 +70,7 @@ class TimeReadCsvSkiprows(BaseReadCsv):
         execute(
             IMPL[ASV_USE_IMPL].read_csv(
                 test_filenames[self.shape_id], skiprows=self.skiprows
-            ),
-            trigger_omnisci_import=True,
+            )
         )
 
 
@@ -94,7 +94,7 @@ class TimeReadCsvTrueFalseValues(BaseReadCsv):
                 true_values=["Yes", "true"],
                 false_values=["No", "false"],
             ),
-            trigger_omnisci_import=True,
+            trigger_omnisci_import=ASV_USE_BACKEND == "omnisci",
         )
 
 
@@ -164,6 +164,5 @@ class TimeReadCsvNamesDtype:
                 header=0,
                 dtype=self.dtype,
                 parse_dates=self.parse_dates,
-            ),
-            trigger_omnisci_import=True,
+            )
         )
--- a/asv_bench/benchmarks/io/csv.py
+++ b/asv_bench/benchmarks/io/csv.py
@@ -77,7 +77,7 @@ class TimeReadCsvTrueFalseValues(BaseReadCsv):
         UNARY_OP_DATA_SIZE[ASV_DATASET_SIZE],
     ]
 
-    # test data file can de created only once
+    # test data file should be created only once
     def setup_cache(self, test_filename="io_test_file"):
         test_filenames = prepare_io_data(test_filename, "true_false_int")
         return test_filenames
--- a/asv_bench/benchmarks/io/csv.py
+++ b/asv_bench/benchmarks/io/csv.py
@@ -26,6 +26,8 @@ from ..utils import (
     get_shape_id,
     prepare_io_data,
 )
+from ..omnisci.utils import trigger_import
+from modin.config import Backend
 
 
 class BaseReadCsv:
@@ -35,6 +37,7 @@ class BaseReadCsv:
         return test_filenames
 
     def setup(self, test_filenames, shape, *args, **kwargs):
+        self.execute = trigger_import if Backend.get == "Omnisci" else execute
         # ray init
         if ASV_USE_IMPL == "modin":
             pd.DataFrame([])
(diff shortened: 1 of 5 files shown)


asv_bench/benchmarks/io/csv.py
```diff
@@ -76,6 +71,27 @@ def time_skiprows(self, test_filenames, shape, skiprows):
         )
 
 
+class TimeReadCsvTrueFalseValues(BaseReadCsv):
```
anmyachev, line 74: The test for OmniSci and Ray is very similar, try to find a way to avoid code duplication.

asv_bench/benchmarks/io/csv.py
```diff
@@ -76,6 +71,27 @@ def time_skiprows(self, test_filenames, shape, skiprows):
...
+    param_names = ["shape"]
+    params = [
+        UNARY_OP_DATA_SIZE[ASV_DATASET_SIZE],
+    ]
+
+    # test data file can de created only once
```
anmyachev, line 80:
```suggestion
    # test data file should be created only once
```

asv_bench/benchmarks/utils.py
```diff
@@ -248,12 +331,23 @@ def gen_data(
...
+
+    Notes
+    -----
+    Returned data type depends on the `data_type` parameter in the next way:
+
+    - `data_type`=="int" - all columns will be contain only integer values
+    - `data_type`=="str_int" some of the columns will be of string type,
+    - `data_type`=="true_false_int" half of the columns will be filed with
+      string values representing "true" and "false" values and another half - with
+      integers.
```
anmyachev, line 343:
```suggestion
    Returned data type depends on the `data_type` parameter in the next way:
    - `data_type`=="int" - all columns will be contain only integer values;
    - `data_type`=="str_int" some of the columns will be of string type;
    - `data_type`=="true_false_int" half of the columns will be filled with
      string values representing "true" and "false" values and another half - with
      integers.
```

asv_bench/benchmarks/utils.py
```diff
@@ -217,6 +217,89 @@ def gen_str_int_data(nrows: int, ncols: int, rand_low: int, rand_high: int) -> d
...
+            nrows, ncols, rand_low, rand_high
+        )
+    )
+    data = gen_int_data(nrows // 2, ncols // 2, rand_low, rand_high).copy()
+    data_true_false = gen_true_false_data(
+        nrows - nrows // 2, ncols - ncols // 2, ["Yes", "true", "No", "false"]
+    ).copy()
+    data.update(data_true_false)
+    data_cache[cache_key] = weakdict(data)
+    return data
```
anmyachev, line 300: This is a rather specific case, it will not be used widely, so I suggest not to divide the functionality into two functions and leave everything in `gen_true_false_int_data`.

Also, given that caching is a fairly general mechanism, then maybe we should move caching into `gen_data` function (to reduce copy-paste). What do you think?

anmyachev (changes requested):

asv_bench/benchmarks/io/csv.py
amyskov: Ok, single implementation was used by specifying `execute` method for different backends.

asv_bench/benchmarks/utils.py
amyskov: Ok, all true/false data generation was moved to the `gen_true_false_int_data` function.

I think it is good idea, implemented in the last commit.

asv_bench/benchmarks/utils.py
```diff
@@ -248,14 +270,32 @@ def gen_data(
...
+    if cache_key in data_cache:
+        return data_cache[cache_key]
+
+    type_to_generator = {
+        "int": gen_int_data,
+        "str_int": gen_str_int_data,
+        "true_false_int": gen_true_false_int_data,
+    }
+    data_generator = type_to_generator.get(data_type, None)
+    assert data_generator
```
anmyachev, line 293: why not just `type_to_generator[data_type]`?)

asv_bench/benchmarks/utils.py
```diff
@@ -248,14 +270,32 @@ def gen_data(
...
+
+    data = data_generator(nrows, ncols, rand_low, rand_high)
+    data_cache[cache_key] = weakdict(data)
+
+    return data.copy()
```
anmyachev, line 298: Why `.copy()`?

asv_bench/benchmarks/utils.py
```diff
@@ -248,14 +270,32 @@ def gen_data(
...
+    cache_key = (data_type, nrows, ncols, rand_low, rand_high)
+    if cache_key in data_cache:
+        return data_cache[cache_key]
+
+    type_to_generator = {
+        "int": gen_int_data,
+        "str_int": gen_str_int_data,
+        "true_false_int": gen_true_false_int_data,
+    }
+    data_generator = type_to_generator.get(data_type, None)
```
anmyachev, line 292: `logging.info` also can be moved from `gen_int_data`, `gen_str_int_data` and `gen_true_false_int_data` to place between 291-292 lines (for example)

asv_bench/benchmarks/utils.py
```diff
@@ -447,3 +487,30 @@ def get_shape_id(shape: tuple) -> str:
...
+    data_type : {"int", "str_int", "true_false_int"}
+        Type of data generation.
+
+    Returns
+    -------
+    test_filenames : dict
+        Dcitionary that maps dataset shape to the file on disk.
+    """
+    test_filenames = {}
+    for shape in UNARY_OP_DATA_SIZE[ASV_DATASET_SIZE]:
```
anmyachev, line 510: It is dangerous enough to use a closure in a situation where the variable being closed is defined in several places.

```suggestion
def prepare_io_data(test_filename: str, data_type: str, shapes: list):
    """
    Prepare data for IO tests with caching.

    Parameters
    ----------
    test_filename : str
        Unique file identifier that is used to distinguish data
        for different tests.
    data_type : {"int", "str_int", "true_false_int"}
        Type of data generation.

    Returns
    -------
    test_filenames : dict
        Dcitionary that maps dataset shape to the file on disk.
    """
    test_filenames = {}
    for shape in shapes:
```

asv_bench/benchmarks/io/csv.py
```diff
@@ -69,13 +67,34 @@ def setup(self, test_filenames, shape, skiprows):
         self.skiprows = self.skiprows_mapping[skiprows] if skiprows else None
 
     def time_skiprows(self, test_filenames, shape, skiprows):
-        execute(
+        self.execute(
             IMPL[ASV_USE_IMPL].read_csv(
                 test_filenames[self.shape_id], skiprows=self.skiprows
             )
         )
```
anmyachev, line 74: What if we go this way?
```suggestion
        execute(
            IMPL[ASV_USE_IMPL].read_csv(
                test_filenames[self.shape_id], skiprows=self.skiprows
            ),
            trigger_omnisci_import=True, 
        )
```

asv_bench/benchmarks/utils.py
amyskov: It was needed to check the correctness of `data_type` parameter, but now your suggestion is implemented.

asv_bench/benchmarks/utils.py
amyskov: Not actually needed, so `.copy()` was removed.

asv_bench/benchmarks/utils.py
amyskov: Ok, implemented.

asv_bench/benchmarks/omnisci/benchmarks.py
```diff
@@ -47,7 +46,7 @@ class TimeJoin:
     def setup(self, shape, how):
         self.df1 = generate_dataframe(ASV_USE_IMPL, "int", *shape, RAND_LOW, RAND_HIGH)
         self.df2 = generate_dataframe(ASV_USE_IMPL, "int", *shape, RAND_LOW, RAND_HIGH)
-        trigger_import(self.df1, self.df2)
+        execute([self.df1, self.df2], trigger_omnisci_import=True)
```
anmyachev, line 49: Please do not change the rest of the benchmarks, as this will lead to not displaying previous results in dashboards.

I propose to transfer `trigger_import` to `asv_bench/benchmarks/utils.py` and use it in `execute` function.

Something like:
```python
    if trigger_omnisci_import:
        trigger_import(dfs)
        return
    if ASV_USE_IMPL == "modin":
        if ASV_USE_BACKEND == "omnisci":
            df._query_compiler._modin_frame._execute()
            return
        partitions = df._query_compiler._modin_frame._partitions
        all(
            map(
                lambda partition: partition.drain_call_queue() or True,
                partitions.flatten(),
```

amyskov: Ok, i see, your suggestion was implemented.

anmyachev (dismissed):
LGTM!

anmyachev:
@vnlitvinov can you review?

asv_bench/benchmarks/utils.py
```diff
@@ -158,30 +154,16 @@ def gen_int_data(nrows: int, ncols: int, rand_low: int, rand_high: int) -> dict:
...
     """
-    cache_key = ("int", nrows, ncols, rand_low, rand_high)
-    if cache_key in data_cache:
-        return data_cache[cache_key]
-
-    logging.info(
-        "Generating int data {} rows and {} columns [{}-{}]".format(
-            nrows, ncols, rand_low, rand_high
-        )
-    )
```
vnlitvinov, line 169: this doesn't seem like it has something to do with benchmarking true/false... can we split this up?

asv_bench/benchmarks/utils.py
```diff
@@ -447,3 +512,32 @@ def get_shape_id(shape: tuple) -> str:
...
+        for different tests.
+    data_type : {"int", "str_int", "true_false_int"}
+        Type of data generation.
+    shapes : list
+        Data shapes to prepare.
+
+    Returns
+    -------
+    test_filenames : dict
+        Dcitionary that maps dataset shape to the file on disk.
```
vnlitvinov, line 534:
```suggestion
        Dictionary that maps dataset shape to the file on disk.
```

vnlitvinov (approved):
LGTM

asv_bench/benchmarks/utils.py
amyskov: Ok, unrelated changes were reverted, corresponding issue was created https://github.com/modin-project/modin/issues/3257.

PR #2936, opened 2021-03-29
FIX-#2936: Improve performance of Modin XGBoost 

### LINKED ISSUE #2935: Improve performance of Modin XGBoost
### Describe the problem

Need to improve performance of Modin XGBoost by adding support of variable number of actors.

prutskov:
## What do these changes do?
This changes add performance improvements to xgboost using several ways:

1. Variable number of actors, instead of `actor per node` approach
2. Evenly distribution with minimal difference of number parts per node(max difference is 1) for nodes in cluster.
3. Additional manipulation with cluster resources using ray placement groups.

- [x] passes `flake8 modin`
- [x] Resolves #2935 
- [x] tests added and passing

### DIFF
--- a/modin/experimental/xgboost/xgboost_ray.py
+++ b/modin/experimental/xgboost/xgboost_ray.py
@@ -14,16 +14,17 @@
 
 import time
 import logging
-from typing import Dict, Optional
-from multiprocessing import cpu_count
-from collections import OrderedDict
+from typing import Dict, List
+import math
+from collections import defaultdict
 
 import xgboost as xgb
 import ray
 from ray.services import get_node_ip_address
+from ray.util.placement_group import placement_group, remove_placement_group
 import pandas
 
-from modin.distributed.dataframe.pandas import unwrap_partitions, from_partitions
+from modin.distributed.dataframe.pandas import from_partitions
 from .utils import RabitContext, RabitContextManager
 
 LOGGER = logging.getLogger("[modin.xgboost]")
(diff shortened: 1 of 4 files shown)


devin-petersohn (approved):
@prutskov LGTM!

modin/experimental/xgboost/xgboost_ray.py
```diff
@@ -339,67 +432,58 @@ def _train(
...
+        for actor in actors
     ]
 
-    results_to_sort = list()
-    for ip, part_res in zip(actors, predictions):
-        results_to_sort.extend(list(zip(part_res, order_of_parts[ip])))
-
-    results = sorted(results_to_sort, key=lambda l: l[1])
-    results = [part_res for part_res, _ in results]
+    ray.wait([part for _, part in predictions], num_returns=len(predictions))
```
YarShev, line 483: Shouldn't we assign returned objects to `partitions` and then create DataFrame from them?

modin/experimental/xgboost/xgboost_ray.py
```diff
@@ -163,146 +209,193 @@ def _split_data_across_actors(
...
+
+            for rank, ip in enumerate(
+                ray.get([actor.get_ip.remote() for actor in actors])
+            ):
+                actor_ips[ip].append(rank)
+
+            # Get distribution of parts between nodes ({ip:[(part, position),..],..})
+            init_parts_distribution = defaultdict(list)
+            for idx, (ip, part_ref) in enumerate(
+                zip(ray.get(list(parts_ips_ref)), parts_ref)
```
YarShev, line 262: `list` call is not necessary because `parts_ips_ref` is a list.

YarShev (commented):
cc @prutskov

PR #2958, opened 2021-04-05
TEST-#2951: add ASV config for omnisci backend

### LINKED ISSUE #2951: [ASV] show performance results on omnisci backend in dashboard
This will require:
* separate ASV configuration (`asv.conf.json`);
* the ability to create an environment from a yml file. This functionality is missing in `asv==0.4.2`, but available from master branch;
* for newer versions, the data storage format has changed, it is argued that it is enough to do `asv update`;
* new approach for trigger computation;

anmyachev:

## What do these changes do?

* Add separate configuration for omnisci backend - `asv_bench/asv.conf.omnisci.json`, since the configuration name is different from the default (`asv.conf.json`) - we should explicitly specify it through the `--config` parameter
* Add trigger of omnisci execution via `frame._execute`
* Add new ASV version which is needed to be able to create an conda environment from yml file
* Benchmarking only working subset of all operations

- [x] passes `flake8 modin`
- [x] Resolves #2951 
- [ ] tests added and passing

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -359,30 +359,30 @@ jobs:
       - uses: actions/checkout@v2
         with:
           fetch-depth: 1
-      - uses: conda-incubator/setup-miniconda@v2
+      - uses: actions/setup-python@v2
         with:
-          activate-environment: modin
-          environment-file: environment-dev.yml
-          python-version: 3.7
-          channel-priority: strict
-          use-only-tar-bz2: true # IMPORTANT: This needs to be set for caching to work properly!
-      - name: Conda environment
-        shell: bash -l {0}
-        run: |
-          conda info
-          conda list
-
+          python-version: "3.7.x"
+          architecture: "x64"
+      - name: ASV installation
+        run: pip install git+https://github.com/airspeed-velocity/asv.git@ef016e233cb9a0b19d517135104f49e0a3c380e9
       - name: Running benchmarks
-        shell: bash -l {0}
         run: |
-          pip install -e .
+          # ASV correctly creates environments for testing only from the branch
+          # with `master` name
+          git checkout -b master
           cd asv_bench
-          asv check -E existing
+          asv check -v
           git remote add upstream https://github.com/modin-project/modin.git
           git fetch upstream
           if git diff upstream/master --name-only | grep -q "^asv_bench/"; then
               asv machine --yes
-              asv run --quick --show-stderr --python=same --launch-method=spawn | sed "/failed$/ s/^/##[error]/" | tee benchmarks.log
+              asv run --quick --show-stderr --launch-method=spawn \
+                | sed "/failed$/ s/^/##[error]/" | tee benchmarks.log
+              # check omnisci backend
+              MODIN_BACKEND=omnisci MODIN_EXPERIMENTAL=true asv run --quick \
+                --show-stderr --launch-method=spawn --config asv.conf.omnisci.json \
+                -b time_head -b time_iloc -b time_loc -b time_columns -b time_index -b time_shape \
+                | sed "/failed$/ s/^/##[error]/" | tee benchmarks.log
               if grep "failed" benchmarks.log > /dev/null ; then
                   exit 1
               fi
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -381,7 +381,7 @@ jobs:
               # check omnisci backend
               MODIN_BACKEND=omnisci MODIN_EXPERIMENTAL=true asv run --quick \
                 -b ^omnisci.benchmarks.TimeHead -b ^omnisci.benchmarks.TimeReadCsvNames \
-                --show-stderr --launch-method=spawn --config asv.conf.omnisci.json -v || echo $?
+                --show-stderr --launch-method=forkserver --config asv.conf.omnisci.json -v || echo $?
 
               conda init bash
               conda activate $(conda env list | grep .asv)
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -417,6 +417,9 @@ jobs:
     env:
       MODIN_ENGINE: ${{matrix.engine}}
       MODIN_MEMORY: 1000000000
+      MKL_NUM_THREADS: 1
+      NUMEXPR_NUM_THREADS: 1
+      OMP_NUM_THREADS: 1
     name: test-ubuntu (engine ${{matrix.engine}}, python ${{matrix.python-version}})
     steps:
       - uses: actions/checkout@v2
@@ -520,6 +523,9 @@ jobs:
       MODIN_ENGINE: "python"
       MODIN_EXPERIMENTAL: "True"
       MODIN_MEMORY: 1000000000
+      MKL_NUM_THREADS: 1
+      NUMEXPR_NUM_THREADS: 1
+      OMP_NUM_THREADS: 1
     name: test cloud
     steps:
       - uses: actions/checkout@v2
@@ -582,6 +588,9 @@ jobs:
     env:
       MODIN_ENGINE: ${{matrix.engine}}
       MODIN_MEMORY: 1000000000
+      MKL_NUM_THREADS: 1
+      NUMEXPR_NUM_THREADS: 1
+      OMP_NUM_THREADS: 1
     name: test-windows
     steps:
       - uses: actions/checkout@v2
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -376,12 +376,11 @@ jobs:
           git fetch upstream
           if git diff upstream/master --name-only | grep -q "^asv_bench/"; then
               asv machine --yes
-              asv run --quick --show-stderr --launch-method=spawn \
+              asv run --quick --show-stderr --launch-method=spawn -b ^benchmarks -b ^io -b ^scalability \
                 | sed "/failed$/ s/^/##[error]/" | tee benchmarks.log
               # check omnisci backend
-              MODIN_BACKEND=omnisci MODIN_EXPERIMENTAL=true asv run --quick \
+              MODIN_BACKEND=omnisci MODIN_EXPERIMENTAL=true asv run --quick -b ^omnisci \
                 --show-stderr --launch-method=spawn --config asv.conf.omnisci.json \
-                -b time_head -b time_iloc -b time_loc -b time_columns -b time_index -b time_shape \
                 | sed "/failed$/ s/^/##[error]/" | tee benchmarks.log
               if grep "failed" benchmarks.log > /dev/null ; then
                   exit 1
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -114,7 +114,7 @@ jobs:
           python-version: "3.7.x"
           architecture: "x64"
       - run: pip install flake8 flake8-print
-      - run: flake8 modin/ asv_bench/benchmarks scripts/doc_checker.py
+      - run: flake8 --enable=T modin/ asv_bench/benchmarks scripts/doc_checker.py
 
   test-api:
     runs-on: ubuntu-latest
@@ -350,6 +350,9 @@ jobs:
   test-asv-benchmarks:
     needs: [lint-commit, lint-flake8, lint-black, test-api, test-headers]
     runs-on: ubuntu-latest
+    defaults:
+      run:
+        shell: bash -l {0}
     env:
       MODIN_ENGINE: ray
       MODIN_MEMORY: 1000000000
@@ -359,12 +362,16 @@ jobs:
       - uses: actions/checkout@v2
         with:
           fetch-depth: 1
-      - uses: actions/setup-python@v2
+      - uses: conda-incubator/setup-miniconda@v2
         with:
-          python-version: "3.7.x"
-          architecture: "x64"
+          auto-activate-base: true
+          activate-environment: ""
       - name: ASV installation
-        run: pip install git+https://github.com/airspeed-velocity/asv.git@ef016e233cb9a0b19d517135104f49e0a3c380e9
+        run: |
+          # FIXME: use the tag or release version of ASV as soon as it appears;
+          # The ability to build a conda environment by specifying yml file has not
+          # yet appeared in the release versions;
+          pip install git+https://github.com/airspeed-velocity/asv.git@ef016e233cb9a0b19d517135104f49e0a3c380e9
       - name: Running benchmarks
         run: |
           # ASV correctly creates environments for testing only from the branch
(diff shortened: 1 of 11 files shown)


anmyachev:
It is necessary to find out why some functions work during testing, but not during benchmarking.

For example, `concat`, `append`, `merge` and so on.

anmyachev:
`append` benchmark failed by timeout if run via ASV

.github/workflows/ci.yml
```diff
@@ -319,30 +319,30 @@ jobs:
...
-      - name: Conda environment
-        shell: bash -l {0}
-        run: |
-          conda info
-          conda list
-
+          python-version: "3.7.x"
+          architecture: "x64"
+      - name: ASV installation
+        run: pip install git+https://github.com/airspeed-velocity/asv.git@ef016e233cb9a0b19d517135104f49e0a3c380e9
```
vnlitvinov, line 327: can we use a tagged something instead of raw commit?..

asv_bench/benchmarks/scalability/scalability_benchmarks.py
```diff
@@ -13,7 +13,11 @@
 
 import modin.pandas as pd
 from modin.pandas.utils import from_pandas
-from modin.utils import to_pandas
+
+try:
+    from modin.utils import to_pandas
+except ImportError:
+    from modin.pandas.utils import to_pandas
```
vnlitvinov, line 20: why?

environment-dev.yml
```diff
@@ -34,9 +34,9 @@ dependencies:
   - rpyc==4.1.5
   - cloudpickle==1.4.1
   - boto3
-  - asv
   - ray-core>=1.0.0,<1.2.0
   - pip:
       - xgboost>=1.3
       - modin-spreadsheet>=0.1.1
       - tqdm
+      - git+https://github.com/airspeed-velocity/asv.git@ef016e233cb9a0b19d517135104f49e0a3c380e9
```
vnlitvinov, line 42: same about using a tag instead of raw commit hash (and in other places)

asv_bench/benchmarks/scalability/scalability_benchmarks.py
anmyachev: This is a compatibility question - to be able to run benchmarks on old Modin commits.

.github/workflows/ci.yml
anmyachev: There is no such possibility - the previous release was about a year ago, it is not known when the next one will be.
https://github.com/airspeed-velocity/asv/issues/967

vnlitvinov: Please add a comment anywhere then describing the reasons (and add a `FIXME` to use a tag when a release is done).

asv_bench/benchmarks/scalability/scalability_benchmarks.py
devin-petersohn: @anmyachev Would you add a short comment here explaining?

devin-petersohn (changes requested):
Looks good @anmyachev, please add the comments requested by @vnlitvinov and me.

anmyachev:
Need:
* add data import trigger into OmniSci;
* revise dimensions of test data;

asv_bench/benchmarks/omnisci/benchmarks.py
```diff
@@ -0,0 +1,163 @@
...
+        OmnisciServer,
+    )
+
+    for df in dfs:
+        df.shape  # to trigger real execution
+        df._query_compiler._modin_frame._partitions[0][
+            0
+        ].frame_id = OmnisciServer().put_arrow_to_omnisci(
+            df._query_compiler._modin_frame._partitions[0][0].get()
+        )  # to trigger real execution
```
vnlitvinov, line 56: @amyskov @ienkovich is this the only way to trigger? don't we have a better (simpler) way? if not I think we should make a protected API call to avoid inserting those ugly constructs...

asv_bench/benchmarks/omnisci/benchmarks.py
```diff
@@ -0,0 +1,163 @@
...
+        self.df1 = generate_dataframe(
+            ASV_USE_IMPL, "int", *shapes[0], RAND_LOW, RAND_HIGH
+        )
+        self.df2 = generate_dataframe(
+            ASV_USE_IMPL, "int", *shapes[1], RAND_LOW, RAND_HIGH
+        )
+        trigger_import((self.df1, self.df2))
+
+    def time_merge(self, shapes, how):
+        # merge dataframes on index to get the predictable shape doesn't support
```
vnlitvinov, line 76: could you please re-phrase the comment? I cannot understand what it means right now :(

asv_bench/benchmarks/omnisci/benchmarks.py
```diff
@@ -0,0 +1,163 @@
...
+    "big": [
+        (1_000_000, 10),
+    ],
+    "small": [
+        (10_000, 10),
+    ],
+}
+
+
+def trigger_import(dfs: list):
```
vnlitvinov, line 45:
```suggestion
def trigger_import(*dfs):
```
IMHO it's easier to just make it accept arbitrary amount of parameters for the future usage of this function

asv_bench/benchmarks/omnisci/benchmarks.py
```diff
@@ -0,0 +1,163 @@
...
+        if indexer_type == "bool":
+            self.indexer = [False, True] * (shape[0] // 2)
+        elif indexer_type == "scalar":
+            self.indexer = shape[0] // 2
+        elif indexer_type == "slice":
+            self.indexer = slice(0, shape[0], 2)
+        elif indexer_type == "list":
+            self.indexer = [x for x in range(shape[0])]
+        elif indexer_type == "function":
+            self.indexer = lambda df: df.index[::-2]
```
vnlitvinov, line 107:
```suggestion
        self.indexer = {
            "bool": [False, True] * (shape[0] // 2),
            "scalar": shape[0] // 2,
            "slice": slice(0, shape[0], 2),
            "list": list(range(shape[0])),
            "function": lambda df: df.index[::-2]
        }[indexer_type]
```
this is shorter and (IMHO) easier to read version with a built-in check for unknown value of `indexer_type`

asv_bench/benchmarks/omnisci/benchmarks.py
```diff
@@ -0,0 +1,163 @@
...
+
+    def time_iloc(self, shape, indexer_type):
+        execute(self.df.iloc[self.indexer])
+
+    def time_loc(self, shape, indexer_type):
+        execute(self.df.loc[self.indexer])
```
vnlitvinov, line 113: why do you accept `shape` and `indexer_type` here if you don't use them?

asv_bench/benchmarks/omnisci/benchmarks.py
```diff
@@ -0,0 +1,163 @@
...
+    def setup(self, shape, head_count):
+        self.df = generate_dataframe(ASV_USE_IMPL, "int", *shape, RAND_LOW, RAND_HIGH)
+        trigger_import((self.df,))
+        self.head_count = (
+            int(head_count * len(self.df.index))
+            if isinstance(head_count, float)
+            else head_count
+        )
+
+    def time_head(self, shape, head_count):
```
vnlitvinov, line 132: why accept `shape` and `head_count` if you don't use them?

asv_bench/benchmarks/omnisci/benchmarks.py
```diff
@@ -0,0 +1,163 @@
...
+
+
+"""
+test = TimeJoin()
+
+import pdb;pdb.set_trace()
+test.setup(((500_000, 20), (1_000_000, 10)), "left", False)
+
+test.time_join(((500_000, 20), (1_000_000, 10)), "left", False)
+"""
```
vnlitvinov, line 163: what's this?

anmyachev: removed

asv_bench/benchmarks/omnisci/benchmarks.py
anmyachev: `setup` and `time_*` function signatures should be equal - ASV limitation

asv_bench/benchmarks/omnisci/benchmarks.py
```diff
@@ -42,7 +42,7 @@
 }
 
 
-def trigger_import(dfs: list):
+def trigger_import(*dfs):
```
vnlitvinov, line 45: you have to fix the callers of `trigger_import` now ;)

vnlitvinov (changes requested):

.github/workflows/ci.yml
anmyachev: Added

asv_bench/benchmarks/scalability/scalability_benchmarks.py
anmyachev: added

anmyachev:
Interesting:
* ASV builds a conda environment correctly only when it is launched from some conda env (at least from `base`).
* Benchmarks running on OmniSci engine cause a timeout error if `--launch-method=spawn` is used; works correctly with `--launch-method=forkserver`

asv_bench/asv.conf.omnisci.json
```diff
@@ -72,9 +72,9 @@
     // the package name by 'pip+', and the package will be installed via
     // pip (with all the conda available packages installed first,
     // followed by the pip installed packages).
-    "matrix": {
-        "omniscidbe4py": "",
-    },
+    //"matrix": {
+    //    "omniscidbe4py": "",
+    //},
```
vnlitvinov, line 77: why?

asv_bench/benchmarks/omnisci/benchmarks.py
```diff
@@ -81,6 +79,192 @@ def time_merge(self, shapes, how):
...
+    ]
+
+    def setup(self, shape, columns_number, ascending_list):
+        self.df = generate_dataframe(ASV_USE_IMPL, "int", *shape, RAND_LOW, RAND_HIGH)
+        trigger_import(self.df)
+        self.columns = random_columns(self.df.columns, columns_number)
+        self.ascending = (
+            random_booleans(columns_number)
+            if ascending_list
+            else bool(random_booleans(1)[0])
```
vnlitvinov, line 163: could you please explain the difference between `then` and `else` clauses of this `if`?

asv_bench/benchmarks/omnisci/benchmarks.py
```diff
@@ -13,47 +13,45 @@
 
 """General Modin on OmniSci backend benchmarks."""
 
+import modin.pandas as pd
```
vnlitvinov, line 16: why? also in Modin before 0.9 this would initialize the engine and backend...

asv_bench/benchmarks/utils.py
```diff
@@ -67,27 +67,21 @@
...
     "big": [
         ((5000, 5000), (5000, 5000)),
-        # the case extremely inefficient
-        # ((20, 500_000), (10, 1_000_000)),
         ((500_000, 20), (1_000_000, 10)),
     ],
     "small": [
         ((250, 250), (250, 250)),
-        ((20, 10_000), (10, 25_000)),
         ((10_000, 20), (25_000, 10)),
```
vnlitvinov, line 74: why do you touch this (seemingly not OmniSci-related) part in OmniSci-related PR?

asv_bench/asv.conf.omnisci.json
anmyachev: omniscidbe4py already in `env_omnisci.yml` file that is used to create a conda environment. `conda install ...` command is called by ASV itself.

vnlitvinov: then remove the section, what's the point of keeping it commented?..

anmyachev: To know which fields can be changed if necessary (without looking for the documentation)

anmyachev: If removed, then all unused fields. What do you think?

asv_bench/benchmarks/omnisci/benchmarks.py
anmyachev: It is possible to define the sort order for each column, if multiple columns are used for sorting.
In `then` clause returns list of bool values, in `else` clause only one bool value defining the order for all columns at once

asv_bench/asv.conf.omnisci.json
vnlitvinov: I don't have a strong preference in this particular case, but usually I don't like comments being used instead of documentation for some third party stuff. So I would prefer we remove all stuff that isn't used.

asv_bench/benchmarks/utils.py
anmyachev: removed

asv_bench/benchmarks/omnisci/benchmarks.py
anmyachev: Now I am using the module defined in the `IMPL` variable.
This side-effects shouldn't affect omnisci benchmarks.

asv_bench/benchmarks/utils.py
```diff
@@ -28,7 +28,7 @@
 from typing import Optional, Union
 
 RAND_LOW = 0
-RAND_HIGH = 1_000_000_000
```
vnlitvinov: why?

vnlitvinov (commented):
Why again are you changing non-omnisci `utils.py` in an omnisci-related PR?..

vnlitvinov (dismissed):
Should be good when CI finally passes.

anmyachev:
> Should be good when CI finally passes.

CI is green (except codecov) :)

vnlitvinov:
@devin-petersohn requested changes as well. **Devin**, are you fine with current version of the PR?

anmyachev:
Merge conflicts fixed. @devin-petersohn ready for review.

vnlitvinov (dismissed):
LGTM, but please wait for @devin-petersohn to confirm he's fine.

devin-petersohn (approved):
I'm okay with these changes. Thanks @anmyachev

.github/workflows/ci.yml
```diff
@@ -411,6 +417,9 @@ jobs:
     env:
       MODIN_ENGINE: ${{matrix.engine}}
       MODIN_MEMORY: 1000000000
+      MKL_NUM_THREADS: 1
```
YarShev, line 420: @anmyachev, which threading technologies are hidden by these env vars except openmp?

anmyachev: I don't know what technologies they use internally, but even if they are all based on the same technology, isn't it better to explicitly specify different environment variables rather than hope that `OMP_NUM_THREADS` will disable parallelization everywhere? [Dask](https://docs.dask.org/en/latest/array-best-practices.html#avoid-oversubscribing-threads) recommends doing just that. (OPENBLAS_NUM_THREADS should be added too)

YarShev: I think If all the packages use the same threading technology, there is no point to export different env vars. However, that is true if we exactly know there are no other threading technologies under the hood.

Issue #2934, opened 2021-03-29
Docs everywhere 📜


vnlitvinov:
## General

We want to have Modin thoroughly documented and every nook and cranny covered up.

This includes creating (or updating, whatever fits each function) docstrings for all functions and modules as well as writing a higher-level documentation on control and data flows and class interactions. Each module has to have a short docstring right beneath the license header section which would describe what this module houses.

Docstrings should be placed in-line (obviously), higher-level "flow" documentation should be placed in `docs/flow/` as separate Markdown documents mimicking the directory layout of modules being documented. Control and data flows are meant to describe _what_ happens to an object (and what objects are inputs and outputs of methods), not _how_ methods transform the objects. Example: "`.squeeze()` takes a Modin Frame and produces a Pandas Series because Series objects are internally stored as 1xN frames" _(the example text might be wrong as I haven't studied the function itself yet, but I hope it conveys the meaning)_.

The agreement on documentation is the following: we make 1 PR per 1 tracker which adds docstrings for one "zone" (zones are listed below) _and_  adds flow documentation. Please tag @aregm and @vnlitvinov when docstrings part is ready for reviewing of flow part.

## Zones
Note that each zone has a priority assigned. These priorities apply for each person individually, not for the global scope.
So you should find your issues and work on them in order indicated by priorities.

| What | Assignees | Size, KB | Size, KLoC | Priority |
| -- | -- | -- | -- | -- |
| asv code, all CI flows (github+teamcity) | @anmyachev | 92.6 | 3 | 1 |
| `modin.data_management` | @dchigarev | 35.9 | 1.2 | 1 |
| `modin.pandas/base.py` | @gshimansky | 77.4 | 3.1 | 1 |
| `modin.pandas/dataframe.py` | @vnlitvinov | 56.9 | 2.2 | 2 |
| `modin.pandas/series.py` | @vnlitvinov | 40.5 | 1.5 | 2 |
| `modin.backends.pandas` | @dchigarev | 96.4 | 3 | 2 |
| `modin.backends.base`, `modin.backends.pyarrow` | @dchigarev @ienkovich  | 55.2 | 1.8 | 3 |
| Other `modin.pandas` modules | @dchigarev @gshimansky | 101.0 | 3.8 | 2 |
| `modin.config` | @vnlitvinov | 10.2 | 0.4 | 3 |
| `modin.distributed` | @YarShev | 4.9 | 0.1 | 2 |
[...]

PR #2968, opened 2021-04-09
TEST-#2966: add script for testing docs

### LINKED ISSUE #2966: Automated tool for checking extra quirks of numpydoc which pydocstyle do not catch
Initial implementation of said tool is here: https://github.com/modin-project/modin/issues/2934#issuecomment-816125767

I think we need it turned into a script accepting multiple filenames or even directories and be placed under `scripts/` (and added into CI).

anmyachev:

## What do these changes do?

- [x] passes `flake8 modin`
- [x] Resolves #2966 
- [ ] tests added and passing

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,9 +45,9 @@ jobs:
         with:
           python-version: "3.7.x"
           architecture: "x64"
-      - run: pip install pytest pytest-cov
+      - run: pip install --no-deps -e .
+      - run: pip install pytest pytest-cov pydocstyle numpydoc
       - run: pytest scripts/test
-      - run: pip install pydocstyle numpydoc
       - run: pip install -e .[all]
       - run: |
           python scripts/doc_checker.py --add-ignore=D101,D102,D105 --disable-numpydoc \
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -48,24 +48,24 @@ jobs:
       - run: pip install pydocstyle numpydoc
       - run: pip install -e .[all]
       - run: |
-          python scripts/doc-checker.py --add-ignore=D101,D102,D105 --no-use-numpydoc \
+          python scripts/doc-checker.py --add-ignore=D101,D102,D105 --disable-numpydoc \
             modin/pandas/dataframe.py modin/pandas/series.py \
             modin/pandas/groupby.py modin/pandas/base.py
       - run: |
-          python scripts/doc-checker.py --add-ignore=D101,D103,D105 --no-use-numpydoc \
+          python scripts/doc-checker.py --add-ignore=D101,D103,D105 --disable-numpydoc \
             modin/pandas/io.py
       - run: |
-          python scripts/doc-checker.py --add-ignore=D101,D102 --no-use-numpydoc \
+          python scripts/doc-checker.py --add-ignore=D101,D102 --disable-numpydoc \
             modin/pandas/series_utils.py
       - run: |
-          python scripts/doc-checker.py --add-ignore=D103 --no-use-numpydoc \
+          python scripts/doc-checker.py --add-ignore=D103 --disable-numpydoc \
             modin/pandas/general.py
       - run: |
-          python scripts/doc-checker.py --no-use-numpydoc \
+          python scripts/doc-checker.py --disable-numpydoc \
             modin/pandas/plotting.py modin/pandas/utils.py \
             modin/pandas/iterator.py modin/pandas/indexing.py
       - run: |
-          python scripts/doc-checker.py --add-ignore=D100,D104 --no-use-numpydoc \
+          python scripts/doc-checker.py --add-ignore=D100,D104 --disable-numpydoc \
             modin/engines/base/frame
       - run: |
           python scripts/doc-checker.py --add-ignore=D100,D101,D102 asv_bench/benchmarks \
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -46,6 +46,7 @@ jobs:
           python-version: "3.7.x"
           architecture: "x64"
       - run: pip install pydocstyle numpydoc click
+      - run: pip install -e .
       - run: |
           python scripts/doc-checker.py --add-ignore=D101,D102,D105 --use-numpydoc False \
             modin/pandas/dataframe.py modin/pandas/series.py \
@@ -66,6 +67,9 @@ jobs:
       - run: |
           python scripts/doc-checker.py --add-ignore=D100,D104 --use-numpydoc False \
             modin/engines/base/frame
+      - run: |
+          python scripts/doc-checker.py --add-ignore=D101,D102 asv_bench/benchmarks \
+            asv_bench/benchmarks/io asv_bench/benchmarks/scalability
 
   lint-flake8:
     name: lint (flake8)
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,8 +45,8 @@ jobs:
         with:
           python-version: "3.7.x"
           architecture: "x64"
-      - run: pip install pytest
-      - run: pytest scripts\test
+      - run: pip install pytest pytest-cov
+      - run: pytest scripts/test
       - run: pip install pydocstyle numpydoc
       - run: pip install -e .[all]
       - run: |
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,7 +32,7 @@ jobs:
           python-version: "3.7.x"
           architecture: "x64"
       - run: pip install black
-      - run: black --check --diff modin/ asv_bench/benchmarks scripts/
+      - run: black --check --diff modin/ asv_bench/benchmarks scripts/doc_checker.py
 
   lint-pydocstyle:
     name: lint (pydocstyle)
@@ -84,8 +84,7 @@ jobs:
           python-version: "3.7.x"
           architecture: "x64"
       - run: pip install flake8 flake8-print
-      - run: flake8 --enable=T modin/ asv_bench/benchmarks
-      - run: flake8 scripts/
+      - run: flake8 --enable=T modin/ asv_bench/benchmarks scripts/doc_checker.py
 
   test-api:
     runs-on: ubuntu-latest
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -51,9 +51,7 @@ jobs:
           python scripts/doc-checker.py --add-ignore=D101,D102,D105 --disable-numpydoc \
             modin/pandas/dataframe.py modin/pandas/series.py \
             modin/pandas/groupby.py modin/pandas/base.py
-      - run: |
-          python scripts/doc-checker.py --add-ignore=D101,D103,D105 --disable-numpydoc \
-            modin/pandas/io.py
+      - run: python scripts/doc-checker.py modin/pandas/io.py
       - run: |
           python scripts/doc-checker.py --add-ignore=D101,D102 --disable-numpydoc \
             modin/pandas/series_utils.py
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -46,6 +46,7 @@ jobs:
           python-version: "3.7.x"
           architecture: "x64"
       - run: pip install pydocstyle numpydoc
+      - run: pip install -e .[all]
       - run: |
           python scripts/doc-checker.py --add-ignore=D101,D102,D105 --use-numpydoc False \
             modin/pandas/dataframe.py modin/pandas/series.py \
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -48,24 +48,24 @@ jobs:
       - run: pip install pydocstyle numpydoc
       - run: pip install -e .[all]
       - run: |
-          python scripts/doc-checker.py --add-ignore=D101,D102,D105 --use-numpydoc False \
+          python scripts/doc-checker.py --add-ignore=D101,D102,D105 --no-use-numpydoc \
             modin/pandas/dataframe.py modin/pandas/series.py \
             modin/pandas/groupby.py modin/pandas/base.py
       - run: |
-          python scripts/doc-checker.py --add-ignore=D101,D103,D105 --use-numpydoc False \
+          python scripts/doc-checker.py --add-ignore=D101,D103,D105 --no-use-numpydoc \
             modin/pandas/io.py
       - run: |
-          python scripts/doc-checker.py --add-ignore=D101,D102 --use-numpydoc False \
+          python scripts/doc-checker.py --add-ignore=D101,D102 --no-use-numpydoc \
             modin/pandas/series_utils.py
       - run: |
-          python scripts/doc-checker.py --add-ignore=D103 --use-numpydoc False \
+          python scripts/doc-checker.py --add-ignore=D103 --no-use-numpydoc \
             modin/pandas/general.py
       - run: |
-          python scripts/doc-checker.py --use-numpydoc False \
+          python scripts/doc-checker.py --no-use-numpydoc \
             modin/pandas/plotting.py modin/pandas/utils.py \
             modin/pandas/iterator.py modin/pandas/indexing.py
       - run: |
-          python scripts/doc-checker.py --add-ignore=D100,D104 --use-numpydoc False \
+          python scripts/doc-checker.py --add-ignore=D100,D104 --no-use-numpydoc \
             modin/engines/base/frame
       - run: |
           python scripts/doc-checker.py --add-ignore=D100,D101,D102 asv_bench/benchmarks \
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,6 +45,8 @@ jobs:
         with:
           python-version: "3.7.x"
           architecture: "x64"
+      - run: pip install pytest
+      - run: pytest scripts\test
       - run: pip install pydocstyle numpydoc
       - run: pip install -e .[all]
       - run: |
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -69,6 +69,7 @@ jobs:
       - run: |
           python scripts/doc-checker.py --add-ignore=D100,D101,D102 asv_bench/benchmarks \
             asv_bench/benchmarks/io asv_bench/benchmarks/scalability
+      - run: python scripts/doc-checker.py scripts/doc-checker.py
 
   lint-flake8:
     name: lint (flake8)
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,13 +45,27 @@ jobs:
         with:
           python-version: "3.7.x"
           architecture: "x64"
-      - run: pip install pydocstyle
-      - run: pydocstyle --convention=numpy --add-ignore=D101,D102,D105 modin/pandas/dataframe.py modin/pandas/series.py modin/pandas/groupby.py modin/pandas/base.py
-      - run: pydocstyle --convention=numpy --add-ignore=D101,D103,D105 modin/pandas/io.py
-      - run: pydocstyle --convention=numpy --add-ignore=D101,D102 modin/pandas/series_utils.py
-      - run: pydocstyle --convention=numpy --add-ignore=D103 modin/pandas/general.py
-      - run: pydocstyle --convention=numpy modin/pandas/plotting.py modin/pandas/utils.py modin/pandas/iterator.py modin/pandas/indexing.py
-      - run: pydocstyle --convention=numpy --add-ignore=D100,D104 modin/engines/base/frame
+      - run: pip install pydocstyle numpydoc click
+      - run: |
+          python scripts/doc-checker.py --add-ignore=D101,D102,D105 --use-numpydoc False \
+            modin/pandas/dataframe.py modin/pandas/series.py \
+            modin/pandas/groupby.py modin/pandas/base.py
+      - run: |
+          python scripts/doc-checker.py --add-ignore=D101,D103,D105 --use-numpydoc False \
+            modin/pandas/io.py
+      - run: |
+          python scripts/doc-checker.py --add-ignore=D101,D102 --use-numpydoc False \
+            modin/pandas/series_utils.py
+      - run: |
+          python scripts/doc-checker.py --add-ignore=D103 --use-numpydoc False \
+            modin/pandas/general.py
+      - run: |
+          python scripts/doc-checker.py --use-numpydoc False \
+            modin/pandas/plotting.py modin/pandas/utils.py \
+            modin/pandas/iterator.py modin/pandas/indexing.py
+      - run: |
+          python scripts/doc-checker.py --add-ignore=D100,D104 --use-numpydoc False \
+            modin/engines/base/frame
 
   lint-flake8:
     name: lint (flake8)
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,7 +45,7 @@ jobs:
         with:
           python-version: "3.7.x"
           architecture: "x64"
-      - run: pip install pydocstyle numpydoc click
+      - run: pip install pydocstyle numpydoc
       - run: |
           python scripts/doc-checker.py --add-ignore=D101,D102,D105 --use-numpydoc False \
             modin/pandas/dataframe.py modin/pandas/series.py \
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -46,7 +46,6 @@ jobs:
           python-version: "3.7.x"
           architecture: "x64"
       - run: pip install pydocstyle numpydoc click
-      - run: pip install -e .
       - run: |
           python scripts/doc-checker.py --add-ignore=D101,D102,D105 --use-numpydoc False \
             modin/pandas/dataframe.py modin/pandas/series.py \
@@ -83,7 +82,8 @@ jobs:
           python-version: "3.7.x"
           architecture: "x64"
       - run: pip install flake8 flake8-print
-      - run: flake8 --enable=T modin/ asv_bench/benchmarks scripts/
+      - run: flake8 --enable=T modin/ asv_bench/benchmarks
+      - run: flake8 scripts/
 
   test-api:
     runs-on: ubuntu-latest
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -48,28 +48,28 @@ jobs:
       - run: pip install pydocstyle numpydoc
       - run: pip install -e .[all]
       - run: |
-          python scripts/doc-checker.py --add-ignore=D101,D102,D105 --disable-numpydoc \
+          python scripts/doc_checker.py --add-ignore=D101,D102,D105 --disable-numpydoc \
             modin/pandas/dataframe.py modin/pandas/series.py \
             modin/pandas/groupby.py modin/pandas/base.py
-      - run: python scripts/doc-checker.py modin/pandas/io.py
+      - run: python scripts/doc_checker.py modin/pandas/io.py
       - run: |
-          python scripts/doc-checker.py --add-ignore=D101,D102 --disable-numpydoc \
+          python scripts/doc_checker.py --add-ignore=D101,D102 --disable-numpydoc \
             modin/pandas/series_utils.py
       - run: |
-          python scripts/doc-checker.py --add-ignore=D103 --disable-numpydoc \
+          python scripts/doc_checker.py --add-ignore=D103 --disable-numpydoc \
             modin/pandas/general.py
       - run: |
-          python scripts/doc-checker.py --disable-numpydoc \
+          python scripts/doc_checker.py --disable-numpydoc \
             modin/pandas/plotting.py modin/pandas/utils.py \
             modin/pandas/iterator.py modin/pandas/indexing.py
       - run: |
-          python scripts/doc-checker.py --add-ignore=D100,D104 --disable-numpydoc \
+          python scripts/doc_checker.py --add-ignore=D100,D104 --disable-numpydoc \
             modin/engines/base/frame
       - run: |
-          python scripts/doc-checker.py asv_bench/benchmarks/utils.py \
+          python scripts/doc_checker.py asv_bench/benchmarks/utils.py \
             asv_bench/benchmarks/__init__.py asv_bench/benchmarks/io/__init__.py \
             asv_bench/benchmarks/scalability/__init__.py
-      - run: python scripts/doc-checker.py scripts/doc-checker.py
+      - run: python scripts/doc_checker.py scripts/doc_checker.py
 
   lint-flake8:
     name: lint (flake8)
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -68,7 +68,7 @@ jobs:
           python scripts/doc-checker.py --add-ignore=D100,D104 --use-numpydoc False \
             modin/engines/base/frame
       - run: |
-          python scripts/doc-checker.py --add-ignore=D101,D102 asv_bench/benchmarks \
+          python scripts/doc-checker.py --add-ignore=D100,D101,D102 asv_bench/benchmarks \
             asv_bench/benchmarks/io asv_bench/benchmarks/scalability
 
   lint-flake8:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,7 +45,6 @@ jobs:
         with:
           python-version: "3.7.x"
           architecture: "x64"
-      - run: pip install --no-deps -e .
       - run: pip install pytest pytest-cov pydocstyle numpydoc
       - run: pytest scripts/test
       - run: pip install -e .[all]
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -66,8 +66,9 @@ jobs:
           python scripts/doc-checker.py --add-ignore=D100,D104 --disable-numpydoc \
             modin/engines/base/frame
       - run: |
-          python scripts/doc-checker.py --add-ignore=D100,D101,D102 asv_bench/benchmarks \
-            asv_bench/benchmarks/io asv_bench/benchmarks/scalability
+          python scripts/doc-checker.py asv_bench/benchmarks/utils.py \
+            asv_bench/benchmarks/__init__.py asv_bench/benchmarks/io/__init__.py \
+            asv_bench/benchmarks/scalability/__init__.py
       - run: python scripts/doc-checker.py scripts/doc-checker.py
 
   lint-flake8:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,7 +32,7 @@ jobs:
           python-version: "3.7.x"
           architecture: "x64"
       - run: pip install black
-      - run: black --check --diff modin/ asv_bench/benchmarks
+      - run: black --check --diff modin/ asv_bench/benchmarks scripts/
 
   lint-pydocstyle:
     name: lint (pydocstyle)
@@ -83,7 +83,7 @@ jobs:
           python-version: "3.7.x"
           architecture: "x64"
       - run: pip install flake8 flake8-print
-      - run: flake8 --enable=T modin/ asv_bench/benchmarks
+      - run: flake8 --enable=T modin/ asv_bench/benchmarks scripts/
 
   test-api:
     runs-on: ubuntu-latest
--- a/asv_bench/benchmarks/benchmarks.py
+++ b/asv_bench/benchmarks/benchmarks.py
@@ -11,7 +11,7 @@
 # ANY KIND, either express or implied. See the License for the specific language
 # governing permissions and limitations under the License.
 
-"""General modin benchmarks."""
+"""General Modin benchmarks."""
 
 # define `MODIN_CPUS` env var to control the number of partitions
 # it should be defined before modin.pandas import (in case of using os.environ)
--- a/asv_bench/benchmarks/benchmarks.py
+++ b/asv_bench/benchmarks/benchmarks.py
@@ -11,7 +11,7 @@
 # ANY KIND, either express or implied. See the License for the specific language
 # governing permissions and limitations under the License.
 
-"""General modin benchmarks"""
+"""General modin benchmarks."""
 
 # define `MODIN_CPUS` env var to control the number of partitions
 # it should be defined before modin.pandas import (in case of using os.environ)
--- a/asv_bench/benchmarks/io/__init__.py
+++ b/asv_bench/benchmarks/io/__init__.py
@@ -11,4 +11,4 @@
 # ANY KIND, either express or implied. See the License for the specific language
 # governing permissions and limitations under the License.
 
-"""IO modin benchmarks"""
+"""IO modin benchmarks."""
--- a/asv_bench/benchmarks/io/__init__.py
+++ b/asv_bench/benchmarks/io/__init__.py
@@ -11,4 +11,4 @@
 # ANY KIND, either express or implied. See the License for the specific language
 # governing permissions and limitations under the License.
 
-"""IO modin benchmarks."""
+"""IO Modin benchmarks."""
--- a/asv_bench/benchmarks/io/csv.py
+++ b/asv_bench/benchmarks/io/csv.py
@@ -26,10 +26,6 @@ from ..utils import (
     get_shape_id,
 )
 
-# ray init
-if ASV_USE_IMPL == "modin":
-    pd.DataFrame([])
-
 
 class BaseReadCsv:
     # test data file can de created only once
@@ -44,6 +40,9 @@ class BaseReadCsv:
         return test_filenames
 
     def setup(self, test_filenames, shape, *args, **kwargs):
+        # ray init
+        if ASV_USE_IMPL == "modin":
+            pd.DataFrame([])
         self.shape_id = get_shape_id(shape)
 
 
@@ -121,6 +120,9 @@ class TimeReadCsvNamesDtype:
         return cache
 
     def setup(self, cache, shape, names, dtype):
+        # ray init
+        if ASV_USE_IMPL == "modin":
+            pd.DataFrame([])
         file_id = self._get_file_id(shape, dtype)
         self.filename, self.names, self.dtype = cache[file_id]
 
--- a/asv_bench/benchmarks/utils.py
+++ b/asv_bench/benchmarks/utils.py
@@ -122,7 +122,7 @@ def translator_groupby_ngroups(groupby_ngroups: Union[str, int], shape: tuple) -
         return groupby_ngroups
 
 
-class weakdict(dict):
+class weakdict(dict):  # noqa: D101
     __slots__ = ("__weakref__",)
 
 
--- a/asv_bench/benchmarks/utils.py
+++ b/asv_bench/benchmarks/utils.py
@@ -12,7 +12,8 @@
 # governing permissions and limitations under the License.
 
 """
-This module contains the functionality that is used when benchmarking modin commits.
+The module contains the functionality that is used when benchmarking modin commits.
+
 In the case of using utilities from the main modin code, there is a chance that when
 benchmarking old commits, the utilities changed, which in turn can unexpectedly affect
 the performance results, hence some utility functions are duplicated here.
@@ -104,12 +105,13 @@ def translator_groupby_ngroups(groupby_ngroups: Union[str, int], shape: tuple) -
 
     Parameters
     ----------
-    groupby_ngroups: str or int
+    groupby_ngroups : str or int
         number of groups that will be used in `groupby` operation
-    shape: tuple
+    shape : tuple
+        same as pandas.Dataframe.shape
 
-    Return
-    ------
+    Returns
+    -------
     int
     """
     if ASV_DATASET_SIZE == "big":
@@ -138,19 +140,19 @@ def gen_int_data(nrows: int, ncols: int, rand_low: int, rand_high: int) -> dict:
 
     Parameters
     ----------
-    nrows: int
+    nrows : int
         number of rows
-    ncols: int
+    ncols : int
         number of columns
-    rand_low: int
+    rand_low : int
         low bound for random generator
-    rand_high:int
+    rand_high : int
         high bound for random generator
 
-    Return
-    ------
+    Returns
+    -------
     dict
-        ncols number of keys, storing numpy.array of nrows length
+        Number of keys - `ncols`, each of them store numpy.array of `nrows` length.
     """
     cache_key = ("int", nrows, ncols, rand_low, rand_high)
     if cache_key in data_cache:
@@ -179,20 +181,20 @@ def gen_str_int_data(nrows: int, ncols: int, rand_low: int, rand_high: int) -> d
 
     Parameters
     ----------
-    nrows: int
+    nrows : int
         number of rows
-    ncols: int
+    ncols : int
         number of columns
-    rand_low: int
+    rand_low : int
         low bound for random generator
-    rand_high:int
+    rand_high : int
         high bound for random generator
 
-    Return
-    ------
+    Returns
+    -------
     dict
-        ncols+1 number of keys, storing numpy.array of nrows length
-        one column with string values
+        Number of keys - `ncols`, each of them store numpy.array of `nrows` length.
+        One of the columns with string values.
     """
     cache_key = ("str_int", nrows, ncols, rand_low, rand_high)
     if cache_key in data_cache:
@@ -204,9 +206,8 @@ def gen_str_int_data(nrows: int, ncols: int, rand_low: int, rand_high: int) -> d
         )
     )
     data = gen_int_data(nrows, ncols, rand_low, rand_high).copy()
-    data["gb_col"] = [
-        "str_{}".format(random_state.randint(rand_low, rand_high)) for i in range(nrows)
-    ]
+    # convert values in `col13` column to string type
+    data["col13"] = [f"str_{x}" for x in data["col13"]]
     data_cache[cache_key] = weakdict(data)
     return data
 
@@ -223,22 +224,22 @@ def gen_data(
 
     Parameters
     ----------
-    data_type: str
+    data_type : {"int", "str_int"}
         type of data generation
-    nrows: int
+    nrows : int
         number of rows
-    ncols: int
+    ncols : int
         number of columns
-    rand_low: int
+    rand_low : int
         low bound for random generator
-    rand_high:int
+    rand_high : int
         high bound for random generator
 
-    Return
-    ------
+    Returns
+    -------
     dict
-        ncols+1 number of keys, storing numpy.array of nrows length
-        one column with string values
+        Number of keys - `ncols`, each of them store numpy.array of `nrows` length.
+        When `data_type`=="str_int" some of the columns will be of string type.
     """
     if data_type == "int":
         return gen_int_data(nrows, ncols, rand_low, rand_high)
@@ -267,31 +268,34 @@ def generate_dataframe(
 
     Parameters
     ----------
-    impl: str
+    impl : str
         implementation used to create the dataframe;
         supported implemetations: {"modin", "pandas"}
-    data_type: str
+    data_type : str
         type of data generation;
         supported types: {"int", "str_int"}
-    nrows: int
+    nrows : int
         number of rows
-    ncols: int
+    ncols : int
         number of columns
-    rand_low: int
+    rand_low : int
         low bound for random generator
-    rand_high: int
+    rand_high : int
         high bound for random generator
-    groupby_ncols: int or None
+    groupby_ncols : int or None
         number of columns for which `groupby` will be called in the future;
         to get more stable performance results, we need to have the same number of values
         in each group every benchmarking time
-    count_groups: int or None
+    count_groups : int or None
         count of groups in groupby columns
 
-    Return
-    ------
-    modin.DataFrame or pandas.DataFrame [and list of groupby columns names if
-        columns for groupby were generated]
+    Returns
+    -------
+    modin.DataFrame or pandas.DataFrame [and list]
+
+    Notes
+    -----
+    the list of groupby columns names returns when groupby columns are generated
     """
     assert not (
         (groupby_ncols is None) ^ (count_groups is None)
@@ -345,8 +349,8 @@ def random_string() -> str:
     """
     Create a 36-character random string.
 
-    Return
-    ------
+    Returns
+    -------
     str
     """
     return str(uuid.uuid4())
@@ -358,13 +362,13 @@ def random_columns(df_columns: list, columns_number: int) -> list:
 
     Parameters
     ----------
-    df_columns: list
+    df_columns : list
         columns to choose from
-    columns_number: int
+    columns_number : int
         how many columns to pick
 
-    Return
-    ------
+    Returns
+    -------
     list
     """
     return list(random_state.choice(df_columns, size=columns_number))
@@ -376,11 +380,11 @@ def random_booleans(number: int) -> list:
 
     Parameters
     ----------
-    number: int
+    number : int
         count of booleans in result list
 
-    Return
-    ------
+    Returns
+    -------
     list
     """
     return list(random_state.choice([True, False], size=number))
@@ -392,7 +396,7 @@ def execute(df: Union[pd.DataFrame, pandas.DataFrame]):
 
     Parameters
     ----------
-    df: modin.DataFrame of pandas.Datarame
+    df : modin.DataFrame or pandas.Datarame
     """
     if ASV_USE_IMPL == "modin":
         partitions = df._query_compiler._modin_frame._partitions
@@ -423,10 +427,11 @@ def get_shape_id(shape: tuple) -> str:
 
     Parameters
     ----------
-    shape: tuple
+    shape : tuple
+        same as pandas.Dataframe.shape
 
-    Return
-    ------
+    Returns
+    -------
     str
     """
     return "_".join([str(element) for element in shape])
--- a/asv_bench/benchmarks/utils.py
+++ b/asv_bench/benchmarks/utils.py
@@ -106,9 +106,9 @@ def translator_groupby_ngroups(groupby_ngroups: Union[str, int], shape: tuple) -
     Parameters
     ----------
     groupby_ngroups : str or int
-        number of groups that will be used in `groupby` operation
+        Number of groups that will be used in `groupby` operation.
     shape : tuple
-        same as pandas.Dataframe.shape
+        Same as pandas.Dataframe.shape.
 
     Returns
     -------
@@ -141,13 +141,13 @@ def gen_int_data(nrows: int, ncols: int, rand_low: int, rand_high: int) -> dict:
     Parameters
     ----------
     nrows : int
-        number of rows
+        Number of rows.
     ncols : int
-        number of columns
+        Number of columns.
     rand_low : int
-        low bound for random generator
+        Low bound for random generator.
     rand_high : int
-        high bound for random generator
+        High bound for random generator.
 
     Returns
     -------
@@ -182,13 +182,13 @@ def gen_str_int_data(nrows: int, ncols: int, rand_low: int, rand_high: int) -> d
     Parameters
     ----------
     nrows : int
-        number of rows
+        Number of rows.
     ncols : int
-        number of columns
+        Number of columns.
     rand_low : int
-        low bound for random generator
+        Low bound for random generator.
     rand_high : int
-        high bound for random generator
+        High bound for random generator.
 
     Returns
     -------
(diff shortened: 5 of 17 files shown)


anmyachev:
There's a challenge with testing documentation of the objects to which decorators are applied.

modin/engines/base/frame/partition_manager.py
```diff
@@ -1007,14 +1009,14 @@ def binary_operation(cls, axis, left, func, right):
...
         ----------
-            axis : int
-                The axis to apply the function over (0 - rows, 1 - columns)
-            left : NumPy array
-                The partitions of left Modin Frame
-            func : callable
-                The function to apply
-            right : NumPy array
-                The partitions of right Modin Frame.
+        axis : int
```
vnlitvinov, line 1012: no need to change that, we're already working on that file in #2955

modin/experimental/xgboost/xgboost_ray.py
```diff
@@ -30,8 +30,17 @@
 LOGGER = logging.getLogger("[modin.xgboost]")
 
 
-@ray.remote
-class ModinXGBoostActor:
+class ModinXGBoost:
```
vnlitvinov, line 33: @prutskov should be covering this in a separate PR

scripts/doc-checker.py
```diff
@@ -0,0 +1,124 @@
...
+"""
+
+import click
+import subprocess
+
+
+def numpydoc_validate(path):
+    import os
+    import ast
+    from numpydoc.validate import validate
```
vnlitvinov, line 17: why those things are not imported at module-level?

scripts/doc-checker.py
```diff
@@ -0,0 +1,124 @@
...
+        *("SS02", "SS04", "SS05", "PR03", "PR04", "PR05", "PR10", "EX04", "RT01"),
+        *("RT04", "RT05", "SA02", "SA03"),
+    }
+
+    # Recursive handle directories
+    if os.path.isdir(path):
+        for f in os.listdir(path):
+            if f.endswith(".py"):
+                numpydoc_validate(os.path.join(path, f))
+        return
```
vnlitvinov: I suggest using `os.walk()` and doing a non-recursive check

scripts/doc-checker.py
```diff
@@ -0,0 +1,124 @@
...
+
+    with open(path) as fd:
+        file_contents = fd.read()
+
+    # using static parsing for collecting module, functions, classes and its methods
+    module = ast.parse(file_contents)
+    functions = [
+        node
+        for node in module.body
+        if isinstance(node, ast.FunctionDef) and not node.name.startswith("__")
```
vnlitvinov, line 63: why don't we check docstrings for `__foo__` methods?.. I think you wanted to check if it's a private one, in this case change the check so it doesn't miss [dunder](https://www.geeksforgeeks.org/dunder-magic-methods-python/#:~:text=Dunder%20or%20magic%20methods%20in%20Python%20are%20the%20methods%20having,__%2C%20__repr__%20etc.) methods

scripts/doc-checker.py
```diff
@@ -0,0 +1,124 @@
...
+    ]
+    classes = [node for node in module.body if isinstance(node, ast.ClassDef)]
+    methods = []
+    for _class in classes:
+        for method in [
+            f"{module_name}.{_class.name}.{node.name}"
+            for node in _class.body
+            if isinstance(node, ast.FunctionDef) and not node.name.startswith("__")
```
vnlitvinov, line 71: Same about public vs private check

Plus iterating using `for` over a result of list comprehension looks... weird at best.
Just do a plain old

```python
for node in _class.body:
    if isinstance(node, ast.FunctionDef) and (not node.name.startswith('__') or node.name.endswith('__')):
        methods.append(f'{module_name}.{_class.name}.{node.name}')
```

scripts/doc-checker.py
```diff
@@ -0,0 +1,124 @@
...
+    # docstrings are taken dynamically
+    f_validates = [validate_object(f"{module_name}.{x.name}") for x in functions]
+    c_validates = [validate_object(f"{module_name}.{x.name}") for x in classes]
+    m_validates = [validate_object(x) for x in methods]
+    if (
+        validate_object(module_name)
+        or any(f_validates)
+        or any(c_validates)
+        or any(m_validates)
+    ):
```
vnlitvinov, line 87: isn't it easier to do stuff like

```python
to_validate = [module_name] + [f'{module_name}.{x.name}' for x in (functions + classes)] + methods

if (any(map(validate, to_validate))):
    exit_status = 1
```
?

scripts/doc-checker.py
```diff
@@ -0,0 +1,124 @@
...
+        text=True,
+        capture_output=True,
+    )
+    if result.returncode:
+        print(f"PYDOCSTYLE OUTPUT FOR {path}\n", result.stdout)
+    return result.returncode
+
+
+@click.command()
+@click.argument("pathes", nargs=-1, required=True, type=click.Path(exists=True))
```
vnlitvinov, line 111:
```suggestion
@click.argument("paths", nargs=-1, required=True, type=click.Path(exists=True))
```
here and in other places

scripts/doc-checker.py
```diff
@@ -0,0 +1,124 @@
...
+        if pydocstyle_validate(path, add_ignore):
+            exit_status = 1
+        if use_numpydoc:
+            if numpydoc_validate(path):
+                exit_status = 1
+    if exit_status:
+        exit(exit_status)
+
+
+if __name__ == "__main__":
```
vnlitvinov, line 127: I see that it might be easier to use `click`, but pulling in a 3rd-party library for _one_ function instead of using built-in `argparse`?.. come on, it would take you like extra 5 (15 if you want a lot of error checks) lines and _no_ extra installation dependency

vnlitvinov (changes requested):
Do not change docstrings here, just enable the tests for _only_ those files which already pass the check.

scripts/doc-checker.py
anmyachev: fixed

asv_bench/benchmarks/io/csv.py
```diff
@@ -26,10 +26,6 @@
     get_shape_id,
 )
 
-# ray init
-if ASV_USE_IMPL == "modin":
-    pd.DataFrame([])
-
```
anmyachev, line 32: Moved to avoid starting the engine while checking the documentation.

asv_bench/benchmarks/utils.py
```diff
@@ -204,9 +206,8 @@ def gen_str_int_data(nrows: int, ncols: int, rand_low: int, rand_high: int) -> d
         )
     )
     data = gen_int_data(nrows, ncols, rand_low, rand_high).copy()
-    data["gb_col"] = [
-        "str_{}".format(random_state.randint(rand_low, rand_high)) for i in range(nrows)
-    ]
+    # convert values in `col13` column to string type
+    data["col13"] = [f"str_{x}" for x in data["col13"]]
```
anmyachev, line 210: Now the size of the dataset is the same for all generating functions. Column name is arbitrary.

modin/engines/base/frame/partition_manager.py
```diff
@@ -39,6 +40,7 @@ def wait_computations_if_benchmark_mode(func):
     """
     if BenchmarkMode.get():
 
+        @wraps(func)
```
anmyachev, line 43: Added to display `__doc__` attribute correctly.

scripts/conda-recipe/run_test.py
```diff
@@ -6,4 +6,4 @@
     import modin.pandas as pd
 
     print(pd.__version__)
-    print(pd.DataFrame([1,2,3]))
+    print(pd.DataFrame([1, 2, 3]))
```
anmyachev, line 9: Very easy change for a separate pull request

.github/workflows/ci.yml
```diff
@@ -45,13 +45,31 @@ jobs:
...
           architecture: "x64"
-      - run: pip install pydocstyle
-      - run: pydocstyle --convention=numpy --add-ignore=D101,D102,D105 modin/pandas/dataframe.py modin/pandas/series.py modin/pandas/groupby.py modin/pandas/base.py
-      - run: pydocstyle --convention=numpy --add-ignore=D101,D103,D105 modin/pandas/io.py
-      - run: pydocstyle --convention=numpy --add-ignore=D101,D102 modin/pandas/series_utils.py
-      - run: pydocstyle --convention=numpy --add-ignore=D103 modin/pandas/general.py
-      - run: pydocstyle --convention=numpy modin/pandas/plotting.py modin/pandas/utils.py modin/pandas/iterator.py modin/pandas/indexing.py
-      - run: pydocstyle --convention=numpy --add-ignore=D100,D104 modin/engines/base/frame
+      - run: pip install pydocstyle numpydoc click
+      - run: pip install -e .
```
vnlitvinov, line 49: instead of doing this I suggest you add modin root directory to `sys.path` from `doc-checker.py` (see below)

scripts/doc-checker.py
```diff
@@ -0,0 +1,123 @@
...
+Notes
+-----
+    * the script can take some pathes to files or to directories
+"""
+
+import click
+import subprocess
+import os
+import ast
+
```
vnlitvinov, line 14: instead of requiring script users to install `modin` locally I suggest doing something like this:
```suggestion
import sys
sys.path.insert(0, os.path.abspath(os.path.join(os.path.dirname(__file__), '..')))

```

vnlitvinov (commented):
Do not change docstrings here, just enable the tests for _only_ those files which already pass the check.

asv_bench/benchmarks/utils.py
vnlitvinov: this seems unrelated, why is it done in this PR?

anmyachev: To simplify the description of the return value for generation function :)

scripts/doc-checker.py
```diff
@@ -22,7 +22,19 @@
 
 
 # code snippet from numpydoc
-def validate_object(import_path):
+def validate_object(import_path: str) -> int:
```
vnlitvinov, line 25: I suggest changing return type to `bool` (this includes returning `True` and `False`, of course). I think original code gave either ratings or at least `OK`/`Warnig`/`Error` while we just need `fine` vs `not fine` outcome.

scripts/doc-checker.py
```diff
@@ -36,55 +48,83 @@ def validate_object(import_path):
...
+
+            current_path = os.path.join(root, f)
+            # get importable name
+            module_name = current_path.replace("/", ".").replace("\\", ".")
+            module_name = module_name.replace(".py", "")
+
+            with open(current_path) as fd:
+                file_contents = fd.read()
+
+            # using static parsing for collecting module, functions, classes and its methods
```
vnlitvinov, line 85:
```suggestion
            # using static parsing for collecting module, functions, classes and their methods
```

scripts/doc-checker.py
```diff
@@ -36,55 +48,83 @@ def validate_object(import_path):
...
+    ----------
+    path : pathlib.Path
+    add_ignore : List[int]
+
+    Returns
+    -------
+    int
+        Return code of pydocstyle subprocess.
+    """
     result = subprocess.run(
```
vnlitvinov, line 128: isn't there a better way of executing `pydocstyle` by importing it and using some API?..

scripts/doc-checker.py
```diff
@@ -102,22 +142,96 @@ def pydocstyle_validate(path, add_ignore):
...
+    Perform pydocstyle and numpydoc checks.
+
+    Parameters
+    ----------
+    paths : List[pathlib.Path]
+        filenames of directories for check
+    add_ignore : List[str]
+        pydocstyle error codes which are not verified
+    use_numpydoc : bool
+        determine if numpydoc checks are not needed
```
vnlitvinov, line 165:
```suggestion
        determine if numpydoc checks are needed
```

scripts/doc-checker.py
```diff
@@ -102,22 +142,96 @@ def pydocstyle_validate(path, add_ignore):
...
+    Returns
+    -------
+    exit_status : int
+    """
     exit_status = 0
     for path in paths:
         if pydocstyle_validate(path, add_ignore):
             exit_status = 1
         if use_numpydoc:
+            monkeypatching()
```
vnlitvinov, line 176: why not call it once from `main` part?

scripts/doc-checker.py
```diff
@@ -102,22 +142,96 @@ def pydocstyle_validate(path, add_ignore):
...
+    Check the obtained values for correctness.
+
+    Raises
+    ------
+    ValueError
+        Occurs in case of non-existent files or directories.
+    """
+    for path in args.paths:
+        if not path.exists():
+            raise ValueError(f"{path} is not exist")
```
vnlitvinov, line 193:
```suggestion
            raise ValueError(f"{path} does not exist")
```
I don't think this check is strictly needed, though, as non-existant files would surface later on... but it's better to fail early

scripts/doc-checker.py
```diff
@@ -102,22 +142,96 @@ def pydocstyle_validate(path, add_ignore):
...
+        help="Filenames or directories; in case of direstories perform recursive check",
+    )
+    parser.add_argument(
+        "--add-ignore",
+        nargs="*",
+        default=[],
+        help="Pydocstyle error codes; for example: D100,D100,D102",
+    )
+    parser.add_argument(
+        "--no-use-numpydoc",
```
vnlitvinov, line 220: maybe use `--disable-numpydoc` for switch name? it's easier to use by a reader

scripts/doc-checker.py
anmyachev: This module contains a public api for checking only the list of files, not directories. Checking directories is available only when the script is explicitly called as the main one.

Bottom line, this can be done using imports from the module, but it will be a little more complicated and more room for errors. If you don't mind, then I would not like to rewrite.

scripts/doc-checker.py
```diff
@@ -0,0 +1,240 @@
...
+import ast
+from typing import List
+
+# error codes that pandas test in CI
+# https://numpydoc.readthedocs.io/en/latest/validation.html#built-in-validation-checks
+NUMPYDOC_BASE_ERR_CODE = {
+    *("GL01", "GL02", "GL03", "GL04", "GL05", "GL06", "GL07", "GL09", "GL10"),
+    *("SS02", "SS04", "SS05", "PR03", "PR04", "PR05", "PR10", "EX04", "RT01"),
+    *("RT04", "RT05", "SA02", "SA03"),
+}
```
anmyachev, line 21: Pay attention to the ERROR CODES that are being tested. The codes were taken from [the file](https://github.com/pandas-dev/pandas/blob/7d4757b4deb851bb44ab6bb20cdc404fa13fffcf/ci/code_checks.sh#L224), but I added two more to them (added for subjective sensations of readability) - GL01 and GL02.

An agreement should be reached. Cc @vnlitvinov @YarShev @devin-petersohn @prutskov .

vnlitvinov: Do you have links describing `GL01` and `GL02` then?

scripts/doc-checker.py
vnlitvinov: Well, you're already validating using an `os.walk()` method... but we can improve later, it's certainly not a blocker.

scripts/doc-checker.py
```diff
@@ -33,22 +33,23 @@ def validate_object(import_path: str) -> int:
 
     Returns
     -------
-    exit_status : int
+    is_failed : bool
```
vnlitvinov, line 36: Using a double-negative thing is usually a bad practice... I suggest changing the meaning to "`True` when object is valid", which seems more logical to me.

Here and in other places, too.

scripts/doc-checker.py
```diff
@@ -22,7 +22,7 @@
```
vnlitvinov, line 22: I don't see any `sys.path` modifications I've suggested.
It means that to run the script correctly the developer would have to install `modin` from its development directory, or you're risking to validate some stale version which just happened to be installed in the environment.

vnlitvinov (changes requested):
Almost there :)

scripts/doc-checker.py
anmyachev: Yes, but `pip install -e .[all]` also install modin. I didn't see a way to disable modin installation but the dependencies installed in the way are needed.

scripts/doc-checker.py
anmyachev: updated

scripts/doc-checker.py
anmyachev: Yes, https://numpydoc.readthedocs.io/en/latest/validation.html#built-in-validation-checks

scripts/doc-checker.py
vnlitvinov: This is irrelevant. What I'm trying to get is to do two things for a developer running this script locally (not on GitHub CI):

1. Make sure the script is validating correct sources, not something random that was installed ages ago (remember that a developer's environment isn't a shiny new clean CI one)
2. Allow the developer to _not_ install Modin whatsoever (I, for once, am running it using `$PYTHONPATH`).

scripts/doc-checker.py
vnlitvinov: Relevant piece:
```
    "GL01": "Docstring text (summary) should start in the line immediately "
            "after the opening quotes (not in the same line, or leaving a "
            "blank line in between)",
    "GL02": "Closing quotes should be placed in the line after the last text "
            "in the docstring (do not close the quotes in the same line as "
            "the text, or leave a blank line between the last text and the "
            "quotes)",
```

I'm fine with adding those checks, we follow them now in most places anyway.

scripts/doc-checker.py
```diff
@@ -141,7 +144,7 @@ def pydocstyle_validate(path: pathlib.Path, add_ignore: List[str]) -> int:
     )
     if result.returncode:
         print(f"PYDOCSTYLE OUTPUT FOR {path}\n", result.stdout)
-    return result.returncode
+    return True if result.returncode == 0 else False
```
vnlitvinov, line 147:
```suggestion
    return result.returncode == 0
```
🤣

scripts/doc-checker.py
anmyachev: Ok, I got it

scripts/doc-checker.py
```diff
@@ -11,6 +11,9 @@
 import os
 import ast
 from typing import List
+import sys
+
+sys.path.insert(0, os.path.abspath(os.path.join(os.path.dirname(__file__), "..")))
```
vnlitvinov, line 16: Hmm, it now came to my mind that this is also wrong... 😆 
As it would test Modin in which this script lives, but not what we asked it to test.

So the "real" fix should be assuring that numpydoc validator is importing the correct things, so we either need to modify `sys.path` according to files being checked or use an import hook.

Or at least validate that files we check live under this script's root repo (and transform paths to them accordingly when composing `module_name` variable down below).

Example that would fail in current version of the script: `python doc-checker.py ../../../modin-dev-2/modin/pandas/dataframe.py` (i.e. we can go outside of the repo that holds this script, and it would break).

.github/workflows/ci.yml
```diff
@@ -45,13 +45,32 @@ jobs:
...
           architecture: "x64"
-      - run: pip install pydocstyle
-      - run: pydocstyle --convention=numpy --add-ignore=D101,D102,D105 modin/pandas/dataframe.py modin/pandas/series.py modin/pandas/groupby.py modin/pandas/base.py
-      - run: pydocstyle --convention=numpy --add-ignore=D101,D103,D105 modin/pandas/io.py
-      - run: pydocstyle --convention=numpy --add-ignore=D101,D102 modin/pandas/series_utils.py
-      - run: pydocstyle --convention=numpy --add-ignore=D103 modin/pandas/general.py
-      - run: pydocstyle --convention=numpy modin/pandas/plotting.py modin/pandas/utils.py modin/pandas/iterator.py modin/pandas/indexing.py
-      - run: pydocstyle --convention=numpy --add-ignore=D100,D104 modin/engines/base/frame
+      - run: pip install pydocstyle numpydoc
+      - run: pip install -e .[all]
```
YarShev, line 49: Why is this done?

asv_bench/benchmarks/benchmarks.py
```diff
@@ -11,7 +11,7 @@
 # ANY KIND, either express or implied. See the License for the specific language
 # governing permissions and limitations under the License.
 
-"""General modin benchmarks"""
+"""General modin benchmarks."""
```
YarShev, line 14:
```suggestion
"""General Modin benchmarks."""
```

https://github.com/modin-project/modin/issues/2934#issuecomment-815494134

asv_bench/benchmarks/io/__init__.py
```diff
@@ -11,4 +11,4 @@
 # ANY KIND, either express or implied. See the License for the specific language
 # governing permissions and limitations under the License.
 
-"""IO modin benchmarks"""
+"""IO modin benchmarks."""
```
YarShev, line 14:
```suggestion
"""IO Modin benchmarks."""
```

asv_bench/benchmarks/utils.py
```diff
@@ -12,7 +12,8 @@
 # governing permissions and limitations under the License.
 
 """
-This module contains the functionality that is used when benchmarking modin commits.
+The module contains the functionality that is used when benchmarking modin commits.
```
YarShev, line 15:
```suggestion
The module contains the functionality that is used when benchmarking Modin commits.
```

YarShev: Please, change `modin` -> `Modin` everywhere

asv_bench/benchmarks/utils.py
```diff
@@ -138,19 +140,19 @@ def gen_int_data(nrows: int, ncols: int, rand_low: int, rand_high: int) -> dict:
...
+    rand_high : int
         high bound for random generator
 
-    Return
-    ------
+    Returns
+    -------
     dict
-        ncols number of keys, storing numpy.array of nrows length
+        Number of keys - `ncols`, each of them store numpy.array of `nrows` length.
```
YarShev, line 155:
```suggestion
        Number of keys - `ncols`, each of them store np.ndarray of `nrows` length.
```

YarShev: We should use either `np.ndarray` or `NumPy array` because `numpy.array` doesn't exist.

asv_bench/benchmarks/utils.py
```diff
@@ -179,20 +181,20 @@ def gen_str_int_data(nrows: int, ncols: int, rand_low: int, rand_high: int) -> d
...
         high bound for random generator
 
-    Return
-    ------
+    Returns
+    -------
     dict
-        ncols+1 number of keys, storing numpy.array of nrows length
-        one column with string values
+        Number of keys - `ncols`, each of them store numpy.array of `nrows` length.
```
YarShev, line 196:
```suggestion
        Number of keys - `ncols`, each of them store np.ndarray of `nrows` length.
```

asv_bench/benchmarks/utils.py
```diff
@@ -223,22 +225,22 @@ def gen_data(
...
```
YarShev, line 242:
```suggestion
        Number of keys - `ncols`, each of them store np.ndarray of `nrows` length.
```

asv_bench/benchmarks/utils.py
```diff
@@ -267,31 +269,34 @@ def generate_dataframe(
...
+    count_groups : int or None
         count of groups in groupby columns
 
-    Return
-    ------
-    modin.DataFrame or pandas.DataFrame [and list of groupby columns names if
-        columns for groupby were generated]
+    Returns
+    -------
+    modin.DataFrame or pandas.DataFrame [and list]
```
YarShev, line 295:
```suggestion
    modin.pandas.DataFrame or pandas.DataFrame [and list]
```

asv_bench/benchmarks/utils.py
```diff
@@ -392,7 +397,7 @@ def execute(df: Union[pd.DataFrame, pandas.DataFrame]):
 
     Parameters
     ----------
-    df: modin.DataFrame of pandas.Datarame
+    df : modin.DataFrame or pandas.Datarame
```
YarShev, line 400:
```suggestion
    df : modin.pandas.DataFrame or pandas.Datarame
```

scripts/doc-checker.py
```diff
@@ -0,0 +1,244 @@
...
+import os
+import ast
+from typing import List
+import sys
+
+sys.path.insert(0, os.path.abspath(os.path.join(os.path.dirname(__file__), "..")))
+
+# error codes that pandas test in CI
+# https://numpydoc.readthedocs.io/en/latest/validation.html#built-in-validation-checks
+NUMPYDOC_BASE_ERR_CODE = {
```
YarShev, line 20:
```suggestion
NUMPYDOC_BASE_ERROR_CODES = {
```

scripts/doc-checker.py
YarShev: 1) `GL04`, `EX04` are not presented in the guide. What is it?
2) Why are some of the numpydoc error codes missed?

.github/workflows/ci.yml
```diff
@@ -45,13 +45,32 @@ jobs:
...
+      - run: |
+          python scripts/doc-checker.py --add-ignore=D101,D102,D105 --disable-numpydoc \
```
YarShev, line 51: Why do we disable numpydoc for modin/pandas?

.github/workflows/ci.yml
anmyachev: To import dependencies that occur in modin code.

.github/workflows/ci.yml
anmyachev: I disabled this check only for files where docstrings from pandas are inherited, because:
1) There is no point in checking someone else's documentation;
2) Our checks are different from pandas checks;

scripts/doc-checker.py
anmyachev: > 1. `GL04`, `EX04` are not presented in the guide. What is it?


Nice catch! These codes are defined by pandas itself and are absent in numpydoc.

> 2. Why are some of the numpydoc error codes missed?


Some codes are contrary to the standard (for example- `GL01`), some will lead to too much strictness in my opinion(for example - `PR08`).

.github/workflows/ci.yml
vnlitvinov: But this leaves out checks for non-inherited methods, which doesn't feel right.

vnlitvinov:
@anmyachev please give a thought at implementing a separate check for `optional, default: foo` part of parameter description (or whatever form we agreed upon) or see if numpydoc already has such a check.

.github/workflows/ci.yml
YarShev: I see, thanks!

scripts/doc-checker.py
YarShev: I suggest adding the following codes to current ones:
- GL08 if it doesn't require a docstring for a nested function.
- SS03
- PR01
- PR02

.github/workflows/ci.yml
anmyachev: Ok, in this case monkeypatching will also help. 

But the work on the documentation also gave me the idea that it is possible to leave modules in which only public functions are defined and docstrings are inherited, and all internal logic can be transferred either to the base class or to utilities.

vnlitvinov: That is refactoring, not documenting. :) Different things.

scripts/doc-checker.py
```diff
@@ -13,16 +13,109 @@
...
 NUMPYDOC_BASE_ERROR_CODES = {
-    *("GL01", "GL02", "GL03", "GL05", "GL06", "GL07", "GL09", "GL10"),
-    *("SS02", "SS04", "SS05", "PR03", "PR04", "PR05", "PR10", "RT01"),
-    *("RT04", "RT05", "SA02", "SA03"),
+    *("GL01", "GL02", "GL03", "GL05", "GL06", "GL07", "GL08", "GL09", "GL10"),
+    *("SS02", "SS03", "SS04", "SS05", "PR01", "PR02", "PR03", "PR04", "PR05"),
+    *("PR10", "RT01", "RT04", "RT05", "SA02", "SA03"),
 }
 
+""" TEST CUSTOM CHECK
```
vnlitvinov, line 27: what's this? 😄

scripts/doc-checker.py
```diff
@@ -151,11 +249,15 @@ def pydocstyle_validate(path: pathlib.Path, add_ignore: List[str]) -> int:
...
 def monkeypatching():
-    """Monkeypatch decorators which incorrectly define __doc__ attribute."""
+    """Monkeypatch decorators which change __doc__ attribute."""
     import ray
+    import modin.utils
 
     ray.remote = lambda *args, **kwargs: args[0]
 
+    modin.utils._inherit_docstrings = lambda *args, **kwargs: lambda cls: cls
+    modin.utils._inherit_func_docstring = lambda *args, **kwargs: lambda func: func
```
vnlitvinov, line 259: I don't understand this change. Why are you turning off these decorators? `ray.remote` has a bug which those do not.
In my opinion skipping the checks if `# noqa` directive is given should happen even with decorators enabled (but we may need to add more `noqa` values in there).

scripts/doc-checker.py
```diff
@@ -199,6 +306,13 @@ def check_args(args: argparse.Namespace):
     for path in args.paths:
         if not path.exists():
             raise ValueError(f"{path} does not exist")
+        abs_path = os.path.abspath(path)
+        if not abs_path.startswith(MODIN_PATH):
+            raise ValueError(
+                f"it is forbidden to use this script on files from another "
```
vnlitvinov, line 312:
```suggestion
                f"it is unsupported to use this script on files from another "
```

scripts/doc-checker.py
```diff
@@ -13,16 +13,109 @@
...
+    source = doc.method_source
+
+    noqa_str = None
+    # find last line of obj definition
+    for line in source.split("\n"):
+        if ")" in line and ":" in line.split(")", 1)[1]:
+            noqa_str = line
+            break
+
+    if noqa_str and "noqa" in noqa_str:
```
vnlitvinov, line 107: I suggest changing how `noqa` is handled by removing the error codes mentioned in `noqa` directive unconditionally.

devin-petersohn (commented):
Since this isn't passing the pydocstyle test, I'm going to go ahead and remove the ready for review flag.

scripts/doc-checker.py
prutskov: I would prefer to add the following codes:

- PR07/RT03 - In my opinion, all parameters/returns should have a description.
- PR08/PR09 - This will ensure uniform style, I think. (It's already use for return values RT04/RT05)

scripts/doc-checker.py
```diff
@@ -0,0 +1,358 @@
...
+                    not node.name.startswith("__") or node.name.endswith("__")
+                )
+
+            functions = [node for node in module.body if is_public_func(node)]
+            classes = [node for node in module.body if isinstance(node, ast.ClassDef)]
+            methods = []
+            for _class in classes:
+                for node in _class.body:
+                    if is_public_func(node):
+                        methods.append(f"{module_name}.{_class.name}.{node.name}")
```
prutskov, line 203: Using generator in this place is also possible, I think.
```python
methods = [
      f"{module_name}.{_class.name}.{node.name}" 
      for _class in classes 
      for node in _class.body if is_public_func(node) ]
```

scripts/doc-checker.py
anmyachev: I turned them off because:
1) we do not need to do what will not be used in the future (by not doing, we save time)
2) we check a lot of errors that Pandas does not check, therefore we will need to add a lot of error codes that can be ignored (this will complicate both support and readability of the code)

scripts/doc-checker.py
anmyachev: What if the comment looks like this: `# noqa: E731`? This code refers to `flake8`

scripts/doc-checker.py
anmyachev: added check for failed case

scripts/doc-checker.py
YarShev: - On PR07/RT03, as we discussed on our previous weekly meeting, if a function is simple and its parameters and return values are intuitive, we can miss their description to not duplicate line of the type and line of the description.
- On PR08/PR09, if these checks do not contrast PR07/RT03 (there is no error that description is missed), we can add them.

.github/workflows/ci.yml
anmyachev: @YarShev now check is disabled only where there is internal undocumented functionality.

scripts/doc-checker.py
anmyachev: >> @anmyachev please give a thought at implementing a separate check for optional, default: foo part of parameter description (or whatever form we agreed upon) or see if numpydoc already has such a check.

There is no such built-in check, but it's easy enough to add your own check and this code demonstrates it. It remains to choose what we want.

My favorite option is where we always write for optional arguments like this: `default [SOME VALUE]`

scripts/doc-checker.py
anmyachev: Added next codes: `GL08`, `SS03`, `PR01`, `PR02`, `PR08`, `PR09`.

scripts/doc-checker.py
vnlitvinov: This error code won't be detected by neither pydocstyle nor numpydoc, so no worries.
The format of `noqa` comment is comma-separated list of ignored checks, so I see no issues in adding new checks to ignore, and even mixing doc checks and flake checks in one `noqa` should probably be fine.

scripts/doc-checker.py
vnlitvinov: Do note that I eventually plan to use those decorators on functions where we _do_ have our own docstrings.
It would append a link to Pandas API, but we still want to check the docstring.

How many error checks we need to ignore if we do not turn off these?..

scripts/doc-checker.py
vnlitvinov: I would prefer `default: VALUE` (i.e. add a colon for more clear delimiting).

My original comment was related to you adding a giant comment with code inside :) Why didn't you just add it as a real code? This is still the check we want to have.

Speaking of additional checks - we should add checks for "modin" vs "Modin", "numpy" vs "NumPy", etc. (ref: https://github.com/modin-project/modin/issues/2934#issuecomment-815494134), but note that this is for standalone words, not for types like `modin.pandas.DataFrame` or `pandas.Series`.

scripts/doc-checker.py
YarShev: I am okay with this.

scripts/doc-checker.py
vnlitvinov: I mean we should be handling a line like `  # noqa: E731, GL08` perfectly fine.

scripts/doc-checker.py
anmyachev: >> Added next codes: GL08, SS03, PR01, PR02, PR08, PR09 and GL01, GL02

Any of these errors may occur.

anmyachev: >>Do note that I eventually plan to use those decorators on functions where we do have our own docstrings.
>>It would append a link to Pandas API, but we still want to check the docstring.

It turns out that your decorator can work in two modes - adding the entire documentation or just a link, respectively, based on some parameter, I can decide when to disable this decorator, and when not

scripts/doc-checker.py
anmyachev: Improved.

scripts/doc-checker.py
anmyachev: Added check on `default: VALUE`

scripts/doc-checker.py
```diff
@@ -0,0 +1,434 @@
...
+    Validate custom modin errors.
+
+    Parameters
+    ----------
+    doc : numpydoc.validate.Docstring
+        Docstring handler.
+
+    Returns
+    -------
+    errors :
```
anmyachev, line 70: numpydoc doesn't handle the case

anmyachev: Maybe add a `RT02` check?

scripts/doc_checker.py
```diff
@@ -24,97 +40,242 @@
...
 
 
-import inspect
+def get_optional_args(doc: Docstring) -> dict:
+    """
+    Get optional parameters for the object for which the docstring is checked.
 
+    Parameters
+    ----------
+    doc : numpydoc.validate.Doctring
```
vnlitvinov, line 55: this isn't a valid type (it has a typo)

scripts/doc_checker.py
```diff
@@ -24,97 +40,242 @@
...
 
-def validate_modin_error(import_path):
-    from numpydoc.validate import Docstring
+def check_optional_args(doc: Docstring) -> list:
+    """
+    Check type description of optional arguments.
+
+    Parameters
+    ----------
+    doc : Docstring
```
vnlitvinov, line 80: pls select one - eihter `Docstring` or `numpy.validate.Docstring`

scripts/doc_checker.py
```diff
@@ -24,97 +40,242 @@
...
-                errors.append(
-                    (
-                        "TE01",
-                        MODIN_ERROR_CODES["TE01"].format(
-                            parameter=parameter, should=should, found=found
-                        ),
-                    )
+    for parameter in optional_args:
+        type_line = doc.doc_parameters[parameter][0]
+        if "default: " not in type_line or "optional" in type_line:
```
vnlitvinov, line 96: I think we can allow either `optional` or `default: VALUE` (but not both at the same time).

Example: a string parameter with default value of `None` meaning "do not consider this parameter". Declaring it as a string and saying that it defaults to `None` looks weird, so in this case I think we should name it just "optional".

scripts/doc_checker.py
```diff
@@ -24,97 +40,242 @@
...
+    "modin" word is treated as the constant.
+    """
+    if not doc.raw_doc:
+        return []
+    components = set(["Modin", "pandas", "NumPy"])
+    check_words = "|".join(x.lower() for x in components)
+
+    # comments work only with re.VERBOSE
+    pattern = r"""
+    (?:\W|^)                # non-capturing group: [^a-zA-Z0-9_] or start
```
vnlitvinov, line 134:
```suggestion
    (?:\W|^)                # non-capturing group of either non-word symbol or line start
```

scripts/doc_checker.py
```diff
@@ -24,97 +40,242 @@
...
+    ({check_words})         # words to check, example - "modin|pandas|numpy"
+    (?:                     # non-capturing group
+        [^"\.\/\w\\]        # any symbol except: '"', '.', '\' and any from [a-zA-Z0-9_]
+        | [\.]\s            # or '.' and any whitespace
```
vnlitvinov, line 138:
```suggestion
        | \.\s            # or '.' and any whitespace
```
I don't get why you'd want to make a "take any of" group with one element inside.

Also does this handle cases like `modin.pandas.DataFrame` correctly?

scripts/doc_checker.py
```diff
@@ -24,97 +40,242 @@
...
+        List of tuples with Modin error code and its description.
+
+    Returns
+    -------
+    dict
+        Updated dict with Modin custom errors.
+    """
+    for error in modin_errors:
+        results["errors"].append(error)
+    return results
```
vnlitvinov, line 203: I'm failing to understand why you want to extract it into two separate functions... I'd merge `validate_modin_error()` and this function into a single one probably

scripts/doc_checker.py
```diff
@@ -24,97 +40,242 @@
...
     """
-    import inspect
-    from numpydoc.validate import Docstring
+    if noqa_checks == ["all"]:
+        return True
 
-    result = False
-    doc = Docstring(import_path)
-    source = doc.method_source
+    if err_code == "GL08":
```
vnlitvinov, line 227: comments explaining the "why" of this weird replacements would be good

scripts/doc_checker.py
```diff
@@ -24,97 +40,242 @@
...
+    if not inspect.ismodule(doc.obj):
+        # find last line of obj definition
+        for line in source.split("\n"):
+            if ")" in line and ":" in line.split(")", 1)[1]:
+                noqa_str = line
+                break
+    else:
+        # pydocstyle only check first line; aling with it
+        noqa_str = source.split("\n", 1)[0]
+    if "noqa" not in noqa_str:
```
vnlitvinov, line 271: I would advise using regexp-s here, as this one currently doesn't check for pound sign (`#`), so it would fail for cases like
```python
def foo(value="noqa: MD01"): # my super function
    return 10
```

scripts/doc_checker.py
```diff
@@ -24,97 +40,242 @@
...
+        return []
+
+    if "noqa:" in noqa_str:
+        noqa_checks = noqa_str.split("noqa:")[1].split(",")
+    elif "noqa" in noqa_str:
```
vnlitvinov, line 276: it's pointless to check it here, you've already checked that at line 271

scripts/doc_checker.py
```diff
@@ -268,13 +269,13 @@ def get_noqa_checks(doc: Docstring) -> list:
     else:
         # pydocstyle only check first line; aling with it
         noqa_str = source.split("\n", 1)[0]
-    if "noqa" not in noqa_str:
-        return []
 
-    if "noqa:" in noqa_str:
+    if "# noqa:" in noqa_str:
         noqa_checks = noqa_str.split("noqa:")[1].split(",")
```
vnlitvinov, line 274:
```suggestion
        noqa_checks = noqa_str.split("# noqa:", 1)[1].split(",")
```
otherwise stuff like
```python
def foo(param="noqa: FOO"): # noqa: hey
    return 10
```
would break

vnlitvinov (commented):
Much better

modin/pandas/io.py
```diff
@@ -120,7 +122,7 @@ def read_csv(
     memory_map=False,
     float_precision=None,
     storage_options: StorageOptions = None,
-):
+):  # noqa: D103
```
YarShev, line 125: Why is `noqa: D103` used?

scripts/doc_checker.py
YarShev: This makes sense but not each person knows about it. We can run into trouble with this. Let's do it so and then will see how it goes.

scripts/doc-checker.py
YarShev: Why do you use `errors :`? Return type should be `list`. The description explains what the list means.

anmyachev: This is one of the available forms, but the user may forget to add a type, and numpydoc will not react to this.

```
-------
err_code : int
    Non-zero value indicates error code, or zero on success.
err_msg : str or None
    Human readable error message, or None on success.
```

scripts/doc_checker.py
anmyachev: Yes, regexp handle that cases.

modin/pandas/io.py
anmyachev: This allows us to disable the check that the function does not have a docstring.

scripts/doc-checker.py
anmyachev: So far I just wrote a type without name.

YarShev: I don't mind on adding `RT02` if this makes sense.

modin/pandas/io.py
YarShev: Okay, I see.

.github/workflows/ci.yml
```diff
@@ -45,9 +45,9 @@ jobs:
         with:
           python-version: "3.7.x"
           architecture: "x64"
-      - run: pip install pytest pytest-cov
+      - run: pip install --no-deps -e .
```
vnlitvinov, line 48: what's the point of doing this and later on doing `pip install -e .[all]`? doesn't make any sense...

scripts/doc_checker.py
```diff
@@ -164,42 +164,25 @@ def check_spelling_words(doc: Docstring) -> list:
...
     Returns
     -------
     dict
         Updated dict with Modin custom errors.
     """
-    for error in modin_errors:
+    errors = check_optional_args(doc)
+    errors += check_spelling_words(doc)
+    for error in errors:
         results["errors"].append(error)
```
vnlitvinov:
```suggestion
    results["errors"].extend(errors)
```
😄

scripts/doc_checker.py
```diff
@@ -0,0 +1,517 @@
...
+    if not doc.doc_parameters:
+        return []
+    optional_args = get_optional_args(doc)
+    if not optional_args:
+        return []
+
+    errors = []
+    for parameter in optional_args:
+        type_line = doc.doc_parameters[parameter][0]
+        if "default: " not in type_line or "optional" in type_line:
```
vnlitvinov, line 96:
```suggestion
        has_default = "default: " in type_line
        has_optional = "optional" in type_line
        if not (has_default ^ has_optional):
```
I think it's a missing `not` in here.
Plus it doesn't check that we specify both, while we should pick one of the two (which is a definition of "exclusive OR" operation).

vnlitvinov (commented):
Last bit of suggestions.

scripts/doc_checker.py
YarShev: So, let's do as @vnlitvinov suggested.

prutskov:
@anmyachev I can't run script on Windows platform. I got the next trace:
```
(modin-docs) C:\Users\aprutsko\Documents\modin>python doc_checker.py modin\experimental\xgboost
  File "doc_checker.py", line 488, in <module>
    if not validate(args.paths, args.add_ignore, not args.disable_numpydoc):
  File "doc_checker.py", line 420, in validate
    if not pydocstyle_validate(path, add_ignore):
  File "doc_checker.py", line 382, in pydocstyle_validate
    capture_output=True,
  File "C:\Users\aprutsko\.conda\envs\modin-docs\lib\subprocess.py", line 488, in run
    with Popen(*popenargs, **kwargs) as process:
  File "C:\Users\aprutsko\.conda\envs\modin-docs\lib\subprocess.py", line 800, in __init__
    restore_signals, start_new_session)
  File "C:\Users\aprutsko\.conda\envs\modin-docs\lib\subprocess.py", line 1148, in _execute_child
    args = list2cmdline(args)
  File "C:\Users\aprutsko\.conda\envs\modin-docs\lib\subprocess.py", line 555, in list2cmdline
    needquote = (" " in arg) or ("\t" in arg) or not arg
TypeError: argument of type 'WindowsPath' is not iterable
``` 

Is it expected behavior?

anmyachev:
> @anmyachev I can't run script on Windows platform. I got the next trace:
> 
> ```
> (modin-docs) C:\Users\aprutsko\Documents\modin>python doc_checker.py modin\experimental\xgboost
> Traceback (most recent call last):
>   File "doc_checker.py", line 488, in <module>
>     if not validate(args.paths, args.add_ignore, not args.disable_numpydoc):
>   File "doc_checker.py", line 420, in validate
>     if not pydocstyle_validate(path, add_ignore):
>   File "doc_checker.py", line 382, in pydocstyle_validate
>     capture_output=True,
>   File "C:\Users\aprutsko\.conda\envs\modin-docs\lib\subprocess.py", line 488, in run
>     with Popen(*popenargs, **kwargs) as process:
>   File "C:\Users\aprutsko\.conda\envs\modin-docs\lib\subprocess.py", line 800, in __init__
>     restore_signals, start_new_session)
>   File "C:\Users\aprutsko\.conda\envs\modin-docs\lib\subprocess.py", line 1148, in _execute_child
>     args = list2cmdline(args)
>   File "C:\Users\aprutsko\.conda\envs\modin-docs\lib\subprocess.py", line 555, in list2cmdline
>     needquote = (" " in arg) or ("\t" in arg) or not arg
> TypeError: argument of type 'WindowsPath' is not iterable
> ```
> 
> Is it expected behavior?

The script was written with the expectation of launching from a specific folder - `scripts/doc_checker.py`.

.github/workflows/ci.yml
anmyachev: This speeds up the case where documentation tests failed.

anmyachev: removed

prutskov:
@anmyachev Are we need to add docstrings for `__init__` method? [Style guide](https://numpydoc.readthedocs.io/en/latest/format.html#documenting-classes) recommends to add docstrings of `__init__` method in class docstring. Currently, I have an error:
```
DMatrix.__init__:45:GL08:The object does not have a docstring
```

anmyachev:
> @anmyachev Are we need to add docstrings for `__init__` method? [Style guide](https://numpydoc.readthedocs.io/en/latest/format.html#documenting-classes) recommends to add docstrings of `__init__` method in class docstring. Currently, I have an error:
> 
> ```
> DMatrix.__init__:45:GL08:The object does not have a docstring
> ```

There is optional now.

scripts/doc_checker.py
```diff
@@ -92,8 +93,13 @@ def check_optional_args(doc: Docstring) -> list:
 
     errors = []
     for parameter in optional_args:
+        # case when not all parameters are listed in "Parameters" section
+        if parameter not in doc.doc_parameters:
+            continue
```
vnlitvinov, line 98: shouldn't this be an error? or is it already handled by `numpydoc`?

scripts/doc_checker.py
```diff
@@ -218,6 +223,10 @@ def skip_check_if_noqa(doc: Docstring, err_code: str, noqa_checks: list) -> bool
     # GL08 - missing docstring in an arbitary object; numpydoc code
     if err_code == "GL08":
         name = doc.name.split(".")[-1]
+        # Numpydoc recommends to add docstrings of __init__ method in class docstring.
+        # So there is no error if docstring missed in __init__
```
vnlitvinov, line 227:
```suggestion
        # So there is no error if docstring is missing in __init__
```

scripts/doc_checker.py
```diff
@@ -255,6 +264,16 @@ def get_noqa_checks(doc: Docstring) -> list:
     If noqa doesn't have any codes - returns ["all"].
     """
     source = doc.method_source
+    # case when property decorator is used; it hides sources
+    if not source and isinstance(doc.obj, property):
+        # readonly property
```
vnlitvinov, line 269: Why do we handle read-only different from read-write? I think a convention is to document everything under `fget` part

scripts/doc_checker.py
```diff
@@ -255,6 +264,16 @@ def get_noqa_checks(doc: Docstring) -> list:
...
+        if doc.obj.fset is None and doc.obj.fdel is None:
+            new_doc = Docstring(doc.name)
+            new_doc.obj = doc.obj.fget
+            source = new_doc.method_source
+        if not source:
+            return source
```
vnlitvinov, line 275:
```suggestion
        if not source:
            return False
```

vnlitvinov (dismissed):
Mostly fine, except question about handling read-write properties.
Is not a blocker, though, so I'm marking as LGTM for now.

scripts/doc_checker.py
anmyachev: Indeed, I did not see the difference.

scripts/doc_checker.py
anmyachev: It's already handled by `numpydoc`.

vnlitvinov (dismissed):
LGTM

anmyachev:
@vnlitvinov thanks for review!

.github/workflows/ci.yml
```diff
@@ -66,6 +85,7 @@ jobs:
           architecture: "x64"
       - run: pip install flake8 flake8-print
       - run: flake8 --enable=T modin/ asv_bench/benchmarks
+      - run: flake8 scripts/
```
devin-petersohn, line 88: I see that this is separate to avoid the `print` checks because there is a print in the test. Is it possible to use `logging` instead of print below?

asv_bench/benchmarks/io/csv.py
```diff
@@ -44,6 +40,9 @@ def setup_cache(self, test_filename="io_test_file"):
         return test_filenames
 
     def setup(self, test_filenames, shape, *args, **kwargs):
+        # ray init
+        if ASV_USE_IMPL == "modin":
+            pd.DataFrame([])
```
devin-petersohn, line 45: We should probably have a `modin.init` instead of using this 😄. In another PR of course.

scripts/doc_checker.py
```diff
@@ -0,0 +1,538 @@
...
+    args = parser.parse_args()
+    check_args(args)
+    return args
+
+
+if __name__ == "__main__":
+    args = get_args()
+    monkeypatching()
+    if not validate(args.paths, args.add_ignore, not args.disable_numpydoc):
+        print("NOT SUCCESSFUL CHECK")
```
devin-petersohn, line 536: A couple of questions here:

* Should this go to stderr instead of stdout?
* Should we show which files are incorrect? I think that would help a lot.
* A minor rewording suggestion:

```suggestion
        print("Invalid Documentation Found")
```

devin-petersohn (commented):
Left a couple of comments. The doc_checker looks really nice. Great job!

.github/workflows/ci.yml
anmyachev: In the latest version I use `logging`, please review.

scripts/doc_checker.py
anmyachev: > A couple of questions here:
> 
> * Should this go to stderr instead of stdout?

I think that errors found in the documentation should go to `stdout`, and errors related to the operation of the script itself should go to `stderr`. That is, in the case of this print, it will be output to the `stdout` with "error" mark.


> * Should we show which files are incorrect? I think that would help a lot.

The output of this script already displays the filepaths. For example:

```python
..\projects\modin>python scripts\doc_checker.py asv_bench\benchmarks\utils.py

INFO:PYDOCSTYLE OUTPUT FOR asv_bench\benchmarks\utils.py
ERROR:asv_bench\benchmarks\utils.py:125 in public class `weakdict`:
        D101: Missing docstring in public class

INFO:NUMPYDOC OUTPUT FOR asv_bench\benchmarks\utils.py - CAN BE EMPTY
ERROR:asv_bench.benchmarks.utils.translator_groupby_ngroups:102:PR01:Parameters {'groupby_ngroups'} not documented    
ERROR:asv_bench.benchmarks.utils.translator_groupby_ngroups:102:PR02:Unknown parameters {'groupby_ngroups: str or int'}
ERROR:asv_bench.benchmarks.utils.translator_groupby_ngroups:102:PR10:Parameter "groupby_ngroups" requires a space before the colon separating the parameter name and type
ERROR:asv_bench.benchmarks.utils.weakdict:125:GL08:The object does not have a docstring
ERROR:INVALID DOCUMENTATION FOUND
```
Are you ok with the output, or we need to add something else?

asv_bench/benchmarks/io/csv.py
anmyachev: I agree:)

Create the issue for this #2989.

scripts/doc_checker.py
```diff
@@ -0,0 +1,546 @@
...
+        capture_output=True,
+    )
+    if result.returncode:
+        logging.info(f"PYDOCSTYLE OUTPUT FOR {path}")
+        logging.error(result.stdout)
+    return True if result.returncode == 0 else False
+
+
+def monkeypatching():
+    """Monkeypatch decorators which change __doc__ attribute."""
```
YarShev, line 440: Doesn't `GL01` say that a docstring should start in the line below?

scripts/doc_checker.py
```diff
@@ -0,0 +1,546 @@
...
+        Docstring handler.
+
+    Returns
+    -------
+    list
+        List of tuples with Modin error code and its description.
+
+    Notes
+    -----
+    "modin" word is treated as the constant.
```
YarShev, line 137: Please, explain this phrase. I do not fully understand it. Also, why `modin`, but not `Modin`?

scripts/doc_checker.py
```diff
@@ -0,0 +1,546 @@
...
+                        found=type_line,
+                    ),
+                )
+            )
+    return errors
+
+
+def check_spelling_words(doc: Docstring) -> list:
+    """
+    Check spelling of chosen words: "Modin", "pandas", "NumPy" in doc.
```
YarShev, line 123: Let's add `Ray` and `Dask` more.

scripts/doc_checker.py
anmyachev: The check is only used for multi-line docstring.

scripts/doc_checker.py
anmyachev: Any special words enclosed in apostrophes(") are treated as python string constants and are not checked for spelling.

scripts/doc_checker.py
YarShev: Okay, I see.

scripts/doc_checker.py
YarShev: Okay, I see.

YarShev (dismissed):
LGTM!

vnlitvinov (approved):
LGTM

prutskov (approved):
thanks @anmyachev, LGTM!

scripts/doc_checker.py
YarShev: Let's add `PyArrow`, `OmniSci` and `XGBoost` more :)

YarShev (approved):

prutskov (approved):

scripts/doc_checker.py
```diff
@@ -358,14 +365,15 @@ def numpydoc_validate(path: pathlib.Path) -> bool:
...
+            if not _file.endswith(".py"):
                 continue
 
-            current_path = os.path.join(root, f)
+            current_path = os.path.join(root, _file)
             # get importable name
             module_name = current_path.replace("/", ".").replace("\\", ".")
-            module_name = module_name.replace(".py", "")
+            # remove ".py"
+            module_name = module_name[:-3]
```
vnlitvinov, line 376:
```suggestion
            module_name = os.path.splitext(module_name)[0]
```
[follow the manual, Luke](https://docs.python.org/3/library/os.path.html#os.path.splitext)

scripts/doc_checker.py
```diff
@@ -442,7 +450,13 @@ def monkeypatching():
     import ray
     import modin.utils
 
-    ray.remote = lambda *args, **kwargs: args[0]
+    def monkeypatch(*args, **kwargs):
+        if len(args) == 1 and len(kwargs) == 0 and callable(args[0]):
+            # This is the case where the decorator is just @ray.remote.
```
vnlitvinov, line 455:
```suggestion
            # This is the case where the decorator is just @ray.remote without parameters.
```

scripts/doc_checker.py
```diff
@@ -120,7 +120,9 @@ def check_optional_args(doc: Docstring) -> list:
 
 def check_spelling_words(doc: Docstring) -> list:
     """
-    Check spelling of chosen words: "Modin", "pandas", "NumPy", "Ray", "Dask" in doc.
+    Check spelling of chosen words in doc.
+
+    Words: "Modin", "pandas", "NumPy", "Ray", "Dask", "PyArrow", "OmniSci", "XGBoost".
```
vnlitvinov, line 125: I don't think we need to specify the words here - it's unnecessary duplication IMHO, they're specified a few dozen lines below.

vnlitvinov (approved):
Minor, easy to fix. Hence flagging "approval".

vnlitvinov (dismissed):
LGTM

scripts/doc_checker.py
```diff
@@ -0,0 +1,559 @@
...
+            functions = [node for node in module.body if is_public_func(node)]
+            classes = [node for node in module.body if isinstance(node, ast.ClassDef)]
+            methods = [
+                f"{module_name}.{_class.name}.{node.name}"
+                for _class in classes
+                for node in _class.body
+                if is_public_func(node)
+            ]
+
+            logging.info(f"NUMPYDOC OUTPUT FOR {current_path} - CAN BE EMPTY")
```
prutskov, line 396: As for me, output like 
```
INFO:NUMPYDOC OUTPUT FOR example.py  CAN BE EMPTY
```
confuses me a little. I would prefer something like this:
```
INFO:NUMPYDOC OUTPUT FOR example.py  SUCCESSFUL
```

anmyachev: Now the message appears only in case of an error.
```python
INFO:PYDOCSTYLE OUTPUT FOR scripts\doc_checker.py
ERROR:scripts\doc_checker.py:341 in public function `numpydoc_validate`:
        D103: Missing docstring in public function

INFO:NUMPYDOC OUTPUT FOR scripts\doc_checker.py
ERROR:scripts.doc_checker.numpydoc_validate:341:GL08:The object does not have a docstring
ERROR:scripts.doc_checker.pydocstyle_validate:397:PR01:Parameters {'add_ignore'} not documented
ERROR:scripts.doc_checker.pydocstyle_validate:397:PR02:Unknown parameters {'add_ignore: List[int]'}
ERROR:scripts.doc_checker.pydocstyle_validate:397:PR10:Parameter "add_ignore" requires a space before the colon separating the parameter name and type
ERROR:INVALID DOCUMENTATION FOUND
```

anmyachev: @devin-petersohn now the script output is like this, see above.

devin-petersohn (approved):
Thanks @anmyachev, looks great!

scripts/doc_checker.py
devin-petersohn: Looks great!

PR #2969, opened 2021-04-09
Improve docstring inheritance

### LINKED ISSUE #2967: Improve docstring inheritance from Pandas
Current inheritance requires explicit exclusion of what docstrings to leave in place.

I propose to just leave non-empty docstrings be, plus to add abilities to point to Pandas API online documentation.

vnlitvinov:
## What do these changes do?
* Remove special function decorator for copying docstring - now both function and class cases are handled by one decorator
* Now decorator adds link to Pandas API if requested
* `modin.pandas.DataFrame` now has said links in docstrings
* Added test that validates that all added URLs exist

- [x] passes `flake8 modin`
- [x] Resolves #2967 
- [x] tests added and passing


vnlitvinov:
cc @gshimansky

gshimansky (approved):

modin/utils.py
```diff
@@ -11,53 +11,141 @@
...
+        and apilink
+        and "Pandas API documentation <" not in target_doc
+        and (not (attr_name or "").startswith("_"))
+    ):
+        if attr_name:
+            token = f"{apilink}.{attr_name}"
+        else:
+            token = apilink
+        url = PANDAS_API_URL_TEMPLATE.format(token)
+        if DocstringUrlTestMode.get():
```
devin-petersohn, line 78: Instead of special casing this for the tests, can we factor out the URL generation independent of this and test that independently?

devin-petersohn (commented):
Thanks @vnlitvinov, great changes. I left one comment.

modin/utils.py
vnlitvinov: I'm not sure I understand your thought.

I could change the test so it would parse URLs out of docstrings (using `re` or smth like this), which would eliminate the need for separate list of URLs.

But I don't like the idea of enumerating every object and its properties...

So I'm not sure what you want to achieve here :)

vnlitvinov: **Gregory** gives me a pointer that what you actually wanted from me here is a separate (unit) test that checks the "add link" functionality (a test for the decorator). This is certainly possible, but it won't replace what I'm testing here - I'm testing that the decorator was used _correctly_ and that the decorator, when used, generated URLs that really _exist_.

devin-petersohn: I see. I prefer to keep special casing testing code out of the implementation. The URL generation test is important, so would it be sufficient to test that part as a unit outside of the logic? We are only testing that the URL is valid, so I believe it should be possible to do.

vnlitvinov: As greed over personal communication, I've removed this testing tidbits and replaced them by a moneky-patchable function for URL generation. I hope it works better for everyone now.

modin/conftest.py
```diff
@@ -426,3 +446,8 @@ def TestReadGlobCSVFixture():
     yield
 
     teardown_test_files(filenames)
+
+
+@pytest.fixture
+def get_generated_doc_urls():
+    return lambda: _generated_doc_urls
```
gshimansky, line 453: Just curious, why do you need `lambda` here instead of just returning `_generated_doc_urls`?

modin/test/test_docstring_urls.py
```diff
@@ -0,0 +1,49 @@
...
+            pass
+    return sorted(get_generated_doc_urls())
+
+
+def test_all_urls_exist(doc_urls):
+    broken = []
+
+    def _test_url(url):
+        try:
+            with urlopen(url):
```
gshimansky, line 41: Are you sure we really need to load all documentation URLs? Won't that get us banned for DDoSing pandas documentation web server?

gshimansky (approved):

modin/conftest.py
vnlitvinov: Because of how pytest fixtures work - they're computed before a function is executed, and the function I'm using it in is actively trying to load everything to load all the URLs (thus I'm expecting a change _after_ the function finishes).

modin/test/test_docstring_urls.py
vnlitvinov: No idea, but I hope it won't.
I don't have any other good way of checking we're not generating some 404 URLs, though, except straight up checking them.

devin-petersohn: Can we run the test on push rather than on PRs? That will reduce the amount of times it is run. We can create a separate test like we do with `ray-master`

vnlitvinov: Great idea, moved.

devin-petersohn (commented):
This looks good to me. @vnlitvinov are there any other changes necessary?

modin/test/test_docstring_urls.py
gshimansky: > Great idea, moved.

I see you moved call to `modin/test/test_docstring_urls.py` into `push.yml`. It runs `on: push` which according to [GitHub actions docs](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions) makes it run on every push to every branch. Not just push to a pull-request branch. It means that this test will run more often, not less often. Is it what you want to happen?

gshimansky (approved):

vnlitvinov:
@devin-petersohn 

> This looks good to me. @vnlitvinov are there any other changes necessary?

I think I got all covered for now. We can always make more PRs later if we need to 😄

devin-petersohn (approved):

modin/test/test_docstring_urls.py
vnlitvinov: huh, where should I have put it then?.. :)

gshimansky: That's a good question, but `ci.yml` is better than `push.yml` because `ci` is run for PRs while `push` is run for every commit.

PR #2975, opened 2021-04-14
Fixed bug in sort_values related to handling ignore_index and key parameters

### LINKED ISSUE #2974: ignore_index in sort_values has no effect
### System information

Ubuntu 20.04.2 LTS

0.9.1+24.gb7e1188f

- **Python version**:

Python 3.8.6

```python
import modin.pandas as pd

df = pd.DataFrame({"col1": [2, 51, 17], "col2": [5.5, 3.3, 4.4], "col3": ["a", "b", "c"]}, index=["abc", "def", "ghi"])
print(df)
print(df.index)
df.sort_values(by=["col1"], inplace=True, ignore_index=True)
print(df)
```

### Describe the problem

With `ignore_index=True` `sort_values` should reset index to default values. But in Modin it doesn't happen.

gshimansky:
## What do these changes do?

- [x] passes `flake8 modin`
- [x] Resolves #2974 ? 
- [x] tests added and passing


devin-petersohn:
@gshimansky Can you rebase this on current master?

devin-petersohn (commented):
How many new tests are added with this PR?

gshimansky:
> @gshimansky Can you rebase this on current master?

I would rather rebase it on #3032 where we have some CI tests acceleration after it is merged.

> How many new tests are added with this PR?

Before this PR we have only 50 tests which I consider inadequate. With this PR there are total 25922 tests, 11520 of which are skipped by default as extra, 8640 are xfailed and 4560 xpassed. Full run (without extra) takes 15 minutes single threaded

gshimansky:
After investigating XPASS tests that now considered as failed, I moved some to extra tests, so the final numbers are here:
`1118 passed, 21600 skipped, 3240 xfailed, 8728 warnings`, so it means 1068 new tests in default mode (without extra).

modin/backends/pandas/query_compiler.py
```diff
@@ -2958,8 +2958,8 @@ def sort_rows_by_column_values(self, columns, ascending=True, **kwargs):
         PandasQueryCompiler
             A new query compiler that contains result of the sort
         """
-        na_position = kwargs.get("na_position", "last")
-        kind = kwargs.get("kind", "quicksort")
+        ignore_index = kwargs.get("ignore_index", False)
```
devin-petersohn, line 2961: Could we handle `ignore_index` at the API layer? Or do you think it is better here?

devin-petersohn: Makes sense.

devin-petersohn (approved):

PR #2977, opened 2021-04-14
FIX-#2976: Update pandas to 1.2.4

### LINKED ISSUE #2976: Update pandas to 1.2.4

YarShev:

## What do these changes do?

- [x] passes `flake8 modin`
- [x] Resolves #2976  
- [x] tests passing

### DIFF
--- a/modin/backends/base/query_compiler.py
+++ b/modin/backends/base/query_compiler.py
@@ -573,32 +573,6 @@ class BaseQueryCompiler(abc.ABC):
         """
         return DataFrameDefault.register(pandas.DataFrame.min)(self, **kwargs)
 
-    def sum(self, squeeze_self, axis, **kwargs):
-        """Returns the sum of each numerical column or row.
-
-        Return:
-            Pandas series with the sum of each numerical column or row.
-        """
-        # TODO: rework to original implementation after pandas issue #41074 resolves if possible.
-        def map_func(df, **kwargs):
-            if squeeze_self:
-                result = df.squeeze(axis=1).sum(**kwargs)
-                if is_scalar(result):
-                    if axis:
-                        return pandas.DataFrame(
-                            [result], index=["__reduced__"], columns=["__reduced__"]
-                        )
-                    else:
-                        return pandas.Series([result], index=[df.columns[0]])
-                else:
-                    return result
-            else:
-                return df.sum(**kwargs)
-
-        return DataFrameDefault.register(
-            map_func,
-        )(self, axis=axis, **kwargs)
-
     def prod(self, squeeze_self, axis, **kwargs):
         """Returns the product of each numerical column or row.
 
@@ -607,6 +581,7 @@ class BaseQueryCompiler(abc.ABC):
         """
         # TODO: rework to original implementation after pandas issue #41074 resolves if possible.
         def map_func(df, **kwargs):
+            """Apply .prod to DataFrame or Series in depend on `squeeze_self.`"""
             if squeeze_self:
                 result = df.squeeze(axis=1).prod(**kwargs)
                 if is_scalar(result):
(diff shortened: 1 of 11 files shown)


vnlitvinov (approved):
LGTM

devin-petersohn (dismissed):
Thanks @YarShev!

devin-petersohn (changes requested):
@YarShev Can we use `xfail` instead of `skipif` in the parameter markers?

vnlitvinov:
> @YarShev Can we use `xfail` instead of `skipif` in the parameter markers?

I know the original author is going to be away for a few days, so I decided to step in and do the requested to speed up the things. I hope it's fine for everyone 😄

devin-petersohn:
@vnlitvinov It is ok with me, but we have some failures because `xfail` should always fail in our tests.

It looks like there are conditions where it still works.

vnlitvinov:
For reference: link to relevant Pandas issue - https://github.com/pandas-dev/pandas/issues/41074

I would suggest waiting for @YarShev to step up then, as I cannot locally reproduce the issues he was marking as `skipif`/`xfail` - if I update `pandas` to 1.2.4, take his branch and remove this `xfail` the tests pass... but he did mark those for some reason.

YarShev:
@devin-petersohn , @vnlitvinov , I got rid of xfail/skipif marks in the tests but it had to make some changes in the code to match pandas behavior.

modin/backends/base/query_compiler.py
```diff
@@ -573,21 +573,57 @@ def min(self, **kwargs):
...
+    def sum(self, squeeze_self, axis, **kwargs):
+        """Returns the sum of each numerical column or row.
 
         Return:
-            Pandas series with the product of each numerical column or row.
+            Pandas series with the sum of each numerical column or row.
         """
-        return DataFrameDefault.register(pandas.DataFrame.prod)(self, **kwargs)
+        # TODO: rework to original implementation after pandas issue #41074 resolves if possible.
+        def map_func(df, **kwargs):
```
vnlitvinov, line 583: pls add one-line summary docstring

modin/backends/base/query_compiler.py
```diff
@@ -573,21 +573,57 @@ def min(self, **kwargs):
...
+    def prod(self, squeeze_self, axis, **kwargs):
+        """Returns the product of each numerical column or row.
 
         Return:
-            Pandas series with the sum of each numerical column or row.
+            Pandas series with the product of each numerical column or row.
         """
-        return DataFrameDefault.register(pandas.DataFrame.sum)(self, **kwargs)
+        # TODO: rework to original implementation after pandas issue #41074 resolves if possible.
+        def map_func(df, **kwargs):
```
vnlitvinov, line 609: pls add one-line docstring

modin/backends/base/query_compiler.py
```diff
@@ -573,21 +573,57 @@ def min(self, **kwargs):
         """
         return DataFrameDefault.register(pandas.DataFrame.min)(self, **kwargs)
 
-    def prod(self, **kwargs):
-        """Returns the product of each numerical column or row.
+    def sum(self, squeeze_self, axis, **kwargs):
+        """Returns the sum of each numerical column or row.
```
vnlitvinov, line 577: what's the reason for swapping the functions?

vnlitvinov (dismissed):
LGTM with some documentation notes

modin/backends/base/query_compiler.py
```diff
@@ -573,21 +573,57 @@ def min(self, **kwargs):
...
+                    if axis:
+                        return pandas.DataFrame(
+                            [result], index=["__reduced__"], columns=["__reduced__"]
+                        )
+                    else:
+                        return pandas.Series([result], index=[df.columns[0]])
+                else:
+                    return result
+            else:
+                return df.sum(**kwargs)
```
vnlitvinov, line 596: I see functions like this in 6 places (`sum` and `prod` here, and `sum`/`prod`/`sum_min_count`/`prod_min_count` in pandas query compiler), can we reduce the amount of copy-pasteness somehow?

modin/backends/base/query_compiler.py
YarShev: There is no any reason.

vnlitvinov: in that case pls swap them back to reduce the diff :)

modin/backends/base/query_compiler.py
YarShev: All our map functions look similarly. The difference is that we register different functions.

modin/backends/base/query_compiler.py
YarShev: added

modin/backends/base/query_compiler.py
YarShev: added

modin/backends/base/query_compiler.py
YarShev: swapped

vnlitvinov (approved):
LGTM

devin-petersohn (approved):
LGTM, thanks @YarShev

PR #3000, opened 2021-04-20
Add documentation to modin/pandas/base.py

gshimansky:
## What do these changes do?

- [x] commit message follows format outlined [here](https://modin.readthedocs.io/en/latest/contributing.html)
- [x] passes `flake8 modin`
- [x] passes `black --check modin`
- [x] signed commit with `git commit -s` 
- [x] Resolves #2942 
- [x] tests added and passing

PR #3001, opened 2021-04-21
Improve doc checker

### LINKED ISSUE #2995: Doc checker quirks
### Describe the problem
There are certain quirks which doc checker from #2968 does not handle.

To reproduce place the code below as `modin/demo.py` and run `python scripts/doc_checker modin/demo.py`.

The issues are:
1. The script does not handle poor monkey-patching (replacing with `lambda` without copying a docstring)
2. `numpydoc.Validate()` reports weird results if identation in `Parameters` section is too big
3. The script does not point to exact line of failure, but rather to object definition start (makes fixing misspelled `moDin` in the example below harder)

We should check identation and report it to the user.

### Source code / logs
<details><summary>modin/demo.py</summary>

```python
def foo(a, b, c):
    """
    Foo mumble mumble.

    moDin is misspelled on purpose.

    Parameters
    ----------
        a : object
            This is description.
        b : object
            More text goes here.
        c : object
            Outta fantasy what to say.

    Returns
    -------
    int
        Random number, generated by fair dice roll.
    """
    return 4


foo = lambda a, b, c: 5
```
</details>

<details><summary>Output when `foo` monkey-patching is active</summary>

```
INFO:PYDOCSTYLE OUTPUT FOR modin\demo.py
ERROR:modin\demo.py:1 at module level:
        D100: Missing docstring in public module
modin\demo.py:1 in public function `foo`:
        D103: Missing docstring in public function

  File "scripts/doc_checker.py", line 557, in <module>
    if not validate(args.paths, args.add_ignore, not args.disable_numpydoc):
  File "scripts/doc_checker.py", line 489, in validate
    if not numpydoc_validate(path):
  File "scripts/doc_checker.py", line 402, in numpydoc_validate
    results = list(map(validate_object, to_validate))
  File "scripts/doc_checker.py", line 328, in validate_object
    noqa_checks = get_noqa_checks(doc)
  File "scripts/doc_checker.py", line 298, in get_noqa_checks
    if "# noqa:" in noqa_str:
TypeError: argument of type 'NoneType' is not iterable
```
</details>

<details><summary>Output when `foo = lambda` is commented out</summary>

```
INFO:PYDOCSTYLE OUTPUT FOR modin\demo.py
ERROR:modin\demo.py:1 at module level:
        D100: Missing docstring in public module

INFO:NUMPYDOC OUTPUT FOR modin\demo.py
ERROR:modin.demo:0:GL08:The object does not have a docstring
ERROR:modin.demo.foo:1:PR01:Parameters {'c', 'b'} not documented
ERROR:modin.demo.foo:1:MD02:Spelling error in line: 3, found: 'moDin', reference: 'Modin'
ERROR:INVALID DOCUMENTATION FOUND
```
</details>

vnlitvinov:
**Note**: this is based on #3034 and should be rebased when that one is merged.

## What do these changes do?
Improve `doc_checker.py` to handle a few specific corner cases.
Also make it skip numpydoc-validating inherited docstrings without requiring to disable numpydoc check.

- [x] passes `flake8 modin`
- [x] Resolves #2995 
- [x] tests added and passing

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,13 +49,16 @@ jobs:
       - run: pytest scripts/test
       - run: pip install -e .[all]
       - run: |
-          python scripts/doc_checker.py \
+          python scripts/doc_checker.py --disable-numpydoc \
             modin/pandas/dataframe.py modin/pandas/series.py \
-            modin/pandas/groupby.py modin/pandas/base.py modin/pandas/io.py \
-            modin/pandas/series_utils.py modin/engines/base/io/io.py \
-            modin/backends/pandas/parsers.py modin/pandas/general.py \
+            modin/pandas/groupby.py modin/pandas/base.py \
+            modin/pandas/series_utils.py modin/pandas/general.py \
             modin/pandas/plotting.py modin/pandas/utils.py \
             modin/pandas/iterator.py modin/pandas/indexing.py \
+      - run: |
+          python scripts/doc_checker.py \
+            modin/pandas/io.py modin/engines/base/io/io.py \
+            modin/backends/pandas/parsers.py  \
             modin/engines/base/frame asv_bench/benchmarks/utils.py \
             asv_bench/benchmarks/__init__.py asv_bench/benchmarks/io/__init__.py \
             asv_bench/benchmarks/scalability/__init__.py \
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,29 +49,14 @@ jobs:
       - run: pytest scripts/test
       - run: pip install -e .[all]
       - run: |
-          python scripts/doc_checker.py --add-ignore=D101,D102,D105 --disable-numpydoc \
+          python scripts/doc_checker.py \
             modin/pandas/dataframe.py modin/pandas/series.py \
-            modin/pandas/groupby.py modin/pandas/base.py
-      - run: python scripts/doc_checker.py modin/pandas/io.py
-      - run: |
-          python scripts/doc_checker.py --add-ignore=D101,D102 --disable-numpydoc \
-            modin/pandas/series_utils.py
-      - run: |
-          python scripts/doc_checker.py --add-ignore=D102 \
-            modin/engines/base/io/io.py
-      - run: |
-          python scripts/doc_checker.py --add-ignore=D101,D102 \
-            modin/backends/pandas/parsers.py
-      - run: |
-          python scripts/doc_checker.py --add-ignore=D103 --disable-numpydoc \
-            modin/pandas/general.py
-      - run: |
-          python scripts/doc_checker.py --disable-numpydoc \
+            modin/pandas/groupby.py modin/pandas/base.py modin/pandas/io.py \
+            modin/pandas/series_utils.py modin/engines/base/io/io.py \
+            modin/backends/pandas/parsers.py modin/pandas/general.py \
             modin/pandas/plotting.py modin/pandas/utils.py \
-            modin/pandas/iterator.py modin/pandas/indexing.py
-      - run: python scripts/doc_checker.py modin/engines/base/frame
-      - run: |
-          python scripts/doc_checker.py asv_bench/benchmarks/utils.py \
+            modin/pandas/iterator.py modin/pandas/indexing.py \
+            modin/engines/base/frame asv_bench/benchmarks/utils.py \
             asv_bench/benchmarks/__init__.py asv_bench/benchmarks/io/__init__.py \
             asv_bench/benchmarks/scalability/__init__.py \
             modin/engines/base/io/column_stores \
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,7 +49,7 @@ jobs:
       - run: pytest scripts/test
       - run: pip install -e .[all]
       - run: |
-          python scripts/doc_checker.py --disable-numpydoc \
+          python scripts/doc_checker.py --add-ignore=D101,D102,D103,D105 --disable-numpydoc \
             modin/pandas/dataframe.py modin/pandas/series.py \
             modin/pandas/groupby.py modin/pandas/base.py \
             modin/pandas/series_utils.py modin/pandas/general.py \
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -220,6 +220,8 @@ jobs:
         run: python -m pytest modin/test/test_envvar_npartitions.py
       - shell: bash -l {0}
         run: python -m pytest -n 2 modin/test/test_partition_api.py
+      - shell: bash -l {0}
+        run: python -m pytest modin/test/test_utils.py
 
   test-defaults:
     needs: [lint-commit, lint-flake8, lint-black, test-api, test-headers]
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,7 +45,7 @@ jobs:
         with:
           python-version: "3.7.x"
           architecture: "x64"
-      - run: pip install pytest pytest-cov pydocstyle numpydoc
+      - run: pip install pytest pytest-cov pydocstyle numpydoc==1.1.0
       - run: pytest scripts/test
       - run: pip install -e .[all]
       - run: |
--- a/modin/utils.py
+++ b/modin/utils.py
@@ -104,7 +104,8 @@ def _replace_doc(
 
     source_doc = source_obj.__doc__ or ""
     target_doc = target_obj.__doc__ or ""
-    doc = source_doc if overwrite or not target_doc else target_doc
+    overwrite = overwrite or not target_doc
+    doc = source_doc if overwrite else target_doc
 
     if parent_cls and not attr_name:
         if isinstance(target_obj, property):
@@ -125,15 +126,19 @@ def _replace_doc(
         else:
             token = apilink
         url = _make_api_url(token)
-        doc += f"\n\n{' ' * _get_indent(doc)}See `Pandas API documentation for {token} <{url}>`_ for more."
+        doc += f"\n{' ' * _get_indent(doc)}See `pandas API documentation for {token} <{url}>`_ for more.\n"
 
     if parent_cls and isinstance(target_obj, property):
+        if overwrite:
+            target_obj.fget.__doc_inherited__ = True
         setattr(
             parent_cls,
             attr_name,
             property(target_obj.fget, target_obj.fset, target_obj.fdel, doc),
         )
     else:
+        if overwrite:
+            target_obj.__doc_inherited__ = True
         target_obj.__doc__ = doc
 
 
--- a/modin/utils.py
+++ b/modin/utils.py
@@ -72,6 +72,11 @@ def _replace_doc(
         Gives the name to `target_obj` if it's an attribute of `parent_cls`.
         Needed to handle some special cases and in most cases could be determined automatically.
     """
+    if isinstance(target_obj, (staticmethod, classmethod)):
+        # we cannot replace docs on decorated objects, we must replace them
+        # on original functions instead
+        target_obj = target_obj.__func__
+
     source_doc = source_obj.__doc__ or ""
     target_doc = target_obj.__doc__ or ""
     doc = source_doc if overwrite or not target_doc else target_doc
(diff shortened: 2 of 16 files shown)


vnlitvinov:
@anmyachev I think this is a starter for fixing #2995, you can take it as a base and push to my branch with other fixes.

modin/utils.py
```diff
@@ -87,23 +113,27 @@ def _replace_doc(
     if (
         source_doc.strip()
         and apilink
-        and "Pandas API documentation <" not in target_doc
+        and "`Pandas API documentation for " not in target_doc
```
anmyachev, line 116:
```suggestion
        and "`pandas API documentation for " not in target_doc
```

scripts/doc_checker.py
```diff
@@ -201,6 +232,8 @@ def validate_modin_error(doc: Docstring, results: dict) -> list:
     """
     errors = check_optional_args(doc)
     errors += check_spelling_words(doc)
+    if any((error_code == "PR01" for error_code, _ in results["errors"])):
+        errors += check_parameters_indention(doc)
```
vnlitvinov, line 236:
```suggestion
    errors += check_parameters_indention(doc)
```
we should check indentation regardless of other checks

scripts/doc_checker.py
```diff
@@ -183,6 +184,36 @@ def check_spelling_words(doc: Docstring) -> list:
     return errors
 
 
+def check_parameters_indention(doc: Docstring) -> list:
```
vnlitvinov, line 187: function name is now wrong

scripts/doc_checker.py
```diff
@@ -50,6 +50,7 @@
 MODIN_ERROR_CODES = {
     "MD01": "'{parameter}' description should be '[type], default: [value]', found: '{found}'",
     "MD02": "Spelling error in line: {line}, found: '{word}', reference: '{reference}'",
+    "MD03": "Section underline is over-indented (in section '{section}')",
```
vnlitvinov, line 53:
```suggestion
    "MD03": "Section contents is over-indented (in section '{section}')",
```

modin/utils.py
vnlitvinov: @anmyachev pls make this change next time you'll be fixing something in there (I think line numbers are still off, so they need to be fixed, right?)

anmyachev: Yes, line numbers need to be fixed.

.github/workflows/ci.yml
```diff
@@ -45,7 +45,7 @@ jobs:
         with:
           python-version: "3.7.x"
           architecture: "x64"
-      - run: pip install pytest pytest-cov pydocstyle numpydoc
+      - run: pip install pytest pytest-cov pydocstyle numpydoc==1.1.0
```
anmyachev, line 48: Internal API is used, which may break in the next release.

anmyachev:
@vnlitvinov ready for review

scripts/doc_checker.py
```diff
@@ -166,6 +167,12 @@ def check_spelling_words(doc: Docstring) -> list:
         for line in doc.raw_doc.splitlines()
     ]
 
+    docstring_start_line = None
+    for idx, line in enumerate(inspect.getsourcelines(doc.code_obj)[0]):
+        if '"""' in line:
```
vnlitvinov, line 172: while we discourage using `'''` as a docstring thing (and `black` replaces it with double quotes), I think it's better to handle both cases:
```suggestion
        if '"""' in line or "'''" in line:
```

vnlitvinov (commented):
Overall looks fine.

YarShev:
Since for now the script throws the error:
```
'foo' description should be '[type], default: [value]', found: '{'a', 'b', 'c'}'
```
It would be great to add possibility of specifying fixed set of values for a parameter if it assumes to be one of these. Please, look at [this](https://github.com/modin-project/modin/issues/2934#issuecomment-815904375).

anmyachev:
> 'foo' description should be '[type], default: [value]', found: '{'a', 'b', 'c'}'

I suppose it is possible to use the form: `foo: {'a', 'b', 'c'}, optional`

YarShev:
> > 'foo' description should be '[type], default: [value]', found: '{'a', 'b', 'c'}'
> 
> I suppose it is possible to use the form: `foo: {'a', 'b', 'c'}, optional`

That doesn't fit numpydoc style. We should be as closer to the guide as possible.

anmyachev:
> > > 'foo' description should be '[type], default: [value]', found: '{'a', 'b', 'c'}'
> > 
> > 
> > I suppose it is possible to use the form: `foo: {'a', 'b', 'c'}, optional`
> 
> That doesn't fit numpydoc style. We should be as closer to the guide as possible.

It seems to me that the format assumed in the numpydoc guide is more error-prone, since there is additional implicitness.

YarShev:
Why the format assumed in the numpydoc guide is more error-prone? What implicitness do you mean?

anmyachev:
> Why the format assumed in the numpydoc guide is more error-prone? What implicitness do you mean?

That the first value in the set is assumed to be the default.

vnlitvinov:
I guess **Anatoly** means that when we use `{'a', 'b', 'c'}` form _without_ specifying default value (or optionality) it's unclear that this parameter has a default value (and if it has, which value it is).

anmyachev:
Also, how will we distinguish this situation from the one where there is also a limited set of accepted values, but there is no default value?

YarShev:
> I guess **Anatoly** means that when we use `{'a', 'b', 'c'}` form _without_ specifying default value (or optionality) it's unclear that this parameter has a default value (and if it has, which value it is).

That is explained in the [guide](https://numpydoc.readthedocs.io/en/latest/format.html#parameters).

YarShev:
> Also, how will we distinguish this situation from the one where there is also a limited set of accepted values, but there is no default value?

Since this is a case of required parameter we would specify just its type and possible values would follow in the description like:
```python
foo : str
  A description of foo. Possible values are {'a', 'b', 'c'}.
```

anmyachev:
> > Also, how will we distinguish this situation from the one where there is also a limited set of accepted values, but there is no default value?
> 
> Since this is a case of required parameter we would specify just its type and possible values would follow in the description like:
> 
> ```python
> foo : str
>   A description of foo. Possible values are {'a', 'b', 'c'}.
> ```

As for me, this is not consistent behavior when the supported values are written either with types or in the parameter description.

YarShev:
We can still put possible values on the same lime with parameter name.
```python
foo : {'a', 'b', 'c'}
  A description of foo.
```
But still if we deal with optional parameter that can take a value of fixed set and it is not clear which default value is used, okay, let's use this form:
```python
foo : {"a", "b", "c"}, default: "a"
```

vnlitvinov (commented):
LGTM (though as I'm partial author of this I cannot put up a green mark).

anmyachev (approved):

anmyachev:
> LGTM (though as I'm partial author of this I cannot put up a green mark).

I approved :)

docs/developer/pandas/base.rst
```diff
@@ -0,0 +1,12 @@
+BasePandasDataset
```
YarShev, line 1: I am not sure of this document should be put here. Why don't we want to put it into `docs/flow/modin/pandas/`?
cc @devin-petersohn

anmyachev: The branch is based on #3000. It is better to discuss it there:)

scripts/doc_checker.py
```diff
@@ -427,6 +452,8 @@ def pydocstyle_validate(path: pathlib.Path, add_ignore: List[str]) -> int:
     pydocstyle = "pydocstyle"
     if not shutil.which(pydocstyle):
         raise ValueError(f"{pydocstyle} not found in PATH")
+    # These check can be done with numpydoc tool, so disable them for pydocstyle.
+    add_ignore.extend(["D100", "D101", "D102", "D103", "D104", "D105"])
```
vnlitvinov, line 456: pls note that this should be disabled _only_ if `numpydoc` check is enabled...

modin/utils.py
```diff
@@ -355,3 +364,13 @@ def get_current_backend():
         Returns <Backend>On<Engine>-like string.
     """
     return f"{'Experimental' if IsExperimental.get() else ''}{Backend.get()}On{Engine.get()}"
+
+
+def instancer(cls):
```
vnlitvinov, line 369: this should have parameters and return value described :)

scripts/doc_checker.py
```diff
@@ -483,6 +479,18 @@ def monkeypatch(*args, **kwargs):
         return lambda cls_or_func: cls_or_func
 
     ray.remote = monkeypatch
+    modin.utils.instancer = lambda cls: cls
+
+    # monkey patch numpydoc for correct work with properties
+    old_load_obj = Docstring._load_obj
+
+    def load_obj(name):
```
vnlitvinov, line 487:
```suggestion
    # monkey-patch numpydoc for working correctly with properties
    def load_obj(name, old_load_obj=Docstring._load_obj):
```
slightly more compact :)

vnlitvinov (commented):
Please look at docstring failures, why there's so many?..

scripts/doc_checker.py
```diff
@@ -307,8 +308,15 @@ def get_noqa_checks(doc: Docstring) -> list:
...
-        # pydocstyle only check first line; aling with it
-        noqa_str = source.split("\n", 1)[0]
+        # noqa string is defined as the first line before the docstring
+        if not doc.raw_doc:
+            # noqa string is meaningless if there is no docstring in module
+            return []
+        lines = source.split("\n")
+        for idx, line in enumerate(lines):
+            if '"""' in line or "'''" in line:
+                noqa_str = lines[idx - 1]
```
vnlitvinov, line 318: you risk taking _last line_ of file if multiline docstring happens to start at the beginning of the file :)

scripts/doc_checker.py
```diff
@@ -307,8 +308,15 @@ def get_noqa_checks(doc: Docstring) -> list:
                 noqa_str = line
                 break
     else:
-        # pydocstyle only check first line; aling with it
-        noqa_str = source.split("\n", 1)[0]
+        # noqa string is defined as the first line before the docstring
+        if not doc.raw_doc:
+            # noqa string is meaningless if there is no docstring in module
+            return []
```
vnlitvinov, line 314: what if I want to disable a check for the whole module (not a check for module docstring)? don't we propagate module-level `noqa` to function level?

scripts/doc_checker.py
```diff
@@ -479,12 +487,15 @@ def monkeypatch(*args, **kwargs):
...
     ray.remote = monkeypatch
-    modin.utils.instancer = lambda cls: cls
 
-    # monkey patch numpydoc for correct work with properties
-    old_load_obj = Docstring._load_obj
+    @functools.wraps(modin.utils.instancer)
+    def instancer(cls):
+        return cls
+
+    modin.utils.instancer = instancer
```
vnlitvinov, line 495:
```suggestion
    modin.utils.instancer = functools.wraps(modin.utils.instancer)(lambda cls: cls)
```

vnlitvinov:
@anmyachev pls also note this: https://github.com/modin-project/modin/pull/3001#discussion_r621209018

modin/utils.py
```diff
@@ -366,11 +366,20 @@ def get_current_backend():
     return f"{'Experimental' if IsExperimental.get() else ''}{Backend.get()}On{Engine.get()}"
 
 
-def instancer(cls):
+def instancer(_class):
```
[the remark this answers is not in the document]
anmyachev: When using `cls` name, an error occurs from the `numpydoc` side.

scripts/doc_checker.py
anmyachev: Agree, but I'm relying on the main use case for docstring usage in Modin modules.

scripts/doc_checker.py
anmyachev: This will change the behavior of `numpydoc` quite a lot. Well, and immediately the first question regarding the implementation - how to distinguish between situations when we need to disable checks only for the module docstring and when we need to disable checks for the entire file. This adds another layer of complexity...

vnlitvinov (commented):
LGTM (but cannot approve own PRs)

vnlitvinov (commented):
LGTM again

devin-petersohn (approved):
LGTM! Thanks @vnlitvinov and @anmyachev

PR #3026, opened 2021-04-27
FIX-#3025: use pandas.DataFrame as return value

### LINKED ISSUE #3025: AttributeError: 'int' object has no attribute 'columns'
### System information
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: `Ubuntu 20.04`
- **Modin version** (`modin.__version__`): 8fe4be960440e2bac952491685c9e86c64715b78
- **Python version**: `Python 3.8.6`

```python
import modin.pandas as pd
import ray
ray.init()
import numpy as np
data = np.random.randint(0,100,(2**8,2**8))
df = pd.DataFrame(data)
df.to_csv('/tmp/test.csv')
```

### Describe the problem

https://github.com/modin-project/modin/issues/3014#issuecomment-827740232

### Source code / logs

```bash
2021-04-27 21:25:09,427	INFO services.py:1171 -- View the Ray dashboard at http://127.0.0.1:8265
2021-04-27 21:25:14,874	ERROR worker.py:980 -- Possible unhandled error from worker: ray::deploy_ray_func() (pid=4823, ip=192.168.31.208)
  File "python/ray/_raylet.pyx", line 463, in ray._raylet.execute_task
  File "/home/pratik/miniconda3/lib/python3.8/site-packages/modin/engines/ray/pandas_on_ray/frame/axis_partition.py", line 114, in deploy_ray_func
    result = func(*args)
  File "/home/pratik/miniconda3/lib/python3.8/site-packages/modin/engines/base/frame/axis_partition.py", line 294, in deploy_axis_func
    if sum(lengths) != len(result.columns):
AttributeError: 'int' object has no attribute 'columns'
```

anmyachev:

## What do these changes do?

- [x] passes `flake8 modin`
- [x] Resolves #3025 
- [x] tests passing

### DIFF
--- a/modin/engines/ray/generic/io.py
+++ b/modin/engines/ray/generic/io.py
@@ -128,7 +128,7 @@ class RayIO(BaseIO):
             queue.put(get_value + 1)
 
             # used for synchronization purposes
-            return 0
+            return pandas.DataFrame()
 
         # signaling that the partition with id==0 can be written to the file
         queue.put(0)


anmyachev:

vnlitvinov (commented):
TBH this change looks very weird, but I see why it helps.

I still think this should be declared a workaround, because current internal API that we use here expects the "partition func" to return the resulting chunk retaining its shape.

cc @YarShev

YarShev:
I am not sure of what the reason to return `0` as synchronization purposes from the apply function but we should probably return pandas DataFrame from it. Moreover, this DataFrame should be `df` that is passed in the function because there might be the error [here](https://github.com/modin-project/modin/blob/master/modin/data_management/utils.py#L103) when iterating over the empty DataFrame.

anmyachev:
> I am not sure of what the reason to return `0` as synchronization purposes from the apply function but we should probably return pandas DataFrame from it. Moreover, this DataFrame should be `df` that is passed in the function because there might be the error [here](https://github.com/modin-project/modin/blob/master/modin/data_management/utils.py#L103) when iterating over the empty DataFrame.

Returning zero is cheaper than a dataframe, due to it still needs to be created.

Returning the same dataframe that was passed in the function is expensive, since it still needs to be written to object storage.

YarShev:
As for performance, we can return something like `pandas.DataFrame([], columns=np.arange(len(df.columns)))`.

anmyachev:
> As for performance, we can return something like `pandas.DataFrame([], columns=np.arange(len(df.columns)))`.

But why? We have tests that don't break when using an empty dataframe.

vnlitvinov:
Because that's the expected behaviour - mapping functions should not change the shape of underlying frames, and here you end up with empty frame.

vnlitvinov:
cc @devin-petersohn 
Maybe we should create a portion of API that would behave just like that (i.e. would allow running some functions over the dataframe, be it block-by-block or axis-wise) that would _not_ expect a `DataFrame` as a result, i.e. effectively create a "scalar" transformation somehow - this is exactly what `to_csv()` does, it transforms our DataFrame into a file and a scalar result (either `None` or a `str` with some details on csv format).

devin-petersohn:
@vnlitvinov Let's create a new issue to track that. The source of the issue here is that we need to block on some return value, and Modin internal APIs expect partitions to be in their specific format. We probably need to have specific utilities for writing the partitions out to some other format.

Since this is blocking existing users, let's do that later.

devin-petersohn (approved):

vnlitvinov:
> Let's create a new issue to track that.

Created #3036, everyone feel free to edit it if you see fit.

PR #3028, opened 2021-04-27
FIX-#3027: Removed unnecessary 'str.repeats' function, fixed test

### LINKED ISSUE #3027: pandas.Series.str.repeats doesn't exist but Modin implements it
### System information

Ubuntu 20.04.2 LTS

0.9.1+37.g1eace767.dirty

- **Python version**:

Python 3.8.6

### Describe the problem

There is no such function as `pandas.Series.str.repeats` but it exists in Modin and has tests for it too. Tests pass accidentally because in Modin this function is also not fully implemented and ultimately raises the same `AttributeError` as in Pandas and this makes tests pass.

gshimansky:

## What do these changes do?

- [x] passes `flake8 modin`
- [x] Resolves #3027 
- [x] tests added and passing

### DIFF
--- a/modin/pandas/series_utils.py
+++ b/modin/pandas/series_utils.py
@@ -203,9 +203,6 @@ class StringMethods(object):
             )
         )
 
-    def repeats(self, repeats):
-        return Series(query_compiler=self._query_compiler.str_repeats(repeats))
-
     def pad(self, width, side="left", fillchar=" "):
         if len(fillchar) != 1:
             raise TypeError("fillchar must be a character, not str")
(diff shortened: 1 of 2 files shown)


vnlitvinov (approved):
LGTM

modin/pandas/series_utils.py
```diff
@@ -203,9 +203,6 @@ def replace(self, pat, repl, n=-1, case=None, flags=0, regex=True):
             )
         )
 
-    def repeats(self, repeats):
-        return Series(query_compiler=self._query_compiler.str_repeats(repeats))
```
devin-petersohn, line 207: Should this be removed from the query compiler as well?

devin-petersohn (approved):

modin/pandas/series_utils.py
devin-petersohn: I see, makes sense.

PR #3032, opened 2021-04-28
FIX-#3031: Added -n 2 to long running pytest tasks and paralellized windows tasks

### LINKED ISSUE #3031: Improve CI performance with xdist pytests

All tests currently executed on GitHub Actions work in single process. This is not optimal since even free GA workers contain 2 CPU cores, so we could speed up CI tasks at least 2x times. More speedup may be possible if we utilize our own GA workers with more CPU cores.

gshimansky:

## What do these changes do?

- [x] passes `flake8 modin`
- [x] Resolves #3031 
- [x] tests added and passing

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,7 +45,7 @@ jobs:
         with:
           python-version: "3.7.x"
           architecture: "x64"
-      - run: pip install pytest pytest-cov pydocstyle numpydoc
+      - run: pip install pytest pytest-cov pytest-xdist pydocstyle numpydoc
       - run: pytest scripts/test
       - run: pip install -e .[all]
       - run: |
@@ -219,7 +219,7 @@ jobs:
       - shell: bash -l {0}
         run: python -m pytest modin/test/test_envvar_npartitions.py
       - shell: bash -l {0}
-        run: python -m pytest modin/test/test_partition_api.py
+        run: python -m pytest -n 2 modin/test/test_partition_api.py
 
   test-defaults:
     needs: [lint-commit, lint-flake8, lint-black, test-api, test-headers]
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -45,7 +45,7 @@ jobs:
         with:
           python-version: "3.7.x"
           architecture: "x64"
-      - run: pip install pytest pytest-cov pydocstyle numpydoc
+      - run: pip install pytest pytest-cov pytest-xdist pydocstyle numpydoc
```
vnlitvinov, line 48: I don't think we need `xdist` here because this is about checking docs, they're fast enough without that :)

.github/workflows/ci.yml
```diff
@@ -406,39 +406,39 @@ jobs:
...
+        run: pytest -n 2 modin/pandas/test/dataframe/test_pickle.py
       - shell: bash -l {0}
-        run: python -m pytest modin/pandas/test/test_series.py
+        run: python -m pytest -n 2 modin/pandas/test/test_series.py
       - shell: bash -l {0}
-        run: python -m pytest modin/pandas/test/test_rolling.py
+        run: python -m pytest -n 2 modin/pandas/test/test_rolling.py
       - shell: bash -l {0}
-        run: python -m pytest modin/pandas/test/test_concat.py
+        run: python -m pytest modin/pandas/test/test_concat.py # Ray version fails with -n 2
```
vnlitvinov, line 435: can we use a `yml if` here somehow to use `-n 2` on non-Ray backends?

vnlitvinov (approved):
LGTM with a few minor comments

vnlitvinov (approved):
LGTM

devin-petersohn (approved):
LGTM, I left a comment here: https://github.com/modin-project/modin/issues/3031#issuecomment-831459406

Not sure this will completely solve the problem, but let's try.

PR #3120, opened 2021-06-03
FIX-#3116: Add variable to be able to set Redis password

### LINKED ISSUE #3116: Fix passing redis password to ray.init
Currently, we pass generated redis password to `ray.init()` that is not exactly correct at all times - this breaks the ability to connect to an already spawned Ray cluster in ordinary, non-cloud workflow.

We need to add an environment variable that is responsible for setting redis password. The value from that variable could be passed to `ray.init()` instead of a generated password.

vnlitvinov:

## What do these changes do?
Add `MODIN_REDIS_PASSWORD` environment variable (and corresponding `RayRedisPassword` configuration object) to allow controlling Redis password to be used during Ray initialization. This should allow connecting to an already spawned cluster in a non-cloud mode.

- [x] passes `flake8 modin`
- [x] Resolves #3116
- [x] tests added and passing

### DIFF
--- a/modin/config/envvars.py
+++ b/modin/config/envvars.py
@@ -16,6 +16,7 @@ import sys
 from textwrap import dedent
 import warnings
 from packaging import version
+import secrets
 
 from .pubsub import Parameter, _TYPE_PARAMS, ExactStr, ValueSource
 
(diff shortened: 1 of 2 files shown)


YarShev (approved):
@vnlitvinov , thanks, LGTM!

devin-petersohn (approved):
Thanks @vnlitvinov, LGTM!

PR #3128, opened 2021-06-04
FIX-#3083: Workaround for Ray import thread race condition

### LINKED ISSUE #3083: [Ray] Work around for Ray's race condition with import thread
We already do this when we initialize a Ray environment, but we need to handle it when ray is already initialized. This line of code needs to be moved to resolve it. 

https://github.com/modin-project/modin/blob/master/modin/engines/ray/utils.py#L184

The issue ray-project/ray#7879 tracks the progress, but after some discussions with the Ray team it seems unlikely that it will be fixed soon due to a lack of consistent reproducer.

devin-petersohn:
Resolves #3083

## What do these changes do?

- [x] passes `flake8 modin`
- [x] Resolves #3083 
- [ ] tests added and passing

### DIFF
--- a/modin/engines/ray/utils.py
+++ b/modin/engines/ray/utils.py
@@ -220,9 +220,6 @@ def initialize_ray(
         ray.worker.global_worker.run_function_on_all_workers(
             _move_stdlib_ahead_of_site_packages
         )
-
-        ray.worker.global_worker.run_function_on_all_workers(_import_pandas)
-
         if Backend.get() == "Cudf":
             from modin.engines.ray.cudf_on_ray.frame.gpu_manager import GPUManager
             from modin.engines.ray.cudf_on_ray.frame.partition_manager import (
@@ -233,7 +230,7 @@ def initialize_ray(
             if not GPU_MANAGERS:
                 for i in range(GpuCount.get()):
                     GPU_MANAGERS.append(GPUManager.remote(i))
-
+    ray.worker.global_worker.run_function_on_all_workers(_import_pandas)
     num_cpus = int(ray.cluster_resources()["CPU"])
     num_gpus = int(ray.cluster_resources().get("GPU", 0))
     if Backend.get() == "Cudf":


vnlitvinov (approved):
LGTM, but let's make sure CI passes first.

modin/engines/ray/utils.py
```diff
@@ -233,7 +230,7 @@ def initialize_ray(
             if not GPU_MANAGERS:
                 for i in range(GpuCount.get()):
                     GPU_MANAGERS.append(GPUManager.remote(i))
-
+    ray.worker.global_worker.run_function_on_all_workers(_import_pandas)
```
AmeerHajAli, line 233: why are you using this private API?

devin-petersohn: This fix was originally contributed by @simon-mo to solve a race condition with Ray importing pandas. This is the only solution that reliably fixes it.

YarShev: Shouldn't we also move the lines 219 and 220?
`_move_stdlib_ahead_of_site_packages()`
`ray.worker.global_worker.run_function_on_all_workers(_move_stdlib_ahead_of_site_packages_`

devin-petersohn: @YarShev It is not related to the race condition. Those lines change the python path so that the local installation of Typing does not interfere. I have not seen recent cases of this, and would prefer to not propagate this change unless absolutely necessary.

YarShev: @devin-petersohn , yes, it is not related to the race condition but I am kind of confused why we call the lines I specified under `if not ray.is_initialized() or override_is_cluster:` construction only. I think we should probably move those outside of `if` structure as part of separate issue even though there hasn't been an issue related to that so far. What do you think?

YarShev (approved):
@devin-petersohn , LGTM! Please, answer this question ([link](https://github.com/modin-project/modin/pull/3128#discussion_r646534537)) not to lose it.

modin/engines/ray/utils.py
devin-petersohn: Ok @YarShev, I will open an issue to track it.

PR #3132, opened 2021-06-07
FIX-#3130: Update Ray version to 1.4

### LINKED ISSUE #3130: Ray version support for >=1.4

devin-petersohn:

## What do these changes do?

- [x] passes `flake8 modin`
- [x] Resolves #3130 
- [ ] tests added and passing

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -427,12 +427,9 @@ jobs:
           conda list
       - name: Install HDF5
         run: sudo apt update && sudo apt install -y libhdf5-dev
-      - run: ray install-nightly
       - run: MODIN_BENCHMARK_MODE=True pytest modin/pandas/test/internals/test_benchmark_mode.py
       - run: pytest modin/experimental/xgboost/test/test_default.py
-      - run: pytest modin/pandas/test/dataframe/test_binary.py -v
-      - run: cat /tmp/ray/session_latest/logs/*
-        if: ${{failure()}}
+      - run: pytest modin/pandas/test/dataframe/test_binary.py
       - run: pytest modin/pandas/test/dataframe/test_default.py
       - run: pytest modin/pandas/test/dataframe/test_indexing.py
       - run: pytest modin/pandas/test/dataframe/test_iter.py
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -234,9 +234,9 @@ jobs:
       - shell: bash -l {0}
         run: python -m pytest modin/test/test_envvar_npartitions.py
       - shell: bash -l {0}
-        run: python -m pytest -n 2 modin/test/test_partition_api.py
-      - shell: bash -l {0}
         run: python -m pytest modin/test/test_utils.py
+      - shell: bash -l {0}
+        run: python -m pytest modin/test/test_partition_api.py
 
   test-defaults:
     needs: [lint-commit, lint-flake8, lint-black, test-api, test-headers]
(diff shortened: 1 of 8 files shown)


.github/workflows/ci.yml
```diff
@@ -270,23 +270,23 @@ jobs:
...
-      - run: pytest -n 2 modin/pandas/test/dataframe/test_reduction.py --backend=${{ matrix.backend }}
-      - run: pytest -n 2 modin/pandas/test/dataframe/test_udf.py --backend=${{ matrix.backend }}
-      - run: pytest -n 2 modin/pandas/test/dataframe/test_window.py --backend=${{ matrix.backend }}
-      - run: pytest -n 2 modin/pandas/test/dataframe/test_pickle.py --backend=${{ matrix.backend }}
-      - run: python -m pytest -n 2 modin/pandas/test/test_series.py --backend=${{ matrix.backend }}
-      - run: python -m pytest -n 2 modin/pandas/test/test_rolling.py --backend=${{ matrix.backend }}
-      - run: python -m pytest -n 2 modin/pandas/test/test_concat.py --backend=${{ matrix.backend }}
-      - run: python -m pytest -n 2 modin/pandas/test/test_groupby.py --backend=${{ matrix.backend }}
-      - run: python -m pytest -n 2 modin/pandas/test/test_reshape.py --backend=${{ matrix.backend }}
-      - run: python -m pytest -n 2 modin/pandas/test/test_general.py --backend=${{ matrix.backend }}
```
vnlitvinov, line 289: so we're forced to run in single-process mode again?..

.github/workflows/ci.yml
```diff
@@ -406,6 +406,8 @@ jobs:
     env:
       MODIN_ENGINE: ${{matrix.engine}}
       MODIN_MEMORY: 1000000000
+      RAY_LOG_LEVEL: DEBUG
```
vnlitvinov, line 409: why?

modin/engines/ray/pandas_on_ray/frame/partition.py
```diff
@@ -17,11 +17,9 @@
 
 from modin.data_management.utils import length_fn_pandas, width_fn_pandas
 from modin.engines.base.frame.partition import PandasFramePartition
-from modin.engines.ray.utils import handle_ray_task_error
```
vnlitvinov, line 20: why we no longer need that?

modin/engines/ray/utils.py
```diff
@@ -164,30 +133,19 @@ def initialize_ray(
...
-                    plasma_directory = gettempdir()
                 # We may have already set the memory from the environment variable, we don't
                 # want to overwrite that value if we have.
                 if object_store_memory is None:
                     # Round down to the nearest Gigabyte.
-                    try:
-                        system_memory = ray._private.utils.get_system_memory()
-                    except AttributeError:  # Compatibility with Ray <= 1.2
-                        system_memory = ray.utils.get_system_memory()
+                    system_memory = ray._private.utils.get_system_memory()
```
vnlitvinov, line 141: isn't there a non-private API to use? this looks unsafe

modin/engines/ray/utils.py
```diff
@@ -164,30 +133,19 @@ def initialize_ray(
 """,
             )
             object_store_memory = Memory.get()
-            plasma_directory = RayPlasmaDir.get()
```
vnlitvinov, line 167: why do we no longer need a configuration for plasma directory?..

also if we don't need it then we should probably remove `RayPlasmaDir` configuration variable

.github/workflows/ci.yml
devin-petersohn: I will try to put it back.

.github/workflows/ci.yml
devin-petersohn: We used this to help debug Ray errors in CI, I will revert.

modin/engines/ray/pandas_on_ray/frame/partition.py
devin-petersohn: Ray throws the correct error Type/Message now instead of wrapping in their own.

modin/engines/ray/utils.py
devin-petersohn: `_plasma_directory` is not public anymore. Setting `_plasma_directory` automatically disables default object spilling (out of core).

If users want a custom setup for Ray, they should initialize themselves. The defaults here should make sense for most users and if someone wants to tweak things they can initialize Ray themselves. I am in favor of simplifying the configurations and initialization in the interest of simplifying the code. If the users need to set 3 or 4 configs they might as well just initialize Ray themselves.

modin/engines/ray/utils.py
devin-petersohn: This is carried over from how we used it before. Should I update as a part of this PR?

modin/engines/ray/pandas_on_ray/frame/partition.py
devin-petersohn: If you look at how we were handling this before, you would cringe. It was a filthy hack.

https://github.com/modin-project/modin/pull/3132/files#diff-72036370d85971ebd29660bf18d7dc4eb75e56609955d5291b23750e2bf1c28aL34

vnlitvinov: I remember that function (even without looking it up), ugh!, and I'm glad it's now gone 😄 was just double-checking.

modin/engines/ray/utils.py
vnlitvinov: I think that if there's stable API available in Ray to get this (not a private one), and it's small enough, then it should be acceptable to call it a part of "migration to ray 1.4" 😉

modin/engines/ray/utils.py
vnlitvinov: Well, less knobs for us (and users) is always better if this doesn't hurt.

In this case we should probably get rid of `RayPlasmaDir` configuration variable in `modin.config` altogether along with removing it here.

devin-petersohn: Done!

modin/engines/ray/utils.py
```diff
@@ -148,8 +119,6 @@ def initialize_ray(
                 address=redis_address or "auto",
                 include_dashboard=False,
                 ignore_reinit_error=True,
-                _redis_password=redis_password,
```
vnlitvinov, line 151: wait, how do we pass redis password for Ray if we're connecting to an already existing cluster?..

modin/engines/ray/utils.py
devin-petersohn: There is not such a stable API that I can find.

modin/engines/ray/utils.py
devin-petersohn: It should be added back

modin/engines/ray/utils.py
vnlitvinov: so be it for now then...

vnlitvinov (approved):
Code should be good assuming CI passes.

Small note for now is we won't be able to install from conda-forge for the time being as they don't have fresh `ray` yet (cc @aregm)

modin/engines/ray/utils.py
YarShev: Since as of Ray 1.3+ object spilling is a default option in Ray shouldn't we update our documentation related to out of core feature?

devin-petersohn: Yes, and we should probably disable the config/flag.

gshimansky: When user initializes Ray on their own, they may miss important settings that Modin does for Ray initialization. For example, if Ray is initialized before Modin, then Ray doesn't get `OMP_NUM_THREADS=1` and all workers spawn as many threads as there are CPU cores leading to oversubscription and sometimes process gets killed for using too many resources.
Maybe we should at least document it somewhere that initializing Ray before Modin requires special handling.

devin-petersohn: Ray also sets `OMP_NUM_THREADS=1` by default. I think Modin started doing it before Ray because the contention was noticeable in Modin very early on.

https://docs.ray.io/en/master/serve/core-apis.html?highlight=omp_num_threads#configuring-parallelism-with-omp-num-threads

Are there other best practices we should include in documentation @gshimansky?

docs/out_of_core.rst
```diff
@@ -1,46 +1,28 @@
...
-.. code-block:: bash
-
-  export MODIN_MEMORY=200000000000 # Set the number of bytes to 200GB
+.. code-block:: python
 
+  import modin.pandas as pd
+  import ray
 
-**The default for Modin is 8x the memory on the machine.**
+  ray.init(_plasma_directory="/tmp")  # setting to disable out of core in Ray
```
YarShev, line 24: Is there a way to disable object spilling in Dask?

devin-petersohn: It seems you have to set a config. 

https://distributed.dask.org/en/latest/worker.html#memory-management

YarShev: Since it is said above `To disable it, start your
preferred compute engine with the appropriate arguments` we should probably provide an example both for Ray and Dask or define separate a section for Ray and Dask where we can show an example how to disable object spilling in Ray and put the link to configure object spilling in Dask if no other ways to do it.

devin-petersohn: Ok added a link there.

YarShev (approved):
@devin-petersohn , thanks, LGTM! Let's wait for other reviewers. May be they have some comments.

modin/engines/ray/utils.py
gshimansky: I wasn't aware that Ray sets OMP_NUM_THREADS now. It is something new to me. I don't have any other objections to ask user to initialize Ray manually.

PR #3134, opened 2021-06-08
Fix case when Index column named 'Unnamed:' after reading DataFrame via read_csv

### LINKED ISSUE #3119: Index column named 'Unnamed: 0' on `.to_csv()` after reading DataFrame with `.read_csv()`
### System information
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: Ubuntu 20.10
- **Modin version** (`modin.__version__`): 0.9.1
- **Python version**: 3.7.4
```
import numpy as np
import pandas as pd
import os
import modin.pandas as pd2

df1 = pd.DataFrame(np.arange(12).reshape(4, 3), columns=list('abc'),
                   index=[f'i{i}' for i in range(1, 5)])
df2 = pd2.DataFrame(np.arange(12).reshape(4, 3), columns=list('abc'),
                    index=[f'i{i}' for i in range(1, 5)])

df1.to_csv('df1.csv')
df2.to_csv('df2.csv')

df1_re = pd.read_csv('df1.csv', index_col=0)
df2_re = pd2.read_csv('df2.csv', index_col=0)

df1_re.to_csv('df1_re.csv')
df2_re.to_csv('df2_re.csv')
```
This gives 
```
$ head *re*csv
==> df1_re.csv <==
,a,b,c
i1,0,1,2
i2,3,4,5
i3,6,7,8
i4,9,10,11

==> df2_re.csv <==
Unnamed: 0,a,b,c
i1,0,1,2
i2,3,4,5
i3,6,7,8
i4,9,10,11
```

### Describe the problem
After reading a CSV and writing it again the resulting file has `Unnamed: 0` as header of the index column, whereas a file that has been written with pandas does not (compare `df1_re.csv` and `df2_re.csv` after running the example). This only occurs with Ray, while Dask gives ``UserWarning: `to_csv` defaulting to pandas implementation.`` and writes the file as expected.

anmyachev:
## What do these changes do?

- [x] passes `flake8 modin`
- [x] Resolves #3119 
- [x] tests added and passing


vnlitvinov:
@anmyachev please make a descriptive PR title

modin/backends/pandas/parsers.py
```diff
@@ -314,6 +314,8 @@ def parse(fname, **kwargs):
             to_read = header + bio.read(end - start)
             bio.close()
             pandas_df = pandas.read_fwf(BytesIO(to_read), **kwargs)
+            if pandas_df.index.name == "Unnamed: 0":
```
vnlitvinov, line 317: what if the index name is `Unnamed: 15` or something similar?..

also why're you changing it in `read_fwf` instead of `read_csv` (and not in generic `_read` or smth - I doubt it's specific to only one text format)

amyskov:
Hi @anmyachev, @vnlitvinov, this PR is ready for review.

`read_csv` implementation was slightly changed - now additional metadata like `names` and `index_col` data is passed to the workers only if it is unavoidable (`pass_names` case), in other cases original parameters and header data will be used by workers. Since there is no need in definition of `names` and `index_col`, code with these values sniffing was removed that slightly improved `read_csv` performance:
```
     [b58663bd]       [0b381118]
     <fix-read-csv~3>       <fix-read-csv>
         179±10ms          181±7ms     1.01  io.csv.TimeReadCsvNamesDtype.time_read_csv_names_dtype((1000000, 10), 'array-like', 'Int64')
         361±30ms         385±20ms     1.06  io.csv.TimeReadCsvNamesDtype.time_read_csv_names_dtype((1000000, 10), 'array-like', 'Int64_Timestamp')
       3.14±0.04s       1.72±0.03s     0.55  io.csv.TimeReadCsvNamesDtype.time_read_csv_names_dtype((5000, 5000), 'array-like', 'Int64')
       3.11±0.08s       1.74±0.02s     0.56  io.csv.TimeReadCsvNamesDtype.time_read_csv_names_dtype((5000, 5000), 'array-like', 'Int64_Timestamp')
         718±30ms         703±20ms     0.98  io.csv.TimeReadCsvSkiprows.time_skiprows((1000000, 10), 'lambda_even_rows')
         755±20ms         740±10ms     0.98  io.csv.TimeReadCsvSkiprows.time_skiprows((1000000, 10), 'range_step2')
          726±5ms          737±7ms     1.02  io.csv.TimeReadCsvSkiprows.time_skiprows((1000000, 10), 'range_uniform')
          214±8ms         201±20ms     0.94  io.csv.TimeReadCsvSkiprows.time_skiprows((1000000, 10), None)
       2.57±0.01s       2.53±0.07s     0.98  io.csv.TimeReadCsvSkiprows.time_skiprows((5000, 5000), 'lambda_even_rows')
       2.57±0.05s       2.60±0.02s     1.01  io.csv.TimeReadCsvSkiprows.time_skiprows((5000, 5000), 'range_step2')
       3.77±0.03s       3.88±0.09s     1.03  io.csv.TimeReadCsvSkiprows.time_skiprows((5000, 5000), 'range_uniform')
       1.97±0.02s       1.57±0.02s     0.79  io.csv.TimeReadCsvSkiprows.time_skiprows((5000, 5000), None)
          119±6ms          117±6ms     0.98  io.csv.TimeReadCsvTrueFalseValues.time_true_false_values((1000000, 10))
       1.83±0.03s       1.43±0.03s     0.78  io.csv.TimeReadCsvTrueFalseValues.time_true_false_values((5000, 5000))
```

Also, looks like, some tests were fixed by this PR, so they were enabled.

modin/backends/pandas/parsers.py
```diff
@@ -210,28 +210,34 @@ def parse(fname, **kwargs):
...
                 fname, "rb", kwargs.pop("compression", "infer")
             )
-            if kwargs.get("encoding", None) is not None:
-                header = b"" + bio.readline()
-            else:
-                header = b""
+            header = b""
+            # In this case we beware that fisrt line can contain BOM, so
+            # adding this line to the `header` for reading and then skip it
+            if kwargs.get("encoding", None) is not None and header_size == 0:
```
vnlitvinov, line 222: could you please explain why do we check for `header_size==0` here?

modin/engines/base/io/text/csv_dispatcher.py
```diff
@@ -220,13 +202,16 @@ def _read_csv_check_support(
         if skiprows is not None and not isinstance(skiprows, int):
             return False
 
+        # can't pickle dialect object to the worker process
+        if read_csv_kwargs.get("dialect", None) is not None:
+            return False
```
vnlitvinov, line 207: are you sure this is relevant to the issue you're stating as being closed by this PR?

modin/engines/base/io/text/csv_dispatcher.py
```diff
@@ -249,11 +232,12 @@ def _define_index(
         row_lengths : list
             Partitions rows lengths.
         """
-        if index_col is None:
-            row_lengths = cls.materialize(index_ids)
-            new_index = pandas.RangeIndex(sum(row_lengths))
+        index_objs = cls.materialize(index_ids)
+        simple_idx = isinstance(index_objs[0], int) if len(index_objs) > 0 else True
+        if simple_idx:
```
vnlitvinov, line 237:
```suggestion
        if len(index_objs) == 0 or isinstance(index_objs[0], int):
```

modin/pandas/test/test_io.py
```diff
@@ -278,6 +275,8 @@ def test_read_csv_col_handling(
             pytest.xfail(
                 reason="The reason of tests fail in `cloud` mode is unknown for now - issue #2340"
             )
+        if names is lib.no_default:
```
vnlitvinov, line 278: shouldn't this be checked for omnisci only or something?

modin/engines/base/io/text/csv_dispatcher.py
```diff
@@ -249,11 +228,11 @@ def _define_index(
         row_lengths : list
             Partitions rows lengths.
         """
-        if index_col is None:
-            row_lengths = cls.materialize(index_ids)
-            new_index = pandas.RangeIndex(sum(row_lengths))
+        index_objs = cls.materialize(index_ids)
+        if len(index_objs) == 0 or isinstance(index_objs[0], int):
```
anmyachev, line 232: Is condition `if len(index_objs) == 0 or isinstance(index_objs[0], int):` exactly equivalent to condition `if index_col is None:`?

modin/backends/pandas/parsers.py
amyskov: In this specific case header row with BOM (that we need for correct encoding) won't be read, in other cases (if `header_size>0`) BOM will be added to to the data for reading anyway.

modin/engines/base/io/text/csv_dispatcher.py
amyskov: Not sure) change was reverted.

modin/pandas/test/test_io.py
amyskov: Not sure, tests with non-default `names` still fail with pandas-on-Ray/Dask backends.

`names` `None` parameter was changed to `lib.no_default` to avoid exceptions `ValueError: Specified named and prefix; you can only specify one.` (that were raised both by pandas and Modin that leads to tests passing by just checking exception types) and it's occurred that only remained failed tests are tests with `names=lib.no_default`, so they were skipped.
Honestly, this change not directly related to PR issue, but enabled tests are checking PR changes, so i prefer leave these changes in this PR.

modin/engines/base/io/text/csv_dispatcher.py
amyskov: I hope so, i can't imagine scenario where these conditions are not equivalent.
In the parsers.py if `index_col is None` we will obtain range index as a result of `pandas.read_csv` that will be replaced with integer.

anmyachev: @amyskov I am thinking that there may be new cases in which numeric Series is replaced by range Series.

amyskov: Can't catch your idea, please provide some example when we use `read_csv` with `index_col=None` and obtain non-range index...
This line just checks if workers returned integers for index (this occurs only if resulting indexes on the workers are range typed), and if so to replace it with RangeIndex.

anmyachev: It turns out that I did not pay attention to that `index_objs` can contain several objects of the type Series. No more questions.

vnlitvinov (approved):
LGTM

Issue #2688, opened 2021-02-04
Modin fails to load csv from s3 with ray client


Bhavya6187:
### System information
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: Ubuntu
- **Modin version** (`modin.__version__`): master (0.8.3+22.ge99b629)
- **Python version**: 3.7
- **Code we can use to reproduce**:

```
import ray
import os
import ray.util
ray.util.connect("<service_ip>:50051")
import modin.pandas as pd
pd.DEFAULT_NPARTITIONS = 10
df = pd.read_csv("s3://<bucket>/HIGGS_100k.csv")

```
### Describe the problem
Modin fails to load csv from s3 with ray client and throws an error.

### Source code / logs
```
---------------------------------------------------------------------------
AssertionError                            Traceback (most recent call last)
<ipython-input-10-9b3c648a226d> in <module>()
----> 1 df = pd.read_csv("s3://<s3_bucket>/HIGGS_100k.csv")

/home/bhavya.agarwal/.local/lib/python3.7/site-packages/modin/pandas/io.py in parser_func(filepath_or_buffer, sep, delimiter, header, names, index_col, usecols, squeeze, prefix, mangle_dupe_cols, dtype, engine, converters, true_values, false_values, skipinitialspace, skiprows, nrows, na_values, keep_default_na, na_filter, verbose, skip_blank_lines, parse_dates, infer_datetime_format, keep_date_col, date_parser, dayfirst, cache_dates, iterator, chunksize, compression, thousands, decimal, lineterminator, quotechar, quoting, escapechar, comment, encoding, dialect, error_bad_lines, warn_bad_lines, skipfooter, doublequote, delim_whitespace, low_memory, memory_map, float_precision)
    114 
    115         kwargs = {k: v for k, v in f_locals.items() if k in _pd_read_csv_signature}
--> 116         return _read(**kwargs)
    117 
    118     return parser_func

/home/bhavya.agarwal/.local/lib/python3.7/site-packages/modin/pandas/io.py in _read(**kwargs)
    133 
    134     Engine.subscribe(_update_engine)
--> 135     pd_obj = EngineDispatcher.read_csv(**kwargs)
    136     # This happens when `read_csv` returns a TextFileReader object for iterating through
    137     if isinstance(pd_obj, pandas.io.parsers.TextFileReader):

[...]

PR #3140, opened 2021-06-09
TEST-#3085: Add testing for Ray Client

### LINKED ISSUE #3085: [Ray] Add unit tests for ray client + modin

Wanted to open discussion about adding unit tests using ray client to this repo's CI. If there are any examples that should be covered that would be handy to know, as well as any guidelines for where to put these tests. Example issue with ray-client on the ray repo https://github.com/ray-project/ray/issues/14857.

@amogkam @devin-petersohn

ckw017:
Signed-off-by: Chris Wong <chriskw.xyz@gmail.com>

## What do these changes do?

On code side there's a new environment variable that can be set to start a ray client connection before the test session starts. 

For push/ci, currently installing a wheel slightly ahead of Ray 1.4 to fix the import deserialization race condition, which seemed to crop up extra frequently when using ray client. Since client is a bit finicky with multiple connections at the moment the `-n 2` flag with pytest doesn't work, so I separated out the tests in the matrix so that they wouldn't bottleneck the completion time (similar to what's being done for test windows). I've also left out the following tests which are failing due to reasons unrelated to ray client (primarily seems like there's been a change with how errors are propagated since 1.1 which the test suite doesn't yet handle)
- `modin/pandas/test/dataframe/test_binary.py`
- `modin/pandas/test/dataframe/test_map_metadata.py`
- `modin/pandas/test/dataframe/test_reduction.py`
- `modin/pandas/test/dataframe/test_udf.py`
- `modin/pandas/test/test_series.py`
- `modin/experimental/pandas/test/test_io_exp.py`

- [x] passes `flake8 modin`
- [x] Resolves #3085 
- [x] tests added and passing

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -665,19 +665,17 @@ jobs:
       matrix:
         python-version: ["3.7", "3.8"]
         test-task:
-          # The following are broken on Ray 1.3+ (Unrelated to ray client)
-          # - modin/pandas/test/dataframe/test_binary.py
-          # - modin/pandas/test/dataframe/test_map_metadata.py
-          # - modin/pandas/test/dataframe/test_reduction.py
-          # - modin/pandas/test/dataframe/test_udf.py
-          # - modin/pandas/test/test_series.py
-          # - modin/experimental/pandas/test/test_io_exp.py
+          - modin/pandas/test/dataframe/test_binary.py
           - modin/pandas/test/dataframe/test_default.py
           - modin/pandas/test/dataframe/test_indexing.py
           - modin/pandas/test/dataframe/test_iter.py
           - modin/pandas/test/dataframe/test_join_sort.py
+          - modin/pandas/test/dataframe/test_map_metadata.py
+          - modin/pandas/test/dataframe/test_reduction.py
+          - modin/pandas/test/dataframe/test_udf.py
           - modin/pandas/test/dataframe/test_window.py
           - modin/pandas/test/dataframe/test_pickle.py
+          - modin/pandas/test/test_series.py
           - modin/pandas/test/test_rolling.py
           - modin/pandas/test/test_concat.py
           - modin/pandas/test/test_groupby.py
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -654,3 +654,68 @@ jobs:
           conda list
       - shell: bash -l {0}
         run: python -m pytest modin/spreadsheet/test/test_general.py
+
+  test-ray-client:
+    needs: [lint-commit, lint-flake8, lint-black, test-api, test-headers]
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        shell: bash -l {0}
+    strategy:
+      matrix:
+        python-version: ["3.7", "3.8"]
+        test-task:
+          # The following are broken on Ray 1.3+ (Unrelated to ray client)
+          # - modin/pandas/test/dataframe/test_binary.py
+          # - modin/pandas/test/dataframe/test_map_metadata.py
+          # - modin/pandas/test/dataframe/test_reduction.py
+          # - modin/pandas/test/dataframe/test_udf.py
+          # - modin/pandas/test/test_series.py
+          # - modin/experimental/pandas/test/test_io_exp.py
+          - modin/pandas/test/dataframe/test_default.py
+          - modin/pandas/test/dataframe/test_indexing.py
+          - modin/pandas/test/dataframe/test_iter.py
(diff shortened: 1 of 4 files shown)


ckw017:
Some alternate ways these tests can be added in are:
- Stick them into push-to-master, where the other nightly tests are
- Instead of spinning up a new machine for each test, the main bottleneck is test_join_sort, so alternatively we can just run that as a separate job and run the remaining tests separately

devin-petersohn (commented):
@ckw017 Thanks for the PR!

We are updating Ray support to >= 1.4 in #3132. That should allow us to enable the commented out tests.

AmeerHajAli:
@devin-petersohn , what is the ETA for getting this merged?

devin-petersohn:
@ckw017 Can you rebase this on current master? Ray 1.4 support is merged.

ckw017:
@devin-petersohn Rebased with master and uncommented the tests that were formerly broken on before the 0.10 updates. LMK if you want anything added or changed!

modin/conftest.py
```diff
@@ -452,3 +452,32 @@ def TestReadGlobCSVFixture():
...
     return lambda: _generated_doc_urls
+
+
+def pytest_sessionstart(session):
+    if TestRayClient.get():
+        import ray
+        import subprocess
+
+        port = "50051"
+        # Clean up any extra processes from previous runs
```
vnlitvinov, line 463: I would say it's a bad idea, who knows which processes are running there from some parallel session (potentially by some other user)?

vnlitvinov (commented):
Do we really want to test this in "normal" CI? Maybe testing on "push to master" would be enough?

I mean how different it should be from Modin's perspective to run on "normal" Ray vs running on "Ray client"?

modin/conftest.py
ckw017: Yeah that's reasonably, I'll get rid of those.

ckw017:
> Do we really want to test this in "normal" CI? Maybe testing on "push to master" would be enough?
> 
> I mean how different it should be from Modin's perspective to run on "normal" Ray vs running on "Ray client"?

The main issues we want to catch were things like #2688, which could be caught pretty quickly with just a few tests. Rather than running through all the tests covered in test-all we can switch to just a subset and only run on one version of Python, if you want to cut down on the clutter in the CI.

.github/workflows/ci.yml
```diff
@@ -654,3 +654,66 @@ jobs:
...
+      - uses: conda-incubator/setup-miniconda@v2
+        with:
+          activate-environment: modin
+          python-version: ${{matrix.python-version}}
+          channel-priority: strict
+          use-only-tar-bz2: true # IMPORTANT: This needs to be set for caching to work properly!
+      - run: pip install -r requirements-dev.txt
+      # Use a ray master commit that includes the fix here: https://github.com/ray-project/ray/pull/16278
+      # Can be changed after a Ray version > 1.4 is released.
+      - run: pip install https://s3-us-west-2.amazonaws.com/ray-wheels/master/c8e3ed9eec30119092ef966ee7b8982c8954c333/ray-2.0.0.dev0-cp37-cp37m-manylinux2014_x86_64.whl
```
devin-petersohn, line 703: Can this (and below) be removed? Or do we need to keep testing on this commit specifically? I think we want Ray 1.4 here.

modin/config/envvars.py
```diff
@@ -239,6 +239,16 @@ class TestDatasetSize(EnvironmentVariable, type=str):
     choices = ("Small", "Normal", "Big")
 
 
+class TestRayClient(EnvironmentVariable, type=bool):
```
devin-petersohn, line 242: As a side note here, we should probably have configs specific for tests elsewhere (not needed to change here).

devin-petersohn (commented):
@vnlitvinov I am okay with adding these on a per-commit basis. The tests don't take that long.

@ckw017 Left a comment, thanks!

.github/workflows/ci.yml
ckw017: We'll probably want to keep that in for now, since it prevents the: `pandas has no attribute 'compat'` issue altogether, which seemed to crop up frequently on Github Actions machines when I was testing this PR out. We can push for that to be in 1.4.1 down the line so that we don't have to keep testing against a specific commit.

ckw017:
@devin-petersohn @vnlitvinov I've opened #3147 which has the same changes as this PR, but only runs on push-to-master. Feel free to merge in whichever one works better on your end, and close the other.

YarShev:
Closing this PR because #3085 was resolved via #3147.

PR #3021, opened 2021-04-26
DOCS-#2996: document `modin.backends.base`

dchigarev:
Signed-off-by: Dmitry Chigarev <dmitry.chigarev@intel.com>

## What do these changes do?

- [x] commit message follows format outlined [here](https://modin.readthedocs.io/en/latest/contributing.html)
- [x] passes `flake8 modin`
- [x] passes `black --check modin`
- [x] signed commit with `git commit -s` 
- [x] Resolves #2996 
- [x] tests added and passing
- This PR adds doc-strings for all of the `BaseQueryCompiler` methods and so defines common query compilers API. PR appears to be really huge but it's required to finally define the first version of backends API which does not depend on pandas (however many function params still don't have any description and so refers to pandas documentation).
- The next step of setting up common query compilers API is to get rid of `*args` and `**kwargs` which hides actual method parameters. In this PR actual method parameters were inferred from high-level API calls signatures and placed into the doc-strings.

Built docs for this PR: https://modin--3021.org.readthedocs.build/en/3021/index.html
Affected sections:
- https://modin--3021.org.readthedocs.build/en/3021/flow/modin/backends/
- https://modin--3021.org.readthedocs.build/en/3021/flow/modin/backends/base/query_compiler.html

PR #3141, opened 2021-06-09
FIX-#2164: Fix tests for BaseOnPython: test___repr__, test_dtype_empty

### LINKED ISSUE #2164: Enable tests which are failing for BaseOnPython
1) [groupby.size](https://github.com/modin-project/modin/blob/master/modin/pandas/test/test_groupby.py#L390)
2) [test___repr__](https://github.com/modin-project/modin/blob/master/modin/pandas/test/test_series.py#L497)
3) [test_dtype_empty](https://github.com/modin-project/modin/blob/master/modin/pandas/test/test_series.py#L1544)
4) [test_copy](https://github.com/modin-project/modin/blob/master/modin/pandas/test/dataframe/test_map_metadata.py#L284)
5) [test_append](https://github.com/modin-project/modin/blob/master/modin/pandas/test/dataframe/test_map_metadata.py#L408)

Rubtsowa:

## What do these changes do?

- [ ] passes `flake8 modin`
- [ ] Resolves #2164 ? 
- [ ] tests added and passing

### DIFF
--- a/modin/engines/base/frame/data.py
+++ b/modin/engines/base/frame/data.py
@@ -2226,10 +2226,7 @@ class PandasFrame(object):
         """
         df = self._partition_mgr_cls.to_pandas(self._partitions)
         if df.empty:
-            if len(self.columns) != 0:
-                df = pandas.DataFrame(columns=self.columns)
-            else:
-                df = pandas.DataFrame(columns=self.columns, index=self.index)
+            df = pandas.DataFrame(columns=self.columns, index=self.index)
         else:
             for axis in [0, 1]:
                 ErrorMessage.catch_bugs_and_request_email(
--- a/modin/pandas/test/test_series.py
+++ b/modin/pandas/test/test_series.py
@@ -19,6 +19,7 @@ import matplotlib
 import modin.pandas as pd
 from numpy.testing import assert_array_equal
 from pandas.core.base import SpecificationError
+from modin.utils import get_current_backend
 import sys
 
 from modin.utils import to_pandas
@@ -486,16 +487,7 @@ def test___pow__(data):
 )
 @pytest.mark.parametrize(
     "data",
-    [
-        *test_data_values,
-        pytest.param(
-            "empty",
-            marks=pytest.mark.xfail_backends(
-                ["BaseOnPython"],
-                reason="Empty Series has a missmatched from Pandas dtype.",
-            ),
-        ),
-    ],
+    [*test_data_values, "empty"],
     ids=[*test_data_keys, "empty"],
 )
 def test___repr__(name, dt_index, data):
@@ -509,7 +501,13 @@ def test___repr__(name, dt_index, data):
             "1/1/2000", periods=len(pandas_series.index), freq="T"
         )
         pandas_series.index = modin_series.index = index
-    assert repr(modin_series) == repr(pandas_series)
+
+    if get_current_backend() == "BaseOnPython" and data == "empty":
+        # TODO: Remove this when default `dtype` of empty Series will be `object` in pandas (see #3142).
+        assert modin_series.dtype == np.object
+        assert pandas_series.dtype == np.float64
+    else:
+        assert repr(modin_series) == repr(pandas_series)
 
 
 @pytest.mark.parametrize("data", test_data_values, ids=test_data_keys)
@@ -1533,12 +1531,14 @@ def test_dropna_inplace(data):
     df_equals(modin_series, pandas_series)
 
 
-@pytest.mark.xfail_backends(
-    ["BaseOnPython"], reason="Empty Series has a missmatched from Pandas dtype."
-)
 def test_dtype_empty():
     modin_series, pandas_series = pd.Series(), pandas.Series()
-    assert modin_series.dtype == pandas_series.dtype
+    if get_current_backend() == "BaseOnPython":
+        # TODO: Remove this when default `dtype` of empty Series will be `object` in pandas (see #3142).
+        assert modin_series.dtype == np.object
+        assert pandas_series.dtype == np.float64
+    else:
+        assert modin_series.dtype == pandas_series.dtype
 
 
 @pytest.mark.parametrize("data", test_data_values, ids=test_data_keys)


modin/pandas/test/test_series.py
```diff
@@ -1533,12 +1537,17 @@ def test_dropna_inplace(data):
...
-    assert modin_series.dtype == pandas_series.dtype
+    try:
+        assert modin_series.dtype == pandas_series.dtype
+    except Exception as e:
+        # TODO delete this except when future version Pandas will be released (Issue #3142)
+        if get_current_backend() == "BaseOnPython":
+            with pytest.raises(AssertionError):
+                assert modin_series.dtype == pandas_series.dtype
+        else:
+            raise e
```
dchigarev, line 1550: I would highlight the difference between modin and pandas dtype:
```suggestion
    except AssertionError as e:
        # TODO delete this except when future version Pandas will be released (Issue #3142)
        if get_current_backend() == "BaseOnPython":
            assert modin_series.dtype == np.object
            assert pandas_series.dtype == np.float64
        else:
            raise e
```

modin/pandas/test/test_series.py
```diff
@@ -488,13 +489,7 @@ def test___pow__(data):
...
     [
         *test_data_values,
-        pytest.param(
-            "empty",
-            marks=pytest.mark.xfail_backends(
-                ["BaseOnPython"],
-                reason="Empty Series has a missmatched from Pandas dtype.",
-            ),
-        ),
+        pytest.param("empty"),
```
prutskov, line 492:
```suggestion
        "empty",
```

modin/pandas/test/test_series.py
```diff
@@ -509,7 +504,16 @@ def test___repr__(name, dt_index, data):
             "1/1/2000", periods=len(pandas_series.index), freq="T"
         )
         pandas_series.index = modin_series.index = index
-    assert repr(modin_series) == repr(pandas_series)
+
+    try:
+        assert repr(modin_series) == repr(pandas_series)
+    except Exception as e:
+        # TODO delete this except when future version Pandas will be released (Issue #3142)
```
prutskov, line 511:
```suggestion
        # TODO: Remove this try-except when default  `dtype` of empty Series will be `object` in pandas (see #3142).
```

modin/pandas/test/test_series.py
```diff
@@ -1533,12 +1537,17 @@ def test_dropna_inplace(data):
...
-@pytest.mark.xfail_backends(
-    ["BaseOnPython"], reason="Empty Series has a missmatched from Pandas dtype."
-)
 def test_dtype_empty():
     modin_series, pandas_series = pd.Series(), pandas.Series()
-    assert modin_series.dtype == pandas_series.dtype
+    try:
+        assert modin_series.dtype == pandas_series.dtype
+    except Exception as e:
+        # TODO delete this except when future version Pandas will be released (Issue #3142)
```
prutskov, line 1545:
```suggestion
        # TODO: Remove this try-except when default  `dtype` of empty Series will be `object` in pandas (see #3142).
```

modin/backends/base/query_compiler.py
```diff
@@ -33,11 +32,10 @@
...
 def _get_axis(axis):
-    def axis_getter(self):
-        ErrorMessage.default_to_pandas(f"DataFrame.get_axis({axis})")
-        return self.to_pandas().axes[axis]
-
-    return axis_getter
+    if axis == 0:
+        return lambda self: self._modin_frame.index
+    else:
+        return lambda self: self._modin_frame.columns
```
prutskov, line 38: The `_get_axis`function with the same implementation is already placed in modin/backedns/pandas/query_compiler.py. Could you, please, exclude duplicating?

modin/backends/pandas/query_compiler.py
```diff
@@ -53,6 +53,8 @@
     groupby_reduce_functions,
 )
 
+from modin.backends.base.query_compiler import _get_axis as _get_axis_
+
 
 def _get_axis(axis):
```
prutskov, line 59: This could be removed, because it duplicates `_get_axis` from import

modin/backends/pandas/query_compiler.py
```diff
@@ -53,6 +53,8 @@
     groupby_reduce_functions,
 )
 
+from modin.backends.base.query_compiler import _get_axis as _get_axis_
```
prutskov, line 56: Propose to move import of `_get_axis` to line
```python
from modin.backends.base.query_compiler import BaseQueryCompiler, _get_axis
```

modin/backends/base/query_compiler.py
```diff
@@ -33,11 +32,10 @@
 
 
 def _get_axis(axis):
-    def axis_getter(self):
-        ErrorMessage.default_to_pandas(f"DataFrame.get_axis({axis})")
-        return self.to_pandas().axes[axis]
-
-    return axis_getter
+    if axis == 0:
```
prutskov, line 35: Add docstring, please.

dchigarev: #3021 adds docstring for this function, I think we shouldn't create conflicts

prutskov: Yes, I agree.

modin/backends/pandas/query_compiler.py
dchigarev: I think we can get rid of `_get_axis` import at all if we would inherit the `index` property from the `BaseQueryCompiler` (since it doesn't defaulting to pandas anymore). `PandasQueryCompiler` overwrites `index` prop at [this line](https://github.com/modin-project/modin/blob/cf259f1cd38c70e246fd5e3ea904b675a5d10bd9/modin/backends/pandas/query_compiler.py#L266), this one should be removed.

dchigarev: I'm wrong, we can't do that because the `index` property also contains `_set_axis(0)` as a setter, which is different in base and pandas query compilers

dchigarev:
[OmniSci query compiler](https://github.com/modin-project/modin/blob/cf259f1cd38c70e246fd5e3ea904b675a5d10bd9/modin/experimental/backends/omnisci/query_compiler.py) also [imports](https://github.com/modin-project/modin/blob/cf259f1cd38c70e246fd5e3ea904b675a5d10bd9/modin/experimental/backends/omnisci/query_compiler.py#L17) base `_get_axis` and uses it in its own axis-getter functions [[1]](https://github.com/modin-project/modin/blob/cf259f1cd38c70e246fd5e3ea904b675a5d10bd9/modin/experimental/backends/omnisci/query_compiler.py#L333) [[2]](https://github.com/modin-project/modin/blob/cf259f1cd38c70e246fd5e3ea904b675a5d10bd9/modin/experimental/backends/omnisci/query_compiler.py#L345) to default to pandas when it's operating with unsupported data (with the changes in this branch default-to-pandas behaviour is identical to the regular OmniSci `_get_index/columns` flow). 

OmniSci backend tests of working with unsupported data located at the [`TestBadData` class](https://github.com/modin-project/modin/blob/cf259f1cd38c70e246fd5e3ea904b675a5d10bd9/modin/experimental/engines/omnisci_on_ray/test/test_dataframe.py#L1620), at the moment it doesn't have explicit tests for `df.index` and `df.columns`, they should be added in order to check whether the changes in `base._get_axis` has no effect on the OmniSci backend. If these tests will be green, then this special-casing in [`_get_index`](https://github.com/modin-project/modin/blob/cf259f1cd38c70e246fd5e3ea904b675a5d10bd9/modin/experimental/backends/omnisci/query_compiler.py#L331) and [`_get_columns`](https://github.com/modin-project/modin/blob/cf259f1cd38c70e246fd5e3ea904b675a5d10bd9/modin/experimental/backends/omnisci/query_compiler.py#L343) should be removed

modin/pandas/test/test_series.py
```diff
@@ -509,7 +504,16 @@ def test___repr__(name, dt_index, data):
             "1/1/2000", periods=len(pandas_series.index), freq="T"
         )
         pandas_series.index = modin_series.index = index
-    assert repr(modin_series) == repr(pandas_series)
+
+    try:
+        assert repr(modin_series) == repr(pandas_series)
+    except Exception as e:
```
dchigarev, line 510: We possibly want to catch only `AssertionError` here, we don't want to enter in this `except` block if `repr(modin_series)` arise some exception.
```suggestion
    except AssertionError as e:
```

modin/pandas/test/test_general.py
```diff
@@ -696,7 +696,6 @@ def test_to_pandas_indices():
     [
         (lambda df: df.mean(level=0), r"DataFrame\.mean"),
         (lambda df: df + df, r"DataFrame\.add"),
-        (lambda df: df.index, r"DataFrame\.get_axis\(0\)"),
```
vnlitvinov, line 699: why are we removing a testcase?

modin/pandas/test/test_series.py
vnlitvinov: Wait, are you seriously suggesting to replace simple `if repr(a) != repr(b): do_stuff()` with this `try.. assert.. except AssertionError.. do_stuff()`?..

If anything I suggest rewriting this piece into smth like
```python
if get_current_backend() == 'BaseOnPython` and data == 'empty':
    # TODO: explain reasons
    assert modin_series.dtype == np.object
    assert pandas_series.dtype == np.float64
else:
    assert repr(modin_series) == repr(pandas_series)
```

modin/pandas/test/test_series.py
```diff
@@ -1533,12 +1537,17 @@ def test_dropna_inplace(data):
...
-    assert modin_series.dtype == pandas_series.dtype
+    try:
+        assert modin_series.dtype == pandas_series.dtype
+    except Exception as e:
+        # TODO: Remove this try-except when default  `dtype` of empty Series will be `object` in pandas (see #3142).
+        if get_current_backend() == "BaseOnPython":
+            assert modin_series.dtype == np.object
+            assert pandas_series.dtype == np.float64
+        else:
+            raise e
```
vnlitvinov, line 1550: I would suggest extracting this check into a separate function to not copy-paste in tests

vnlitvinov (changes requested):

vnlitvinov:
Also, to add to my review above - why are we adding some special corner-case handling in the tests instead of fixing the bugs which make those test fail?..

Rubtsowa:
@vnlitvinov We adding some special corner-case because this problem will be solved with future version pandas. Just now default Series in Pandas is float64, but in Modin - object. More details in this issue - https://github.com/modin-project/modin/issues/3142

modin/pandas/test/test_general.py
Rubtsowa: This test is no longer needed because we changed the _get_axis for BaseOnPython.

Rubtsowa:
@dchigarev I added tests in modin/experimental/engines/omnisci_on_ray/test/test_dataframe.py and they pass.

dchigarev:
> @dchigarev I added tests in modin/experimental/engines/omnisci_on_ray/test/test_dataframe.py and they pass.

That means that [this special casing](https://github.com/modin-project/modin/blob/cf259f1cd38c70e246fd5e3ea904b675a5d10bd9/modin/experimental/backends/omnisci/query_compiler.py#L332-L333) at `_get_index` and `_get_columns` was redundant and so can be safely removed. Even more, now we can completely remove these methods. They only used as the getters of the [`index` and `columns` properties](https://github.com/modin-project/modin/blob/cf259f1cd38c70e246fd5e3ea904b675a5d10bd9/modin/experimental/backends/omnisci/query_compiler.py#L680-L681), I believe we can replace `_get_index/_get_columns -> get_axis(0/1)` here.

modin/pandas/test/test_general.py
vnlitvinov: have we added the test for `_get_axis` in base implementation then?

modin/backends/base/query_compiler.py
```diff
@@ -33,11 +32,10 @@
 
 
 def _get_axis(axis):
-    def axis_getter(self):
```
YarShev, line 36: I don't think we should change the implementation of this method because the query compiler located here is responsible for default to pandas execution. I suggest keeping `_get_axis` methods here and in pandas QC as those were.

YarShev (changes requested):

modin/backends/base/query_compiler.py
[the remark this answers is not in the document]
Rubtsowa: @vnlitvinov yes.In this case, the indices for empty dataframes are assigned the ones that they should, and not just the default.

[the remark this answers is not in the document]
YarShev: We should fix other backends (different from `BaseOnPython`) using `BaseOnPython` backend but not vice versa. That's why I requested changes.

The issue was discussed offline and it was decided to create an issue in pandas repo. @Rubtsowa , put a link to the pandas issue here, please.

modin/backends/base/query_compiler.py
prutskov: In that case we should fix https://github.com/modin-project/modin/blob/e633a32c2706f75fbbdf852deace9e3384b31086/modin/engines/base/frame/data.py#L2219-L2242
because, I think, root cause of issue is placed here. In `PandasFrame.to_pandas` method in case of empty df we update only `columns`, but not `index`, because for empty `Series` `PandasFrame.columns` is equal to [`__reduced__`]. In this case, when we created empty `Series` with specific type of `index` (datetime), we have correct type for internal `PandasFrame.index`, but we will have default type of `index` for `pandas.DataFrame`, that will returned from `PandasFrame.to_pandas`. Possible, this is actual for the other backends.

vnlitvinov (commented):
~~Changes now look fine to me, left some questions though.~~ _accidental premature post, ignore_

modin/engines/base/frame/data.py
```diff
@@ -2226,10 +2226,7 @@ def to_pandas(self):
         """
         df = self._partition_mgr_cls.to_pandas(self._partitions)
         if df.empty:
-            if len(self.columns) != 0:
-                df = pandas.DataFrame(columns=self.columns)
-            else:
-                df = pandas.DataFrame(columns=self.columns, index=self.index)
+            df = pandas.DataFrame(columns=self.columns, index=self.index)
```
vnlitvinov, line 2229: Could you please explain what this change does? It looks okay, but I'm wondering why the original code was like it was, and why this was needed to fix the tests...

modin/pandas/test/test_series.py
```diff
@@ -961,15 +959,17 @@ def test_asof(where):
     # With NaN:
     values = [1, 2, np.nan, 4]
     index = [10, 20, 30, 40]
-    modin_series, pandas_series = pd.Series(values, index=index), pandas.Series(
-        values, index=index
+    modin_series, pandas_series = (
+        pd.Series(values, index=index),
+        pandas.Series(values, index=index),
```
vnlitvinov, line 964: why do you want to change this in this PR? I agree the changed version looks cleaner when formatted, but it's purely aesthetical and doesn't correlate with the issue being fixed (so looks like an unrelated change being squeezed into).

vnlitvinov (commented):
Changes now look fine to me, left some questions though.

prutskov:
As described here https://github.com/modin-project/modin/pull/3141#discussion_r662146492, the root cause is bug in `PandasFrame.to_pandas` method. It affected all engines (reproducer is below, tested on Ray engine):
```python
import modin.pandas as pd
import pandas

ser = pd.Series()
index = pandas.date_range(
    "1/1/2000", periods=len(ser.index), freq="T"
)
ser.index = index

print(f'Index: {ser.index}')
print(f'Index after to_pandas: {ser._query_compiler.to_pandas().index}')
```
Output
```bash
Index: DatetimeIndex([], dtype='datetime64[ns]', freq='T')
Index after to_pandas: Index([], dtype='object')
```
Output after Maria's fix:
```bash
Index: DatetimeIndex([], dtype='datetime64[ns]', freq='T')
Index after to_pandas: DatetimeIndex([], dtype='datetime64[ns]', freq='T')
```

Possible, need to add testing of such cases.

modin/pandas/test/test_series.py
Rubtsowa: @vnlitvinov This reformatting happened during the use of black. Otherwise, black and flake8 in ci would be red.

modin/engines/base/frame/data.py
Rubtsowa: @vnlitvinov len (columns) always! = 0, because there the column '__reduced__' is written in the case of an empty dataframe, because of which the length will not be equal to zero. And we will always have a default index.

vnlitvinov:
I was mostly wondering why we dropped the index for non-empty columns in an empty dataframe... there has to be a reason. :)

Also, adding this

> Output after Maria's fix:
> 
> ```shell
> Index: DatetimeIndex([], dtype='datetime64[ns]', freq='T')
> Index after to_pandas: DatetimeIndex([], dtype='datetime64[ns]', freq='T')
> ```

as a test would be a great addition to this PR 😄

vnlitvinov (approved):
LGTM

YarShev (approved):
@Rubtsowa , LGTM, thanks!

prutskov:
Why did we solve not to add test described here https://github.com/modin-project/modin/pull/3141#issuecomment-872258328?

YarShev:
@prutskov , good notice! Please, add this one-line change in a separate PR.

PR #3175, opened 2021-06-22
DOCS-#3168: Fix inconsistencies in the `modin.config` docs

### LINKED ISSUE #3168: Fix inconsistencies in the `modin.config` docs
Comments https://github.com/modin-project/modin/pull/3129#discussion_r655444346 and https://github.com/modin-project/modin/pull/3129#discussion_r655445196 should be addressed.

amyskov:

## What do these changes do?

- [x] passes `flake8 modin`
- [x] Resolves #3168  

### DIFF
--- a/modin/config/__main__.py
+++ b/modin/config/__main__.py
@@ -56,7 +56,7 @@ def export_config_help(filename: str):
         if isinstance(obj, type) and issubclass(obj, Parameter) and not obj.is_abstract:
             data = {
                 "Config Name": obj.__name__,
-                "Env. Variable Name": obj.varname,
+                "Env. Variable Name": getattr(obj, "varname", None),
                 "Default Value": obj._get_default()
                 if obj.__name__ != "RayRedisPassword"
                 else "random string",
@@ -71,7 +71,7 @@ def export_config_help(filename: str):
 if __name__ == "__main__":
     parser = argparse.ArgumentParser()
     parser.add_argument(
-        "-export_path",
+        "--export_path",
         dest="export_path",
         type=str,
         required=False,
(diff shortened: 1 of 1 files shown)


modin/config/__main__.py
```diff
@@ -71,7 +71,7 @@ def export_config_help(filename: str):
 if __name__ == "__main__":
     parser = argparse.ArgumentParser()
     parser.add_argument(
-        "-export_path",
+        "--export_path",
```
vnlitvinov, line 74:
```suggestion
        "--export-path",
```
plus note that you should change the calling line in `conf.py` now, otherwise the variables' table won't be regenerated properly

modin/config/__main__.py
```diff
@@ -56,7 +56,7 @@ def export_config_help(filename: str):
         if isinstance(obj, type) and issubclass(obj, Parameter) and not obj.is_abstract:
             data = {
                 "Config Name": obj.__name__,
-                "Env. Variable Name": obj.varname,
+                "Env. Variable Name": getattr(obj, "varname", None),
```
vnlitvinov, line 59:
```suggestion
                "Env. Variable Name": getattr(obj, "varname", "not backed by environment"),
```

vnlitvinov (changes requested):

modin/config/__main__.py
amyskov: In the `conf.py`, `export_config_help` called directly (when we used separate process to call `python -m modin.config --export-path` file with configs didn't generated during build on the readthedocs website for some reason, so function direct call was used), so it shouldn't affect docs generating process.

vnlitvinov (approved):
LGTM

PR #3189, opened 2021-06-27
TEST-#3188: remove Dask pin

### LINKED ISSUE #3188: Remove Dask pin in setup.py
https://github.com/conda-forge/modin-feedstock/pull/17

anmyachev:

## What do these changes do?

- [ ] passes `flake8 modin`
- [x] Resolves #3188 
- [ ] tests added and passing

### DIFF
--- a/environment-dev.yml
+++ b/environment-dev.yml
@@ -6,8 +6,8 @@ dependencies:
   - numpy>=1.16.5
   - pyarrow>=1.0.0
   - libthrift<0.14.1  # ref: modin gh-2840
-  - dask[complete]>=2.12.0,<=2.19.0
-  - distributed>=2.12.0,<=2.19.0
+  - dask[complete]>=2.12.0
+  - distributed>=2.12.0
   - xarray
   - Jinja2
   - pathlib
--- a/environment-dev.yml
+++ b/environment-dev.yml
@@ -6,8 +6,8 @@ dependencies:
   - numpy>=1.16.5
   - pyarrow>=1.0.0
   - libthrift<0.14.1  # ref: modin gh-2840
-  - dask[complete]==2.21.0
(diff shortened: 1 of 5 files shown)


anmyachev:
> @anmyachev LGTM, is it ready to move from draft?

No, with the current versions of dask and distributed(2.21.0), some CI jobs usually work much longer (up to 6 hours). Maybe deadlock. I will try another versions.

environment-dev.yml
```diff
@@ -6,8 +6,8 @@ dependencies:
   - numpy>=1.16.5
   - pyarrow>=1.0.0
   - libthrift<0.14.1  # ref: modin gh-2840
-  - dask[complete]>=2.12.0,<=2.19.0
-  - distributed>=2.12.0,<=2.19.0
+  - dask[complete]>=2.22.0
```
devin-petersohn, line 9:
```suggestion
  - dask[complete]>=2020
```

devin-petersohn (commented):
They changed their version numbers to follow year & month recently. Maybe we can try starting with 2020.

It might be good to put a timeout on the jobs too so they don't eat our minutes.

environment-dev.yml
vnlitvinov: As we've removed the upper pin, it now installs whatever is recent enough, doesn't it? I see `dask 2021.6.2` being used right now in CI.

vnlitvinov (approved):
LGTM

anmyachev:
> They changed their version numbers to follow year & month recently. Maybe we can try starting with 2020.
> 
> It might be good to put a timeout on the jobs too so they don't eat our minutes.

`2.22.0` version works, so we can leave it.
We can choose not to set timeouts until the deadlock occurs frequently.

Issue #3142, opened 2021-06-09
Unskip tests with Future version pandas


Rubtsowa:
Currently we have problem with tests: test___repr__ - https://github.com/modin-project/modin/blob/master/modin/pandas/test/test_series.py#L501 , test_dtype_empty - https://github.com/modin-project/modin/blob/master/modin/pandas/test/test_series.py#L1539 . 

I found this problem when I was working on #2164.  

These tests creating empty Series. Modin have default type for Series ``object``. But currently Pandas have default type for Series ``float64``. Therefore now tests failed. 

Currently version pandas have next DeprecationWarning:
`` DeprecationWarning: The default dtype for empty Series will be 'object' instead of 'float64' in a future version. Specify a dtype explicitly to silence this warning.``

More details about this you can see here: https://pandas.pydata.org/pandas-docs/stable/whatsnew/v1.0.0.html

Therefore tests will be passed after new version Pandas.

YarShev:
@Rubtsowa , I am not sure that I understand the issue. Example:

```python
import pandas
import modin.pandas as pd
ps = pandas.Series()
Series([], dtype: float64)
ms = pd.Series()
Series([], dtype: float64)
ps.dtypes == ms.dtypes
True
```
Can you elaborate?

Related to #3141 .

YarShev:
@Rubtsowa , what pandas version do you have installed?

Rubtsowa:
@YarShev 1.2.4

Rubtsowa:
This problem occurs when we call dtype. Because the dtype is assigned internally from an empty DataFrame.

YarShev:
[42078](https://github.com/pandas-dev/pandas/issues/42078)

Garra1980:
so what is the plan here then?

Garra1980:
We need to wait Pandas 2.0 for test unskipping

RehanSD:
Hi @Garra1980 @dchigarev @YarShev has this issue been resolved?

anmyachev:
Fixed in https://github.com/modin-project/modin/pull/5995.

PR #3212, opened 2021-07-02
TEST-#3214: Add test for df/series.index after PandasFrame.to_pandas is called

### LINKED ISSUE #3214: Add test for df/series.index after PandasFrame.to_pandas is called
Function `PandasFrame.to_pandas` was fixed in #3141 but test wasn't added.

Rubtsowa:

## What do these changes do?

- [x] passes `flake8 modin`
- [x] Resolves #3214  
- [x] tests added and passing

### DIFF
--- a/modin/pandas/test/test_general.py
+++ b/modin/pandas/test/test_general.py
@@ -674,9 +674,10 @@ def test_to_numeric(data, errors, downcast):
     df_equals(modin_result, pandas_result)
 
 
-def test_to_pandas_indices():
-    data = test_data_values[0]
-
+@pytest.mark.parametrize(
+    "data", [test_data_values[0], []], ids=["test_data_values[0]", "[]"]
+)
+def test_to_pandas_indices(data):
     md_df = pd.DataFrame(data)
     index = pandas.MultiIndex.from_tuples(
         [(i, i * 2) for i in np.arange(len(md_df) + 1)], names=["A", "B"]
--- a/modin/pandas/test/test_series.py
+++ b/modin/pandas/test/test_series.py
@@ -506,6 +506,7 @@ def test___repr__(name, dt_index, data):
         # TODO: Remove this when default `dtype` of empty Series will be `object` in pandas (see #3142).
         assert modin_series.dtype == np.object
         assert pandas_series.dtype == np.float64
+        df_equals(modin_series.index, pandas_series.index)
     else:
         assert repr(modin_series) == repr(pandas_series)
 


modin/pandas/test/test_series.py
```diff
@@ -1905,6 +1905,30 @@ def test_index(data):
...
+@pytest.mark.parametrize("name", ["Dates", None])
+@pytest.mark.parametrize(
+    "dt_index", [True, False], ids=["dt_index_true", "dt_index_false"]
+)
+@pytest.mark.parametrize(
+    "data",
+    [*test_data_values, "empty"],
+    ids=[*test_data_keys, "empty"],
+)
+def test_index_dtype(name, dt_index, data):
```
YarShev, line 1917: Why did you copy `test___repr` to test the case? You only needed to add the line `df_equals(modin_series.index, pandas_series.index)` after that line because the issue persisted for BaseOnPython only. https://github.com/modin-project/modin/blob/c6393b54639b75ad38f028ba06c6eb1e43bfb790/modin/pandas/test/test_series.py#L508

Also, create an issue for this PR, please.

Rubtsowa: @YarShev We have issue about default dtypes: https://github.com/modin-project/modin/issues/3142

YarShev: I know. We follow the following practice: for each PR should be created an issue to be resolved.

Rubtsowa: @YarShev Thanks!

modin/pandas/test/test_series.py
```diff
@@ -509,6 +509,8 @@ def test___repr__(name, dt_index, data):
     else:
         assert repr(modin_series) == repr(pandas_series)
 
+    df_equals(modin_series.index, pandas_series.index)
```
prutskov, line 512: Why do we add this check? Original plan was in adding check on equality of `df.index` and `df._query_compiler.to_pandas().index`, described here https://github.com/modin-project/modin/pull/3141#issuecomment-872258328

YarShev: @Rubtsowa , this check should go right after 508 line as I specified [here](https://github.com/modin-project/modin/pull/3212#discussion_r663693092) because to create an empty DF `to_pandas` is called for BaseOnPython only.
@prutskov , this check covers that case.

prutskov: This check doesn't cover that case for the other backends, because original issue was in missing correct index of pandas DataFrame after `PandasFrame.to_pandas()` method. The other backends hided `to_pandas` issue, because `df.index` don't provoke call of `to_pandas` underneath.

YarShev: The issue was related to `to_pandas`. We need to verify that Modin Dataframe index after `to_pandas` call is equal to pandas Dataframe index. The call `modin_df._query_compiler.to_pandas().index` for each backend is the same if we call `df.index` for `BaseOnPython` since `to_pandas` is called underneath.

prutskov: I agree with you, that "The call `modin_df._query_compiler.to_pandas().index` for each backend is the same if we call `df.index` for `BaseOnPython` for empty frames since the same `to_pandas` is called underneath." My suggestion is mainly directed at highlighting this issue for other backends.

YarShev: Okay, let's add the test. @Rubtsowa, can you do it, please?

YarShev: Also, fix your commit message (remove spaces in "TEST - #3214").

modin/pandas/test/test_general.py
```diff
@@ -690,6 +690,8 @@ def test_to_pandas_indices():
 
     pd_df = md_df._to_pandas()
 
+    df_equals(md_df.index, md_df._query_compiler.to_pandas().index)
```
YarShev, line 693: This check doesn't test our case because md_df is not empty. We should add a parameter named `data`, which would have two values (`test_data_values[0]` and `[]`, to this test. After that you can remove this line and leave the body of the test as it is.

YarShev (approved):
@Rubtsowa , thanks, LGTM!

PR #3240, opened 2021-07-14
TEST-#3254: add check asv benchmarks on pure pandas

### LINKED ISSUE #3254: Add check ASV benchmarks on pure Pandas
A run of benchmarks on the Pandas was added, the results in the dashboard can be viewed using `machine=xeon-e5-pandas` filter. This check is needed to find errors at an earlier stage (Modin CI).

Example:
http://modin.org/modin-bench/#benchmarks.TimeAppend.time_append?machine=xeon-e5-pandas&python=3.8&p-shapes=((500000%2C%2020)%2C%20(1000000%2C%2010))

anmyachev:

## What do these changes do?

- [x] passes `flake8 modin`
- [x] Resolves #3254 
- [x] tests added and passing

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -388,7 +388,7 @@ jobs:
           asv check -v
           git remote add upstream https://github.com/modin-project/modin.git
           git fetch upstream
-          if [ "test" = "test" ]; then
+          if git diff upstream/master --name-only | grep -q "^asv_bench/"; then
               asv machine --yes
 
               # check Modin on Ray
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -392,21 +392,17 @@ jobs:
               asv machine --yes
 
               # check Modin on Ray
-              asv run --quick --show-stderr --launch-method=spawn -b ^benchmarks -b ^io -b ^scalability \
-                | sed "/failed$/ s/^/##[error]/" | tee benchmarks.log
+              asv run --quick --strict --show-stderr --launch-method=spawn \
+                -b ^benchmarks -b ^io -b ^scalability | tee benchmarks.log
 
               # check pure pandas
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -390,15 +390,19 @@ jobs:
           git fetch upstream
           if git diff upstream/master --name-only | grep -q "^asv_bench/"; then
               asv machine --yes
-              asv run --quick --show-stderr --launch-method=spawn -b ^benchmarks -b ^io -b ^scalability \
-                | sed "/failed$/ s/^/##[error]/" | tee benchmarks.log
-              # check omnisci backend
-              MODIN_BACKEND=omnisci MODIN_EXPERIMENTAL=true asv run --quick -b ^omnisci \
-                --show-stderr --launch-method=forkserver --config asv.conf.omnisci.json \
-                | sed "/failed$/ s/^/##[error]/" | tee benchmarks.log
-              if grep "failed" benchmarks.log > /dev/null ; then
```
vnlitvinov, line 399: why `sed` and `grep` are no longer there? were they not needed?

anmyachev: Previously, ASV would always return 0, even if the benchmarks failed. In this case, we used parsing of logs in order to return an error code other than 0. The new version of the program has `--strict` option that returns an error code other than 0 if benchmarks fail, which eliminates the need to parse the logs.

vnlitvinov: Where is ASV version update then?..

anmyachev: This feature already in ASV that we take from master https://github.com/modin-project/modin/blob/7727c23adc58c470b6008dc38519da7d21c3ac0b/environment-dev.yml#L42

vnlitvinov (approved):
LGTM

PR #3286, opened 2021-07-27
FIX-#3276: fix ZeroDivision error during data import operation call

### LINKED ISSUE #3276: Data import operations fail if empty frame should be read
### System information
- **Modin version**: d03ed30868fb042bbc7282b9c9574145e1228eeb
- **Python version**: 3.8.10
```
import modin.pandas as pd
from io import StringIO
import os

test_filename = "test.csv"
data = """col1,col2
1,2
3,4
"""

try :
    with open(test_filename, "w") as f:
        f.write(data)
    
    pd.read_csv(test_filename, usecols=["col1"], index_col="col1")
finally:
    os.remove(test_filename)
```

### Describe the problem

`read_csv` fails during metadata calculation.

### Source code / logs

```
  File "zero_div.py", line 15, in <module>
    pd.read_csv(test_filename, usecols=["col1"], index_col="col1")
  File "/modin/modin/pandas/io.py", line 131, in read_csv
    return _read(**kwargs)
  File "/modin/modin/pandas/io.py", line 59, in _read
    pd_obj = FactoryDispatcher.read_csv(**kwargs)
  File "/modin/modin/data_management/factories/dispatcher.py", line 172, in read_csv
    return cls.__factory._read_csv(**kwargs)
  File "/modin/modin/data_management/factories/factories.py", line 206, in _read_csv
  File "/modin/modin/engines/base/io/file_dispatcher.py", line 67, in read
    query_compiler = cls._read(*args, **kwargs)
  File "/modin/modin/engines/base/io/text/csv_dispatcher.py", line 112, in _read
    column_widths, num_splits = cls._define_metadata(
  File "/modin/modin/engines/base/io/text/text_file_dispatcher.py", line 422, in _define_metadata
    column_chunksize = compute_chunksize(df, num_splits, axis=1)
  File "/modin/modin/data_management/utils.py", line 70, in compute_chunksize
    col_chunksize = get_default_chunksize(len(df.columns), num_splits)
  File "/modin/modin/data_management/utils.py", line 37, in get_default_chunksize
    length // num_splits if length % num_splits == 0 else length // num_splits + 1
ZeroDivisionError: integer division or modulo by zero
```

amyskov:

## What do these changes do?

- [x] passes `flake8 modin`
- [x] Resolves #3276   
- [x] tests added and passing

### DIFF
--- a/modin/backends/pandas/parsers.py
+++ b/modin/backends/pandas/parsers.py
@@ -153,8 +153,11 @@ class PandasParser(object):
             pandas.Series where index is columns names and values are
             columns dtypes.
         """
+        partitions_dtypes = cls.materialize(dtypes_ids)
+        if all([len(dtype) == 0 for dtype in partitions_dtypes]):
+            return None
         return (
-            pandas.concat(cls.materialize(dtypes_ids), axis=1)
+            pandas.concat(partitions_dtypes, axis=1)
             .apply(lambda row: find_common_type_cat(row.values), axis=1)
             .squeeze(axis=0)
         )
--- a/modin/data_management/utils.py
+++ b/modin/data_management/utils.py
@@ -59,6 +59,8 @@ def compute_chunksize(df, num_splits, default_block_size=32, axis=None):
         If axis is 1 or 0, returns an integer number of rows/columns to split the
         DataFrame. If axis is None, returns a tuple containing both.
     """
+    # `num_splits` should be a non-zero value
+    num_splits = num_splits if num_splits > 0 else 1
     if axis == 0 or axis is None:
         row_chunksize = get_default_chunksize(len(df.index), num_splits)
         # Take the min of the default and the memory-usage chunksize first to avoid a
--- a/modin/data_management/utils.py
+++ b/modin/data_management/utils.py
@@ -59,8 +59,6 @@ def compute_chunksize(df, num_splits, default_block_size=32, axis=None):
         If axis is 1 or 0, returns an integer number of rows/columns to split the
         DataFrame. If axis is None, returns a tuple containing both.
     """
-    # `num_splits` should be a non-zero value
-    num_splits = num_splits if num_splits > 0 else 1
     if axis == 0 or axis is None:
         row_chunksize = get_default_chunksize(len(df.index), num_splits)
         # Take the min of the default and the memory-usage chunksize first to avoid a
--- a/modin/engines/base/io/text/csv_dispatcher.py
+++ b/modin/engines/base/io/text/csv_dispatcher.py
@@ -107,7 +107,7 @@ class CSVDispatcher(TextFileDispatcher):
         # Max number of partitions available
         num_partitions = NPartitions.get()
         # This is the number of splits for the columns
-        num_splits = min(len(column_names), num_partitions)
+        num_splits = min(len(column_names) or 1, num_partitions)
         # Metadata definition
         column_widths, num_splits = cls._define_metadata(
             pd_df_metadata, num_splits, column_names
--- a/modin/engines/base/io/text/csv_dispatcher.py
+++ b/modin/engines/base/io/text/csv_dispatcher.py
@@ -103,15 +103,7 @@ class CSVDispatcher(TextFileDispatcher):
             **dict(kwargs, nrows=1, skipfooter=0, index_col=index_col),
         )
         column_names = pd_df_metadata.columns
-
-        # Max number of partitions available
-        num_partitions = NPartitions.get()
-        # This is the number of splits for the columns
-        num_splits = min(len(column_names) or 1, num_partitions)
-        # Metadata definition
-        column_widths, num_splits = cls._define_metadata(
-            pd_df_metadata, num_splits, column_names
-        )
+        column_widths, num_splits = cls._define_metadata(pd_df_metadata, column_names)
 
         # kwargs that will be passed to the workers
         partition_kwargs = dict(
@@ -129,7 +121,7 @@ class CSVDispatcher(TextFileDispatcher):
         with cls.file_open(filepath_or_buffer_md, "rb", compression_infered) as f:
             splits = cls.partitioned_file(
                 f,
-                num_partitions=num_partitions,
+                num_partitions=NPartitions.get(),
                 nrows=kwargs.get("nrows", None),
                 skiprows=skiprows,
                 quotechar=quotechar,
--- a/modin/engines/base/io/text/csv_dispatcher.py
+++ b/modin/engines/base/io/text/csv_dispatcher.py
@@ -107,8 +107,7 @@ class CSVDispatcher(TextFileDispatcher):
         # Max number of partitions available
         num_partitions = NPartitions.get()
         # This is the number of splits for the columns
-        cols_number = len(column_names)
-        num_splits = min(cols_number if cols_number else 1, num_partitions)
+        num_splits = min(len(column_names) or 1, num_partitions)
         # Metadata definition
         column_widths, num_splits = cls._define_metadata(
             pd_df_metadata, num_splits, column_names
--- a/modin/engines/base/io/text/csv_dispatcher.py
+++ b/modin/engines/base/io/text/csv_dispatcher.py
@@ -107,7 +107,8 @@ class CSVDispatcher(TextFileDispatcher):
         # Max number of partitions available
         num_partitions = NPartitions.get()
         # This is the number of splits for the columns
-        num_splits = min(len(column_names), num_partitions)
+        cols_number = len(column_names)
+        num_splits = min(cols_number if cols_number else 1, num_partitions)
         # Metadata definition
         column_widths, num_splits = cls._define_metadata(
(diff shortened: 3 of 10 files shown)


modin/engines/base/io/text/csv_dispatcher.py
```diff
@@ -107,7 +107,8 @@ def _read(cls, filepath_or_buffer, **kwargs):
         # Max number of partitions available
         num_partitions = NPartitions.get()
         # This is the number of splits for the columns
-        num_splits = min(len(column_names), num_partitions)
+        cols_number = len(column_names)
+        num_splits = min(cols_number if cols_number else 1, num_partitions)
```
anmyachev, line 111:
```suggestion
        num_splits = min(len(column_names) or 1, num_partitions)
```

anmyachev: Why is the minimum number of columns - 1?

amyskov: This is done to avoid division by zero [here](https://github.com/modin-project/modin/blob/master/modin/data_management/utils.py#L37) (for full traceback see issue header).

modin/engines/base/io/text/csv_dispatcher.py
```diff
@@ -107,7 +107,7 @@ def _read(cls, filepath_or_buffer, **kwargs):
         # Max number of partitions available
         num_partitions = NPartitions.get()
         # This is the number of splits for the columns
-        num_splits = min(len(column_names), num_partitions)
+        num_splits = min(len(column_names) or 1, num_partitions)
```
YarShev, line 110: Is this adding really required?

modin/backends/pandas/parsers.py
```diff
@@ -153,8 +153,11 @@ def get_dtypes(cls, dtypes_ids):
             pandas.Series where index is columns names and values are
             columns dtypes.
         """
+        partitions_dtypes = cls.materialize(dtypes_ids)
+        if all([len(dtype) == 0 for dtype in partitions_dtypes]):
```
YarShev, line 157: Can you elaborate why this change fixes the issue?

modin/pandas/test/test_io.py
```diff
@@ -1115,6 +1115,22 @@ def get_internal_df(df):
         for name1, name2 in zip(get_internal_df(read_df).index.names, [None, "a"]):
             assert name1 == name2
 
+    def test_read_csv_empty_frame(
+        self,
+        request,
+    ):
+        if request.config.getoption("--simulate-cloud").lower() != "off":
```
YarShev, line 1122: Can you consider this comment?
https://github.com/modin-project/modin/issues/3295#issuecomment-889047542

modin/backends/pandas/parsers.py
amyskov: Yes, this change is needed because when empty frame should be read, each worker will read empty frame and it's `dtypes` will be empty `Series` that corresponds to `dtypes=None`. If we skip this check, `dtypes` will values will be calculated incorrectly because `partitions_dtypes` contains no columns.

modin/engines/base/io/text/csv_dispatcher.py
amyskov: Yes, we need to set non-zero `num_splits` value to avoid `ZeroDivision` error [here](https://github.com/modin-project/modin/blob/master/modin/data_management/utils.py#L37).

modin/pandas/test/test_io.py
amyskov: Yes, fixed.

YarShev (dismissed):
@amyskov , LGTM assuming tests pass. @anmyachev , any other comments?

anmyachev:
@amyskov you are changing the way define partitions for `read_csv` function only, but this problem is more general. It is necessary to generalize changes at least to all I/O functions.

YarShev:
@anmyachev , good notice! We have some places where num_splits is computed in depend on the length of columns. There might be an issue when reading empty frames from other files. @amyskov , can you check it?

amyskov:
> @anmyachev , good notice! We have some places where num_splits is computed in depend on the length of columns. There might be an issue when reading empty frames from other files. @amyskov , can you check it?

Yes, `num_splits` check was moved to the higher level function, so now it affects all IO functions.

modin/data_management/utils.py
```diff
@@ -59,6 +59,8 @@ def compute_chunksize(df, num_splits, default_block_size=32, axis=None):
         If axis is 1 or 0, returns an integer number of rows/columns to split the
         DataFrame. If axis is None, returns a tuple containing both.
     """
+    # `num_splits` should be a non-zero value
+    num_splits = num_splits if num_splits > 0 else 1
```
anmyachev, line 63: `num_splits` is also used in functions:
* `_define_metadata`;
* `split_result_of_axis_func_pandas`;
* Based on this variable, a `column_widths` variable for `build_partition` function is determined (For example, in `JSONDispatcher._read`)

Should we also change these places for consistency?

It would be nice to add examples to some other I/O functions (For example, `read_json`).

amyskov: In the case of empty frame reading, `num_splits` will be redefined to 1 in the `_define_metadata` and then processed normally further, but i think it will be more correct to set correct `num_splits` value in the moment of `num_splits` definition, so previous approach was used.
I suggest to fix in this PR only `read_csv` function as it mentioned in the issue. Actually, what you mean under "examples"?

anmyachev: Oh, I wanted to say `add more IO tests` :)

anmyachev: There are a few more places where similar lines can be found:
* https://github.com/modin-project/modin/blob/54898ef78e83209fba3233606841ca988e531aee/modin/engines/base/io/text/excel_dispatcher.py#L163
* https://github.com/modin-project/modin/blob/54898ef78e83209fba3233606841ca988e531aee/modin/engines/base/io/text/fwf_dispatcher.py#L157

I believe that the same error exists for them. Why make changes for them separately if the problematic mechanism (`num_splits` definition) is the same?

amyskov: Yes, you are right, the same issue can occur for another dispatchers, but for excel dispatcher we wouldn't be capable to verify this fix because of https://github.com/modin-project/modin/issues/3305, so additional PR should be done anyway. But, i think, it should be fine to add these fixes anyway, so they were added.

modin/engines/base/io/text/csv_glob_dispatcher.py
```diff
@@ -184,7 +184,7 @@ def _read(cls, filepath_or_buffer, **kwargs):
             # Max number of partitions available
             num_partitions = NPartitions.get()
             # This is the number of splits for the columns
-            num_splits = min(len(column_names), num_partitions)
+            num_splits = min(len(column_names) or 1, num_partitions)
```
anmyachev, line 187: Is it possible to move this code into a separate function so that this code can be reused?

amyskov: Yes, this and other related code was moved to the `_define_metadata` function.

modin/engines/base/io/text/json_dispatcher.py
```diff
@@ -119,3 +108,44 @@ def _read(cls, path_or_buf, **kwargs):
         )
         new_frame.synchronize_labels(axis=0)
         return cls.query_compiler_cls(new_frame)
+
+    @classmethod
+    def _define_metadata(
```
anmyachev, line 113: We already have `_define_metadata` function in `text_file_dispatcher.py`.

Why create another function with the same name in `json_dispatcher.py`?
Shouldn't they be doing the same thing?

anmyachev: It seems to me that there is a bug in `_define_metadata` function in `json_dispatcher.py`.

Just try with next parameters: `num_splits == 4, len(column_names) == 80 and column_chunksize == 32`.

It turns out that `column_widths=[32, 32, 32, -16]`, but should be equals to `[32, 32, 16, 0]`

anmyachev: So we can use `_define_metadata` from `text_file_dispatcher.py` in all place.

I suggest first fixing the found bug in a separate pull request, and then merge this one.

anmyachev:
@amyskov please don't forget to update issue, since you are fixing a more global problem.

YarShev:
@amyskov , please, ping me when the PR is ready for review.

modin/engines/base/io/text/json_dispatcher.py
amyskov: You are right, it's a bug, corresponding issue was created https://github.com/modin-project/modin/issues/3313.

amyskov:
@YarShev , PR is ready for review.

YarShev:
@amyskov , please, consider the discussion from #3314.

YarShev (approved):
@amyskov , LGTM, thanks!

PR #3290, opened 2021-07-28
FIX-#3289: Update pandas to 1.3.1

### LINKED ISSUE #3289: Update pandas to 1.3.1

YarShev:

## What do these changes do?

- [x] passes `flake8 modin`
- [x] Resolves #3289  
- [x] tests added and passing

### DIFF
--- a/modin/backends/base/query_compiler.py
+++ b/modin/backends/base/query_compiler.py
@@ -2066,7 +2066,7 @@ class BaseQueryCompiler(abc.ABC):
             df.insert(loc, column, value)
             return df
 
-        return DataFrameDefault.register(inserter)(
+        return DataFrameDefault.register(inserter, inplace=True)(
             self, loc=loc, column=column, value=value
         )
 
--- a/modin/pandas/dataframe.py
+++ b/modin/pandas/dataframe.py
@@ -1113,13 +1113,9 @@ class DataFrame(BasePandasDataset):
         Insert column into ``DataFrame`` at specified location.
         """
         if isinstance(value, (DataFrame, pandas.DataFrame)):
-            # DataFrame considered as an array-like objects and so value have to be
-            # retrieved via DataFrame.__iter__ which iterates on the frame column names.
-            # This behavior may be changed in the future, you can follow the changes here:
-            # https://github.com/pandas-dev/pandas/issues/42403
-            value = value.columns
-            if len(value) == 1:
-                value = value[0]
+            if len(value.columns) != 1:
+                raise ValueError("Wrong number of items passed 2, placement implies 1")
+            value = value.squeeze(axis=1)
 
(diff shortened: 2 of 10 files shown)


modin/pandas/dataframe.py
```diff
@@ -1113,13 +1113,9 @@ def insert(self, loc, column, value, allow_duplicates=False):  # noqa: PR01, D20
...
         if isinstance(value, (DataFrame, pandas.DataFrame)):
-            # DataFrame considered as an array-like objects and so value have to be
-            # retrieved via DataFrame.__iter__ which iterates on the frame column names.
-            # This behavior may be changed in the future, you can follow the changes here:
-            # https://github.com/pandas-dev/pandas/issues/42403
-            value = value.columns
-            if len(value) == 1:
-                value = value[0]
+            if len(value.columns) != 1:
+                raise ValueError("Wrong number of items passed 2, placement implies 1")
```
vnlitvinov, line 1117:
```suggestion
                raise ValueError(f"Wrong number of items passed ({len(value.columns)}), placement implies 1")
```
or just delete that `2`

modin/pandas/test/dataframe/test_udf.py
```diff
@@ -160,11 +160,6 @@ def apply_func(series, y):
     )
 
 
-@pytest.mark.xfail(
-    reason="Modin's 'apply' produces DataFrame instead of Series and so insertion"
-    "of the apply result to the source frame is failed. Issue to track this bug:"
-    "https://github.com/modin-project/modin/issues/3219"
-)
```
vnlitvinov, line 167: if that is fixed why the issue is still open?

modin/backends/base/query_compiler.py
```diff
@@ -2066,7 +2066,7 @@ def inserter(df, loc, column, value):
             df.insert(loc, column, value)
             return df
 
-        return DataFrameDefault.register(inserter)(
+        return DataFrameDefault.register(inserter, inplace=True)(
```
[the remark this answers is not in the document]
YarShev: Since insert is an in-place operation we should return an object to which function was applied but not the result of the function. Even though we define inserter above we should always keep things explicit.

modin/pandas/test/dataframe/test_udf.py
YarShev: That is not fixed. We still have a problem described in #3219. This test checks kind of another thing.

vnlitvinov: Is enabling the test related to pandas migration? 😛

YarShev: Yep :)

YarShev: I didn't look into it.

vnlitvinov (approved):
LGTM

PR #3314, opened 2021-08-05
FIX-#3313: Fix `read_json` metadata definition

### LINKED ISSUE #3313: `read_json` defines metadata incorrectly
### System information
- **Modin version**: bded2e968a0057b1d448aec1957986fdc650d906
- **Python version**: 3.8.10
```
import pandas
import numpy as np
import os

import modin.pandas as pd
NPartitions.put(4)

filename = "test_json.json"
row_size = 256
col_size = 80

try:
    df = pandas.DataFrame(
        {f"col{x}": np.arange(row_size) for x in range(col_size)}
    )
    df.to_json(filename, lines=True, orient="records")

    df_pd = pd.read_json(filename, lines=True, orient="records")
    parts_width_cached = df_pd._query_compiler._modin_frame._column_widths_cache
    num_splits = len(df_pd._query_compiler._modin_frame._partitions[0])
    parts_width = [len(df_pd._query_compiler._modin_frame._partitions[0][i].get().columns) for i in range(num_splits)]

    print(parts_width_cached, parts_width) # [32, 32, 32, -16] [32, 32, 16, 0]

finally:
    os.remove(filename)
```

amyskov:

## What do these changes do?

- [x] passes `flake8 modin`
- [x] Resolves #3313  
- [x] tests added and passing

### DIFF
--- a/modin/engines/base/io/text/json_dispatcher.py
+++ b/modin/engines/base/io/text/json_dispatcher.py
@@ -75,12 +75,15 @@ class JSONDispatcher(TextFileDispatcher):
                 column_widths = [len(columns)]
                 num_splits = 1
             else:
-                column_widths = [
-                    column_chunksize
-                    if i != num_splits - 1
-                    else len(columns) - (column_chunksize * (num_splits - 1))
-                    for i in range(num_splits)
-                ]
+                column_widths = []
+                for i in range(num_splits):
+                    if len(columns) > (column_chunksize * i):
+                        if len(columns) > (column_chunksize * (i + 1)):
+                            column_widths.append(column_chunksize)
+                        else:
+                            column_widths.append(len(columns) - (column_chunksize * i))
+                    else:
+                        column_widths.append(0)
 
             args = {"fname": path_or_buf, "num_splits": num_splits, **kwargs}
 
--- a/modin/pandas/test/test_io.py
+++ b/modin/pandas/test/test_io.py
@@ -92,12 +92,12 @@ def setup_json_file(filename, row_size=NROWS, force=True):
         df.to_json(filename)
 
 
-def setup_json_lines_file(filename, row_size=NROWS, force=True):
+def setup_json_lines_file(filename, row_size=NROWS, col_size=2, force=True):
     if os.path.exists(filename) and not force:
         pass
     else:
         df = pandas.DataFrame(
-            {"col1": np.arange(row_size), "col2": np.arange(row_size)}
+            {f"col{x + 1}": np.arange(row_size) for x in range(col_size)}
         )
         df.to_json(filename, lines=True, orient="records")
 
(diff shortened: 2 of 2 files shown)


amyskov:
@YarShev , please, take a look at this PR, it's a subpart of https://github.com/modin-project/modin/pull/3286.

modin/engines/base/io/text/json_dispatcher.py
```diff
@@ -75,12 +75,15 @@ def _read(cls, path_or_buf, **kwargs):
                 column_widths = [len(columns)]
                 num_splits = 1
             else:
-                column_widths = [
-                    column_chunksize
-                    if i != num_splits - 1
-                    else len(columns) - (column_chunksize * (num_splits - 1))
-                    for i in range(num_splits)
-                ]
+                column_widths = []
```
YarShev, line 78: Can we use cls._define_metadata? This code is identical to the one that is in the method.

YarShev: Btw, we can probably rewrite _define_metadata as [here](https://github.com/modin-project/modin/blob/master/modin/engines/base/io/text/fwf_dispatcher.py#L166) and use it everywhere it is possible.

amyskov: Yes, we can use `_define_metadata` function in all text format dispatchers, but [plan](https://github.com/modin-project/modin/pull/3286#discussion_r682754751) is to fix json dispatcher first in this PR and then use `_define_metadata` for all text dispatchers in https://github.com/modin-project/modin/pull/3286.
Also version of definition of `column_widths ` in `_define_metadata` a little bit longer, but mush more readable, so i suggest don't change `_define_metadata` implementation)

YarShev: Ok, let's start using `_define_metadata` in #3286. As to the changes for `_define_metadata`, we should probably look at perf for the both approaches.

YarShev (approved):
@amyskov , LGTM, thanks!

modin/engines/base/io/text/json_dispatcher.py
amyskov: Not sure, that understood your point about `_define_metadata` changes - both `_define_metadata` implementations should deliver the same performance since these implementations differ only by formatting, see example below:
```
num_splits = 4
column_chunksize = 32
column_names = [f"col{x}" for x in range(80)]

# Current `_define_metadata` implementation
def func1(num_splits, column_chunksize, column_names):
    column_widths = []
    for i in range(num_splits):
        if len(column_names) > (column_chunksize * i):
            if len(column_names) > (column_chunksize * (i + 1)):
                column_widths.append(column_chunksize)
            else:
                column_widths.append(len(column_names) - (column_chunksize * i))
        else:
            column_widths.append(0)

    return column_widths

# `fwf_dispatcher` implementation
def func2(num_splits, column_chunksize, column_names):
    column_widths = [
        column_chunksize
        if len(column_names) > (column_chunksize * (i + 1))
        else 0
        if len(column_names) < (column_chunksize * i)
        else len(column_names) - (column_chunksize * i)
        for i in range(num_splits)
    ]

    return column_widths

ans1 = func1(num_splits, column_chunksize, column_names) # [32, 32, 16, 0]
ans2 = func2(num_splits, column_chunksize, column_names) # [32, 32, 16, 0]

assert ans1 == ans2
```

YarShev: https://www.linkedin.com/pulse/list-comprehension-python-always-faster-than-alex-falkovskiy/

amyskov: I see, but i believe we can neglect this effect in this case, i have measured `func1` and `func2` (from comparison above) with `len(column_names) == 10_000`, `num_splits == 300` and `column_chunksize == 32` and obtained 0.6 and 0.3 microseconds execution time for `func1` and `func2` respectively, that is much lower than other Modin data import overhead.

YarShev: I understand that is much lower than other Modin data import overhead. However, if we don't care of such a kind of things, we can get significant overhead summarized out of each small slowdown. Moreover, it happens in the driver process. So I wouldn't neglect it.

YarShev: We can always leave a comment clarifying the code.

amyskov: Ok, let's use list comprehension here.

PR #3337, opened 2021-08-10
FEAT-#3336: support nunique aggregate in OmniSci backend

### LINKED ISSUE #3336: [OmniSci] Support 'nunique' aggregate
We need to support aggregation like `df.groupby('col1').agg('col2' : 'nunique')`

ienkovich:
## What do these changes do?
Support `nunique` aggregate in `groupby.agg` method.

- [x] passes `flake8 modin`
- [x] Resolves #3336 
- [x] tests added and passing

### DIFF
--- a/modin/experimental/backends/omnisci/query_compiler.py
+++ b/modin/experimental/backends/omnisci/query_compiler.py
@@ -359,7 +359,7 @@ class DFAlgQueryCompiler(BaseQueryCompiler):
             raise NotImplementedError(
                 f"OmniSci's nunique does not support such set of parameters: axis={axis}, dropna={dropna}."
             )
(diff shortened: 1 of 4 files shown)


YarShev (commented):
@ienkovich , can you explain what is the difference between previous and current implementation? It seems to me they are identical but implemented on different layers.

ienkovich:
> @ienkovich , can you explain what is the difference between previous and current implementation? It seems to me they are identical but implemented on different layers.

I prefer to make the translation of `nunique` to `distinct count` on a lower level to avoid multiple places where we have to make such a translation. With this approach, we don't need additional changes to have `nunique` aggregation in `groupby.agg`.

YarShev (approved):

Issue #3341, opened 2021-08-16
Expansion DMatrix


Rubtsowa:
Need add new parameters and functions in class DMatrix

PR #3344, opened 2021-08-16
FEAT-#3341: Expansion DMatrix

Rubtsowa:

## What do these changes do?

- [ ] passes `flake8 modin`
- [ ] Resolves #? 
- [ ] tests added and passing


prutskov:
@Rubtsowa I don't see the practical usage of added parameters in internal implementation of training/prediction for Ray engine. Is it in progress? As example of parameters you can see on [Dask implementation](https://github.com/dmlc/xgboost/blob/master/python-package/xgboost/dask.py) of DMatrix

PR #3397, opened 2021-09-02
FIX-#3395: Make `label` param of DMatrix optional

### LINKED ISSUE #3395: Make parameter `label` of Modin DMatrix optional
### Describe the problem

To support drop-in-replacement API for DMatrix

prutskov:

## What do these changes do?

- [x] passes `flake8 modin`
- [x] Resolves #3395  
- [ ] tests added and passing

### DIFF
--- a/modin/experimental/xgboost/xgboost.py
+++ b/modin/experimental/xgboost/xgboost.py
@@ -35,7 +35,7 @@ class DMatrix:
     ----------
     data : modin.pandas.DataFrame
         Data source of DMatrix.
-    label : modin.pandas.DataFrame or modin.pandas.Series
+    label : modin.pandas.DataFrame or modin.pandas.Series, optional
         Labels used for training.
 
     Notes
@@ -43,16 +43,20 @@ class DMatrix:
     Currently DMatrix supports only `data` and `label` parameters.
     """
 
-    def __init__(self, data, label):
+    def __init__(self, data, label=None):
         assert isinstance(
             data, pd.DataFrame
         ), f"Type of `data` is {type(data)}, but expected {pd.DataFrame}."
-        assert isinstance(
-            label, (pd.DataFrame, pd.Series)
-        ), f"Type of `data` is {type(label)}, but expected {pd.DataFrame} or {pd.Series}."
+
+        if label is not None:
+            assert isinstance(
+                label, (pd.DataFrame, pd.Series)
+            ), f"Type of `data` is {type(label)}, but expected {pd.DataFrame} or {pd.Series}."
+            self.label = unwrap_partitions(label, axis=0)
+        else:
+            self.label = None
 
         self.data = unwrap_partitions(data, axis=0, get_ip=True)
-        self.label = unwrap_partitions(label, axis=0)
 
         self.metadata = (
             data.index,


modin/experimental/xgboost/xgboost.py
```diff
@@ -25,7 +25,7 @@
 LOGGER = logging.getLogger("[modin.xgboost]")
 
 
-class DMatrix(xgb.DMatrix):
```
anmyachev, line 28: @prutskov why don't you inherit anymore?

prutskov: Inheritance doesn't include any benefits, because Modin DMatrix provided very limited set of supported properties/methods. Also, inheritance includes additional problems for `doc_checker` for `labels` parameter of `__init__` dunder method.

anmyachev: ок

anmyachev (approved):

anmyachev (approved):

modin/experimental/xgboost/xgboost.py
anmyachev, line 28: Please make this change in a separate pull request, as it does not apply to the current one.

anmyachev (changes requested):

modin/experimental/xgboost/xgboost.py
prutskov: If I leave only changes related to fix issue, the new issue will occur - `doc_checker` will found errors for `label` parameter, because in that case docstring from native `xgb.DMatrix` (has differences from our style) will be used. Does it make sense to move inheritance to a separate RR?

anmyachev: In this case, `doc_checker` should have found errors before this pull request. How did it work before?

prutskov: In my assumption, inheritance of docstrings happens only in case `label=None` (when we aligned with parent class). Before it was positional parameter with own docstring.

anmyachev: I think we should fix `doc_checker` in this case as users might rely on inheritance.

prutskov: The root cause of issue is in incorrect docstring of `__init__` method in parent class (`xgb.DMatrix`) for `label` parameter. In our code we skip the check of correctness of inherited docstrings with decorator `_inherit_docstrings`. I don't see the point to add this decorator to the method `__init__`, because we don't require separate docstrings for `__init__` methods in our docstyle. So, I just added `noqa` to skip the check of doc here.

anmyachev: If the documentation does inherit for `__init__`, then it's a problem because it [doesn't match](https://xgboost.readthedocs.io/en/latest/python/python_api.html#xgboost.DMatrix) the function signature. We need to delete it.

prutskov: I added decorator `_remove_docstring`, that removes docstring for object (for `__init__` method in our case).

modin/experimental/xgboost/xgboost.py
```diff
@@ -25,6 +25,23 @@
...
+    Parameters
+    ----------
+    obj : any
+        Object from which need to remove docstring.
+
+    Returns
+    -------
+    `obj` without docstring.
+    """
+    setattr(obj, "__doc__", "")
```
anmyachev, line 41:
```suggestion
    setattr(obj, "__doc__", None)
```

anmyachev (dismissed):
@vnlitvinov take a look please.

modin/experimental/xgboost/xgboost.py
prutskov: If we will set `__doc__ to `None`, inherited docstring will be used.

anmyachev (approved):

PR #3398, opened 2021-09-02
FEAT-#2006: support local variables in query and eval

### LINKED ISSUE #2006: Local variables not yet supported in eval.
I encounter this message when trying to use modin 
Local variables not yet supported in eval.
To request implementation, send an email to feature_requests@modin.org

here's a code 
Screenshot (314)

anmyachev:

## What do these changes do?

- [x] passes `flake8 modin`
- [x] Resolves #2006 
- [x] tests added and passing

### DIFF
--- a/modin/pandas/dataframe.py
+++ b/modin/pandas/dataframe.py
@@ -1672,11 +1672,41 @@ class DataFrame(BasePandasDataset):
     product = prod
     radd = add
 
+    def _update_var_dicts_in_kwargs(self, expr, **kwargs):
+        import re
+
+        frame = sys._getframe()
+        try:
+            f_locals = frame.f_back.f_back.f_locals
+            f_globals = frame.f_back.f_back.f_globals
+        finally:
+            del frame
+        local_names = re.findall(r"@([\w]*)", expr)
+        local_dict = {}
+        global_dict = {}
+
+        for name in local_names:
+            if name in f_locals:
+                local_dict[name] = f_locals[name]
+            elif name in f_globals:
+                global_dict[name] = f_globals[name]
+
+        if local_dict:
+            if kwargs.get("local_dict"):
+                local_dict.update(kwargs["local_dict"])
+            kwargs["local_dict"] = local_dict
+        if global_dict:
+            if kwargs.get("global_dict"):
+                global_dict.update(kwargs.get("global_dict"))
+            kwargs["global_dict"] = global_dict
+        return kwargs
+
     def query(self, expr, inplace=False, **kwargs):  # noqa: PR01, RT01, D200
         """
         Query the columns of a ``DataFrame`` with a boolean expression.
         """
         ErrorMessage.non_verified_udf()
+        kwargs = self._update_var_dicts_in_kwargs(expr, **kwargs)
         self._validate_eval_query(expr, **kwargs)
         inplace = validate_bool_kwarg(inplace, "inplace")
         new_query_compiler = self._query_compiler.query(expr, **kwargs)
@@ -2854,9 +2884,6 @@ class DataFrame(BasePandasDataset):
         """
         if isinstance(expr, str) and expr == "":
             raise ValueError("expr cannot be an empty string")
-
-        if isinstance(expr, str) and "@" in expr:
-            ErrorMessage.not_implemented("Local variables not yet supported in eval.")
 
         if isinstance(expr, str) and "not" in expr:
             if "parser" in kwargs and kwargs["parser"] == "python":
--- a/modin/pandas/dataframe.py
+++ b/modin/pandas/dataframe.py
@@ -26,6 +26,7 @@ from pandas.io.formats.printing import pprint_thing
 from pandas._libs.lib import no_default
 from pandas._typing import StorageOptions
 
+import re
 import itertools
 import functools
 import numpy as np
@@ -821,7 +822,7 @@ class DataFrame(BasePandasDataset):
 
     def _update_var_dicts_in_kwargs(self, expr, **kwargs):
         """
-        Copy variables with "@" prefix in `local_dict` and `global_dict`.
+        Copy variables with "@" prefix in `local_dict` and `global_dict` keys of kwargs.
 
         Parameters
         ----------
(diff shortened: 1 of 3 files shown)


modin/pandas/test/dataframe/test_udf.py
```diff
@@ -305,6 +299,21 @@ def test_query(data, funcs):
         df_equals(modin_result, pandas_result)
 
 
+@pytest.mark.parametrize("data", test_data_values, ids=test_data_keys)
+def test_query_with_local_vars(data):
+    modin_df = pd.DataFrame(data)
+    pandas_df = pandas.DataFrame(data)
+    x = 2  # noqa F841
```
gshimansky, line 306: There is no test for a global variable, is there? What about a variable that is a function argument?

anmyachev: Test with global variable is added.

Can you give me example when variable is a function argument? I'm not sure that I understand this.

gshimansky: Something like this:
```python
def func(arg):
    modin_df.query("@arg")
```

modin/pandas/dataframe.py
```diff
@@ -819,12 +819,57 @@ def explode(
...
+        """
+        import re
+
+        frame = sys._getframe()
+        try:
+            f_locals = frame.f_back.f_back.f_locals
+            f_globals = frame.f_back.f_back.f_globals
+        finally:
+            del frame
+        local_names = re.findall(r"@([\w]*)", expr)
```
vnlitvinov, line 846:
```suggestion
        local_names = set(re.findall(r"@([\w]+)", expr))
```
to remove duplicates to speed up iteration and to require at least one symbol in the name :) (otherwise it would pick empty string in e.g. `" hey @ someone!"`)

modin/pandas/dataframe.py
```diff
@@ -819,12 +819,57 @@ def explode(
...
+        local_dict = {}
+        global_dict = {}
+
+        for name in local_names:
+            if name in f_locals:
+                local_dict[name] = f_locals[name]
+            elif name in f_globals:
+                global_dict[name] = f_globals[name]
```
vnlitvinov, line 854:
```suggestion
            for dct_out, dct_in in ((local_dict, f_locals), (global_dict, f_globals)):
                try:
                    dct_out[name] = dct_in[name]
                except KeyError:
                    pass
```
this works a little faster (hopefully; if `for` loop slows stuff down one could always unwind it to two separate `try..except` blocks) and supports the case of same variable being present in both locals and globals (not that it matters much in this case TBH, but still is _more_ technically correct)...

modin/pandas/dataframe.py
```diff
@@ -819,12 +819,57 @@ def explode(
...
+
+        if local_dict:
+            if kwargs.get("local_dict"):
+                local_dict.update(kwargs["local_dict"])
+            kwargs["local_dict"] = local_dict
+        if global_dict:
+            if kwargs.get("global_dict"):
+                global_dict.update(kwargs.get("global_dict"))
```
vnlitvinov, line 862: Squash the ifs like this:
```suggestion
        if local_dict and kwargs.get("local_dict"):
            local_dict.update(kwargs["local_dict"])
            kwargs["local_dict"] = local_dict
        if global_dict and kwargs.get("global_dict"):
            global_dict.update(kwargs["global_dict"])
```

modin/pandas/dataframe.py
```diff
@@ -819,12 +819,57 @@ def explode(
...
+            kwargs["global_dict"] = global_dict
+        return kwargs
```
vnlitvinov, line 864: you're modifying `kwargs` in-place, what's the point of returning those?..

modin/pandas/test/dataframe/test_udf.py
```diff
@@ -305,6 +310,24 @@ def test_query(data, funcs):
         df_equals(modin_result, pandas_result)
 
 
+TEST_VAR = 2
```
vnlitvinov, line 313: you have this defined at line 230, what's the point of re-defining?..

modin/pandas/test/dataframe/test_udf.py
```diff
@@ -305,6 +310,24 @@ def test_query(data, funcs):
...
+
+
+@pytest.mark.parametrize("data", test_data_values, ids=test_data_keys)
+@pytest.mark.parametrize("local_var", [2])
+def test_query_with_local_and_global_var(data, local_var):
```
vnlitvinov, line 318: this looks exactly the same as `test_eval_with_local_and_global_var` except the method called on the dataframe... could we collapse those in a single test and just parametrize by the method name? or at least put those next to each other.

modin/pandas/dataframe.py
```diff
@@ -819,12 +819,57 @@ def explode(
...
+            The expression string to search variables with "@" prefix.
+        **kwargs : dict
+            See the documentation for eval() for complete details on the keyword arguments accepted by query().
+
+        Returns
+        -------
+        kwargs : dict
+            Dict with updated `local_dict` and `global_dict` keys.
+        """
+        import re
```
vnlitvinov, line 838: Imports should be done in the top of the file unless there's very good reason not to.
They implicitly acquire import lock which could cause a deadlock in some very funny code, plus it might confuse some IDEs for completion etc.

modin/pandas/dataframe.py
```diff
@@ -819,12 +819,57 @@ def explode(
...
+
+        frame = sys._getframe()
```
vnlitvinov, line 840:
```suggestion
        if "@" not in expr:
            return kwargs
        frame = sys._getframe()
```
let's insert some quick check for a hot path (I would expect that _most_ queries do not reference local variables).

modin/pandas/dataframe.py
anmyachev: In case when `kwargs.get("local_dict")==None` and `local_dict=={"local_var": 10}` we should make the assignment: `kwargs["local_dict"] = local_dict` but with these changes it won't.

modin/pandas/dataframe.py
anmyachev: left over from previous approaches

modin/pandas/test/dataframe/test_udf.py
anmyachev: removed

modin/pandas/dataframe.py
vnlitvinov: Right... in this case I suggest to change it to
```python
        if local_dict:
            local_dict.update(kwargs.get("local_dict") or {})
            kwargs["local_dict"] = local_dict
        if global_dict:
            global_dict.update(kwargs.get("global_dict") or {})
```
(note the `or {}` part - this trick pushes the creation of an empty dictionary to the moment in time when a key is missing in `kwargs` - a lazier version of `.get(key, {})`)

modin/pandas/dataframe.py
```diff
@@ -861,15 +858,15 @@ def _update_var_dicts_in_kwargs(self, expr, **kwargs):
             if kwargs.get("global_dict"):
                 global_dict.update(kwargs.get("global_dict"))
             kwargs["global_dict"] = global_dict
-        return kwargs
+        return
```
vnlitvinov: just delete the statement, a lone `return` in the end of function is not necessary :)

vnlitvinov (approved):
LGTM

vnlitvinov (approved):
LGTM

PR #3408, opened 2021-09-08
FIX-#3407: change type of series shape

### LINKED ISSUE #3407: ASV: all shapes should be of list type
There are two ways to define shapes, by default through a dictionary or through a `json` file. Due to the fact that the `json` format does not support the use of `tuples`, we have to use the `list` type. But since in both cases the hash from the benchmark source code is considered and it must be the same, then the `list` type must also be used in dictionaries.

Here the shape are of `tuple` type: https://github.com/modin-project/modin/blob/600fa2632973167176333cb03c8a8c3b278be4ca/asv_bench/benchmarks/utils/data_shapes.py#L50
https://github.com/modin-project/modin/blob/600fa2632973167176333cb03c8a8c3b278be4ca/asv_bench/benchmarks/utils/data_shapes.py#L53

anmyachev:

## What do these changes do?

- [ ] passes `flake8 modin`
- [x] Resolves #3407 
- [ ] tests added and passing


vnlitvinov (commented):
Code changes are good... but the original issue lacks the _why_ explanation - what's wrong with shapes being tuples?.. Tuples consume less memory so are IMHO preferred whenever possible...

anmyachev:
> Code changes are good... but the original issue lacks the _why_ explanation - what's wrong with shapes being tuples?.. Tuples consume less memory so are IMHO preferred whenever possible...

The issue is updated with more details https://github.com/modin-project/modin/issues/3407#issue-991040871.

vnlitvinov (approved):
LGTM

PR #3416, opened 2021-09-09
FEAT-#2301: Add `storage_options` param for `read_parquet`

### LINKED ISSUE #2301: Add support of parameter `storage_options` in functions `modin.pandas.read_*`
### Describe the problem

When pandas will release `storage_options` for `read_*` functions (which are currently supported in [master](https://github.com/pandas-dev/pandas/tree/4aa41b8fee4f4d5095d12bfbf644495df16c53d5)) we should include this feature in modin.

prutskov:

## What do these changes do?

- [ ] passes `flake8 modin`
- [ ] Resolves #2301 
- [ ] tests added and passing


modin/engines/base/io/column_stores/parquet_dispatcher.py
```diff
@@ -79,58 +78,22 @@ def _read(cls, path, engine, columns, **kwargs):
...
-                except NoCredentialsError:
-                    fs = s3fs.S3FileSystem(anon=True)
-                    pd = ParquetDataset(path, filesystem=fs)
-                meta = pd.metadata
-                column_names = pd.schema.to_arrow_schema().names
-            else:
-                meta = ParquetFile(path).metadata
-                column_names = meta.schema.to_arrow_schema().names
+            fs, path_ = (
+                fsspec.core.url_to_fs(path, **(kwargs["storage_options"] or {}))
```
vnlitvinov, line 87:
```suggestion
                fsspec.core.url_to_fs(path, **(kwargs.get("storage_options") or {}))
```

modin/engines/base/io/column_stores/parquet_dispatcher.py
```diff
@@ -79,58 +78,22 @@ def _read(cls, path, engine, columns, **kwargs):
...
-                    # the pyarrow storage has no concept of row labels/index.
-                    # This ensures that our metadata lines up with the partitions without
-                    # extra communication steps once we have done all the remote
-                    # computation.
-                    index_columns = json.loads(pandas_metadata.decode("utf8")).get(
-                        "index_columns", []
-                    )
-                    column_names = [c for c in column_names if c not in index_columns]
+            if dataset.schema.pandas_metadata is not None:
+                index_columns = dataset.schema.pandas_metadata.get("index_columns", [])
```
vnlitvinov, line 96:
```suggestion
                index_columns = set(dataset.schema.pandas_metadata.get("index_columns", []))
```
to make next line a trifle faster :)

vnlitvinov:
@prutskov I don't think this PR resolves #2301 completely, as it only adds `storage_options` to `read_parquet`, not to all `read_*` methods. Please add a separate tracker and fix the PR description.

prutskov:
> @prutskov I don't think this PR resolves #2301 completely, as it only adds `storage_options` to `read_parquet`, not to all `read_*` methods. Please add a separate tracker and fix the PR description.

@vnlitvinov  Looks like, the only one of `read_*` functions which is needed in additional fixes for support `storage_options` is `read_parquet`. The other functions already support this parameter in `kwargs`. But it would be useful to have tests on `storage_options` for other functions in this PR. What do you think?

modin/engines/base/io/column_stores/parquet_dispatcher.py
prutskov: What's benefit in this case?

anmyachev:
> > @prutskov I don't think this PR resolves #2301 completely, as it only adds `storage_options` to `read_parquet`, not to all `read_*` methods. Please add a separate tracker and fix the PR description.
> 
> @vnlitvinov Looks like, the only one of `read_*` functions which is needed in additional fixes for support `storage_options` is `read_parquet`. The other functions already support this parameter in `kwargs`. But it would be useful to have tests on `storage_options` for other functions in this PR. What do you think?

@prutskov please add these tests.

modin/engines/base/io/column_stores/parquet_dispatcher.py
anmyachev: Everything is the same, except that no exceptions will be thrown if `storage_options` key is not present.

modin/engines/base/io/column_stores/parquet_dispatcher.py
prutskov: `dataset.schema.pandas_metadata.get("index_columns", [])` is `dict` or `list`. In case `dict` we can't use that, because `dict` is unhashable.

prutskov:
@anmyachev, @vnlitvinov This PR will only fix issues for `read_parquet`

anmyachev (approved):

modin/engines/base/io/column_stores/parquet_dispatcher.py
vnlitvinov: Uhm... :) unhashable means something if you want to use it as an _element_, but here it's being `converted` from a dictionary to a set. Works just fine:
```
>>> set({'a':1,'b':2})
{'b', 'a'}
```

vnlitvinov (approved):
LGTM, the other suggestion is a nitpick anyway.

modin/engines/base/io/column_stores/parquet_dispatcher.py
prutskov: I introduced some misunderstanding. The result of `dataset.schema.pandas_metadata.get("index_columns", [])` is:
1. `list` of values. For example `['col0', 'col1']`. We can easy create `set` from this list.
2. `list` of dicts. For example `[{'kind': 'range', 'name': None, 'start': 0, 'stop': 256, 'step': 1}]`. In that case we cann't create set:
```python
l = [{'a': 1}]
s = set(l) # <-- TypeError: unhashable type: 'dict'
```

PR #3419, opened 2021-09-09
FIX-#3417: Fix read_csv with `skiprows` and `header` parameters

### LINKED ISSUE #3417: `read_csv` fails with `skiprows` and `header` parameters passed
### System information
- **Modin version** (`modin.__version__`): 203b9a59d2be1622241c2e0eaa949a42dc0f99d8
- **Python version**: 3.8.10
```
import os
import modin.pandas as pd
import pandas

NPartitions.put(4)

test_filename = "test.csv"
data = """col1,col2
1,2
3,4
5,6
"""

skiprows = [2]

try :
    with open(test_filename, "w") as f:
        f.write(data)
    
    df_pandas = pandas.read_csv(test_filename, skiprows=[0, 2], header=0)
    df_pd = pd.read_csv(test_filename, skiprows=[0, 2], header=0)

    df_equals(df_pd, df_pandas)

finally:
    os.remove(test_filename)
```

### Describe the problem

Errors occurs if `skiprows` and `header` parameters have intersections.

### Source code / logs

```
AssertionError: DataFrame.iloc[:, 0] (column name="1") are different

DataFrame.iloc[:, 0] (column name="1") values are different (100.0 %)
[index]: [0]
[left]:  [1]
[right]: [5]
```

amyskov:

## What do these changes do?
PR fixes corner cases of `read_csv` function with `skiprows` and `header` parameters, which values have intersections. Also manipulations with `skiprows` parameter were localized in `_manage_skiprows_parameter` function.

- [x] passes `flake8 modin`
- [x] Resolves #3417  
- [x] tests added and passing


amyskov:
PR is ready for review.

Changes performance measurements (no performance degradation was obtained):
```
     [0a642753]       [4c747de1]
     <master>         <fix_read_csv_issue>
         316±10ms          319±6ms     1.01  io.csv.TimeReadCsvSkiprows.time_skiprows([1000000, 10], 'lambda_even_rows')
         478±20ms         475±10ms     0.99  io.csv.TimeReadCsvSkiprows.time_skiprows([1000000, 10], 'range_step2')
         140±10ms         139±10ms     1.00  io.csv.TimeReadCsvSkiprows.time_skiprows([1000000, 10], 'range_uniform')
         143±10ms         142±10ms     0.99  io.csv.TimeReadCsvSkiprows.time_skiprows([1000000, 10], None)
         933±30ms         927±40ms     0.99  io.csv.TimeReadCsvSkiprows.time_skiprows([5000, 5000], 'lambda_even_rows')
         948±40ms         918±10ms     0.97  io.csv.TimeReadCsvSkiprows.time_skiprows([5000, 5000], 'range_step2')
         924±30ms         942±40ms     1.02  io.csv.TimeReadCsvSkiprows.time_skiprows([5000, 5000], 'range_uniform')
         907±20ms         913±10ms     1.01  io.csv.TimeReadCsvSkiprows.time_skiprows([5000, 5000], None)
```

anmyachev:
@amyskov if we need so much code for each edge case, then the code base for IO functions will be very difficult to maintain.
Maybe look for other ways of implementation?

amyskov:
> @amyskov if we need so much code for each edge case, then the code base for IO functions will be very difficult to maintain.
> Maybe look for other ways of implementation?

You are right, such edge cases bring logic complication. These edge cases occurs because of complex interconnections of `pandas.read_csv` parameters and parallelism brought by Modin, not sure we can find significantly better solution.

anmyachev:
What if we delete rows after the dataframe is created?

amyskov:
> What if we delete rows after the dataframe is created?

We can do that, but we still should know what to skip. See pandas example below:
```
import pandas
from io import StringIO

data = """0,0
1,1
2,2
3,3
4,4
5,5
6,6
7,7
8,8
"""

df = pandas.read_csv(StringIO(data), skiprows=[0,2,4], header=1)

print(df)
```
Output:
```
   3  3.1
0  5    5
1  6    6
2  7    7
3  8    8
```
In this case rows `[0,2,4]` were skipped and than the second remaining row was selected as header (the first remaining row was skipped). If we want to get the same result in Modin we need to skip the first five rows (header inferred separately). So some pandas behavior emulation should be done (since we can't just use unchanged `skiprows` parameter) and i think it should be done in the `_manage_skiprows_parameter` function. Does this make sense?

amyskov:
Hi @modin-project/modin-core please take a look at this PR.

modin/engines/base/io/text/csv_dispatcher.py
```diff
@@ -331,21 +331,25 @@ def _get_new_qc(
...
                 )
             elif callable(skiprows_md):
-                mod_index = skiprows_md(index_range + header_size)
+                try:
+                    # dicrect `skiprows_md` is more efficient than using of
+                    # map method, but in some cases it can work incorrectly, e.g.
+                    # when `skiprows_md` contains `in` operator
+                    mod_index = skiprows_md(index_range)
+                    assert is_list_like(mod_index)
+                except Exception:
```
devin-petersohn, line 343: Why catch all exception types? This is not ideal

modin/engines/base/io/text/csv_dispatcher.py
```diff
@@ -376,33 +380,113 @@ def _manage_skiprows_parameter(
...
+            Original skiprows parameter of read_csv function.
         header_size : int, default: 0
-                Number of rows that are used by header.
+            Number of rows that are used by header.
 
         Returns
         -------
-        skiprows : int, array or callable
-                Updated skiprows parameter.
+        skiprows_md : int, array or callable
```
devin-petersohn, line 397: What is `md`?

modin/engines/base/io/text/csv_dispatcher.py
amyskov: Ok, exception types were clarified.

modin/engines/base/io/text/csv_dispatcher.py
amyskov: `md` stands for Modin. In the context of this and `csv_dispatcher._read` functions `skiprows` is original `pandas.read_csv` `skiprows` parameter, while `skiprows_md` is parameter version modified for Modin specific purposes (sorting of list-like parameter, etc.)

devin-petersohn (commented):
I tend to agree with @anmyachev about the complexity of the code.

pandas `read_csv` has a lot of edge cases and parameter combinations. It would be good to try to solve more generally or in a more compact and maintainable way. 

I know this is blocking other issues, I am ok to proceed with this one, but adding 100+ lines for each combination of parameters will make the code unmaintainable.

amyskov:
> I tend to agree with @anmyachev about the complexity of the code.
> 
> pandas `read_csv` has a lot of edge cases and parameter combinations. It would be good to try to solve more generally or in a more compact and maintainable way.
> 
> I know this is blocking other issues, I am ok to proceed with this one, but adding 100+ lines for each combination of parameters will make the code unmaintainable.

I'm proposing to use pandas implementation in this corner case (i suppose it is really rare corner case). More general implementation are possible, but i'm not sure it will be effective in all `skiprows/header` combinations.
Also this PR contains some refactoring code, so it can be moved into separate. What do you think?

amyskov:
> > I tend to agree with @anmyachev about the complexity of the code.
> > pandas `read_csv` has a lot of edge cases and parameter combinations. It would be good to try to solve more generally or in a more compact and maintainable way.
> > I know this is blocking other issues, I am ok to proceed with this one, but adding 100+ lines for each combination of parameters will make the code unmaintainable.
> 
> I'm proposing to use pandas implementation in this corner case (i suppose it is really rare corner case). More general implementation are possible, but i'm not sure it will be effective in all `skiprows/header` combinations.
> Also this PR contains some refactoring code, so it can be moved into separate. What do you think?

@anmyachev @devin-petersohn what do you think about the new proposal?

anmyachev:
> > > I tend to agree with @anmyachev about the complexity of the code.
> > > pandas `read_csv` has a lot of edge cases and parameter combinations. It would be good to try to solve more generally or in a more compact and maintainable way.
> > > I know this is blocking other issues, I am ok to proceed with this one, but adding 100+ lines for each combination of parameters will make the code unmaintainable.
> > 
> > 
> > I'm proposing to use pandas implementation in this corner case (i suppose it is really rare corner case). More general implementation are possible, but i'm not sure it will be effective in all `skiprows/header` combinations.
> > Also this PR contains some refactoring code, so it can be moved into separate. What do you think?
> 
> @anmyachev @devin-petersohn what do you think about the new proposal?

Yeh, let's default to pandas in the case until we found new maintainable way of support different Pandas's corner cases.

Also please move refactoring code in separate PR.

devin-petersohn (commented):
@amyskov Please also clear conflict in addition to @anmyachev comments. Thanks!

amyskov:
> @amyskov Please also clear conflict in addition to @anmyachev comments. Thanks!

@devin-petersohn we can detach another one [PR](https://github.com/modin-project/modin/pull/3511), please take a look.

amyskov:
@anmyachev @devin-petersohn blocking issue was fixed, so this PR is ready for review now.

devin-petersohn (commented):
It seems a little strange to add another return value here `need_fallback_impl` instead of creating a utility that checks for the condition. Is there a reason it needs to be added here?

modin/core/io/text/text_file_dispatcher.py
```diff
@@ -759,14 +785,19 @@ def _manage_skiprows_parameter(
...
+
         elif callable(skiprows):
 
             def skiprows_func(x):
                 return skiprows(x + header_size)
 
             skiprows_md = skiprows_func
+        elif skiprows is not None:
+            raise TypeError(
+                f"Not acceptable type of `skiprows` parameter: {type(skiprows)}"
```
anmyachev, line 799: The error message is equal to pandas?

modin/core/io/text/text_file_dispatcher.py
```diff
@@ -641,6 +649,24 @@ def check_parameters_support(
...
+            is_list_like(skiprows_md) and skiprows_md[0] < header_size
+        ) or (
+            callable(skiprows)
+            and any(
+                map(
+                    bool,
+                    cls._get_skip_mask(pandas.RangeIndex(header_size), skiprows),
+                )
+            )
+        )
```
anmyachev, line 662: This part of the condition is hard to read. You can, for example, save the result of `any(...)` function to a variable.

modin/core/io/text/text_file_dispatcher.py
```diff
@@ -641,6 +649,24 @@ def check_parameters_support(
         if read_kwargs["chunksize"] is not None:
             return False
 
+        skiprows_unsupported = (
+            is_list_like(skiprows_md) and skiprows_md[0] < header_size
+        ) or (
+            callable(skiprows)
```
anmyachev, line 655: Why are you use `skiprows` and `skiprowd_md` variables at the same time?

modin/core/io/text/text_file_dispatcher.py
amyskov: Ok, code fragment was rewritten.

modin/core/io/text/text_file_dispatcher.py
amyskov: This is done in that way because in the case of list-like `skiprows` we need already sorted array to get the smallest element (the first element in the `skiprows_md` array) but in the case of callable `skiprows` we can't get `skiprows_md` since it's value already shifted on `header_size` value and we can't get intersection information from it (so we need unshifted `skiprows` callable).

modin/core/io/text/text_file_dispatcher.py
```diff
@@ -641,6 +649,24 @@ def check_parameters_support(
         if read_kwargs["chunksize"] is not None:
             return False
 
+        skiprows_unsupported = False
```
anmyachev, line 652: Double negatives are hard to read.

```suggestion
        skiprows_supported = True
```

anmyachev (approved):

amyskov:
@anmyachev @devin-petersohn do you have some additional comments?

anmyachev (dismissed):

modin/core/io/text/text_file_dispatcher.py
```diff
@@ -759,8 +785,9 @@ def _manage_skiprows_parameter(
                 )
                 skiprows_partitioning = len(skiprows_md)
                 skiprows_md = 0
-            else:
+            elif skiprows_md[0] > header_size:
                 skiprows_md = skiprows_md - header_size
+
```
devin-petersohn, line 790:
```suggestion
```

modin/core/io/text/text_file_dispatcher.py
```diff
@@ -641,6 +649,24 @@ def check_parameters_support(
...
+            is_intersection = any(
+                cls._get_skip_mask(pandas.RangeIndex(header_size), skiprows)
+            )
+            if is_intersection:
+                skiprows_supported = False
+
+        if not skiprows_supported:
+            ErrorMessage.single_warning(
+                "Values of `header` and `skiprows` parameters have intersections. "
+                "This case doesn't supported by Modin, so pandas implementation will be used!"
```
devin-petersohn, line 666:
```suggestion
                "This case is unsupported by Modin, so pandas implementation will be used"
```

YarShev:
@amyskov, I think we can merge this PR after @devin-petersohn's comments are addressed and conflicts are resolved.

YarShev (approved):
LGTM

YarShev:
@amyskov, please update release notes.

amyskov:
> @amyskov, please update release notes.

Right. Release notes doc was updated.

YarShev (approved):

PR #3434, opened 2021-09-16
REFACTOR-#3176: Use ray from conda instead of pip

### LINKED ISSUE #3176: Move testing back to Ray 1.4.0 using conda-forge
Ray 1.4.0 is now [available in conda-forge](https://anaconda.org/conda-forge/ray-core/files?version=1.4.0), so we should switch testing back to that channel from pip for consistency.

vnlitvinov:

## What do these changes do?

- [x] passes `flake8 modin`
- [x] Resolves #3176
- [x] tests added and passing

### DIFF
--- a/environment-dev.yml
+++ b/environment-dev.yml
@@ -31,6 +31,7 @@ dependencies:
   - packaging
   - coverage<5.0
   - pygithub==1.53
+  - ray-default
   - rpyc==4.1.5
   - cloudpickle
   - boto3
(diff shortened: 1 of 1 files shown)


devin-petersohn (commented):
The main issue I see here is that Ray from pip is how most users will install. I prefer to develop and test the code how users will run/install it.

vnlitvinov:
We're installing `modin` by `conda` anyway. `conda` itself recommends to **not** mix it and `pip` :)
That said, I don't think users would install `modin` from conda and `ray` from pip - most likely they would install both from the same source.

devin-petersohn (approved):
Agree, should not mix both. This is fine, then.

PR #3288, opened 2021-07-27
FIX-#3278: `read_csv` with heterogeneous data

amyskov:
## What do these changes do?

- [ ] commit message follows format outlined [here](https://modin.readthedocs.io/en/latest/contributing.html)
- [ ] passes `flake8 modin`
- [ ] passes `black --check modin`
- [ ] signed commit with `git commit -s` 
- [ ] Resolves #3278  
- [ ] tests added and passing
- [ ] module layout described at `docs/developer/architecture.rst` is up-to-date

PR #3438, opened 2021-09-17
FIX-#3437: Warn user about heterogeneous data presence during read_csv call and propose workaround

### LINKED ISSUE #3437: Warn user about heterogeneous data presence during `read_csv` call and propose workaround
Subpart of https://github.com/modin-project/modin/issues/3278.

amyskov:

## What do these changes do?

- [x] passes `flake8 modin`
- [x] Resolves #3437  
- [ ] tests added and passing

### DIFF
--- a/modin/backends/pandas/parsers.py
+++ b/modin/backends/pandas/parsers.py
@@ -149,9 +149,10 @@ class PandasParser(object):
 
         Returns
         -------
-        pandas.Series
-            pandas.Series where index is columns names and values are
-            columns dtypes.
+        frame_dtypes : pandas.Series or dtype
+            Resulting dtype or pandas.Series where column names are used as
+            index and types of columns are used as values for full resulting
+            frame.
         """
         # each element in `partitions_dtypes` is a Series, where column names are
         # used as index and types of columns for different partitions are used as values
@@ -159,24 +160,28 @@ class PandasParser(object):
         if all([len(dtype) == 0 for dtype in partitions_dtypes]):
             return None
 
-        # concat all elements of `partitions_dtypes` and find common dtype
-        # for each of the column among all partitions
-        combined_dtypes = pandas.concat(partitions_dtypes, axis=1).apply(
-            lambda row: find_common_type_cat(row.values),
-            axis=1,
-        )
+        combined_part_dtypes = pandas.concat(partitions_dtypes, axis=1)
+        frame_dtypes = combined_part_dtypes.iloc[:, 0]
 
-        for part_dtype in partitions_dtypes:
-            if not combined_dtypes.equals(part_dtype):
+        for part_dtype_col in combined_part_dtypes.columns:
+            if not combined_part_dtypes[part_dtype_col].equals(frame_dtypes):
                 ErrorMessage.missmatch_with_pandas(
-                    operation="read_csv",
+                    operation="read_*",
                     message="Data types of partitions are different! "
                     "Please refer to the troubleshooting section of the Modin documentation "
                     "to fix this issue",
                 )
+
+                # concat all elements of `partitions_dtypes` and find common dtype
+                # for each of the column among all partitions
+                frame_dtypes = combined_part_dtypes.apply(
+                    lambda row: find_common_type_cat(row.values),
+                    axis=1,
+                ).squeeze(axis=0)
+
                 break
 
-        return combined_dtypes.squeeze(axis=0)
+        return frame_dtypes
 
     @classmethod
     def single_worker_read(cls, fname, **kwargs):
--- a/modin/backends/pandas/parsers.py
+++ b/modin/backends/pandas/parsers.py
@@ -138,7 +138,7 @@ class PandasParser(object):
     """Base class for parser classes with pandas backend."""
 
     @classmethod
-    def get_dtypes(cls, dtypes_ids):
+    def get_dtypes(cls, dtypes_ids, check_homogeneity=False):
         """
         Get common for all partitions dtype for each of the columns.
 
@@ -146,6 +146,10 @@ class PandasParser(object):
         ----------
         dtypes_ids : list
             Array with references to the partitions dtypes objects.
+        check_homogeneity : bool, default: False
+            Whether or not to check data homogeneity.
+            If data is not homogeneous return dict columns and types
+            it should be casted.
 
         Returns
         -------
@@ -153,13 +157,32 @@ class PandasParser(object):
             pandas.Series where index is columns names and values are
             columns dtypes.
         """
+        is_data_homogeneous = True
+        # each element in `partitions_dtypes` is a Series, where index is
+        # a column name and value is dtype of this coulumn in the concreate
+        # partition
         partitions_dtypes = cls.materialize(dtypes_ids)
+
         if all([len(dtype) == 0 for dtype in partitions_dtypes]):
             return None
+
+        # concat all elements of `partitions_dtypes` and find common dtype
+        # for each of the column among all partitions
+        combined_dtypes = pandas.concat(partitions_dtypes, axis=1).apply(
+            lambda row: find_common_type_cat(row.values),
+            axis=1,
+        )
+        if check_homogeneity:
+            for part_dtype in partitions_dtypes:
+                if not combined_dtypes.equals(part_dtype):
+                    is_data_homogeneous = False
+                    break
+        combined_dtypes = combined_dtypes.squeeze(axis=0)
+
         return (
-            pandas.concat(partitions_dtypes, axis=1)
-            .apply(lambda row: find_common_type_cat(row.values), axis=1)
-            .squeeze(axis=0)
+            (combined_dtypes, is_data_homogeneous)
+            if check_homogeneity
+            else combined_dtypes
         )
 
     @classmethod
--- a/modin/backends/pandas/parsers.py
+++ b/modin/backends/pandas/parsers.py
@@ -138,7 +138,7 @@ class PandasParser(object):
     """Base class for parser classes with pandas backend."""
 
     @classmethod
-    def get_dtypes(cls, dtypes_ids, check_homogeneity=False):
+    def get_dtypes(cls, dtypes_ids):
         """
         Get common for all partitions dtype for each of the columns.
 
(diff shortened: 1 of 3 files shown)


modin/backends/pandas/parsers.py
```diff
@@ -162,6 +162,46 @@ def get_dtypes(cls, dtypes_ids):
             .squeeze(axis=0)
         )
 
+    @classmethod
+    def get_dtypes_check_homogen(cls, dtypes_ids):
```
anmyachev, line 166: What is the performance impact of this check?

Is it possible to enable it only when the types are not explicitly specified by the user?

modin/backends/pandas/parsers.py
```diff
@@ -162,6 +162,46 @@ def get_dtypes(cls, dtypes_ids):
...
+
+        if all([len(dtype) == 0 for dtype in partitions_dtypes]):
+            return None, True
+
+        # concat all elements of `partitions_dtypes` and find common dtype
+        # for each of the column among all partitions
+        combined_dtypes = pandas.concat(partitions_dtypes, axis=1).apply(
+            lambda row: find_common_type_cat(row.values),
+            axis=1,
+        )
```
anmyachev, line 197: Can `get_dtypes` function be reused here?

modin/backends/pandas/parsers.py
amyskov: The complexity of this check linearly depends on the number of partitions/CPUs, so it shouldn't affect `read_csv` performance. Measurements results are next (no performance degradation is obtained, scalability issue for 5000x5000 dataset should be investigated separately):
```
MODIN_CPUS=8
     [119a8be2]       [e92bfc04]
     <master>         <read_csv_warn_het_data>
          139±9ms          138±1ms     0.99  io.csv.TimeReadCsvSkiprows.time_skiprows([1000000, 10], None)
         927±20ms          908±8ms     0.98  io.csv.TimeReadCsvSkiprows.time_skiprows([5000, 5000], None)


MODIN_CPUS=112
     [119a8be2]       [e92bfc04]
     <master>         <read_csv_warn_het_data>
          159±3ms          158±4ms     0.99  io.csv.TimeReadCsvSkiprows.time_skiprows([1000000, 10], None)
       5.66±0.01s       5.69±0.01s     1.00  io.csv.TimeReadCsvSkiprows.time_skiprows([5000, 5000], None)
```

Yes, we can enable this check only if `dtype` is not set, but even in this case partitions dtypes can be different (for example if the user defined `dtype` not for all columns with heterogeneous data), so i suggest to leave check even if `dtype` is set (since this check is cheap).

anmyachev: Performance is indeed linearly dependent on the number of partitions, but the dependence of the number of partitions on the number of cores is quadratic :)

modin/engines/base/io/text/csv_dispatcher.py
```diff
@@ -358,6 +363,13 @@ def _get_new_qc(
             return new_query_compiler[new_query_compiler.columns[0]]
         if index_col is None:
             new_query_compiler._modin_frame.synchronize_labels(axis=0)
+        if not is_data_homogeneous:
+            ErrorMessage.missmatch_with_pandas(
+                operation="read_csv",
+                message="Data types of partitions are different! "
+                "Please refer to the troubleshooting section of the Modin documentation "
+                "to fix this issue",
+            )
```
anmyachev, line 372: It is best to issue warnings as soon as possible.

modin/backends/pandas/parsers.py
amyskov: Not sure that understood your point(
If we run something like this 
`MODIN_CPUS=8 python -c "import modin.pandas as pd; df = pd.read_csv('data_5000_5000.csv')"`
and check `partitions_dtypes` variable, it length will be equals to `MODIN_CPUS`:
```
> modin/modin/backends/pandas/parsers.py(199)get_dtypes_check_homogen()
-> for part_dtype in partitions_dtypes:
(Pdb) len(partitions_dtypes)
8
(Pdb) partitions_dtypes[0]
col0       int64
col1       int64
col2       int64
col3       int64
col4       int64
           ...
col4995    int64
col4996    int64
col4997    int64
col4998    int64
col4999    int64
Length: 5000, dtype: object
```
P.S. If quadratic appears from column partitioning, `dtypes` objects [not spitted](https://github.com/modin-project/modin/blob/master/modin/backends/pandas/parsers.py#L244-L247) even if number of columns is big enough.

modin/engines/base/io/text/csv_dispatcher.py
amyskov: Yes, i agree. Warning was replaced.

modin/backends/pandas/parsers.py
anmyachev: > P.S. If quadratic appears from column partitioning, dtypes objects not spitted even if number of columns is big enough.

Oh, I see.

modin/backends/pandas/parsers.py
anmyachev: Then let's leave it as it is until #3288.

anmyachev (approved):

anmyachev:
@devin-petersohn please take a look.

modin/engines/base/io/text/csv_dispatcher.py
```diff
@@ -301,7 +302,18 @@ def _get_new_qc(
         # reported dtypes from differing rows can be different based on the inference in
         # the limited data seen by each worker. We use pandas to compute the exact dtype
         # over the whole column for each column. The index is set below.
-        dtypes = cls.get_dtypes(dtypes_ids) if len(dtypes_ids) > 0 else None
+        dtypes, is_data_homogeneous = (
+            cls.get_dtypes_check_homogen(dtypes_ids)
+            if len(dtypes_ids) > 0
+            else (None, None)
+        )
+        if not is_data_homogeneous:
```
devin-petersohn, line 310: Can we just move this check to `cls.get_dtypes` to handle all of the mismatched `dtype`s? We can throw the exception or warning there, right? That way we don't have to add a new method.

amyskov: Yes, let's use single `get_dtypes` for now. If it will be needed, separate function can be added in https://github.com/modin-project/modin/pull/3288.

modin/backends/pandas/parsers.py
```diff
@@ -153,15 +153,32 @@ def get_dtypes(cls, dtypes_ids):
             pandas.Series where index is columns names and values are
             columns dtypes.
         """
+        # each element in `partitions_dtypes` is a Series, where index is
+        # a column name and value is dtype of this coulumn in the concreate
+        # partition
```
anmyachev, line 158:
```suggestion
        # each element in `partitions_dtypes` is a Series, where column names are 
        # used as index and types of columns for different partitions are used as values
```

modin/backends/pandas/parsers.py
```diff
@@ -153,15 +153,31 @@ def get_dtypes(cls, dtypes_ids):
...
+        # for each of the column among all partitions
+        combined_dtypes = pandas.concat(partitions_dtypes, axis=1).apply(
+            lambda row: find_common_type_cat(row.values),
+            axis=1,
         )
 
+        for part_dtype in partitions_dtypes:
+            if not combined_dtypes.equals(part_dtype):
+                ErrorMessage.missmatch_with_pandas(
+                    operation="read_csv",
```
devin-petersohn, line 172: Will it always be `read_csv` here? It could be the other readers, correct?

modin/backends/pandas/parsers.py
```diff
@@ -153,15 +153,31 @@ def get_dtypes(cls, dtypes_ids):
...
-            .squeeze(axis=0)
+
+        # concat all elements of `partitions_dtypes` and find common dtype
+        # for each of the column among all partitions
+        combined_dtypes = pandas.concat(partitions_dtypes, axis=1).apply(
+            lambda row: find_common_type_cat(row.values),
+            axis=1,
         )
 
+        for part_dtype in partitions_dtypes:
```
devin-petersohn, line 169: This will be really expensive for really large clusters and for very wide data.

What about a fastpath for all equal data types:

```python
combined_part_dtypes = pandas.concat(partitions_dtypes, axis=1)
if combined_part_dtypes.eq(combined_part_dtypes.iloc[0], axis=1).all(axis=None):
    combined_dtypes = combined_part_dtypes.iloc[0]
else:
    ErrorMessage.missmatch_with_pandas(...)

```

Does this make sense?

modin/backends/pandas/parsers.py
amyskov: Right, changed to `read_*`.

modin/backends/pandas/parsers.py
amyskov: Yes, for very wide data and large number of partitions it potentially can be expensive. Implemented suggestion with couple corrections, so now common for all partitions dtypes `combined_part_dtypes.iloc[0]` is calculated only if heterogeneous data is detected.

modin/backends/pandas/parsers.py
```diff
@@ -149,18 +149,39 @@ def get_dtypes(cls, dtypes_ids):
...
-        return (
-            pandas.concat(partitions_dtypes, axis=1)
-            .apply(lambda row: find_common_type_cat(row.values), axis=1)
-            .squeeze(axis=0)
-        )
+
+        combined_part_dtypes = pandas.concat(partitions_dtypes, axis=1)
+        frame_dtypes = combined_part_dtypes.iloc[:, 0]
+
+        for part_dtype_col in combined_part_dtypes.columns:
```
devin-petersohn, line 166: This for loop will still be costly for the common case where all are equal. I prefer to avoid the for loop with something like:

```python
if combined_part_dtypes.eq(combined_part_dtypes.iloc[:, 0], axis=1).all(axis=None):
    pass
```

This does the check in a non-looping way. The `eq` will do a column-wise comparison and `all(axis=None)` will check for all values being `True`.

amyskov: Ok, `combined_part_dtypes.eq` method was used.

devin-petersohn (commented):
Thanks @amyskov it looks good. Can we get a quick benchmark from asv on this change?

amyskov:
> Thanks @amyskov it looks good. Can we get a quick benchmark from asv on this change?

Yes, sure, results with `MODIN_CPUS=8` are next:
```
     [e97e2e62]       [b3bc0eab]
     <master>         <read_csv_warn_het_data>
         236±20ms          239±8ms     1.01  io.csv.TimeReadCsvSkiprows.time_skiprows([1000000, 10], None)
       1.51±0.04s       1.46±0.04s     0.97  io.csv.TimeReadCsvSkiprows.time_skiprows([5000, 5000], None)
```
P.S. Improvement for 5000x5000 dataset is inside of measurement error boundaries, but this improvement persists across several asv runs.

devin-petersohn (approved):

devin-petersohn:
Thanks @amyskov!

PR #3452, opened 2021-09-21
FEAT-#3451: Support `__partitioned__` protocol

### LINKED ISSUE #3451: Support `__partitioned__` protocol
https://github.com/IntelPython/DPPY-Spec/blob/draft/partitioned/Partitioned.md
IntelPython/DPPY-Spec#3

YarShev:

## What do these changes do?

- [x] passes `flake8 modin`
- [x] Resolves #3451  
- [x] tests passing

### DIFF
--- a/modin/pandas/base.py
+++ b/modin/pandas/base.py
@@ -43,7 +43,7 @@ from typing import Optional, Union
 import warnings
 import pickle as pkl
 
-from modin.utils import try_cast_to_pandas, _inherit_docstrings
+from modin.utils import try_cast_to_pandas, _inherit_docstrings, get_current_backend
 from modin.error_message import ErrorMessage
 from modin.pandas.utils import is_scalar
 from modin.config import IsExperimental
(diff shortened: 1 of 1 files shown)


Garra1980:
yeah, let's not consider this for 0.11

fschlimb:
@devin-petersohn You're raising a chicken-and-egg issue.

Generally, adding support as a consumer is more involved but at the same time adds more value. Adding this at the producer side is less involved. Like here it is very isolated - it has no effect on anything else.

Current Implementation status is
* https://github.com/ray-project/xgboost_ray/pull/153
* https://github.com/helmholtz-analytics/heat/pull/788
* https://github.com/fschlimb/daal4py/tree/feature/partitioned_interface
* https://github.com/modin-project/modin/pull/3452

More to come, but we need producers! I'd appreciate seeing modin as a producer.

To add more value for modin, we can add support for consuming this API. A simple "from_partitioned" would provide the basic functionality. "Automatic" detection when accepting data at construction time would be nice to have and could be added later.

YarShev:
> To add more value for modin, we can add support for consuming this API. A simple "from_partitioned" would provide the basic functionality. "Automatic" detection when accepting data at construction time would be nice to have and could be added later.

It would also help to match partitions of the left object (Modin DF/Series) and partitions of the right object (an instance supporting `__partitioned__`) when performing operations that may require right object(s).

devin-petersohn:
>Generally, adding support as a consumer is more involved but at the same time adds more value. Adding this at the producer side is less involved. Like here it is very isolated - it has no effect on anything else.

It's not a big deal to add something, but I don't want code bloat for things that won't be useful. I think it's a good idea to have a protocol like this, but who would use it? What library has said that they actually need an interface like this to be able to work with Modin? We at least need to know if it would be useful to a meaningful consumer before we add it.

Typically these protocols have significant amounts of input from multiple interested parties (producers and consumers) to answer these questions before they are even designed. I just want to make sure we aren't adding something that won't be used. I don't deny the **usefulness** of this protocol, but whether or not it _will_ be used is still not answered to me.

fschlimb:
> Typically these protocols have significant amounts of input from multiple interested parties (producers and consumers) to answer these questions before they are even designed.

Yes, we are open to feedback and suggestions for different designs and features. As mentioned above, your feedback/suggestions is/are highly appreciated. Talking about something concrete is usually easier than keeping discussions in the abstract.

> I just want to make sure we aren't adding something that won't be used. I don't deny the **usefulness** of this protocol, but whether or not it _will_ be used is still not answered to me.

That's of course a valid requrest.

One issue in the process of ramping this up is that we are targeting something that is currently less of a concern but we want to make sure we can avoid running into a situation where implementations of distributed features become messy.

> It's not a big deal to add something, but I don't want code bloat for things that won't be useful. I think it's a good idea to have a protocol like this, but who would use it? What library has said that they actually need an interface like this to be able to work with Modin? We at least need to know if it would be useful to a meaningful consumer before we add it.

The idea is to allow a packages to consume various distributed containers. It is not meant to be an enabler for modin specifically.  Of course packages like xgboost_ray can support modin/RayDataSetl/MLDataSet/HeAT/... by implementing dedicated code for each. Avoiding such specialization for each (in particular upcoming) structure is the major motivation for this.

YarShev:
@vnlitvinov , yes, we can mark the PR as a draft for now. I'll update it once the protocol settles.

YarShev:
Closing this PR as no plans on pushing this forward.

PR #3453, opened 2021-09-21
Read json glob

anmyachev:
## What do these changes do?

- [ ] passes `flake8 modin`
- [ ] Resolves #? 
- [ ] tests added and passing


modin/conftest.py
```diff
@@ -342,6 +348,133 @@ def make_csv_file():
     teardown_test_files(filenames)
 
 
+def create_and_call(filenames, filename, force, nrows, ncols, func: str, func_kw=None):
```
amyskov, line 351: It's a little bit confusing function name, will be great if docstring describing it possible purpose will be added.

modin/conftest.py
```diff
@@ -342,6 +348,133 @@ def make_csv_file():
...
+    if force or not os.path.exists(filename):
+        df = pandas.DataFrame({f"col{x + 1}": np.arange(nrows) for x in range(ncols)})
+        getattr(df, func)(filename, **func_kw if func_kw else {})
+        filenames.append(filename)
+
+
+def make_default_file(file_type: str):
```
amyskov, line 358: I think it is possible to reduce code duplication of this function. Also this function can be moved to the `utils` module.

modin/conftest.py
```diff
@@ -342,6 +348,133 @@ def make_csv_file():
...
+    filenames = []
+    func = None
+
+    if file_type == "json":
+
+        def _make_json_file(filename, nrows=NROWS, ncols=2, lines=False, force=True):
+            func_kw = {"lines": lines, "orient": "records"} if lines else {}
+            create_and_call(
```
amyskov, line 366: Looks like test DataFrame will be created for each of the tests, can it be cached somehow?

amyskov (commented):
Also looks like test files don't properly deleted after tests execution, while on master clean-up works fine (checked on `TestJson` tests).

anmyachev:
> Also looks like test files don't properly deleted after tests execution, while on master clean-up works fine (checked on `TestJson` tests).

You are right, the bug has already been fixed.

modin/conftest.py
anmyachev: Renamed and moved this function inside `make_default_file`, as it is very narrowly focused. Do I need to add documentation?

modin/conftest.py
anmyachev: Reduced and moved.

modin/conftest.py
anmyachev: Creating a dataframe with default sizes takes no more than 0.02 seconds. I think there is no need yet.

modin/conftest.py
amyskov: Yes, i think docstring with note will be fine (something describing what should be created and why then call `func`).

modin/conftest.py
amyskov: Ok, let's leave it as is.

anmyachev:
@amyskov ready for review

modin/conftest.py
```diff
@@ -362,32 +434,32 @@ def _make_parquet_file(
...
-            pass
-        elif directory:
-            if os.path.exists(filename):
-                shutil.rmtree(filename)
+        if force or not os.path.exists(filename):
+            df = pandas.DataFrame(
+                {f"col{x + 1}": np.arange(nrows) for x in range(ncols)}
+            )
+            if directory:
+                # ???
```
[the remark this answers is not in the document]
anmyachev: In other cases, if the file exists, then we do not delete it and do not create a new one, but in the case we delete - why?

anmyachev:
@amyskov ready for review

modin/conftest.py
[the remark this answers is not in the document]
amyskov: Interesting) Does it works now?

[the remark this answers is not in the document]
anmyachev: I see.
Yeh it works.

amyskov (approved):
LGTM! Thanks @anmyachev

PR #3466, opened 2021-09-23
Fix usage of modin.HDFStore in modin.read_hdf func

### LINKED ISSUE #3393: HDFStore import creates invalid DataFrame
### System information
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: Arch Linux
- **Modin version** (`modin.__version__`): 0.10.2
- **Python version**: 3.9.6

Code to reproduce:
```python
#!/usr/bin/env python
import modin.pandas as pd
import numpy as np

if __name__ == "__main__":
    # Create a test dataframe
    dfin = pd.DataFrame(np.random.rand(8, 8))
    # and write it to a HDF5 file...
    dfin.to_hdf("hdf.h5", "/key")

    # Open up a HDFStore and read it back in
    with pd.HDFStore("hdf.h5") as h:
        df = pd.read_hdf(h, "/key")
    # Trying to print it fails entirely
    print(df)
```

### Describe the problem
When reading in a HDF5 via a `pandas.HDFStore`, the resulting dataframe is unusable - it appears to have been doubly-wrapped somehow, as the error shows a modin DataFrame being returned by the `.to_pandas()` method.

### Source code / logs
Output of above test script (run in a new venv with `modin[ray]` and `tables` installed):
```

    import ray
    ray.init()

UserWarning: The size of /dev/shm is too small (8178835456 bytes). The required size at least half of RAM (8349171712 bytes). Please, delete files in /dev/shm or increase size of /dev/shm with --shm-size in Docker. Also, you can set the required memory size for each Ray worker in bytes to MODIN_MEMORY environment variable.
UserWarning: `DataFrame.to_hdf` defaulting to pandas implementation.
UserWarning: Distributing <class 'pandas.core.frame.DataFrame'> object. This may take some time.
UserWarning: `open` defaulting to pandas implementation.
UserWarning: `read_hdf` defaulting to pandas implementation.
UserWarning: `select` defaulting to pandas implementation.
UserWarning: `get_node` defaulting to pandas implementation.
UserWarning: `close` defaulting to pandas implementation.
  File "/home/jake/./test.py", line 13, in <module>
    print(df)
  File "/home/jake/modin-env/lib/python3.9/site-packages/modin/pandas/base.py", line 3014, in __str__
    return repr(self)
  File "/home/jake/modin-env/lib/python3.9/site-packages/modin/pandas/dataframe.py", line 209, in __repr__
    result = repr(self._build_repr_df(num_rows, num_cols))
  File "/home/jake/modin-env/lib/python3.9/site-packages/modin/pandas/base.py", line 195, in _build_repr_df
    return self.iloc[indexer]._query_compiler.to_pandas()
  File "/home/jake/modin-env/lib/python3.9/site-packages/modin/backends/pandas/query_compiler.py", line 255, in to_pandas
    return self._modin_frame.to_pandas()
  File "/home/jake/modin-env/lib/python3.9/site-packages/modin/engines/base/frame/data.py", line 2227, in to_pandas
    df = self._partition_mgr_cls.to_pandas(self._partitions)
  File "/home/jake/modin-env/lib/python3.9/site-packages/modin/engines/base/frame/partition_manager.py", line 678, in to_pandas
    retrieved_objects = [[obj.to_pandas() for obj in part] for part in partitions]
  File "/home/jake/modin-env/lib/python3.9/site-packages/modin/engines/base/frame/partition_manager.py", line 678, in <listcomp>
    retrieved_objects = [[obj.to_pandas() for obj in part] for part in partitions]
  File "/home/jake/modin-env/lib/python3.9/site-packages/modin/engines/base/frame/partition_manager.py", line 678, in <listcomp>
    retrieved_objects = [[obj.to_pandas() for obj in part] for part in partitions]
  File "/home/jake/modin-env/lib/python3.9/site-packages/modin/engines/ray/pandas_on_ray/frame/partition.py", line 194, in to_pandas
    assert type(dataframe) is pandas.DataFrame or type(dataframe) is pandas.Series
AssertionError
```

anmyachev:
## What do these changes do?

- [ ] passes `flake8 modin`
- [x] Resolves #3393 
- [ ] tests added and passing

### DIFF
--- a/modin/engines/base/io/io.py
+++ b/modin/engines/base/io/io.py
@@ -29,6 +29,7 @@ from pandas.util._decorators import doc
 from modin.error_message import ErrorMessage
 from modin.backends.base.query_compiler import BaseQueryCompiler
 from modin.utils import _inherit_docstrings
+from modin.pandas.io import HDFStore
 
 _doc_default_io_method = """
 {summary} using pandas.
@@ -519,21 +520,26 @@ class BaseIO(object):
         **kwargs,
     ):  # noqa: PR01
         ErrorMessage.default_to_pandas("`read_hdf`")
-        return cls.from_pandas(
-            pandas.read_hdf(
-                path_or_buf,
-                key=key,
-                mode=mode,
-                columns=columns,
-                errors=errors,
-                where=where,
-                start=start,
-                stop=stop,
-                iterator=iterator,
-                chunksize=chunksize,
-                **kwargs,
-            )
+        modin_store = isinstance(path_or_buf, HDFStore)
+        if modin_store:
+            path_or_buf._return_modin_dataframe = False
+        df = pandas.read_hdf(
+            path_or_buf,
+            key=key,
+            mode=mode,
+            columns=columns,
+            errors=errors,
+            where=where,
+            start=start,
+            stop=stop,
+            iterator=iterator,
+            chunksize=chunksize,
+            **kwargs,
         )
+        if modin_store:
+            path_or_buf._return_modin_dataframe = True
+
+        return cls.from_pandas(df)
 
     @classmethod
     @_inherit_docstrings(pandas.read_feather, apilink="pandas.read_feather")
--- a/modin/engines/base/io/io.py
+++ b/modin/engines/base/io/io.py
@@ -29,7 +29,6 @@ from pandas.util._decorators import doc
 from modin.error_message import ErrorMessage
 from modin.backends.base.query_compiler import BaseQueryCompiler
 from modin.utils import _inherit_docstrings
-from modin.pandas.io import HDFStore
 
 _doc_default_io_method = """
 {summary} using pandas.
@@ -519,6 +518,8 @@ class BaseIO(object):
         chunksize=None,
         **kwargs,
     ):  # noqa: PR01
+        from modin.pandas.io import HDFStore
+
         ErrorMessage.default_to_pandas("`read_hdf`")
         modin_store = isinstance(path_or_buf, HDFStore)
         if modin_store:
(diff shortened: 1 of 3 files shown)


modin/engines/base/io/io.py
```diff
@@ -519,21 +520,26 @@ def read_hdf(
...
-                stop=stop,
-                iterator=iterator,
-                chunksize=chunksize,
-                **kwargs,
-            )
+        modin_store = isinstance(path_or_buf, HDFStore)
+        if modin_store:
+            path_or_buf._return_modin_dataframe = False
+        df = pandas.read_hdf(
+            path_or_buf,
```
anmyachev, line 527: In some cases, `modin.HDFStore` object passed in `pandas.read_hdf` function that does not count on it.

As a solution for the case, it is proposed to disable conversion to modin objects.

modin/engines/base/io/io.py
```diff
@@ -518,22 +518,29 @@ def read_hdf(
         chunksize=None,
         **kwargs,
     ):  # noqa: PR01
+        from modin.pandas.io import HDFStore
```
anmyachev, line 521: Using import at the beginning of the file leads to a change in the import order, which in turn breaks `pydocstyle` test.

modin/engines/base/io/io.py
devin-petersohn: I would prefer to avoid return any pandas objects. Modin and pandas objects do not play well together. Worst case we should throw an error, but ideally this is fixable to return a Modin dataframe.

anmyachev: Pandas object will be return only before `cls.from_pandas(df)` call, in which `df` will be cast to Modin dataframe.

devin-petersohn (approved):
Thanks @anmyachev, LGTM!

PR #3473, opened 2021-09-24
TEST-#3474: fix pip install commmand in push-to-master

### LINKED ISSUE #3474: New pip resolver takes a very long time to complete
Issue on pip side https://github.com/pypa/pip/issues/9187.

Problem in our CI (takes more that 5 hours) https://github.com/modin-project/modin/runs/3695414357.

Good comment about possible solutions https://github.com/pypa/pip/issues/9187#issuecomment-736000492.

I propose the easiest way to support - to use the old resolver, until the pip optimizes the new resolver.

anmyachev:

## What do these changes do?

- [ ] passes `flake8 modin`
- [x] Resolves #3474 
- [ ] tests added and passing

### DIFF
--- a/.github/workflows/push-to-master.yml
+++ b/.github/workflows/push-to-master.yml
@@ -1,79 +1,6 @@
 name: push-to-master
-on:
-  push:
-    branches:
-      - master
+on: pull_request
 jobs:
-  test-ray-master:
-    runs-on: ubuntu-latest
-    defaults:
-      run:
-        # `shell: bash -l {0}` - special way to activate modin environment
-        shell: bash -l {0}
-    steps:
-      - uses: actions/checkout@v2
-        with:
-          fetch-depth: 1
-      - uses: conda-incubator/setup-miniconda@v2
-        with:
-          activate-environment: modin
-          environment-file: requirements/requirements-no-engine.yml
-          python-version: 3.7
-          channel-priority: strict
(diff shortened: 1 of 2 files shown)


.github/workflows/push-to-master.yml
```diff
@@ -1,79 +1,6 @@
 name: push-to-master
-on:
-  push:
-    branches:
-      - master
+on: pull_request
 jobs:
-  test-ray-master:
-    runs-on: ubuntu-latest
```
vnlitvinov, line 8: why change this?

anmyachev: For test purpose; this should be reverted before merge.

.github/workflows/push-to-master.yml
```diff
@@ -115,7 +42,7 @@ jobs:
           python-version: ${{matrix.python-version}}
           channel-priority: strict
           use-only-tar-bz2: true # IMPORTANT: This needs to be set for caching to work properly!
-      - run: pip install -r requirements-dev.txt
+      - run: pip install -r requirements-dev.txt --use-deprecated=legacy-resolver
```
anmyachev, line 45: This will be the only change.

vnlitvinov: okay, revert the unnecessary changes then

vnlitvinov (approved):
Not looking good, but would solve as a stopgap I think...

Issue #3480, opened 2021-09-24
[omnisci] Various problems in datetime processing


gshimansky:
### System information
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**:

Ubuntu 20.04.1 LTS

- **Modin version** (`modin.__version__`):

0.10.2+39.g15f168ca

- **Python version**:

Python 3.9.7

- **Code we can use to reproduce**:

```python
import modin.pandas as pd

df = pd.read_csv("datetime.csv", parse_dates=["pickup_datetime", "dropoff_datetime"])

print(df.dtypes)
print(df["pickup_datetime"].astype("int64"))

df["dropoff_datetime"][0] = pd.to_datetime("2014-01-09 20:52:31")
df["dropoff_datetime"][1] = pd.to_datetime("2014-01-09 20:55:12")
# If you uncomment this line you get a different exception
# df["dropoff_datetime"][3] = pd.to_datetime("2014-01-09 20:59:46")

print(df["dropoff_datetime"].astype("int64"))
```

### Describe the problem

Code above finds three different problems:
1. When datetime is read from CSV file, it is kept in seconds, not in nanoseconds as type `datetime64[ns]` implies. When this code runs on Pandas it shows numbers like `1389300325000000000` while the same code on Modin on omnisci produces numbers like `1389300325`.
2. When datetime is assigned to column from `pd.to_datetime` it produces an `AssertionError: unaligned dtypes (dropoff_datetime    int64`.
3. If you uncomment a 3rd assignment line this code produces another exception `ValueError: cannot reindex from a duplicate axis`.

Attached is the data file. [datetime.csv](https://github.com/modin-project/modin/files/7227994/datetime.csv)

Garra1980:
What else do we need here?

Garra1980:
Let’s close this then

YarShev:
@ienkovich , did #3483 fully covered this issue?

ienkovich:
I worked on the second problem only.

YarShev:
Let's put the rest part to be resolved in 0.13 then.

YarShev:
HDK engine is deprecated and will be removed in a future version.

PR #3483, opened 2021-09-27
FIX-#3480: Fix dtypes for astype result.

ienkovich:
## What do these changes do?
Fix unaligned dtypes issue.

- [x] passes `flake8 modin`
- [ ] Partly covers #3480? 
- [x] tests added and passing

### DIFF
--- a/modin/experimental/engines/omnisci_on_native/frame/data.py
+++ b/modin/experimental/engines/omnisci_on_native/frame/data.py
@@ -723,12 +723,12 @@ class OmnisciOnNativeFrame(PandasFrame):
             The new frame.
         """
         columns = col_dtypes.keys()
-        new_dtypes = self.dtypes.copy()
+        new_dtypes = self._dtypes.copy()
         for column in columns:
             dtype = col_dtypes[column]
             if (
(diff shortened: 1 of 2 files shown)


Garra1980 (approved):

dchigarev (approved):

modin/experimental/engines/omnisci_on_native/frame/data.py
```diff
@@ -723,12 +723,12 @@ def astype(self, col_dtypes, **kwargs):
             The new frame.
         """
         columns = col_dtypes.keys()
-        new_dtypes = self.dtypes.copy()
+        new_dtypes = self._dtypes.copy()
```
vnlitvinov, line 726: why do we want to use `_dtype` here? pls explain

ienkovich: `_dtypes` is an internal attribute holding data type for all columns including index ones. `dtypes` is a public property holding data type for `columns`, which doesn't include index columns. Until recent times `dtypes` property worked incorrectly and always returned `_dtypes`. After `dtypes` property fix, this code started to work incorrectly.

PR #3489, opened 2021-09-27
REFACTOR-3488: Refactor `read_csv` `skiprows` parameter processing

### LINKED ISSUE #3488: Refactor processing of `skiprows` parameter in `modin.engines.base.io.text.csv_dispatcher`
It is needed to localize processing of `skiprows` parameter in the csv dispatcher.

amyskov:

## What do these changes do?

- [x] passes `flake8 modin`
- [x] Resolves #3488  

### DIFF
--- a/modin/engines/base/io/text/csv_dispatcher.py
+++ b/modin/engines/base/io/text/csv_dispatcher.py
@@ -367,22 +367,42 @@ class CSVDispatcher(TextFileDispatcher):
         Manage read_csv `skiprows` parameter.
 
         Change `skiprows` parameter in the way Modin could more optimally
-        process it. If `skiprows` is an array, this array will be sorted and
-        then, if array is uniformly distributed, `skiprows` will be "squashed"
-        into integer value and `pre_reading` parameter will be set if needed
-        (in this case fastpath can be done).
+        process it. `csv_dispatcher` have two mechanisms of rows skipping:
+
+        1) During file partitioning (setting of file limits that should be read
+        by each partition) exact rows can be excluded from partitioning scope,
+        thus they won't be read at all and can be considered as skipped. This is
+        the most effective way of rows skipping (since it doesn't require any
+        actual data reading and postprocessing), but in this case `skiprows`
+        parameter can be an integer only. When it possible Modin always uses
+        this approach by setting of `skiprows_partitioning` return value.
+
+        2) Rows for skipping can be dropped after full dataset import. This is
+        more expensive way since it requires extra IO work and postprocessing
+        afterwards, but `skiprows` parameter can be of any non-integer type
+        supported by `pandas.read_csv` `skiprows` parameter.
+
+        In some cases, if `skiprows` is uniformly distributed array (e.g. [1,2,3]),
+        `skiprows` can be "squashed" and represented as integer to make a fastpath.
+        If there is a gap between the first row for skipping and the last line of
+        the header (that will be skipped too), then assign to read this gap first
+        (assign the first partition to read these rows be setting of `pre_reading`
+        return value). See `Examples` section for details.
 
         Parameters
         ----------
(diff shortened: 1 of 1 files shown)


modin/engines/base/io/text/csv_dispatcher.py
```diff
@@ -86,7 +86,11 @@ def _read(cls, filepath_or_buffer, **kwargs):
             header,
             names,
         )
-        skiprows_md, pre_reading = cls._manage_skiprows_parameter(skiprows, header_size)
+        (
+            skiprows_md,
+            pre_reading,
+            skiprows_partitioning,
```
anmyachev, line 92: I'm starting to get confused between these options. Why do we need so many of them?

amyskov: Ok, i see, explanation was added to the `_manage_skiprows_parameter` docstring.

modin/engines/base/io/text/csv_dispatcher.py
```diff
@@ -371,38 +367,89 @@ def _manage_skiprows_parameter(
...
+            splitting for further reading (the number of rows for
+            the first partition).
+        skiprows_partitioning : int
+            The number of rows that should be skipped virtually (skipped during
+            data file partitioning).
+
+        Examples
+        --------
+        Let's consider case when `header` = "infer" and `skiprows`=[3,4,5]. In
+        this specific case fastpath can be done, so `pre_reading` = 2 and
```
devin-petersohn, line 417: Sorry, what is `pre_reading` and why does it get set to 2 in this example? This example is very helpful, but I would like to understand more about why it was designed this way.

amyskov: @devin-petersohn example section was expanded, please take a look.

devin-petersohn (commented):
The updated documentation makes sense, thanks!

@anmyachev does this approach make sense to you?

modin/engines/base/io/text/csv_dispatcher.py
```diff
@@ -371,38 +367,97 @@ def _manage_skiprows_parameter(
...
+        more expensive way since it requires extra IO work and postprocessing
+        afterwards, but `skiprows` parameter can be of any non-integer type
+        supported by `pandas.read_csv` `skiprows` parameter. These rows is
+        specified by setting of `skiprows_md` return value.
+
+        In some cases, if `skiprows` is uniformly distributed array (e.g. [1,2,3]),
+        `skiprows` can be "squashed" and represented as integer to make a fastpath.
+        If there is a gap between the first row for skipping and the last line of
+        the header (that will be skipped too), then assign to read this gap first
+        (assign the first partition to read these rows be setting of `pre_reading`
```
anmyachev, line 390: Can we just pass several pairs (start position, end position) into one parser, and then make the resulting dataframe with `concat` function. This will get rid of situations where partitions are not balanced in terms of the number of rows.

anmyachev: This is outside of this pull request, but still I wonder what you think about it.

anmyachev:
> The updated documentation makes sense, thanks!
> 
> @anmyachev does this approach make sense to you?

Yes

anmyachev (approved):

devin-petersohn (approved):
LGTM

modin/engines/base/io/text/csv_dispatcher.py
amyskov: Yes, you are right, we can balance the load between partitions in this way, but it will cause code complication and it is hard to estimate the speed-up we can achieve using this approach (i beware speed up will not be big enough), but anyway let's try. Corresponding issue was created https://github.com/modin-project/modin/issues/3506.

PR #3500, opened 2021-09-30
FIX-#3499: Add s3fs to deps in setup.py

### LINKED ISSUE #3499: Add `s3fs` to deps in setup.py

prutskov:
## What do these changes do?

- [x] passes `flake8 modin`
- [x] Resolves #3499  
- [ ] tests added and passing


devin-petersohn (commented):
See my comment on #3499, I don't think this should necessarily be added.

PR #3540, opened 2021-10-13
REFACTOR-#3539: align read_fwf with read_csv

### LINKED ISSUE #3539: Remove code duplication in read_csv and read_fwf implementations

anmyachev:

## What do these changes do?

- [x] passes `flake8 modin`
- [x] Resolves #3539 
- [x] tests added and passing


anmyachev:
@vnlitvinov please review

vnlitvinov (dismissed):
LGTM, thanks!

anmyachev:
@devin-petersohn can we merge the PR?

modin/core/io/text/text_file_dispatcher.py
```diff
@@ -607,6 +614,54 @@ def _launch_tasks(cls, splits: list, **partition_kwargs) -> Tuple[list, list, li
 
         return partition_ids, index_ids, dtypes_ids
 
+    @classmethod
+    def _read_csv_fwf_check_support(
+        cls,
+        filepath_or_buffer: FilePathOrBuffer,
+        read_kwargs: ReadCsvKwargsType,
+        compression_infered: str,
+        is_for_fwf_reader: bool = False,
```
devin-petersohn, line 623: It feels wrong to handle this condition like this. Why not just split the two?

modin/core/io/text/text_file_dispatcher.py
```diff
@@ -630,3 +685,380 @@ def _validate_usecols_arg(cls, usecols):
...
+                new_query_compiler = new_query_compiler.view(
+                    pandas.RangeIndex(len(new_query_compiler.index))[:nrows]
+                )
+        if index_col is None:
+            new_query_compiler._modin_frame.synchronize_labels(axis=0)
+
+        return new_query_compiler
+
+    @classmethod
+    def _generic_read(cls, filepath_or_buffer, callback, is_for_fwf_reader, **kwargs):
```
devin-petersohn, line 934: A few comments here:

- Ideally this would be `_read` and the parameters like `callback` would be fields of the class. 
- Why are we handling `is_for_fwf_reader` like this? I don't like passing a boolean flag for special casing the code in this way. I would prefer to just override it (even if it's more code).
- Can we add type hints here?

devin-petersohn (commented):
I think this is a great PR, but I prefer to restructure the code a bit. I don't like handling fwf files by passing around a boolean flag, and if we need to handle them separately we should just handle them separately instead of mixing that logic in with the other readers.

I left some comments for how it could be structured so we aren't passing unnecessary state around. Let me know if it makes sense!

anmyachev:
@devin-petersohn I rewrote the code based on your comments :) please review

anmyachev:
Blocked by #3684

modin/core/io/text/csv_dispatcher.py
```diff
@@ -14,44 +14,38 @@
...
-    Class handles utils for reading `.csv` files.
+    """Class handles utils for reading `.csv` files."""
 
-    Inherits some common for text files util functions from `TextFileDispatcher` class.
-    """
+    callback = pandas.read_csv
 
     @classmethod
-    def _read(cls, filepath_or_buffer, **kwargs):
+    def check_parameters_support(
```
vnlitvinov, line 28: There's no point of overriding parent's method if you change nothing. Just remove this method here 🙃

modin/core/io/text/fwf_dispatcher.py
```diff
@@ -14,44 +14,41 @@
...
+        use_modin_impl = TextFileDispatcher.check_parameters_support(
             filepath_or_buffer,
-            callback=pandas.read_fwf,
-            is_for_fwf_reader=True,
-            **kwargs
+            read_kwargs,
         )
+        # If infer_nrows is a significant portion of the number of rows, pandas may be
+        # faster.
+        return use_modin_impl and not read_kwargs["infer_nrows"] > 100
```
vnlitvinov, line 54: Move the check for `infer_nrows` before you call parent's function to speed up things - there's no need for extra checks if `infer_nrows > 100` already. Something like
```python
    if read_kwargs["infer_nrows"] > 100:
        # If infer_nrows is a significant portion of the number of rows, pandas may be
        # faster.
        return False
    return super().check_parameters_support(filepath_or_buffer, read_kwargs)
```

modin/core/io/text/text_file_dispatcher.py
```diff
@@ -37,18 +37,11 @@
...
 IndexColType = Union[int, str, bool, Sequence[int], Sequence[str], None]
 
 
 class TextFileDispatcher(FileDispatcher):
-    """
-    Class handles utils for reading text formats files.
-
-    Inherits some util functions for processing files from `FileDispatcher` class.
-    """
+    """Class handles utils for reading text formats files."""
```
vnlitvinov, line 44: pls declare `callback = None` here with a comment

modin/core/io/text/fwf_dispatcher.py
```diff
@@ -14,44 +14,41 @@
...
 
 
 class FWFDispatcher(TextFileDispatcher):
-    """
-    Class handles utils for reading of tables with fixed-width formatted lines.
+    """Class handles utils for reading of tables with fixed-width formatted lines."""
 
-    Inherits some common for text files util functions from `TextFileDispatcher` class.
-    """
+    callback = pandas.read_fwf
```
vnlitvinov, line 25: I suggest renaming it to `read_callback` for clarity

devin-petersohn (commented):
Thanks for the updates @anmyachev ! It looks pretty good, but I think there are some conflicts to clear.

modin/core/io/text/text_file_dispatcher.py
```diff
@@ -43,6 +43,8 @@
 class TextFileDispatcher(FileDispatcher):
     """Class handles utils for reading text formats files."""
 
+    read_callback = None
```
vnlitvinov, line 46: would be awesome if you could add a comment on what this field is expected to do

vnlitvinov (commented):
@anmyachev CI is red, pls look at it

modin/core/storage_formats/pandas/parsers.py
```diff
@@ -152,7 +152,12 @@ def generic_parse(fname, **kwargs):
             return callback(fname, **kwargs)
 
         # pop "compression" from kwargs because bio is uncompressed
-        with OpenFile(fname, "rb", kwargs.pop("compression", "infer")) as bio:
+        with OpenFile(
+            fname,
+            "rb",
+            kwargs.pop("compression", "infer"),
+            **(kwargs.pop("storage_options", None) or {}),
```
vnlitvinov, line 159: I wonder why is this done in refactoring PR...

anmyachev: It's a issue after rebase. It's [already done](https://github.com/modin-project/modin/blob/8d67cd3a467d1eb2262898f6f5e2cf159c27d7b0/modin/core/storage_formats/pandas/parsers.py#L238) for CSV and FWF parsers in master branch.

vnlitvinov (approved):
LGTM, thanks for bearing with me!

devin-petersohn (approved):
Thanks @anmyachev, LGTM!

modin/core/io/text/text_file_dispatcher.py
YarShev: same comment as above

modin/pandas/test/test_io.py
```diff
@@ -1864,7 +1864,19 @@ def test_fwf_file_chunksize(self, make_fwf_file):
...
 
-    @pytest.mark.parametrize("nrows", [13, None])
+    @pytest.mark.parametrize(
+        "nrows",
+        [
+            pytest.param(
+                13,
+                marks=pytest.mark.xfail(
+                    Engine.get() == "Ray",
+                    reason="read_fwf bug on pandas side: pandas-dev/pandas#44021",
```
YarShev, line 1874: Why does this test pass on Dask?

modin/core/io/text/csv_dispatcher.py
```diff
@@ -13,442 +13,12 @@
...
-        elif callable(skiprows):
-
-            def skiprows_func(x):
-                return skiprows(x + header_size)
-
-            skiprows_md = skiprows_func
+    """Class handles utils for reading `.csv` files."""
 
-        return skiprows_md, pre_reading, skiprows_partitioning
+    read_callback = pandas.read_csv
```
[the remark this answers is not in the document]
anmyachev: Why introduce another level of distinction? If necessary, `read_callback` can be used for pyarrow.

Globally, the idea is to describe mechanisms more generally so that different (engines+storage_formats) combinations can reuse them.

modin/pandas/test/test_io.py
anmyachev: I don't know for sure, some branch of code using pandas may differ.

Issue #3505, opened 2021-10-01
Throw custom errors with message for optional packages instead of `ModuleNotFoundError`


prutskov:
Currently in the list of optional deps we have:

- fsspec
- s3fs

We should check the package availability before usage and throw message to user in the absence. Possible example is here: https://github.com/modin-project/modin/issues/3504#issuecomment-931540502

anmyachev:
@prutskov `fsspec` is currently a required dependency

devin-petersohn:
Resolved by #3652

PR #3542, opened 2021-10-14
FIX-#3541: add fsspec to modin dependencies

### LINKED ISSUE #3541: `fsspec` should be explicitly stated in setup.py and env files
`fsspec` package became required dependency after https://github.com/modin-project/modin/pull/3529

anmyachev:

## What do these changes do?

`fsspec` is needed for almost all IO functions (including `read_csv`) so it should be considered as required dependency(after #3529) and added into `setup.py` and env `.yml` files.

- [ ] passes `flake8 modin`
- [x] Resolves #3541 
- [ ] tests added and passing

### DIFF
--- a/environment-dev.yml
+++ b/environment-dev.yml
@@ -8,6 +8,7 @@ dependencies:
   - libthrift<0.14.1  # ref: modin gh-2840
   - dask[complete]>=2.22.0
   - distributed>=2.22.0
+  - fsspec
   - xarray
   - Jinja2
   - pathlib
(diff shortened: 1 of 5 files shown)


YarShev:
I see, `fsspec` looks as a required dependency to me for now and I am okay with the changes. @devin-petersohn , what are your thoughts?

vnlitvinov:
> #3505 may also be fixed by the PR.

No, that one is about giving user appropriate feedback. It won't be closed by making something a strict requirement instead of optional 😄

PR #3553, opened 2021-10-15
TEST-#3561: split frequently failing tests

### LINKED ISSUE #3561: We spend a lot of time when restarting CI
This happens because `github actions` does not have the ability to restart separate jobs (only all jobs).

I propose to move the most frequently failing tests to a separate workflow - IO tests.

anmyachev:

## What do these changes do?

This approach allows us to reduce the time for restarting tests, but we have code duplication that is difficult to get rid of.

- [ ] passes `flake8 modin`
- [x] Resolves #3561 
- [ ] tests added and passing

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -465,7 +465,6 @@ jobs:
       - run: python -m pytest -n 2 modin/pandas/test/test_groupby.py
       - run: python -m pytest -n 2 modin/pandas/test/test_reshape.py
       - run: python -m pytest -n 2 modin/pandas/test/test_general.py
-      - run: python -m pytest modin/pandas/test/test_io.py --verbose
       - run: python -m pytest modin/experimental/pandas/test/test_io_exp.py
       - run: pip install "dfsql>=0.4.2" && pytest modin/experimental/sql/test/test_sql.py
       - run: |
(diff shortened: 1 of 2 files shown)


anmyachev:
> @anmyachev , do we really spend a lot of time when restarting CI? How much time do we save with the changes in this PR? Not sure I like this code duplication and about similar names of the CI jobs.

Every time when we restart checks - we increase the total time spent on CI checks by >2 hours. Sometimes we have to restart CI checks several times.

Total duration of `io-ci` checks with failure is ~40 minutes. I think we will spend **at least three times less** time.

Note that we already have a fair amount of code duplication when comparing `ci.yml` and `push.yml` files.

Garra1980:
In a couple of weeks. hopefully

anmyachev:
> > UPD: in addition, we can remove lint checks from `io-ci.yml` file, in the case there will be much less duplication, but this also means that the remaining IO checks will _always_ run.
> 
> Yes, I think the lint checks can be excluded out of `io-ci.yml`. After that I would like to look at CI and its jobs.

Ready for review

vnlitvinov:
@anmyachev what is the status of this?

anmyachev:
> @anmyachev what is the status of this?

I think it makes no sense to separate `io` tests now, because after https://github.com/modin-project/modin/pull/3575, they will become more stable.

PR #3560, opened 2021-10-17
FIX-#3559: remove support of 'S3://...' like paths

### LINKED ISSUE #3559: BUG: Modin currently support s3 paths that start with "S3" and "s3" prefixes
The main problem is that this behavior is different from `pandas` and `fsspec`.

I suggest removing the logic to maintain this behavior in `Modin`. If necessary, we can add this logic to `fsspec` library, but take into account the need for compatibility with `pandas`.

anmyachev:

## What do these changes do?

- [ ] passes `flake8 modin`
- [x] Resolves #3559 
- [ ] tests added and passing

### DIFF
--- a/modin/core/io/file_dispatcher.py
+++ b/modin/core/io/file_dispatcher.py
@@ -24,7 +24,7 @@ import re
 from modin.config import Backend
 import numpy as np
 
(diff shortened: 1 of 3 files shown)


YarShev:
> I don't think we need this if we're fixing #3572

Oh, exactly.

Issue #2462, opened 2020-11-19
read_parquet  : pyarrow.lib.ArrowInvalid: Field named 'item' not found


raphaelauv:
### System information
ubuntu 20.04
Python 3.8.5
modin 0.8.2

```python
import modin.pandas as pd

def test_get_dataframe():
    test_df = pd.DataFrame({
        'col1': [['ad_1', 'ad_2'], ['ad_3']],
    })

    with tempfile.TemporaryDirectory() as directory:
        path = f'{directory}/data'
        os.makedirs(path)
        Path(f'{path}/_SUCCESS').touch()
        test_df.to_parquet(path + '/part-00000.parquet')
        read_df = pd.read_parquet(path)
        pandas.testing.assert_frame_equal(test_df, read_df)
```

throw 

`E   pyarrow.lib.ArrowInvalid: Field named 'item' not found or not unique in the schema.
`

It work well if you replace import `modin.pandas as pd `by import `pandas as pd`

PR #3573, opened 2021-10-19
FEAT-#626: Add `to_parquet` for Ray execution engine

### LINKED ISSUE #626: to_parquet is not supported
### System information

```
$ cat /etc/lsb-release 
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=18.04
DISTRIB_CODENAME=bionic
DISTRIB_DESCRIPTION="Ubuntu 18.04.2 LTS"
$ conda --version
conda 4.6.14
$ python --version
Python 3.7.3
$ pip --version
pip 19.1 from /home/dlweber/miniconda3/envs/gis-dataprocessing/lib/python3.7/site-packages/pip (python 3.7)

$ pip freeze | grep modin
modin==0.5.0
$ pip freeze | grep pandas
pandas==0.24.2
$ pip freeze | grep numpy
numpy==1.16.3
```

miniconda3 was used to install most of the sci-py stack, with a pip clause to add modin, e.g.
```
# environment.yaml
channels:
  - conda-forge
  - defaults

dependencies:
  - python>=3.7
  - affine
  - configobj
  - dask
  - numpy
  - pandas
  - pyarrow
  - rasterio
  - s3fs
  - scikit-learn
  - scipy
  - shapely
  - xarray
  - pip
  - pip:
    - modin
```

### Describe the problem

https://modin.readthedocs.io/en/latest/pandas_supported.html says `to_parquet` is supported, but maybe not:

```
import numpy as np
import modin.pandas as pd
size = (1, 10 * 10)
column_ij = ["%04d_%04d" % (i, j) for i in range(10) for j in range(10)]
data = np.random.randint(0, 10000, size=size, dtype="uint16")
df = pd.DataFrame(data, columns=column_ij)
df.to_parquet('/tmp/tmp.parquet')
```

```
UserWarning: `DataFrame.to_parquet` defaulting to pandas implementation.
```

More details:
```
2019-05-21 16:03:46,207	WARNING worker.py:1337 -- WARNING: Not updating worker name since `setproctitle` is not installed. Install this with `pip install setproctitle` (or ray[debug]) to enable monitoring of worker processes.
2019-05-21 16:03:46,207	INFO node.py:469 -- Process STDOUT and STDERR is being redirected to /tmp/ray/session_2019-05-21_16-03-46_18437/logs.
2019-05-21 16:03:46,310	INFO services.py:407 -- Waiting for redis server at 127.0.0.1:55558 to respond...
2019-05-21 16:03:46,418	INFO services.py:407 -- Waiting for redis server at 127.0.0.1:41726 to respond...
2019-05-21 16:03:46,420	INFO services.py:804 -- Starting Redis shard with 2.1 GB max memory.
2019-05-21 16:03:46,426	INFO node.py:483 -- Process STDOUT and STDERR is being redirected to /tmp/ray/session_2019-05-21_16-03-46_18437/logs.
2019-05-21 16:03:46,427	WARNING services.py:1304 -- WARNING: The object store is using /tmp instead of /dev/shm because /dev/shm has only 5238738944 bytes available. This may slow down performance! You may be able to free up space by deleting files in /dev/shm or terminating any running plasma_store_server processes. If you are inside a Docker container, you may need to pass an argument with the flag '--shm-size' to 'docker run'.
2019-05-21 16:03:46,427	INFO services.py:1427 -- Starting the Plasma object store with 6.0 GB memory using /tmp.
UserWarning: `DataFrame.to_parquet` defaulting to pandas implementation.
```

Maybe modin could be added to conda-forge so that conda can help with resolving version dependencies?

naren-ponder:

## What do these changes do?

**DRAFT PR IN PROGRESS**

- [x] passes `flake8 modin`
- [x] Resolves #626  
- [x] tests added and passing

### DIFF
--- a/modin/core/execution/ray/generic/io/io.py
+++ b/modin/core/execution/ray/generic/io/io.py
@@ -196,3 +196,100 @@ class RayIO(BaseIO):
         for rows in result:
             for partition in rows:
                 wait([partition.oid])
+
+    @staticmethod
+    def _to_parquet_check_support(kwargs):
+        """
+        Check if parallel version of ``to_parquet`` could be used.
+
+        Parameters
+        ----------
+        kwargs : dict
+            Keyword arguments passed to ``.to_parquet()``.
+
+        Returns
+        -------
+        bool
+            Whether parallel version of ``to_parquet`` is applicable.
+        """
+        path = kwargs["path"]
+        engine = kwargs["engine"]
+        compression = kwargs["compression"]
+        if not isinstance(path, str):
+            return False
+        if any((path.endswith(ext) for ext in [".gz", ".bz2", ".zip", ".xz"])):
+            return False
+        if engine != "auto":
+            return False
+        if compression is None or not compression == "snappy":
+            return False
+        return True
+
+    @classmethod
+    def to_parquet(cls, qc, **kwargs):
+        """
+        Write a ``DataFrame`` to the binary parquet format.
+
+        Parameters
+        ----------
+        qc : BaseQueryCompiler
+            The query compiler of the Modin dataframe that we want to run ``to_parquet`` on.
+        **kwargs : dict
+            Parameters for ``pandas.to_parquet(**kwargs)``.
+        """
+        if not cls._to_parquet_check_support(kwargs):
+            return BaseIO.to_parquet(qc, **kwargs)
+
+        # The partition id will be added to the queue, for which the moment
+        # of writing to the file has come
+        queue = Queue(maxsize=1)
+
+        def func(df, **kw):
+            """
+            Dump a chunk of rows as parquet, then save them to target maintaining order.
+
+            Parameters
+            ----------
+            df : pandas.DataFrame
+                A chunk of rows to write to a parquet file.
+            **kw : dict
+                Arguments to pass to ``pandas.to_parquet(**kw)`` plus an extra argument
+                `partition_idx` serving as chunk index to maintain rows order.
+            """
+            from modin.distributed.dataframe.pandas import unwrap_partitions
+            partitions = unwrap_partitions(df, axis=0)
+            for i, part in enumerate(partitions):
+                # TODO: Test changing kwargs["partition_cols"]
+                df.to_parquet(**kwargs)
+                # TODO for naren-ponder: Figure out this part for to_parquet
+
+            # each process waits for its turn to write to a file;
+            # in case of violation of the order of receiving messages from the queue,
+            # the message is placed back
+            while True:
+                get_value = queue.get(block=True)
+                if get_value == kw["partition_idx"]:
+                    break
+                queue.put(get_value)
+
+            # signal that the next process can start writing to the file
+            queue.put(get_value + 1)
+
+            # used for synchronization purposes
+            return pandas.DataFrame()
+
+        # signaling that the partition with id==0 can be written to the file
+        queue.put(0)
+        result = qc._modin_frame._partition_mgr_cls.map_axis_partitions(
+            axis=0,
+            partitions=qc._modin_frame._partitions,
+            map_func=func,
+            keep_partitioning=True,
+            lengths=None,
+            enumerate_partitions=True,
+        )
+
+        # pending completion
+        for rows in result:
+            for partition in rows:
+                wait([partition.oid])
--- a/modin/core/execution/ray/generic/io/io.py
+++ b/modin/core/execution/ray/generic/io/io.py
@@ -256,7 +256,8 @@ class RayIO(BaseIO):
             partitions = unwrap_partitions(df, axis=0)
             for i, _ in enumerate(partitions):
                 output_path = kwargs["path"]
-                kwargs["path"] = f"{output_path}/part-{i:04d}.snappy.parquet"
+                compression = kwargs["compression"]
+                kwargs["path"] = f"{output_path}/part-{i:04d}.{compression}.parquet"
                 df.to_parquet(**kwargs)
 
         # signaling that the partition with id==0 can be written to the file
--- a/modin/core/execution/ray/generic/io/io.py
+++ b/modin/core/execution/ray/generic/io/io.py
@@ -261,7 +261,6 @@ class RayIO(BaseIO):
             df.to_parquet(**kwargs)
             return pandas.DataFrame()
 
-        # signaling that the partition with id==0 can be written to the file
         result = qc._modin_frame._partition_mgr_cls.map_axis_partitions(
             axis=1,
             partitions=qc._modin_frame._partitions,
--- a/modin/core/execution/ray/generic/io/io.py
+++ b/modin/core/execution/ray/generic/io/io.py
@@ -255,8 +255,11 @@ class RayIO(BaseIO):
             partition_idx = kw["partition_idx"]
             if not os.path.exists(output_path):
                 os.makedirs(output_path)
-            kwargs["path"] = f"{output_path}/part-{partition_idx:04d}.{compression}.parquet"
+            kwargs[
+                "path"
+            ] = f"{output_path}/part-{partition_idx:04d}.{compression}.parquet"
             df.to_parquet(**kwargs)
+            return pandas.DataFrame()
 
         # signaling that the partition with id==0 can be written to the file
         qc._modin_frame._partition_mgr_cls.map_axis_partitions(
--- a/modin/core/execution/ray/generic/io/io.py
+++ b/modin/core/execution/ray/generic/io/io.py
@@ -201,17 +201,17 @@ class RayIO(BaseIO):
     @staticmethod
     def _to_parquet_check_support(kwargs):
         """
-        Check if parallel version of ``to_parquet`` could be used.
+        Check if parallel version of `to_parquet` could be used.
 
         Parameters
         ----------
         kwargs : dict
-            Keyword arguments passed to ``.to_parquet()``.
+            Keyword arguments passed to `.to_parquet()`.
 
         Returns
         -------
         bool
-            Whether parallel version of ``to_parquet`` is applicable.
+            Whether parallel version of `to_parquet` is applicable.
         """
         path = kwargs["path"]
         compression = kwargs["compression"]
@@ -231,9 +231,9 @@ class RayIO(BaseIO):
         Parameters
         ----------
         qc : BaseQueryCompiler
-            The query compiler of the Modin dataframe that we want to run ``to_parquet`` on.
+            The query compiler of the Modin dataframe that we want to run `to_parquet` on.
         **kwargs : dict
-            Parameters for ``pandas.to_parquet(**kwargs)``.
+            Parameters for `pandas.to_parquet(**kwargs)`.
         """
         if not cls._to_parquet_check_support(kwargs):
             return BaseIO.to_parquet(qc, **kwargs)
@@ -247,7 +247,7 @@ class RayIO(BaseIO):
             df : pandas.DataFrame
                 A chunk of rows to write to a parquet file.
             **kw : dict
-                Arguments to pass to ``pandas.to_parquet(**kw)`` plus an extra argument
+                Arguments to pass to ``pandas.to_parquet(**kwargs)`` plus an extra argument
                 `partition_idx` serving as chunk index to maintain rows order.
             """
             output_path = kwargs["path"]
--- a/modin/core/execution/ray/generic/io/io.py
+++ b/modin/core/execution/ray/generic/io/io.py
@@ -14,6 +14,7 @@
 """The module holds base class implementing required I/O over Ray."""
 
 import io
+import os
 import pandas
 
 from modin.core.io import BaseIO
@@ -213,13 +214,10 @@ class RayIO(BaseIO):
             Whether parallel version of ``to_parquet`` is applicable.
         """
         path = kwargs["path"]
-        engine = kwargs["engine"]
         compression = kwargs["compression"]
         if not isinstance(path, str):
             return False
         if any((path.endswith(ext) for ext in [".gz", ".bz2", ".zip", ".xz"])):
-            return False
-        if engine != "auto":
             return False
         if compression is None or not compression == "snappy":
             return False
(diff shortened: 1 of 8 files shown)


modin/core/execution/ray/generic/io/io.py
```diff
@@ -196,3 +197,77 @@ def func(df, **kw):
         for rows in result:
             for partition in rows:
                 wait([partition.oid])
+
+    @staticmethod
+    def _to_parquet_check_support(kwargs):
+        """
+        Check if parallel version of ``to_parquet`` could be used.
```
YarShev, line 204:
```suggestion
        Check if parallel version of `to_parquet` could be used.
```
We use one back stick for methods and two ones for classes.

modin/core/execution/ray/generic/io/io.py
```diff
@@ -196,3 +197,77 @@ def func(df, **kw):
...
+
+        Parameters
+        ----------
+        kwargs : dict
+            Keyword arguments passed to ``.to_parquet()``.
```
YarShev, line 209:
```suggestion
            Keyword arguments passed to `.to_parquet()`.
```

modin/core/execution/ray/generic/io/io.py
```diff
@@ -196,3 +197,77 @@ def func(df, **kw):
...
+
+        Returns
+        -------
+        bool
+            Whether parallel version of ``to_parquet`` is applicable.
```
YarShev, line 214:
```suggestion
            Whether parallel version of `to_parquet` is applicable.
```

modin/core/execution/ray/generic/io/io.py
```diff
@@ -196,3 +197,77 @@ def func(df, **kw):
...
+        if not isinstance(path, str):
+            return False
+        if any((path.endswith(ext) for ext in [".gz", ".bz2", ".zip", ".xz"])):
+            return False
+        if compression is None or not compression == "snappy":
+            return False
+        return True
+
+    @classmethod
+    def to_parquet(cls, qc, **kwargs):
```
YarShev, line 227: We should probably add a parallel version of `to_parquet` for Dask engine too since we have parallel `read_parquet` both for Ray and Dask.

modin/core/execution/ray/generic/io/io.py
```diff
@@ -196,3 +197,77 @@ def func(df, **kw):
...
+        """
+        Write a ``DataFrame`` to the binary parquet format.
+
+        Parameters
+        ----------
+        qc : BaseQueryCompiler
+            The query compiler of the Modin dataframe that we want to run ``to_parquet`` on.
```
YarShev, line 234:
```suggestion
            The query compiler of the Modin dataframe that we want to run `to_parquet` on.
```

modin/core/execution/ray/generic/io/io.py
```diff
@@ -196,3 +197,77 @@ def func(df, **kw):
...
+        **kwargs : dict
+            Parameters for ``pandas.to_parquet(**kwargs)``.
```
YarShev, line 236:
```suggestion
            Parameters for `pandas.to_parquet(**kwargs)`.
```

modin/core/execution/ray/generic/io/io.py
```diff
@@ -196,3 +197,77 @@ def func(df, **kw):
...
+        def func(df, **kw):
+            """
+            Dump a chunk of rows as parquet, then save them to target maintaining order.
+
+            Parameters
+            ----------
+            df : pandas.DataFrame
+                A chunk of rows to write to a parquet file.
+            **kw : dict
+                Arguments to pass to ``pandas.to_parquet(**kw)`` plus an extra argument
```
YarShev, line 250:
```suggestion
                Arguments to pass to `pandas.to_parquet(**kwargs)` plus an extra argument
```

modin/core/execution/ray/generic/io/io.py
```diff
@@ -196,3 +197,77 @@ def func(df, **kw):
...
+        ----------
+        qc : BaseQueryCompiler
+            The query compiler of the Modin dataframe that we want to run ``to_parquet`` on.
+        **kwargs : dict
+            Parameters for ``pandas.to_parquet(**kwargs)``.
+        """
+        if not cls._to_parquet_check_support(kwargs):
+            return BaseIO.to_parquet(qc, **kwargs)
+
+        def func(df, **kw):
```
YarShev, line 241: Where are `**kw`` used?

modin/pandas/dataframe.py
```diff
@@ -2195,23 +2195,26 @@ def to_parquet(
         compression="snappy",
         index=None,
         partition_cols=None,
-        storage_options: StorageOptions = None,
```
YarShev, line 2198: Why is this removed?

modin/pandas/test/test_io.py
```diff
@@ -95,6 +96,33 @@ def setup_clipboard(row_size=NROWS):
     df.to_clipboard()
 
 
+def parquet_eval_to_file(modin_obj, pandas_obj, fn, extension, **fn_kwargs):
+    """Helper function to test `to_<extension>` methods.
```
YarShev, line 100: The docstring doesn't correspond to numpydoc.

modin/pandas/test/test_io.py
```diff
@@ -1259,7 +1287,7 @@ def test_to_parquet(self):
         reason="The reason of tests fail in `cloud` mode is unknown for now - issue #3264",
     )
     def test_read_parquet_2462(self):
-        test_df = pd.DataFrame(
+        test_df = pandas.DataFrame(
```
YarShev, line 1290: Why is this replaced to pandas usage?

modin/core/execution/ray/generic/io/io.py
```diff
@@ -196,3 +197,77 @@ def func(df, **kw):
...
+            partition_idx = kw["partition_idx"]
+            if not os.path.exists(output_path):
+                os.makedirs(output_path)
+            kwargs[
+                "path"
+            ] = f"{output_path}/part-{partition_idx:04d}.{compression}.parquet"
+            df.to_parquet(**kwargs)
+            return pandas.DataFrame()
+
+        # signaling that the partition with id==0 can be written to the file
```
YarShev, line 264: I guess you copied the code of recent implementation of `read_csv`. Please, make sure that parallel `to_parquet` is correct.

modin/core/execution/ray/generic/io/io.py
naren-ponder: It is used to get the partition_idx for the output file name.

modin/pandas/dataframe.py
naren-ponder: I see that the default is None is the Pandas docs here (https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.to_parquet.html?highlight=to_parquet#pandas.DataFrame.to_parquet), did not understand the need for adding the `StorageOptions`

modin/pandas/test/test_io.py
naren-ponder: We are testing `read_parquet` here, so it follows as a test that we test the read on a Pandas Dataframe because the `read_parquet` method is Modin's.

modin/core/execution/ray/generic/io/io.py
naren-ponder: Yes, `to_parquet` is writing to files in parallel and I have tested both manually and the ray tests on TeamCity pass.

modin/core/execution/ray/generic/io/io.py
devin-petersohn: @YarShev it can happen in a different PR. I think it is probably better. to separate them

modin/pandas/test/test_io.py
devin-petersohn: To expand on what @naren-ponder said, the new Modin `to_parquet` does not reproduce the bug that we fixed in #2462. It was incorrect written at first, but not caught because Modin was using pandas `to_parquet`

modin/core/execution/ray/generic/io/io.py
devin-petersohn: @YarShev it is well tested at this point.

modin/core/execution/ray/generic/io/io.py
devin-petersohn: It is documented below, I think it is ok. Do you have any issues @YarShev?

modin/pandas/dataframe.py
devin-petersohn: @naren-ponder We can probably add it back. The actual implementation of pandas uses it as a type hint: https://github.com/pandas-dev/pandas/blob/v1.3.4/pandas/core/frame.py#L2586

I think it probably should be `Optional[StorageOptions]` in pandas, but they choose to make it optional all the time which is where I think the confusion is coming from: https://github.com/pandas-dev/pandas/blob/v1.3.4/pandas/_typing.py#L177

modin/core/execution/ray/generic/io/io.py
YarShev: I don't mind, but the PR title needs to be corrected.

modin/core/execution/ray/generic/io/io.py
YarShev: Okay, I see

modin/core/execution/ray/generic/io/io.py
YarShev: What is the comment above for if we don't perform any synchronization. Actually, why don't we need to perform the synchronization?

modin/pandas/dataframe.py
YarShev: Suggest adhering to pandas.

modin/pandas/test/test_io.py
```diff
@@ -95,6 +96,33 @@ def setup_clipboard(row_size=NROWS):
     df.to_clipboard()
 
 
+def parquet_eval_to_file(modin_obj, pandas_obj, fn, extension, **fn_kwargs):
+    """Helper function to test `to_parquet` method.
```
YarShev, line 100: The docstring is not numpydoc-like style yet. Please, fix it. You can look at how you document `to_parquet` as an example. https://github.com/modin-project/modin/pull/3573/files#diff-d6ab859ea39dde23e5d8b784b25150c9cf1b52b7f7d9233b99fd355145127130R228

modin/pandas/test/test_io.py
YarShev: ok, makes sense to me

modin/core/execution/ray/generic/io/io.py
naren-ponder: Ah that comment is not necessary, removed it. Parquet is a serialized way of storing a Pandas DataFrame (i.e. you are writing the exact representation from the dataframe to disk so the types of columns and indices are the same). So a synchronization process for different partitions is not necessary.

YarShev: Okay, I see

modin/pandas/test/test_io.py
YarShev: We have `eval_to_file` to check `to_*` functions. Why do we need this specific function for `to_parquet`?

modin/pandas/dataframe.py
```diff
@@ -2197,21 +2197,24 @@ def to_parquet(
...
+
+        config = {
+            "path": path,
+            "engine": engine,
+            "compression": compression,
+            "index": index,
+            "partition_cols": partition_cols,
+            "storage_options": storage_options,
+        }
+        new_query_compiler = self._query_compiler
```
YarShev, line 2210: What is this assignment for?

modin/pandas/test/test_io.py
naren-ponder: Because modin to_parquet implementation creates multiple `.parquet` files (one for each partition) that needs to be checked against pandas which creates a single `.parquet` file. The other `to_*` modin implementations create a single file that can be checked against pandas, but for `to_parquet` we have to have specific code to test reading and writing entire directories and checking them against a single. Additionally, tearing these directories/files down at the end of the test requires specific code not handled by `eval_to_file`.

modin/pandas/dataframe.py
naren-ponder: This is similar to to_csv where we get the query compiler from the class itself then pass it to the Factory Dispatcher. In theory, `self._query_compiler` could be passed directly to the factory dispatcher without this assignment, but I think this is a bit cleaner and is similar to other portions of the codebase.

YarShev: Okay, I see

modin/pandas/test/test_io.py
YarShev: makes sense

modin/pandas/test/test_io.py
```diff
@@ -96,6 +97,38 @@ def setup_clipboard(row_size=NROWS):
...
+    ----------
+        modin_obj: pd.DataFrame
+            Modin DataFrame or Series to test `to_parquet` method.
+        pandas_obj: pandas.DataFrame
+            Pandas DataFrame or Series to test `to_parquet` method.
+        fn: str
+            Name of the method, that should be tested.
+        extension: str
+            Extension of the test file.
+    """
```
YarShev, line 113:
```suggestion
    """
    Helper function to test `to_parquet` method.

    Parameters
    ----------
    modin_obj : pd.DataFrame
        Modin DataFrame or Series to test `to_parquet` method.
    pandas_obj: pandas.DataFrame
        Pandas DataFrame or Series to test `to_parquet` method.
    fn : str
        Name of the method, that should be tested.
    extension : str
        Extension of the test file.
    """
```
Since we do not check test files for docstrings CI passes but the docstring doesn't correspond yet to our guide. We should probably fix all docstrings in tests at some point.

YarShev (commented):
Left a comment. All the rest looks good.

YarShev (approved):
@naren-ponder , LGTM, thanks!

devin-petersohn (approved):
LGTM, thanks for getting this in @naren-ponder!

modin/pandas/test/test_io.py
anmyachev: @devin-petersohn @YarShev if the behavior of the Modin function differs from the behavior of the pandas function, we should add it to the experimental mode, just as it was with `read_pickle_distributed` and `to_pickle_distributed` functions. Why is it done differ in this case?

devin-petersohn: This is still consistent with parquet format, where `read_pickle_distributed` is not consistent with the format. We can still use pandas to read the output of this `parquet` writer, but it wouldn't work to read the directory of pickle files created by `read_pickle_distributed`.

anmyachev: got it, thank you

PR #3575, opened 2021-10-20
FIX-#3551: fix hanging behaviour for to_csv

### LINKED ISSUE #3551: Modin stuck at to_csv operation
### System information
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: macOS Big sur, also tested on WSL with windows 10
- **Modin version** (`modin.__version__`): 0.10.2
- **Python version**: Python 3.8.11

### Describe the problem
I have a post-processing function to arrange forecast results from gluonts to a dataframe.
The function below works well and computes a dataframe. The function output is then saved as a csv file where raylet is getting stuck. (Not even defaulting to pandas)

### Source code / logs
I first initialize ray with a certain number of CPU count. I'm testing Modin with 1 CPU (8 cores)
```
import os
os.environ["MODIN_ENGINE"] = "ray"  # Modin will use Ray
import ray
ray.init(num_cpus=6)
```

Here the post-processing function
```
def transform_outputs(
    kpi_name: str,
    prediction_list: List[gluonts.model.forecast.SampleForecast],
    confidence_intervals: List[float] = [0.5, 0.9]
):
    
    list_dfs = []
    for forecast in tqdm(prediction_list):
        ns, horz = forecast.samples.shape
        dates = pd.date_range(forecast.start_date, freq=forecast.freq, periods=horz)
        forecast_serie = {}
        forecast_serie[f"{kpi_name}_predict"] = forecast.quantile(0.5)
        df_forecast = pd.DataFrame(forecast_serie)
        
        intervals_dfs = []
        for interval in confidence_intervals:
            lower_p = (1.0 - interval) / 2
            upper_p = (1.0 + interval) / 2

            percentage = str(round(interval*100))
            series={}
            series[f"{kpi_name}_lower_predict_{percentage}%"] = forecast.quantile(q=lower_p)
            series[f"{kpi_name}_upper_predict_{percentage}%"] = forecast.quantile(q=upper_p)
            intervals_dfs.append(pd.DataFrame(series))

        df_intervals = pd.concat(intervals_dfs, axis=1)
        df = pd.concat([df_forecast, df_intervals], axis=1)
        df.index = dates
        df['info'] = forecast.item_id
        list_dfs.append(df)

    return pd.concat(list_dfs, axis=0)

df_forecast_group = transform_outputs(ts_name=ts_name, prediction_list=list_forecasts, confidence_intervals=[0.5, 0.8, 0.9])
logger.info('finnished post processing and start saving')
df_forecast_group.to_csv(f"{output_filepath}/{file_name.replace('.csv', '_forecast.csv')}",
                             index=True,
                             date_format='%Y-%m-%d'
                             )
```

anmyachev:

## What do these changes do?

If an exception occurs in one of the tasks, we must complete the rest, which are waiting for messages from `Queue` (communication channel). When using `get` function, this is done automatically.

Another reason for the hang could be the implementation of communication between tasks using one queue. Situations could arise when the same task put and took a message from the queue many times in a row, preventing other tasks from taking the message. To avoid this situation, I propose to use its own communication channel for each task (a separate events).

- [ ] passes `flake8 modin`
- [x] Resolves #3551 
- [ ] tests added and passing

**Performance**:

Version| Master| PR
-|-|-
Pandas| 40,5 sec| 40,5 sec
Modin - 4 cores| 12,4 sec| 12,7 sec
Modin - 8 cores| 7,0 sec| 7,1 sec
Modin - 16 cores| 4,6 sec| 4,5 sec

<details><summary>Benchmark</summary>

```python
from time import time
import os
import numpy as np
import modin.pandas as pd
#import pandas as pd

df = pd.DataFrame(np.random.randint(10**6, size=(375 * 10**4, 32)))

filename = "test_to_csv.csv"
for i in range(5):
    start = time()
    df.to_csv(filename)
    print(f"TO_CSV TIME: {time()-start}")
    os.remove(filename)
```

</details>

### DIFF
--- a/modin/core/execution/ray/common/utils.py
+++ b/modin/core/execution/ray/common/utils.py
@@ -17,6 +17,9 @@ import os
 import sys
 import psutil
 import warnings
+import asyncio
+
+import ray
 
 from modin.config import (
     StorageFormat,
@@ -107,8 +110,6 @@ def initialize_ray(
         What password to use when connecting to Redis.
         If not specified, ``modin.config.RayRedisPassword`` is used.
     """
-    import ray
-
     if not ray.is_initialized() or override_is_cluster:
         cluster = override_is_cluster or IsRayCluster.get()
         redis_address = override_redis_address or RayRedisAddress.get()
@@ -196,3 +197,54 @@ def initialize_ray(
         NPartitions._put(num_gpus)
     else:
         NPartitions._put(num_cpus)
+
+
+@ray.remote
+class SignalActor:  # pragma: no cover
+    """
+    Help synchronize across tasks and actors on cluster.
+
+    For details see: https://docs.ray.io/en/latest/advanced.html?highlight=signalactor#multi-node-synchronization-using-an-actor
+
+    Parameters
+    ----------
+    event_count : int
+        Number of events required for synchronization.
+    """
+
+    def __init__(self, event_count: int):
+        self.events = [asyncio.Event() for _ in range(event_count)]
+
+    def send(self, event_idx: int):
+        """
+        Indicate that event with `event_idx` has occured.
+
+        Parameters
+        ----------
+        event_idx : int
+        """
+        self.events[event_idx].set()
+
+    async def wait(self, event_idx: int):
+        """
+        Wait until event with `event_idx` has occured.
+
+        Parameters
+        ----------
+        event_idx : int
+        """
+        await self.events[event_idx].wait()
+
+    def is_set(self, event_idx: int) -> bool:
+        """
+        Check that event with `event_idx` had occured or not.
+
+        Parameters
+        ----------
+        event_idx : int
+
+        Returns
+        -------
+        bool
+        """
+        return self.events[event_idx].is_set()
--- a/modin/core/execution/ray/generic/io/io.py
+++ b/modin/core/execution/ray/generic/io/io.py
@@ -13,13 +13,13 @@
 
 """The module holds base class implementing required I/O over Ray."""
 
+import asyncio
 import io
 import os
 import pandas
 
 from modin.core.io import BaseIO
-from ray.util.queue import Queue
-from ray import get
+import ray
 
 
 class RayIO(BaseIO):
@@ -118,9 +118,23 @@ class RayIO(BaseIO):
         if not cls._to_csv_check_support(kwargs):
             return BaseIO.to_csv(qc, **kwargs)
 
-        # The partition id will be added to the queue, for which the moment
-        # of writing to the file has come
-        queue = Queue(maxsize=1)
+        @ray.remote(num_cpus=0)
+        class SignalActor:
+            def __init__(self, event_count):
+                self.events = [asyncio.Event() for _ in range(event_count)]
+
+            def send(self, event_idx):
+                self.events[event_idx].set()
+
+            async def wait(self, event_idx, should_wait=True):
+                if should_wait:
+                    await self.events[event_idx].wait()
+
+            def is_set(self, event_idx):
+                return self.events[event_idx].is_set()
+
+        signal_count = len(qc._modin_frame._partitions)
+        signals = SignalActor.remote(signal_count + 1)
 
         def func(df, **kw):
             """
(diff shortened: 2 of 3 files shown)


anmyachev:
@vnlitvinov ready to merge

anmyachev:
@devin-petersohn I think we shouldn't worry about the actor's resources. From: https://docs.ray.io/en/latest/actors.html#resources-with-actors  - "If you do not specify any resource requirements in the actor class’s remote decorator, then by default, the actor will not acquire any resources for its lifetime."

To remove an actor we can rely on https://docs.ray.io/en/latest/actors.html#automatic-termination.

This pull request fixes the hang, but from time to time there are errors https://github.com/modin-project/modin/pull/3575#issuecomment-954932691 that I think relate to Ray. I think it's better for users to get the error as quickly as possible than faced with hang. I think we should merge the PR, and try to fix the rest errors before 0.12 release.

devin-petersohn:
@anmyachev Ok, sounds good.

.github/workflows/ci.yml
```diff
@@ -466,7 +466,12 @@ jobs:
       - run: python -m pytest -n 2 modin/pandas/test/test_groupby.py
       - run: python -m pytest -n 2 modin/pandas/test/test_reshape.py
       - run: python -m pytest -n 2 modin/pandas/test/test_general.py
-      - run: python -m pytest modin/pandas/test/test_io.py --verbose
+      - timeout-minutes: 30
+        run: |
+          python -m pytest modin/pandas/test/test_io.py --verbose
+          while [ $? -ne 0 ]; do
+            python -m pytest modin/pandas/test/test_io.py --verbose
+          done
```
vnlitvinov, line 474: this feels mightily wrong... this would mask _any_ unstable error we might have

I would rather we continue to restart things from time to time (this would serve as the reminder to fix the issues) rather than mask and forget about such issues

vnlitvinov (changes requested):

.github/workflows/ci.yml
anmyachev: you are right, we can also catch timeouts instead of errors

vnlitvinov (approved):
LGTM!

vnlitvinov:
> I think it's better for users to get the error as quickly as possible than faced with hang.

While that is true, I think the behaviour in current PR is actually worse - it would almost silently produce a broken csv file (well, technically it would spit something in the stdout, but who cares?..).

I think we need to verify that csv file is fine somehow or the actors worked. Waiting (or checking) for the `N+1` event which would be fired after last chunk is written could serve as such check.

anmyachev:
> > I think it's better for users to get the error as quickly as possible than faced with hang.
> 
> While that is true, I think the behaviour in current PR is actually worse - it would almost silently produce a broken csv file (well, technically it would spit something in the stdout, but who cares?..).
> 
> I think we need to verify that csv file is fine somehow or the actors worked. Waiting (or checking) for the `N+1` event which would be fired after last chunk is written could serve as such check.

If an exception occurs, it can be seen in the console, in addition, the code execution will stop. In the case when the worker dies unexpectedly, we will still see an exception and the code execution will also stop.

There is a small detail here that the Ray' worker, by default, is **restarted 3 times in case of a crash**, which can lead to a situation where data from one partition will be written to the file several times. To get rid of this, the simplest thing we can do is remove the restart option for functions that write information to files.

vnlitvinov (commented):
Technically this looks correct, but I would request you to split this PR now in two, where the first one adds `remote_options` and this one utilizes it.

vnlitvinov:
@anmyachev please do not pin `aiobotocore` like this, ninja-patching dependencies is bad.

I've filed a separate issue (https://github.com/modin-project/modin/issues/3648) to address s3fs-related failures, I believe it's a better way of handling the problem.

anmyachev:
Blocked by #3659

anmyachev:
Blocked by #3659

modin/core/execution/ray/generic/io/io.py
```diff
@@ -270,4 +276,4 @@ def func(df, **kw):
             lengths=None,
             enumerate_partitions=True,
         )
-        wait([part.oid for row in result for part in row])
+        ray.wait([part.oid for row in result for part in row])
```
anmyachev, line 279: Here, we should use `get` function to handle errors correctly (In a separate PR)

modin/core/execution/ray/generic/io/io.py
```diff
@@ -118,9 +118,23 @@ def to_csv(cls, qc, **kwargs):
...
+        class SignalActor:  # pragma: no cover
+            def __init__(self, event_count):
+                self.events = [asyncio.Event() for _ in range(event_count)]
+
+            def send(self, event_idx):
+                self.events[event_idx].set()
+
+            async def wait(self, event_idx, should_wait=True):
+                if should_wait:
+                    await self.events[event_idx].wait()
```
vnlitvinov, line 131:
```suggestion
            async def wait(self, event_idx):
                await self.events[event_idx].wait()
```
I don't see any point of passing this argument - if it should be `False` one can just _not_ call the method.
Plus I don't see it used anywhere, so let's keep it a little cleaner.

modin/core/execution/ray/generic/io/io.py
```diff
@@ -118,9 +118,23 @@ def to_csv(cls, qc, **kwargs):
         if not cls._to_csv_check_support(kwargs):
             return BaseIO.to_csv(qc, **kwargs)
 
-        # The partition id will be added to the queue, for which the moment
-        # of writing to the file has come
-        queue = Queue(maxsize=1)
+        @ray.remote
+        class SignalActor:  # pragma: no cover
```
vnlitvinov, line 122: I think we should extract it to `core/execution/ray/common/utils.py` as it's not specific to `to_csv()`.

modin/core/execution/ray/generic/io/io.py
```diff
@@ -118,9 +118,23 @@ def to_csv(cls, qc, **kwargs):
...
+
+            async def wait(self, event_idx, should_wait=True):
+                if should_wait:
+                    await self.events[event_idx].wait()
+
+            def is_set(self, event_idx):
+                return self.events[event_idx].is_set()
+
+        signal_count = len(qc._modin_frame._partitions)
+        signals = SignalActor.remote(signal_count + 1)
```
vnlitvinov, line 137:
```suggestion
        signals = SignalActor.remote(len(qc._modin_frame._partitions) + 1)
```
what is the point of declaring a variable to be used once in next line only?

modin/core/execution/ray/generic/io/io.py
```diff
@@ -134,56 +148,51 @@ def func(df, **kw):
                 Arguments to pass to ``pandas.to_csv(**kw)`` plus an extra argument
                 `partition_idx` serving as chunk index to maintain rows order.
             """
-            if kw["partition_idx"] != 0:
+            idx = kw["partition_idx"]
+            _kwargs = kwargs.copy()
```
vnlitvinov, line 152:
```suggestion
            csv_kwargs = kwargs.copy()
```
I suggest renaming the variable for clarity.

Also pls add a small comment on why copy.

modin/core/execution/ray/generic/io/io.py
```diff
@@ -134,56 +148,51 @@ def func(df, **kw):
...
-
-            # each process waits for its turn to write to a file;
-            # in case of violation of the order of receiving messages from the queue,
-            # the message is placed back
-            while True:
-                get_value = queue.get(block=True)
-                if get_value == kw["partition_idx"]:
-                    break
-                queue.put(get_value)
+                _kwargs["path_or_buf"] = io.StringIO()
```
vnlitvinov, line 167:
```python
_kwargs["path_or_buf"] = io.BytesIO() if is_binary else io.StringIO()
```

modin/core/execution/ray/generic/io/io.py
```diff
@@ -134,56 +148,51 @@ def func(df, **kw):
...
-                kwargs["mode"] if is_binary else kwargs["mode"] + "t",
-                encoding=kwargs["encoding"],
-                errors=kwargs["errors"],
-                compression=kwargs["compression"],
-                storage_options=kwargs["storage_options"],
+                _kwargs["mode"] if is_binary else _kwargs["mode"] + "t",
+                encoding=_kwargs["encoding"],
+                errors=_kwargs["errors"],
+                compression=_kwargs["compression"],
+                storage_options=_kwargs["storage_options"],
```
vnlitvinov, line 184: you haven't changed any of those used parameters in `_kwargs`, why don't you just use original `kwargs`?..

modin/core/execution/ray/generic/io/io.py
```diff
@@ -134,56 +148,51 @@ def func(df, **kw):
                 Arguments to pass to ``pandas.to_csv(**kw)`` plus an extra argument
                 `partition_idx` serving as chunk index to maintain rows order.
             """
-            if kw["partition_idx"] != 0:
+            idx = kw["partition_idx"]
```
vnlitvinov, line 151:
```suggestion
            partition_idx = kw["partition_idx"]
```
for the sake of better understanding

modin/pandas/test/test_io.py
```diff
@@ -143,7 +152,11 @@ def eval_to_file(modin_obj, pandas_obj, fn, extension, **fn_kwargs):
     unique_filename_pandas = get_unique_filename(extension=extension)
 
     try:
-        getattr(modin_obj, fn)(unique_filename_modin, **fn_kwargs)
+        try:
+            getattr(modin_obj, fn)(unique_filename_modin, **fn_kwargs)
+        except EXCEPTIONS:
+            getattr(modin_obj, fn)(unique_filename_modin, **fn_kwargs)
```
vnlitvinov, line 158: Pls add a comment on why is this needed.
Also I suggest turning this in a `for _ in range(3)` (or some other number) loop.

vnlitvinov (changes requested):
Mostly fine, just lots of small notes, but I'd like those to be fixed.

anmyachev:
Blocked by #3695

modin/core/execution/ray/generic/io/io.py
anmyachev: https://github.com/modin-project/modin/issues/3695

modin/core/execution/ray/generic/io/io.py
```diff
@@ -148,51 +132,50 @@ def func(df, **kw):
...
+                errors=kwargs["errors"],
+                compression=kwargs["compression"],
+                storage_options=kwargs["storage_options"],
                 is_text=False,
             ) as handles:
                 handles.handle.write(content)
 
             # signal that the next process can start writing to the file
-            ray.get(signals.send.remote(idx + 1))
+            get(signals.send.remote(partition_idx + 1))
```
vnlitvinov, line 173: I think `ray.get()` is clearer, `get()` by itself feels too generic

modin/pandas/test/test_io.py
```diff
@@ -152,10 +152,15 @@ def eval_to_file(modin_obj, pandas_obj, fn, extension, **fn_kwargs):
...
-            getattr(modin_obj, fn)(unique_filename_modin, **fn_kwargs)
+        # parameter `max_retries=0` is set for `to_csv` function on Ray engine,
+        # in order to increase the stability of tests, we repeat the call of
+        # the entire function manually
+        for _ in range(3):
+            try:
+                getattr(modin_obj, fn)(unique_filename_modin, **fn_kwargs)
+            except EXCEPTIONS:
+                continue
+            break
```
vnlitvinov, line 163: please add `else:` clause to the `for` loop and raise an exception in it if all three attempts to create the file failed

vnlitvinov (approved):
LGTM

anmyachev:
@devin-petersohn can we merge the PR?

devin-petersohn (approved):
LGTM

modin/core/execution/ray/common/utils.py
```diff
@@ -196,3 +197,54 @@ def initialize_ray(
         NPartitions._put(num_gpus)
     else:
         NPartitions._put(num_cpus)
+
+
+@ray.remote
+class SignalActor:  # pragma: no cover
```
YarShev, line 203: I would prefer this was in task_wrapper.py or class_wrapper.py file. Maybe, some generic file should be created to put content from task_wrapper.py and `SignalActor` there.

anmyachev: https://github.com/modin-project/modin/issues/3861

Issue #1901, opened 2020-08-07
Groupby mean performance


devin-petersohn:
The performance of `groupby().mean` can be improved.

pyrito:
Already in this epic: https://github.com/modin-project/modin/issues/4315

PR #1902, opened 2020-08-07
REFACTOR-#1901: Improve performance of `groupby.mean`

devin-petersohn:
Signed-off-by: Devin Petersohn <devin.petersohn@gmail.com>

## What do these changes do?

- [ ] commit message follows format outlined [here](https://modin.readthedocs.io/en/latest/contributing.html)
- [ ] passes `flake8 modin`
- [ ] passes `black --check modin`
- [ ] signed commit with `git commit -s` 
- [ ] Resolves #? 
- [ ] tests added and passing

PR #3586, opened 2021-10-21
Attempt to accelerate mean aggregation

### LINKED ISSUE #3585: Improve performance of "mean" groupby aggregation

Currently "mean" aggregation uses `_apply_agg_function` which later calls `broadcast_apply_full_axis`. Its performance on NYC Taxi benchmark is worse than of a single threaded Pandas "mean" implementation. Instead a man-reduce approach could be used to better parallelize the calculations.

gshimansky:
This PR accelerates mean aggregation on NYC Taxi benchmark from 55 seconds to 6 seconds on my workstation.

## What do these changes do?

- [x] Resolves #3585, #1901 
- [x] tests added and passing

### DIFF
--- a/modin/core/storage_formats/pandas/query_compiler.py
+++ b/modin/core/storage_formats/pandas/query_compiler.py
@@ -2495,16 +2495,28 @@ class PandasQueryCompiler(BaseQueryCompiler):
         kwargs["min_count"] = 1
         result = dfgb.sum(**kwargs)
         divisor = dfgb.count()
-        divisor.set_axis(["__mean_agg_size_column__" + x for x in divisor.columns], axis=1, inplace=True)
+        divisor.set_axis(
+            ["__mean_agg_size_column__" + x for x in divisor.columns],
+            axis=1,
+            inplace=True,
+        )
         result = pandas.concat([result, divisor], axis=1, copy=False)
         return result
 
     def _mean_agg_reduce(dfgb, **kwargs):
         kwargs["min_count"] = 1
         result = dfgb.sum(**kwargs)
-        divirgent = result[[x for x in result.columns if not x.startswith("__mean_agg_size_column__")]]
-        divisor = result[[x for x in result.columns if x.startswith("__mean_agg_size_column__")]]
-        divisor.set_axis([x[len("__mean_agg_size_column__"):] for x in divisor.columns], axis=1, inplace=True)
+        divirgent = result[
+            [x for x in result.columns if not x.startswith("__mean_agg_size_column__")]
+        ]
+        divisor = result[
+            [x for x in result.columns if x.startswith("__mean_agg_size_column__")]
+        ]
+        divisor.set_axis(
+            [x[len("__mean_agg_size_column__") :] for x in divisor.columns],
+            axis=1,
+            inplace=True,
+        )
         # Following line makes sure we exclude any "by" columns that could be carried in via "__mean_agg_size_column__"
         # while they shouldn't be present and was removed since map phase was done.
         divisor = divisor[divirgent.columns]
--- a/modin/pandas/groupby.py
+++ b/modin/pandas/groupby.py
@@ -131,7 +131,7 @@ class DataFrameGroupBy(object):
     def mean(self, *args, **kwargs):
         fallback = False
         converted_columns = {}
-        numeric_only=kwargs.get("numeric_only", False)
+        numeric_only = kwargs.get("numeric_only", False)
         if not numeric_only:
             for col, dt in self._query_compiler.dtypes.items():
                 if is_datetime64_any_dtype(dt):
--- a/modin/pandas/test/test_groupby.py
+++ b/modin/pandas/test/test_groupby.py
@@ -1465,6 +1465,7 @@ def test_agg_exceptions(operation):
 
     eval_aggregation(*create_test_dfs(data), operation=operation)
 
+
 @pytest.mark.parametrize("numeric_only", [True, False])
 def test_mean_agg_different_types(numeric_only):
     N = 200
@@ -1477,14 +1478,7 @@ def test_mean_agg_different_types(numeric_only):
             np.datetime64("2009-01-01"),
         ]
         * (N // 5),
-        "int_only": [
-            2000,
-            2010,
-            2011,
-            2012,
-            2009
-        ]
-        * (N // 5),
+        "int_only": [2000, 2010, 2011, 2012, 2009] * (N // 5),
         "int_and_nan": [
             2000,
             2010,
@@ -1523,7 +1517,9 @@ def test_mean_agg_different_types(numeric_only):
 
     data = {**data1, **data2}
 
-    eval_aggregation(*create_test_dfs(data), operation="mean", numeric_only=numeric_only)
+    eval_aggregation(
+        *create_test_dfs(data), operation="mean", numeric_only=numeric_only
+    )
 
 
 @pytest.mark.skip(


modin/pandas/groupby.py
```diff
@@ -128,7 +129,46 @@ def sem(self, ddof=1):
...
+        converted_columns = {}
+        if not kwargs.get("numeric_only", False):
+            for col, dt in self._query_compiler.dtypes.items():
+                if is_datetime64_any_dtype(dt):
+                    if self._df[col].hasnans:
+                        fallback = True
+                        break
+                    else:
+                        converted_columns[col] = dt
+        if fallback:
```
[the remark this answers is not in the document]
gshimansky: I think renaming into `groupby_mean_numeric` is the best approach. Otherwise it is not clear how to handle `mean` aggregation in query_compiler separately from all other kinds of aggregations.

gshimansky:
With `count` used in map operations computation is slower than when `size` was used. I am getting the following timings on Taxi Q2:
|                              |  20M records | 100M records |
|---------------------|----------------|----------------|
|Modin master        |  39.15             |            231.05 |
|Modin accelerated |   6.97             |              26.80 |
|Pandas                   |    6.60             |            45.48   |

modin/core/storage_formats/pandas/query_compiler.py
```diff
@@ -2491,6 +2491,44 @@ def _callable_func(self, func, axis, *args, **kwargs):
     groupby_prod = GroupByReduce.register("prod")
     groupby_sum = GroupByReduce.register("sum")
 
+    def _mean_agg_map(dfgb, **kwargs):
+        kwargs["min_count"] = 1
+        result = dfgb.sum(**kwargs)
+        divisor = dfgb.count()
+        divisor.set_axis(
+            ["__mean_agg_size_column__" + x for x in divisor.columns],
```
dchigarev, line 2499: this breaks on a MultiIndex columns in `divisor`
```python
import modin.pandas as pd

df = pd.DataFrame(
    {
        ("foo", "a"): [1, 1, 2],
        ("foo", "b"): [3, 4, 5],
        ("bar", "a"): [10, 20, 30],
        ("bar", "b"): [40, 50, 60],
    }
)

print(df.groupby(("foo", "a")).mean())

```

<details><summary>Error traceback</summary>

```
  File "/localdisk1/dchigare/repos/modin_bp/modin/core/execution/ray/implementations/pandas_on_ray/partitioning/partition.py", line 479, in apply_list_of_funcs
    partition = func(partition, *args, **kwargs)
  File "/localdisk1/dchigare/repos/modin_bp/modin/core/execution/ray/implementations/pandas_on_ray/partitioning/partition_manager.py", line 169, in map_func
    return apply_func(df, **{other_name: other})
  File "/localdisk1/dchigare/repos/modin_bp/modin/core/dataframe/algebra/groupby.py", line 415, in _map
    result = wrapper(df, other)
  File "/localdisk1/dchigare/repos/modin_bp/modin/core/dataframe/algebra/groupby.py", line 402, in wrapper
    return cls.map(
  File "/localdisk1/dchigare/repos/modin_bp/modin/core/dataframe/algebra/groupby.py", line 139, in map
    result = apply_func(
  File "/localdisk1/dchigare/repos/modin_bp/modin/core/storage_formats/pandas/query_compiler.py", line 2499, in _mean_agg_map
    ["__mean_agg_size_column__" + x for x in divisor.columns],
  File "/localdisk1/dchigare/repos/modin_bp/modin/core/storage_formats/pandas/query_compiler.py", line 2499, in <listcomp>
    ["__mean_agg_size_column__" + x for x in divisor.columns],
TypeError: can only concatenate str (not "tuple") to str

```

</details>

I would recommend using `add_prefix` here instead of manual renaming: `divisor.add_prefix(prefix)`.

dchigarev: Also, I would like the prefix to be stored in some variable to simplify its usage

modin/core/storage_formats/pandas/query_compiler.py
```diff
@@ -2491,6 +2491,44 @@ def _callable_func(self, func, axis, *args, **kwargs):
...
+            [x for x in result.columns if not x.startswith("__mean_agg_size_column__")]
+        ]
+        divisor = result[
+            [x for x in result.columns if x.startswith("__mean_agg_size_column__")]
+        ]
+        divisor.set_axis(
+            [x[len("__mean_agg_size_column__") :] for x in divisor.columns],
+            axis=1,
+            inplace=True,
+        )
```
dchigarev, line 2519: The order of columns among dividend and divisor frames should be the same, so can we just do: `divisor.columns = divirgent.columns.copy()`?

modin/pandas/groupby.py
```diff
@@ -128,7 +129,46 @@ def sem(self, ddof=1):
...
+            # We cannot use map-reduce approach because non-numeric columns are present and
+            # are requested to be processed. It happens because in map-reduce we use "sum"
+            # aggregation which always drops non-numeric columns unconditionally, no matter
+            # what arguments are specified. It is different from how "mean" handles non-numeric
+            # columns (in particular datetime types). We could use a workaround to convert
+            # datetime to int64 but it works only when NaNs aren't present in datetime columns.
+            # NaNs converted to int64 produce wrong results and have to be handled differently,
+            # so we have to resort to less efficient approach of broadcast full axis in
```
dchigarev, line 151: why does `NaNs` converted to int produce wrong results? As far as I know, we can convert `NaN` to `float` only. Does backward converting of `float` -> `int` -> `datetime` causing precision loss and so incorrect results?

modin/pandas/groupby.py
```diff
@@ -128,7 +129,46 @@ def sem(self, ddof=1):
...
+            # _apply_agg_function.
+            result = self._apply_agg_function(lambda df: df.mean(*args, **kwargs))
+        else:
+            if len(converted_columns) > 0:
+                # Convert all datetime64 types to int64 to allow pandas "sum" to work.
+                self._df = self._df.astype(
+                    {col: "int64" for col in converted_columns.keys()}
+                )
```
dchigarev, line 159: I still think that it's pandas-backend specific because of its MapReduce implementation and so should be moved to the QC. Let's wait for other reviewers' opinions though.

modin/pandas/test/test_groupby.py
```diff
@@ -1466,6 +1466,62 @@ def test_agg_exceptions(operation):
     eval_aggregation(*create_test_dfs(data), operation=operation)
 
 
+@pytest.mark.parametrize("numeric_only", [True, False])
+def test_mean_agg_different_types(numeric_only):
```
dchigarev, line 1470: let's also add a `mean` test to the [`test_groupby_multiindex`](https://github.com/modin-project/modin/blob/41213581b974760928ed36f05c479937dcdfea55/modin/pandas/test/test_groupby.py#L1266) to check whether our intermediate columns renaming works with them

dchigarev (changes requested):

dchigarev:
> With `count` used in map operations computation is slower than when `size` was used. I am getting the following timings on Taxi Q2:
> 
> 20M records	100M records
> Modin master	39.15	231.05
> Modin accelerated	6.97	26.80
> Pandas	6.60	45.48

@gshimansky is checking for non-nan columns in order to use a common divisor computed via `size` is also slower than `count`?

modin/pandas/groupby.py
```diff
@@ -128,7 +129,46 @@ def sem(self, ddof=1):
...
+        converted_columns = {}
+        numeric_only = kwargs.get("numeric_only", False)
+        if not numeric_only:
+            for col, dt in self._query_compiler.dtypes.items():
+                if is_datetime64_any_dtype(dt):
+                    if self._df[col].hasnans:
+                        fallback = True
+                        break
+                    else:
+                        converted_columns[col] = dt
```
dchigarev, line 142: in order to minimize backend-API calls I would replace it with something like this:
```python
datetime_cols = self._df[[col for col, dt in self._df.dtypes.items() if is_datetime64_any_dtype(dt)]]
has_nans_mask = datetime_cols.isna().any()
```
I've compared these two approaches and it appeared that the proposed one almost always faster:

<details><summary>Micro-benchmark and its results</summary>

```python
import modin.pandas as pd
import numpy as np
import timeit


def is_dtype_to_pick(dtype):
    return dtype == "float"


def fn1(df):
    to_convert = {}
    fallback = False
    for col, dt in df.dtypes.items():
        if is_dtype_to_pick(dt):
            if df[col].hasnans:
                fallback = True
                break
            else:
                to_convert[col] = dt


def fn2(df):
    cols_to_select = [col for col, dt in df.dtypes.items() if is_dtype_to_pick(dt)]
    if len(cols_to_select) == 0:
        return
    selected_df = df[cols_to_select]
    fallback = selected_df.isna().any().any()
    if not fallback:
        to_convert = selected_df.dtypes.items()


NROWS = 1_000_000
NCOLS = 10

df_no_floats = pd.DataFrame({f"col{i}": np.arange(NROWS) for i in range(NCOLS)})
df_has_half_float_no_nans = df_no_floats.astype({col: "float" for col in df_no_floats.columns[:len(df_no_floats.columns) // 2]})

df_has_half_float_nans_first = df_has_half_float_no_nans.copy()
df_has_half_float_nans_first.loc[0, df_has_half_float_nans_first.select_dtypes("float").columns[0]] = np.nan

df_has_half_float_nans_last = df_has_half_float_no_nans.copy()
df_has_half_float_nans_first.loc[0, df_has_half_float_nans_first.select_dtypes("float").columns[-1]] = np.nan

NRUNS = 5

print("df_no_float:")
print("for-loop:", timeit.timeit(lambda: fn1(df_no_floats), number=NRUNS))
print("non-for-loop:", timeit.timeit(lambda: fn2(df_no_floats), number=NRUNS))

print("\ndf_has_half_float_no_nans:")
print("for-loop:", timeit.timeit(lambda: fn1(df_has_half_float_no_nans), number=NRUNS))
print("non-for-loop:", timeit.timeit(lambda: fn2(df_has_half_float_no_nans), number=NRUNS))

print("\ndf_has_half_float_nans_first:")
print("for-loop:", timeit.timeit(lambda: fn1(df_has_half_float_nans_first), number=NRUNS))
print("non-for-loop:", timeit.timeit(lambda: fn2(df_has_half_float_nans_first), number=NRUNS))

print("\ndf_has_half_float_nans_last:")
print("for-loop:", timeit.timeit(lambda: fn1(df_has_half_float_nans_last), number=NRUNS))
print("non-for-loop:", timeit.timeit(lambda: fn2(df_has_half_float_nans_last), number=NRUNS))

```

Results:
```
df_no_float:
for-loop: 8.46749753691256e-05
non-for-loop: 6.431201472878456e-05

df_has_half_float_no_nans:
for-loop: 5.488216175988782
non-for-loop: 1.856114284018986

df_has_half_float_nans_first:
for-loop: 1.3251252140034921
non-for-loop: 1.966320100997109

df_has_half_float_nans_last:
for-loop: 5.407154903979972
non-for-loop: 1.9547889120294712

```

</details>

YarShev:
@gshimansky, what is the status here?

modin/pandas/groupby.py
gshimansky: Consider the following example:
```
df = pd.DataFrame({"a": [1, 1, 2, 2], "b": [np.datetime64("2001-01-01"), np.NaN, np.datetime64("2001-01-01"), np.datetime64("2002-01-01")]})
```
Here if we convert column "b" to int64 as `df.astype({"b": "int64"})` you can see that NaN in column "b" is converted into value `-9223372036854775808` which affects the final sum of values. If NaN was converted into zero, we could use this approach because effectively NaN values would be ignored. We cannot convert into float64 because pandas gives an error.

gshimansky:
> @gshimansky, what is the status here?

No progress so far. Handling all corner cases introduces too much complexity and slows down the overall performance. I am not sure that I should continue to push this PR.

modin/core/storage_formats/pandas/query_compiler.py
gshimansky: Except divisor and divirgent have different sets of columns because divisor also has "by" columns prefixed with "__mean_agg_size_column__". The following line (with comment above it) makes sure that they are excluded.

YarShev:
> > @gshimansky, what is the status here?
> 
> No progress so far. Handling all corner cases introduces too much complexity and slows down the overall performance. I am not sure that I should continue to push this PR.

IIRC, @mvashishtha was digging into it. Would you be able to continue this work?

Also, I think @dchigarev is always glad to look at groupby 😄

mvashishtha:
@YarShev I believe @naren-ponder was looking at this. Do you have any updates about it?

modin/core/storage_formats/pandas/query_compiler.py
```diff
@@ -2491,6 +2491,44 @@ def _callable_func(self, func, axis, *args, **kwargs):
...
+        result = pandas.concat([result, divisor], axis=1, copy=False)
+        return result
+
+    def _mean_agg_reduce(dfgb, **kwargs):
+        kwargs["min_count"] = 1
+        result = dfgb.sum(**kwargs)
+        divirgent = result[
+            [x for x in result.columns if not x.startswith("__mean_agg_size_column__")]
```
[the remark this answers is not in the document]
gshimansky: Thinking again about this, probably a best approach would be to add a new multiindex level which would distinguish between original columns and result of `.size()` operation. What do you think? I am not sure whether to do it this way in all cases or only in case of MultIndex columns.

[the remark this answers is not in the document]
devin-petersohn: `mean` is `df.sum() / df.count()` (or `df.groupby.sum()`/`df.groupby.count()`), and `size` is just `len` so I don't think it would be correct. Each column has an independent `count`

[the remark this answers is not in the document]
gshimansky: Yes, it should be `count()` instead of `size()` in my comment above (I edited it too for clarity). The comment itself was not about using `size()`, but was about distinguishing new columns from original without adding a string prefix to their names in a multiindex case.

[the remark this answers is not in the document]
devin-petersohn: I see, I have a prototype of this here: https://github.com/modin-project/modin/pull/1902/files

dchigarev:
> Handling all corner cases introduces too much complexity and slows down the overall performance.

I think we should default to pandas on such cases. As far as I understand MultiIndex handling is the hardest thing here, we can just call `self.default_to_pandas(...)` at the QC level if we detect such index. Handling MultiIndex could be a part of a separate issue to not over-complicate the PR.

gshimansky:
> > Handling all corner cases introduces too much complexity and slows down the overall performance.
> 
> I think we should default to pandas in such cases then. As far as I understand MultiIndex handling is the hardest thing here, we can just call `self.default_to_pandas(...)` at the QC level if we detect such index. Handling MultiIndex could be a part of a separate issue to not over-complicate this PR.

MultiIndex doesn't look like a big problem to me. In a comment above I suggest to add an extra layer of MultiIndex which would contain two columns: original df columns and output of `.size()` columns instead of using prefix `__mean_agg_size_column__`. In fact, this approach could be used in case of a single layer index too, this way it won't be necessary to manipulate string names of the columns.

What is slowing down the algorithm is handling of NaNs because `DataFrameGroupBy.sum()` handles them differently from `DataFrameGroupBy.mean()`. Handling of NaNs is the whole reason why `.size()` has to be used instead of `.count()`. Handling of NaNs is the reason why conversion of datimetime type to int64 is required.

modin/pandas/groupby.py
jbrockmendel: in pandas this astype would make a copy, so we'd do `.view('i8')` instead.  is that different in modin?

YarShev:
@prutskov, should we close this PR in favor #4591, which has been merged?

modin/pandas/groupby.py
mvashishtha: > in pandas this astype would make a copy, so we'd do .view('i8') instead. is that different in modin?

No, `astype` creates a copy in Modin, too. Unfortunately, in Modin, `Series.view` does not have the same semantics as in pandas. While it creates a dataframe with the correct type, the new series doesn't share the same underlying values as the original one. Not only that, Modin `view` also creates a copy of all the data under the hood. I've created #4650 to track the differences in behavior.

modin/pandas/groupby.py
jbrockmendel: what would it take to make this case (or just all datetimelike cases) just use the pandas implementation? https://github.com/pandas-dev/pandas/blob/main/pandas/_libs/groupby.pyx#L721

modin/pandas/groupby.py
jbrockmendel: the `datetime_cols = ...` here would make a copy in pandas (not sure about modin).  I suggest `select_dtypes`

modin/pandas/groupby.py
```diff
@@ -128,7 +129,46 @@ def sem(self, ddof=1):
         return self._default_to_pandas(lambda df: df.sem(ddof=ddof))
 
     def mean(self, *args, **kwargs):
-        return self._apply_agg_function(lambda df: df.mean(*args, **kwargs))
+        fallback = False
+        converted_columns = {}
+        numeric_only = kwargs.get("numeric_only", False)
+        if not numeric_only:
+            for col, dt in self._query_compiler.dtypes.items():
+                if is_datetime64_any_dtype(dt):
```
jbrockmendel, line 137: will timedelta64 be affected too?

YarShev:
@dchigarev, would you take a look at this PR? What should we do next, continue or close it?

dchigarev:
> @dchigarev, would you take a look at this PR? What should we do next, continue or close it?

we already did the same changes in #4591

PR #3660, opened 2021-11-06
FIX-#3669: use `kwargs` way to pass additional parameters to the engines for `map_axis_partitions` function

### LINKED ISSUE #3659: Ray tasks are restarted when workers died, this behavior is not always desirable
### System information
- **Modin version** (`modin.__version__`): 8acad95ed2dea2c7671d6dc3518d4b0b12c8b682
- **Link to Ray docs**: https://docs.ray.io/en/latest/fault-tolerance.html#tasks

### Describe the problem

An example of undesirable behavior is when tasks write to disk. In this case, duplication of data may occur. This behavior can be observed for `to_csv` function on Ray engine.

For such cases, we should use `max_retries=0` parameter of `ray.remote` function.

### LINKED ISSUE #3669: Add a way to pass additional parameters to the engines for `map_axis_partitions` function
Through the way, we should be able to pass additional parameters to the engines, for example, `max_retries` for Ray engine, which will help get rid of problems in the implementation of `to_csv` function. See details in #3659.

anmyachev:
## What do these changes do?

Main change in the PR: `**kwargs` that added to `map_axis_partitions` and `broadcast_axis_partitions` functions.

Through this parameter, we can pass additional parameters to the engines. For example, `max_retries` for Ray engine that would be set via `.options` attribute when do remote call https://docs.ray.io/en/latest/advanced.html#dynamic-remote-parameters.

- [x] Resolves #3669 
- [ ] tests added and passing


anmyachev:
@devin-petersohn please review

modin/core/execution/ray/generic/io/io.py
```diff
@@ -190,6 +190,7 @@ def func(df, **kw):
             keep_partitioning=True,
             lengths=None,
             enumerate_partitions=True,
+            max_retries=0,
```
vnlitvinov, line 193: This is a separate change. Please limit this PR to only adding _ability_ to pass extra keywords, and use them separately (you could use that in the `to_csv`-hang-fixing PR after this one is merged).

vnlitvinov (changes requested):

vnlitvinov (approved):
LGTM!

anmyachev:
@devin-petersohn friendly reminder about review :)

devin-petersohn (approved):
LGTM

Issue #3653, opened 2021-11-04
test_io.py flakily times out in Github CI


mvashishtha:
### Describe the problem

test_io.py sometimes times out on CI for PRs that shouldn't be making it much slower, e.g. [here](https://github.com/modin-project/modin/runs/4107837313?check_suite_focus=true) for #3651 and [here](https://github.com/modin-project/modin/runs/4012089345?check_suite_focus=true) for #3573. 

Alternatives are:

1) Make the test faster, e.g. by deleting tests, speeding up tests, or increasing parallelism
2) Increase the timeout from 30 minutes [here](https://github.com/modin-project/modin/blob/master/.github/workflows/ci.yml#L612).

I'll make a PR for option 2) without looking into 1).

mvashishtha:
#3551 is the better fix.

jeffreykennethli:
@anmyachev is there any update to this issue? I'd like to help resolve this task (can take it on if appropriate).

mvashishtha:
@anmyachev mentioned [here](https://github.com/modin-project/modin/pull/3575#issuecomment-962652220) in the fix PR that the fix is blocked by #3659. He is working on #3659 as well.

mvashishtha:
It's too late to officially link https://github.com/modin-project/modin/pull/3575 to this issue, but it's supposed to have fixed this issue.

PR #3663, opened 2021-11-08
FIX-#3616: Remove setup.py platform-specific logic to make distribution pure Python

### LINKED ISSUE #3616: Change setup.py to be pure Python
At one point, we needed to have different distributions for Windows and Unix/Linux. Now, we no longer need that, so it would be good if we could remove the distribution specific logic in the setup.py and in the future create only one wheel (since Modin is pure python).

jeffreykennethli:

## What do these changes do?

Removes distclass argument to setup() in setup.py so that the wheel created is a [Pure Python wheel](https://packaging.python.org/guides/distributing-packages-using-setuptools/#pure-python-wheels) that is platform/os-agnostic.

- [ ] Resolves #3616? 
- [ ] tests added and passing


jeffreykennethli:
Failing check is a ci test that is likely a duplicate of #3653.

PR #3681, opened 2021-11-16
FIX-#3616: Remove setup.py platform-specific logic to make distribution pure Python

### LINKED ISSUE #3616: Change setup.py to be pure Python

jeffreykennethli:
## What do these changes do?

Deletes platform-specific logic in setup.py so that the Python wheel file is [pure python](https://packaging.python.org/guides/distributing-packages-using-setuptools/#pure-python-wheels) so that all platforms can use the same wheel.

- [x] Resolves #3616?
- [ ] tests added and passing

### DIFF
--- a/docs/release-procedure.md
+++ b/docs/release-procedure.md
@@ -58,7 +58,7 @@ Push the commit, make a PR _against `release-X.Y.Z`_ branch and, when it's merge
   * Push the tag to master: `git push upstream X.Y.Z`
     * If you're re-pushing a tag (beware! you shouldn't be doing that, no, _really_!), you can remove remote tag and push a local one by `git push upstream :refs/tags/X.Y.Z`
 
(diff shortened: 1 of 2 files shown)


devin-petersohn (dismissed):
LGTM, will wait for @vnlitvinov to also take a look

vnlitvinov:
@jeffreykennethli please update the "how to release" document as well: https://github.com/modin-project/modin/blob/master/docs/release-procedure.md

vnlitvinov (approved):
LGTM, thanks!

devin-petersohn (approved):

PR #3682, opened 2021-11-16
FIX-#3659: add 'max_retries=0' for to_csv on Ray engine

### LINKED ISSUE #3659: Ray tasks are restarted when workers died, this behavior is not always desirable

anmyachev:

## What do these changes do?

- [x] Resolves #3659 
- [ ] tests added and passing

### DIFF
--- a/modin/core/execution/ray/generic/io/io.py
+++ b/modin/core/execution/ray/generic/io/io.py
@@ -191,6 +191,7 @@ class RayIO(BaseIO):
             keep_partitioning=True,
             lengths=None,
             enumerate_partitions=True,
+            max_retries=0,
         )
 
         # pending completion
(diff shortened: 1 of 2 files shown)


vnlitvinov (approved):
LGTM

PR #3685, opened 2021-11-16
REFACTOR-#3684: remove compression check in '_read_csv_check_support'

### LINKED ISSUE #3684: REFACTOR: remove compression check from ` _read_csv_check_support` function
This [check](https://github.com/modin-project/modin/blob/6178a5764489e05aa85b4bf47a6fecac2c89c53e/modin/core/io/text/csv_dispatcher.py#L205) is used to identify situations where Modin does not support any compression format in order to use the pandas code in these cases. However we can rely on `infer_compression` [function](https://github.com/modin-project/modin/blob/6178a5764489e05aa85b4bf47a6fecac2c89c53e/modin/core/io/text/csv_dispatcher.py#L75) from the pandas(called before checks), which will throw an exception if the compression algorithm is not supported.

anmyachev:

## What do these changes do?

- [x] Resolves #3684 
- [ ] tests added and passing

### DIFF
--- a/modin/core/io/text/csv_dispatcher.py
+++ b/modin/core/io/text/csv_dispatcher.py
@@ -21,7 +21,6 @@ from modin.core.io.text.text_file_dispatcher import (
 import pandas
 from pandas.core.dtypes.common import is_list_like
 from csv import QUOTE_NONE, Dialect
-import sys
 import io
 from typing import Union, Sequence, Callable, Dict, Tuple
 from pandas._typing import FilePathOrBuffer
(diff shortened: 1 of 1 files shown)


vnlitvinov (commented):
What if we use pyarrow to read (or some other library) which might be supporting _less_ potential compressions than pandas?..

anmyachev:
> What if we use pyarrow to read (or some other library) which might be supporting _less_ potential compressions than pandas?..

Modin deals with the issue of opening archives itself(we use `fsspec` for this), so `pyarrow` or some another library will work with unpacked data.

vnlitvinov (approved):
LGTM

vnlitvinov:
It just came to me - what about OmniSci?

anmyachev:
> It just came to me - what about OmniSci?

OmniSci has it' own version of `_read_csv_check_support` function: https://github.com/modin-project/modin/blob/9498b927ab0e1fd31b20ebc3b79dc41dff00fae0/modin/experimental/core/execution/native/implementations/omnisci_on_native/io/io.py#L409

Issue #3531, opened 2021-10-11
Create Head and Tail for ProgressBar estimation and optimized Repr operations


todd-yu:
**Is your feature request related to a problem? Please describe.**

Create Head and Tail for ProgressBar estimation and optimized Repr operations

todd-yu:
Assigned to @todd-yu

RehanSD:
Hi @todd-yu has this issue been resolved?

mvashishtha:
@RehanSD I think that as of modin version 88f7b27d32b099c508e93251cafe503f2908d426, we still don't have progress bars for `head`, `tail`, or `repr`.

todd-yu:
Hey all, nope this was largely a research implementation and isn't fit to be merged into the Modin codebase.

RehanSD:
@todd-yu do we plan to implement this, or should we close this as won't do?

todd-yu:
Let's close this for now, thanks!

PR #3699, opened 2021-11-17
[INCOMPLETE DO-NOT-MERGE] FEAT-#3531: Added MicroPartition Config Variable and variable micro-p…

todd-yu:
Add Configurable Micro Partitions

Signed-off-by: Todd Yu <toddyu@berkeley.edu>

## What do these changes do?

Added environment variable `MicroPartitions` to allow users to configure micro-partitions. Added micropartition support for .parquet files.

- [x] Resolves #? 
- [ ] tests added and passing


vnlitvinov:
Why do we need a separate option? Can't we just use [pandas option](https://pandas.pydata.org/pandas-docs/stable/user_guide/options.html#available-options) `display.min_rows` and its friends for this purpose?

devin-petersohn:
@vnlitvinov We could potentially use some multiple of `display.min_rows`, however this is implemented at the physical partitioning layer, and we should try our best to avoid dependencies on pandas at that layer.

vnlitvinov:
Makes sense. In this case I suggest setting default value of the introduced parameter to a multiple of pandas option - in this case our low level is pandas-independent but "surface" behaviour is aligned.

PR #3707, opened 2021-11-19
FIX-#3706: Deflake TestHdf.test_HDFStore.

### LINKED ISSUE #3706: TestHdf.test_HDFStore is flaky
### System information
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: MacOS Big Sur 11.5.2
- **Modin version** (`modin.__version__`): 0.11.1+57.g75f698cd
- **Python version**: 3.8.8

```
 pytest modin/pandas/test/test_io.py -k test_HDFStore --count 100
```

### Describe the problem
On my mac, the `test_HDFStore` test case fails in 4 out of 100 runs. There have been multiple flakes in github CI, e.g. [here](https://github.com/modin-project/modin/runs/3973959291?check_suite_focus=true), [here](https://github.com/modin-project/modin/runs/4141062378?check_suite_focus=true), and [here](https://github.com/mvashishtha/modin/runs/4242590530?check_suite_focus=true).

mvashishtha:
The HDF files that Modin and Pandas write sometimes have different bytes.
Instead of checking that they are exactly equal, check that they contain
the same dataframes.

I ran:

```
pytest modin/pandas/test/test_io.py -k test_HDFStore --count 100
```

locally and got 4 failures for 100 attempts. After this commit, I got no
failures.

## What do these changes do?

- [x] Resolves #3706 
- [x] tests added and passing

### DIFF
--- a/modin/pandas/test/test_io.py
+++ b/modin/pandas/test/test_io.py
@@ -1622,7 +1622,9 @@ class TestHdf:
 
             modin_store.close()
             pandas_store.close()
-            assert assert_files_eq(unique_filename_modin, unique_filename_pandas)
+            modin_df = pandas.read_hdf(unique_filename_modin, key="foo", mode="r")
+            pandas_df = pandas.read_hdf(unique_filename_pandas, key="foo", mode="r")
+            df_equals(modin_df, pandas_df)
             assert isinstance(modin_store, pd.HDFStore)
 
             handle, hdf_file = tempfile.mkstemp(suffix=".hdf5", prefix="test_read")


modin/pandas/test/test_io.py
```diff
@@ -1622,7 +1622,9 @@ def test_HDFStore(self):
 
             modin_store.close()
             pandas_store.close()
-            assert assert_files_eq(unique_filename_modin, unique_filename_pandas)
```
anmyachev, line 1625: Here, most likely the same problem with duplicate data in case of worker crash: https://github.com/modin-project/modin/issues/3659. In this way, we will simply hide the error.

anmyachev (changes requested):

modin/pandas/test/test_io.py
mvashishtha: Do you mean that the tests are flaky currently because the ray workers sometimes crash while writing these files to disk? I would expect that to cause the checks I'm adding on the `foo` dataframe to fail because the test wouldn't be able to parse the file.

Github says you are requesting a change, but I don't see a change. What do you propose instead to fix the flakiness?

devin-petersohn: @anmyachev There are no worker crashes in this particular test failure, it is a quirk of HDF5 file format. I think this change is a good one, we just need to make sure that the tests will pass now.

devin-petersohn (approved):
LGTM @mvashishtha, thanks!

anmyachev (approved):

modin/pandas/test/test_io.py
anmyachev: ok

devin-petersohn (approved):
LGTM!

modin/pandas/test/test_io.py
YarShev: Why do we read data using pandas but not Modin in this line?

mvashishtha: I used pandas to read the CSV here because this part of the test was checking that modin writes the HDF file correctly, not that it reads the HDF file correctly.

YarShev: I am just wondering why pandas is used to read the data here but below (on 1636 line) Modin is used to read the data?

mvashishtha: Here, the test checks that Modin correctly **writes** an HDF file. It is not checking that Modin reads HDF files correctly. We need to check that the HDF file has been written correctly whether we use Modin or pandas, so we use pandas to read both files.

In line 1636, the test checks that Modin correctly **reads** an HDF file. It is not checking that Modin writes HDF files correctly. We first write a file with pandas, then read it with both Modin and pandas and check that we get the same result.

YarShev: I see, makes sense. @mvashishtha , thanks!

PR #3733, opened 2021-11-24
FIX-#3732: Change doc for native+omnisci

### LINKED ISSUE #3732: Poor work description native+omnisci
Currently doc is a bit confusing for other users.

Rubtsowa:

## What do these changes do?

- [x] Resolves #3732? 
- [x] tests added and passing

### DIFF
--- a/docs/UsingOmnisci/index.rst
+++ b/docs/UsingOmnisci/index.rst
@@ -1,27 +1,6 @@
 OmniSci
 =======
 
-This section describes usage related documents for the OmniSciDB-based engine of Modin.
-
-This engine uses analytical database OmniSciDB_ to obtain high single-node scalability for
-specific set of dataframe operations.
-To enable this engine you could set the following environment variables:
-
-.. code-block:: bash
-
-   export MODIN_ENGINE=native
-   export MODIN_STORAGE_FORMAT=omnisci
-   export MODIN_EXPERIMENTAL=true
-
-or turn it on in source code:
-
-.. code-block:: python
-
-   import modin.config as cfg
-   cfg.Engine.put('native')
-   cfg.StorageFormat.put('omnisci')
-   cfg.IsExperimental.put(True)
-
 To enable ``OmniSci`` engine launch export :
 
 .. code-block:: bash
--- a/docs/UsingOmnisci/index.rst
+++ b/docs/UsingOmnisci/index.rst
@@ -1,7 +1,11 @@
 OmniSci
 =======
 
-To enable ``OmniSci`` engine launch export :
+This section describes usage related documents for the OmniSciDB-based engine of Modin.
+
+This engine uses analytical database OmniSciDB_ to obtain high single-node scalability for
+specific set of dataframe operations.
+To enable this engine you can set the following environment variable:
 
 .. code-block:: bash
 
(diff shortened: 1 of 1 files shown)


docs/UsingOmnisci/index.rst
```diff
@@ -1,27 +1,6 @@
 OmniSci
 =======
 
-This section describes usage related documents for the OmniSciDB-based engine of Modin.
```
YarShev, line 4: My first comment about this file is to rename it to `docs/UsingOmnisciOnNative/index.rst`.
Second, I am still thinking we should provide info on setting both omnisci storage format and native engine. Below we can say
```
There is a more simple way to enable ``OmniSci`` engine. You can only export `MODIN_STORAGE_FORMAT` environment variable and Modin itself defines the rest ones.
```

@Garra1980 , @vnlitvinov , @devin-petersohn , what do you think?

YarShev: @vnlitvinov , @devin-petersohn , what do you think?

vnlitvinov (commented):
I don't understand why something is removed but nothing added...

docs/UsingOmnisci/index.rst
vnlitvinov: I'm siding with @Garra1980 here - there's no need to define both when it's just more verbose way of achieving the same result.

YarShev: We could put a section related only to setting omnisci storage format first and further provide more info on setting `IsExperimental`, `Engine` and `StorageFormat` to give users more understanding about the stuff and why/how it works.

Garra1980: Yes, but that info about setting is not about Omnisci engine(or whatever it is called now), it looks as related to some general part of documentation so we can just put a link to that part in Omnisci section

vnlitvinov: It should go to the docs on how `modin.config` works I think:, maybe even here: https://modin.readthedocs.io/en/latest/flow/modin/config.html#usage-guide

docs/UsingOmnisci/index.rst
```diff
@@ -1,27 +1,6 @@
...
-or turn it on in source code:
-
-.. code-block:: python
-
-   import modin.config as cfg
-   cfg.Engine.put('native')
-   cfg.StorageFormat.put('omnisci')
-   cfg.IsExperimental.put(True)
-
 To enable ``OmniSci`` engine launch export :
```
YarShev, line 4:
```suggestion
This section describes usage related documents for the OmniSciDB-based engine of Modin.

This engine uses analytical database OmniSciDB_ to obtain high single-node scalability for
specific set of dataframe operations.
To enable this engine you can set the following environment variable:
```

YarShev: Also, at the bottom of this page I would add the following:
```
Since OmniSci is run through its native engine Modin itself sets ``MODIN_ENGINE=Native`` and you might not specify it explicitly.
```

docs/UsingOmnisci/index.rst
YarShev: @Garra1980 , not sure what you are saying is related to general part, info about setting respective env vars relates to this page.

@vnlitvinov, I don't quite get what you mean. Please, elaborate.

I proposed my suggestions. What do you think on those?

Garra1980: I meant what @vnlitvinov said - we can put link to usage-guide and not mention explicitly all that env setting stuff on Omnisci page

vnlitvinov (approved):
LGTM, thanks @Rubtsowa and @YarShev!

YarShev (approved):

PR #3735, opened 2021-11-24
DOCS-#3734: remove obvious sentences from dispatcher docstrings

### LINKED ISSUE #3734: remove obvious sentences from dispatchers docstrings
For example: `Inherits some common...`.

Initial found in https://github.com/modin-project/modin/pull/3537#discussion_r737664091.

anmyachev:

## What do these changes do?

- [x] Resolves #3734 
- [ ] tests added and passing

### DIFF
--- a/modin/core/io/pickle/pickle_dispatcher.py
+++ b/modin/core/io/pickle/pickle_dispatcher.py
@@ -21,7 +21,7 @@ from modin.config import NPartitions
 
 
 class PickleExperimentalDispatcher(FileDispatcher):
-    """Class handles utils for reading `.pkl` files."""
+    """Class handles utils for reading pickle files."""
 
     @classmethod
     def _read(cls, filepath_or_buffer, **kwargs):
--- a/modin/core/io/pickle/pickle_dispatcher.py
+++ b/modin/core/io/pickle/pickle_dispatcher.py
@@ -21,11 +21,7 @@ from modin.config import NPartitions
 
 
 class PickleExperimentalDispatcher(FileDispatcher):
-    """
-    Class handles utils for reading `.pkl` files.
-
-    Inherits some common from `FileDispatcher` class.
-    """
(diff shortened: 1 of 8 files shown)


YarShev (commented):
I would prefer to extend docstrings for classes, instead of reducing them.

modin/core/io/pickle/pickle_dispatcher.py
```diff
@@ -21,11 +21,7 @@
 
 
 class PickleExperimentalDispatcher(FileDispatcher):
-    """
-    Class handles utils for reading `.pkl` files.
-
-    Inherits some common from `FileDispatcher` class.
-    """
+    """Class handles utils for reading `.pkl` files."""
```
vnlitvinov, line 24:
```suggestion
    """Class handles utils for reading pickle files."""
```
I don't think `.pkl` is the standard extension, unlinke `.csv` or `.parquet`.

vnlitvinov (approved):
LGTM with minor nit pick.

YarShev:
> I would prefer to extend docstrings for classes, instead of reducing them.

@vnlitvinov , thoughts?

vnlitvinov:
> > I would prefer to extend docstrings for classes, instead of reducing them.
> 
> @vnlitvinov , thoughts?

In general I would agree, but stating "this inherits common features from Foo" where `Foo` is plainly stated as a parent class is "thank you, captain Obvious" kind of useless thing.

I mean, I'm all in for bigger docstrings, but they should be bigger by bringing value, not by re-phrasing `class Bar(Foo)` statement.

vnlitvinov (approved):
LGTM

YarShev:
At some point we may want to generate docs for those objects. By default sphinx doesn't expose base classes and those lines of docstrings might be helpful. An alternative option might be `show-inheritance` to use in such a case. So we'll probably be relying on the latter.

YarShev (approved):

PR #3738, opened 2021-11-25
FEAT-#3724: Converting self.resample_args for Resampler into a dict

### LINKED ISSUE #3724: Converting self.resample_args for Resampler into a dict
self.resample_args for Resampler should be converted into a dict to ensure simplicity of handling resample parameters at the query compiler level.

Rubtsowa:

## What do these changes do?
self.resample_args for Resampler converted into a dict to ensure simplicity of handling resample parameters at the query compiler level.

- [x] Resolves #3724 
- [x] tests added and passing

### DIFF
--- a/modin/pandas/base.py
+++ b/modin/pandas/base.py
@@ -3111,21 +3111,21 @@ class Resampler(object):
         axis = self._dataframe._get_axis_number(axis)
         # FIXME: this should be converted into a dict to ensure simplicity
         # of handling resample parameters at the query compiler level.
-        self.resample_args = [
-            rule,
-            axis,
-            closed,
-            label,
-            convention,
-            kind,
-            loffset,
-            base,
-            on,
-            level,
-            origin,
-            offset,
-        ]
-        self.__groups = self.__get_groups(*self.resample_args)
+        self.resample_args = {
+            "rule": rule,
+            "axis": axis,
+            "closed": closed,
+            "label": label,
+            "convention": convention,
+            "kind": kind,
+            "loffset": loffset,
+            "base": base,
+            "on": on,
+            "level": level,
+            "origin": origin,
+            "offset": offset,
+        }
+        self.__groups = self.__get_groups(**self.resample_args)
 
     def __getitem__(self, key):
         """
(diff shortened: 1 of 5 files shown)


modin/pandas/base.py
```diff
@@ -3111,21 +3111,21 @@ def __init__(
...
-            kind,
-            loffset,
-            base,
-            on,
-            level,
-            origin,
-            offset,
-        ]
-        self.__groups = self.__get_groups(*self.resample_args)
+        self.resample_args = {
```
vnlitvinov, line 3114: pls rename the attribute to be consistent to `self.resample_kwargs`

modin/pandas/base.py
```diff
@@ -3144,9 +3144,7 @@ def __getitem__(self, key):
         """
 
         def _get_new_resampler(key):
-            subset = self._dataframe[key]
-            resampler = type(self)(subset, *self.resample_args)
-            return resampler
+            return Resampler(self._dataframe[key], **self.resample_args)
```
vnlitvinov, line 3147: I think @YarShev was against my simplification - this indeed makes it harder to make a subclass if we create `Resampler` explicitly

vnlitvinov (changes requested):

modin/pandas/base.py
```diff
@@ -3111,21 +3111,21 @@ def __init__(
...
-            kind,
-            loffset,
-            base,
-            on,
-            level,
-            origin,
-            offset,
-        ]
-        self.__groups = self.__get_groups(*self.resample_args)
+        self.resample_kwargs = {
```
YarShev, line 3114: The comment above should be removed.

vnlitvinov (approved):
LGTM

YarShev:
@Rubtsowa , we should probably add the changes to BaseQueryCompiler too since it seems the CI job failed because of that.

vnlitvinov:
> @Rubtsowa , we should probably add the changes to BaseQueryCompiler too since it seems the CI job failed because of that.

I don't see any CI failures here, what are you talking about?

vnlitvinov (approved):
LGTM

YarShev:
All entries `resample_args : list` -> `resample_kwargs : dict` need to be replaced in base QC.

vnlitvinov (approved):
LGTM assuming CI passes!

vnlitvinov:
> All entries `resample_args : list` -> `resample_kwargs : dict` need to be replaced in base QC.

Good catch!

YarShev (approved):

Issue #3741, opened 2021-11-26
IndexError: index 0 is out of bounds for axis 0 with size 0


Lozovskii-Aleksandr:
### System information
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: Windows 10
- **Modin version** (`modin.__version__`): 0.12.0
- **Python version**: 3.8.11
- **Code we can use to reproduce**:

```
set MODIN_EXPERIMENTAL=1
set MODIN_ENGINE=Python
python -m pytest --capture=no --simulate-cloud=normal modin/pandas/test/dataframe/test_default.py::test_kurt_kurtosis[kurtosis-True-True-over_rows_int]
```

### Describe the problem

Occurs in 165 tests.

- modin/pandas/test/dataframe/test_default.py::test_kurt_kurtosis[kurtosis-True-True-over_rows_int]
- modin/pandas/test/dataframe/test_default.py::test_kurt_kurtosis[kurtosis-True-True-over_rows_str]
- modin/pandas/test/dataframe/test_default.py::test_kurt_kurtosis[kurtosis-True-False-over_rows_int]
- modin/pandas/test/dataframe/test_default.py::test_kurt_kurtosis[kurtosis-True-False-over_rows_str]
- modin/pandas/test/dataframe/test_default.py::test_kurt_kurtosis[kurtosis-True-None-over_rows_int]

And also in 160 tests.

### Source code / logs

```
E           ========= Remote Traceback (1) =========
E           Traceback (most recent call last):
E             File "C:\Users\alozovsk\Anaconda3\envs\modin\lib\site-packages\rpyc\core\protocol.py", line 338, in _dispatch_request
E               res = self._HANDLERS[handler](self, *args)
E             File "C:\Users\alozovsk\Anaconda3\envs\modin\lib\site-packages\rpyc\core\protocol.py", line 630, in _handle_call
E               return obj(*args, **dict(kwargs))
E             File "C:\prog\modin\modin\pandas\base.py", line 1554, in kurt
E               return self._reduce_dimension(
E             File "C:\prog\modin\modin\pandas\dataframe.py", line 2878, in _reduce_dimension
E               return Series(query_compiler=query_compiler)
E             File "C:\prog\modin\modin\pandas\series.py", line 160, in __init__
E               self._query_compiler = query_compiler.columnarize()
E             File "C:\prog\modin\modin\core\storage_formats\pandas\query_compiler.py", line 661, in columnarize
E               return self.transpose()
[...]

PR #3742, opened 2021-11-27
FIX-#3741: Added delivery of the data argument to df

Lozovskii-Aleksandr:
Signed-off-by: aleksandr.lozovskii <aleksandr.lozovskii@intel.com>

## What do these changes do?

- [X] commit message follows format outlined [here](https://modin.readthedocs.io/en/latest/contributing.html)
- [X] passes `flake8 modin/ asv_bench/benchmarks scripts/doc_checker.py`
- [X] passes `black --check modin/ asv_bench/benchmarks scripts/doc_checker.py`
- [X] signed commit with `git commit -s` 
- [X] Resolves [#3741](https://github.com/modin-project/modin/issues/3741)
- [X] tests added and passing
- [X] module layout described at `docs/developer/architecture.rst` is up-to-date 

The problem was that in the `cloud` mode modin was created a proxy `DataFrame`, where the element type was not `float64`, as intended, but a proxy object. To solve this problem for `DataFrame`, we will pass the data parameter from the local side to the remote side.

### DIFF
--- a/modin/experimental/cloud/rpyc_proxy.py
+++ b/modin/experimental/cloud/rpyc_proxy.py
@@ -428,9 +428,13 @@ def make_proxy_cls(
 
         def __init__(self, *a, __remote_end__=None, **kw):
             if __remote_end__ is None:
-                if hasattr(self, "_preprocess_init_args"):
-                    a, kw = self._preprocess_init_args(*a, **kw)
-            if __remote_end__ is None:
+                try:
+                    preprocess = object.__getattribute__(self, "_preprocess_init_args")
+                except AttributeError:
+                    pass
+                else:
+                    a, kw = preprocess(*a, **kw)
+
                 __remote_end__ = remote_cls(*a, **kw)
             while True:
                 # unwrap the object if it's a wrapper
--- a/modin/experimental/cloud/rpyc_proxy.py
+++ b/modin/experimental/cloud/rpyc_proxy.py
@@ -428,6 +428,9 @@ def make_proxy_cls(
 
         def __init__(self, *a, __remote_end__=None, **kw):
             if __remote_end__ is None:
+                if hasattr(self, "_preprocess_init_args"):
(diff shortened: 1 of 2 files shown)


modin/experimental/cloud/rpyc_proxy.py
```diff
@@ -427,6 +427,9 @@ class Wrapper(override, origin_cls, metaclass=ProxyMeta):
         __wrapper_remote__ = remote_cls
 
         def __init__(self, *a, __remote_end__=None, **kw):
+            if __remote_end__ is None:
+                if hasattr(self, "_preprocess_init_args"):
+                    a, kw = self._preprocess_init_args(*a, **kw)
             if __remote_end__ is None:
```
vnlitvinov, line 433: do this under single `if` statement, not double

vnlitvinov:
@Lozovskii-Aleksandr what makes this a draft?

Also please add tests in cloud mode which start to pass so they stay passing from now on.

Lozovskii-Aleksandr:
> @Lozovskii-Aleksandr what makes this a draft?

I ran my tests, which found several bugs in this fix.  I would like to fix this before merging. That's why I marked it as a draft.

modin/experimental/cloud/rpyc_proxy.py
```diff
@@ -622,6 +629,27 @@ def iteritems(self):
...
+            cls,
+            data=None,
+            index=None,
+            columns=None,
+            dtype=None,
+            copy=None,
+            query_compiler=None,
+        ):
+
+            (data,) = conn.deliver((data,), {})[0]
```
Lozovskii-Aleksandr, line 643: I specified only `data` because if I add other parameters (except `copy` and `query_compiler`), errors may occur, the essence of which is not yet clear.

vnlitvinov (approved):
LGTM

PR #3751, opened 2021-11-30
TEST-#3750: Create a variable for minimal partition size.

### LINKED ISSUE #3750: Create a variable for minimal partition size.
### Describe the problem
The tests [here](https://github.com/modin-project/modin/blob/72a3e51aeeae3465a1a3076d962aef95af689dde/modin/pandas/test/utils.py#L208-L214) and [here](https://github.com/modin-project/modin/blob/72a3e51aeeae3465a1a3076d962aef95af689dde/modin/pandas/test/dataframe/test_udf.py#L184) check that some functions work when all partitions are full. Per @YarShev [here](https://github.com/modin-project/modin/pull/3693#discussion_r759218409), we should create a common variable for the minimum partition size required to start a new partition.

mvashishtha:

## What do these changes do?

Create a variable for minimal partition size.

- [x] Resolves #3750 
- [x] tests added and passing

### DIFF
--- a/modin/config/envvars.py
+++ b/modin/config/envvars.py
@@ -404,7 +404,7 @@ class OmnisciLaunchParameters(EnvironmentVariable, type=dict):
         return result
 
 
-class MinNumElementsToStartNewPandasPartition(EnvironmentVariable, type=int):
+class MinElementsInPartition(EnvironmentVariable, type=int):
     """
     Minimum number of rows/columns in a single pandas partition split.
 
--- a/modin/config/envvars.py
+++ b/modin/config/envvars.py
@@ -405,10 +405,12 @@ class OmnisciLaunchParameters(EnvironmentVariable, type=dict):
 
 
 class MinNumElementsToStartNewPandasPartition(EnvironmentVariable, type=int):
-    """Minimum number of rows/columns in a single Pandas partition split.
+    """
+    Minimum number of rows/columns in a single pandas partition split.
 
-    Once a partition for a Pandas dataframe has more than this many elements,
-    Modin adds another partition."""
+    Once a partition for a pandas dataframe has more than this many elements,
+    Modin adds another partition.
+    """
 
     varname = "MODIN_MIN_NUM_ELEMENTS_TO_START_NEW_PANDAS_PARTITION"
     default = 32
--- a/modin/config/envvars.py
+++ b/modin/config/envvars.py
@@ -412,7 +412,7 @@ class MinNumElementsToStartNewPandasPartition(EnvironmentVariable, type=int):
     Modin adds another partition.
     """
 
-    varname = "MODIN_MIN_NUM_ELEMENTS_TO_START_NEW_PANDAS_PARTITION"
+    varname = "MODIN_MIN_ELEMENTS_IN_PARTITION"
     default = 32
 
 
--- a/modin/config/envvars.py
+++ b/modin/config/envvars.py
@@ -404,6 +404,16 @@ class OmnisciLaunchParameters(EnvironmentVariable, type=dict):
         return result
 
 
+class MinNumElementsToStartNewPandasPartition(EnvironmentVariable, type=int):
+    """Minimum number of rows/columns in a single Pandas partition split.
+
+    Once a partition for a Pandas dataframe has more than this many elements,
+    Modin adds another partition."""
+
+    varname = "MODIN_MIN_NUM_ELEMENTS_TO_START_NEW_PANDAS_PARTITION"
+    default = 32
+
+
 def _check_vars():
     """
     Check validity of environment variables.
--- a/modin/config/envvars.py
+++ b/modin/config/envvars.py
@@ -404,7 +404,7 @@ class OmnisciLaunchParameters(EnvironmentVariable, type=dict):
         return result
 
 
-class MinElementsInPartition(EnvironmentVariable, type=int):
+class MinPartitionSize(EnvironmentVariable, type=int):
     """
     Minimum number of rows/columns in a single pandas partition split.
 
@@ -412,7 +412,7 @@ class MinElementsInPartition(EnvironmentVariable, type=int):
     Modin adds another partition.
     """
 
-    varname = "MODIN_MIN_ELEMENTS_IN_PARTITION"
+    varname = "MODIN_MIN_PARTITION_SIZE"
     default = 32
 
 
--- a/modin/core/storage_formats/pandas/utils.py
+++ b/modin/core/storage_formats/pandas/utils.py
@@ -13,6 +13,7 @@
 
 """Contains utility functions for frame partitioning."""
 
+from modin.config.envvars import MinNumElementsToStartNewPandasPartition
 import numpy as np
 import pandas
 
@@ -38,7 +39,12 @@ def get_default_chunksize(length, num_splits):
     )
 
 
-def compute_chunksize(df, num_splits, default_block_size=32, axis=None):
+def compute_chunksize(
+    df,
+    num_splits,
+    default_block_size=MinNumElementsToStartNewPandasPartition.get(),
+    axis=None,
+):
     """
     Compute the number of rows and/or columns to include in each partition.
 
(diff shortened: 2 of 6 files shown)


modin/pandas/test/utils.py
```diff
@@ -54,6 +54,10 @@
 # Directory for storing I/O operations test data
 IO_OPS_DATA_DIR = os.path.join(os.path.dirname(__file__), "io_tests_data")
 
+# Once a partition has more than this many elements, Modin uses another
+# partition.
+min_num_elements_to_start_new_partition = 32
```
YarShev, line 59: What I had in mind is we should have a variable to specify default/minimal partition size, put it in `modin/core/storage_formats/pandas/utils.py` and use it where possible. One of the places would be [here](https://github.com/modin-project/modin/blob/72a3e51aeeae3465a1a3076d962aef95af689dde/modin/core/storage_formats/pandas/utils.py#L41) as well, for example.

vnlitvinov: I wonder if we should have this (the parameter in the splitting function) as a configurable variable in `modin.config` so users could tune Modin behaviour specifically to their workloads...

YarShev: Yes, it would probably be a better way.

modin/config/envvars.py
```diff
@@ -404,6 +404,18 @@ def get(self):
...
 
+class MinNumElementsToStartNewPandasPartition(EnvironmentVariable, type=int):
+    """
+    Minimum number of rows/columns in a single pandas partition split.
+
+    Once a partition for a pandas dataframe has more than this many elements,
+    Modin adds another partition.
+    """
+
+    varname = "MODIN_MIN_NUM_ELEMENTS_TO_START_NEW_PANDAS_PARTITION"
```
vnlitvinov, line 415:
```suggestion
    varname = "MODIN_MIN_ELEMENTS_IN_PARTITION"
```
I don't see how users would set such a long name... let's make it shorter.

This should probably be reflected in the class name, too.

modin/core/storage_formats/pandas/utils.py
```diff
@@ -38,7 +39,12 @@ def get_default_chunksize(length, num_splits):
     )
 
 
-def compute_chunksize(df, num_splits, default_block_size=32, axis=None):
+def compute_chunksize(
+    df,
+    num_splits,
+    default_block_size=MinNumElementsToStartNewPandasPartition.get(),
```
vnlitvinov, line 45:
```suggestion
    default_block_size=None,
```
and down below in the function body do smth like
```python
default_block_size = MinNumElementsToStartNewPandasPartition.get() if default_block_size else default_block_size
```
that way it would allow the user to tune this variable during their workloads and to have a dynamic split.

modin/config/envvars.py
mvashishtha: Done.

modin/core/storage_formats/pandas/utils.py
mvashishtha: Done, except I swapped the `if` and `else` in your suggestion for the function body. I think you meant that we should fall back to the environment variable if `default_block_size` isn't specified.

modin/core/storage_formats/pandas/utils.py
```diff
@@ -65,6 +65,9 @@ def compute_chunksize(
         If axis is 1 or 0, returns an integer number of rows/columns to split the
         DataFrame. If axis is None, returns a tuple containing both.
     """
+    default_block_size = (
+        default_block_size if default_block_size else MinElementsInPartition.get()
```
vnlitvinov, line 69: either use `if default_block_size is None` or just simply use `default_block_size = default_block_size or MinElementsInPartition.get()` if `default_block_size=0` should be replaced, too

mvashishtha: I switched to `if default_block_size is None`

vnlitvinov (dismissed):
LGTM!

modin/config/envvars.py
```diff
@@ -404,6 +404,18 @@ def get(self):
         return result
 
 
+class MinElementsInPartition(EnvironmentVariable, type=int):
```
YarShev, line 407: I am still thinking this name might be confusing because it may be understood as minimal amount of all elements held in a partition. I would suggest to use `MinPartitionBlockSize`, `MinPartitionSize` or just `PartitionSize`.

YarShev: I am more tend to the latter.

vnlitvinov: I'm against dropping `Min` here, as it implies that partitions would be of exact this size which is wrong. `MinPartitionSize` is fine by me, though.

modin/pandas/test/dataframe/test_udf.py
```diff
@@ -15,6 +15,7 @@
 import numpy as np
 import pandas
 import matplotlib
+from modin.config.envvars import MinElementsInPartition
```
prutskov, line 18:
```suggestion
from modin.config import MinElementsInPartition
```
All objects from `envvars` are available from `modin.config`

modin/pandas/test/utils.py
```diff
@@ -21,7 +21,7 @@
     assert_index_equal,
     assert_extension_array_equal,
 )
-from modin.config.envvars import NPartitions
+from modin.config.envvars import MinElementsInPartition, NPartitions
```
prutskov, line 24:
```suggestion
from modin.config import MinElementsInPartition, NPartitions
```
All objects from `envvars` are available from `modin.config`

modin/core/storage_formats/pandas/utils.py
```diff
@@ -13,6 +13,7 @@
 
 """Contains utility functions for frame partitioning."""
 
+from modin.config.envvars import MinElementsInPartition
```
prutskov, line 16:
```suggestion
from modin.config import MinElementsInPartition
```
All objects from `envvars` are available from `modin.config`

modin/config/envvars.py
prutskov: I agree with @YarShev, `PartitionSize` will be clearer definition.

devin-petersohn: `MinPartitionSize` is the most clear to me 😄

prutskov (approved):
LGTM assuming tests passed!

modin/core/storage_formats/pandas/utils.py
```diff
@@ -38,7 +39,12 @@ def get_default_chunksize(length, num_splits):
     )
 
 
-def compute_chunksize(df, num_splits, default_block_size=32, axis=None):
+def compute_chunksize(
+    df,
+    num_splits,
+    default_block_size=None,
```
YarShev, line 45: Since default value for this parameter was changed we should correct corresponding line for this parameter in the docstring.
`default_block_size : int, default: 32` -> `default_block_size : int, optional`

vnlitvinov (changes requested):
Please fix docstring per https://github.com/modin-project/modin/pull/3751#discussion_r762604657, then LGTM

modin/core/storage_formats/pandas/utils.py
```diff
@@ -48,7 +54,7 @@ def compute_chunksize(df, num_splits, default_block_size=32, axis=None):
         DataFrame to split.
     num_splits : int
         Number of splits to separate the DataFrame into.
-    default_block_size : int, default: 32
+    default_block_size : int, optional
         Minimum number of rows/columns in a single split.
```
vnlitvinov, line 58:
```suggestion
        Minimum number of rows/columns in a single split.
        If not specified, the value is assumed equal to ``MinPartitionSize``.
```
or something along these lines to indicate where we fill the default from.

mvashishtha:
@vnlitvinov 

> Please fix docstring per [#3751 (comment)](https://github.com/modin-project/modin/pull/3751#discussion_r762604657), then LGTM

Done.

vnlitvinov (approved):
LGTM assuming green CI, thanks!

YarShev (approved):

prutskov (approved):
LGTM

PR #3769, opened 2021-12-02
REFACTOR-#3768: change 'compute_chunksize' signature

### LINKED ISSUE #3768: Update `compute_chunksize` interface; no need to pass the whole DataFrame
Current implementation: https://github.com/modin-project/modin/blob/7a7ae184ff822d8a97b8ba38bcd75c50be06194a/modin/core/storage_formats/pandas/utils.py#L41

If we change the first argument to shape, then this function can be used in the case when the dataframe has not yet been created. In particular, this will help when implementing repartition in https://github.com/modin-project/modin/pull/3720.

anmyachev:

## What do these changes do?

This change increases the number of situations where this feature is applicable.

- [x] Resolves #3768 


vnlitvinov (approved):
LGTM

modin/core/io/column_stores/column_store_dispatcher.py
```diff
@@ -128,9 +128,7 @@ def build_index(cls, partition_ids):
         else:
             index = index_len
             index_len = len(index)
-        index_chunksize = compute_chunksize(
-            pandas.DataFrame(index=index), num_partitions, axis=0
-        )
+        index_chunksize = compute_chunksize((len(index), None), num_partitions, axis=0)
```
devin-petersohn, line 131: This feels a little hacky, instead of passing a tuple should we pass the values separately?

devin-petersohn (commented):
I really like this change @anmyachev, left a comment for discussion.

modin/core/io/column_stores/column_store_dispatcher.py
vnlitvinov: I guess this is done to be able to pass a simple `df.shape` without unpacking it in the caller.

anmyachev: @vnlitvinov exactly

devin-petersohn: I don't think it would be very different to have them as separate variables that can default to `None`. It would be `*df.shape` to unpack.

anmyachev: Suppose these arguments are as follows: `def compute_chunksize(row_count=None, column_count=None, num_splits, ...)`. `num_splits` is a non-optional variable that need to be placed before optional arguments. What do you think?

YarShev: I am okay with this interface. Also, we could make `num_splits` optional too and if it is not specified, `NPartitions.get()` is used by default.

Side note: I wonder should we have `axis` parameter in that case (`row_count`/`col_count` could be enough)?

anmyachev: @devin-petersohn @YarShev the signature is updated.

Previous version is better IMHO.

devin-petersohn (approved):
Thanks for the changes @anmyachev, looks great.

As a side note, should we create some unit tests for these kinds of utilities? (In a separate issue/PR of course)

anmyachev:
> Thanks for the changes @anmyachev, looks great.
> 
> As a side note, should we create some unit tests for these kinds of utilities? (In a separate issue/PR of course)

Highly desirable! :)

modin/core/dataframe/pandas/partitioning/partition_manager.py
```diff
@@ -712,7 +712,7 @@ def update_bar(pbar, f):
 
         num_splits = NPartitions.get()
         put_func = cls._partition_class.put
-        row_chunksize, col_chunksize = compute_chunksize(df, num_splits)
+        row_chunksize, col_chunksize = compute_chunksize(*df.shape, num_splits)
```
YarShev, line 715: Since `compute_chunksize` parameters are keyword arguments I would prefer to pass them explicitly.

modin/core/storage_formats/pandas/utils.py
```diff
@@ -38,40 +40,48 @@ def get_default_chunksize(length, num_splits):
...
     Compute the number of rows and/or columns to include in each partition.
 
     Parameters
     ----------
-    df : pandas.DataFrame
-        DataFrame to split.
-    num_splits : int
-        Number of splits to separate the DataFrame into.
+    row_count : int, optional
+        Rows count.
```
YarShev, line 52:
```suggestion
        Row count.
```

modin/core/storage_formats/pandas/utils.py
```diff
@@ -38,40 +40,48 @@ def get_default_chunksize(length, num_splits):
...
+    col_count : int, optional
+        Columns count.
```
YarShev, line 54:
```suggestion
        Column count.
```

modin/core/storage_formats/pandas/utils.py
```diff
@@ -38,40 +40,48 @@ def get_default_chunksize(length, num_splits):
...
 
     Returns
     -------
-    int if axis was specified, tuple of ints otherwise,
-        If axis is 1 or 0, returns an integer number of rows/columns to split the
-        DataFrame. If axis is None, returns a tuple containing both.
+    int
+        - `row_count` and `col_count` are specified: returns tuple of ints otherwise,
+        - `row_count`/`col_count` are specified: returns an integer number of rows/columns to split the
+        DataFrame.
```
YarShev, line 65:
```suggestion
        - if `row_count` and `col_count` are specified, tuple of ints will be returned
        - if `row_count`/`col_count` is specified, an integer number of rows/columns to split the
        DataFrame will be returned.
```

YarShev:
@anmyachev, also resolve conflicts please.

modin/core/dataframe/pandas/partitioning/partition_manager.py
anmyachev: Why? Current version looks shorter and more intuitive.

YarShev: Someone may change parameter order and then we would get incorrect flow.

modin/core/storage_formats/pandas/utils.py
```diff
@@ -40,48 +42,51 @@ def get_default_chunksize(length, num_splits):
...
-    axis : int, optional
-        Axis to split across. If not specified - split accros both axes.
 
     Returns
     -------
-    int if axis was specified, tuple of ints otherwise,
-        If axis is 1 or 0, returns an integer number of rows/columns to split the
-        DataFrame. If axis is None, returns a tuple containing both.
+    int
+        - if `row_count` and `col_count` are specified, tuple of ints will be returned
```
YarShev, line 65:
```suggestion
        - if `row_count` and `col_count` are specified, tuple of ints will be returned.
```

modin/core/dataframe/pandas/partitioning/partition_manager.py
anmyachev: Arguments can be passed to any function in `keyword` form, instead of `positional` form. We don't do that everywhere, do we?

YarShev: Yes, we do not do that but I would prefer. If the function had positional arguments only, we could pass the arguments as it is now.

modin/core/storage_formats/pandas/utils.py
```diff
@@ -38,40 +40,48 @@ def get_default_chunksize(length, num_splits):
...
-        If axis is 1 or 0, returns an integer number of rows/columns to split the
-        DataFrame. If axis is None, returns a tuple containing both.
+    int
+        - `row_count` and `col_count` are specified: returns tuple of ints otherwise,
+        - `row_count`/`col_count` are specified: returns an integer number of rows/columns to split the
+        DataFrame.
     """
-    if axis == 0 or axis is None:
-        row_chunksize = get_default_chunksize(df_shape[0], num_splits)
+    assert row_count is not None or col_count is not None
```
vnlitvinov, line 69:
```suggestion
    assert not (row_count is None and col_count is None)
```
I think this looks more intuitive, but it's totally up to you to accept or not this proposal - it's pure cosmetics (your version works slightly faster, but I think the difference is negligible)

modin/core/storage_formats/pandas/utils.py
```diff
@@ -38,40 +40,48 @@ def get_default_chunksize(length, num_splits):
     )
 
 
-def compute_chunksize(df_shape, num_splits, default_block_size=32, axis=None):
+def compute_chunksize(
```
vnlitvinov, line 44: Given how this function is now written and used, I just wonder if we should change the logic to just accept _one_ count (by one axis), and in the case when we need two splits (both rows and columns) just call it twice...

I think it would make code simpler.

anmyachev: I thought about it, if you think the same, then it is definitely worth a try :)

modin/core/storage_formats/pandas/utils.py
```diff
@@ -38,40 +40,48 @@ def get_default_chunksize(length, num_splits):
...
+    row_count=None, col_count=None, num_splits=None, default_block_size=32
```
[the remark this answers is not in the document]
anmyachev: fixed

modin/core/storage_formats/pandas/utils.py
```diff
@@ -38,40 +40,48 @@ def get_default_chunksize(length, num_splits):
...
-    axis : int, optional
-        Axis to split across. If not specified - split accros both axes.
 
     Returns
     -------
-    int if axis was specified, tuple of ints otherwise,
-        If axis is 1 or 0, returns an integer number of rows/columns to split the
-        DataFrame. If axis is None, returns a tuple containing both.
+    int
+        - `row_count` and `col_count` are specified: returns tuple of ints otherwise,
```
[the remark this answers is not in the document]
anmyachev: no need more

modin/core/storage_formats/pandas/utils.py
anmyachev: @vnlitvinov ready for review

modin/core/storage_formats/pandas/utils.py
```diff
@@ -18,73 +18,36 @@
...
+    if index_len % num_splits == 0:
+        chunksize += 1
     # Take the min of the default and the memory-usage chunksize first to avoid a
     # large amount of small partitions.
-    col_chunksize = max(1, col_chunksize, default_block_size)
-    if axis == 1:
-        return col_chunksize
-
-    return row_chunksize, col_chunksize
+    return max(1, chunksize, default_block_size)
```
vnlitvinov, line 50: this looks weird - comment says "take the min" but code uses `max()`... also why do we pass `1` to `max()`?..

anmyachev: Updated

modin/core/storage_formats/pandas/utils.py
```diff
@@ -18,73 +18,36 @@
...
+    if default_block_size is None:
+        default_block_size = MinPartitionSize.get()
 
-    return row_chunksize, col_chunksize
+    chunksize = index_len // num_splits
+    if index_len % num_splits:
+        chunksize += 1
+    # chunksize shouldn't be less that `default_block_size` to avoid a
+    # large amount of small partitions.
+    return max(chunksize, default_block_size)
```
devin-petersohn, line 50: This can be 0 if user sets something silly for `MinPartitionSize` and `chunksize` is less than `num_splits`. That is why we had a 1 value here before, so that the chunksize is always 1 or greater. It needs a comment, I agree, but I think we should keep the 1 here.

```suggestion
    return max(1, chunksize, default_block_size)
```

modin/core/storage_formats/pandas/utils.py
```diff
@@ -18,73 +18,36 @@
 import pandas
 
 
-def get_default_chunksize(length, num_splits):
+def compute_chunksize(index_len, num_splits, default_block_size=None):
```
devin-petersohn, line 21: Minor rename, `index` is too overloaded in my opinion.
```suggestion
def compute_chunksize(axis_len, num_splits, default_block_size=None):
```

devin-petersohn (commented):
New structure looks good to me. Left a couple of comments.

modin/core/storage_formats/pandas/utils.py
anmyachev: Maybe we should add a check for assigned values for `MinPartitionSize`? For example, check if `value > 0`.

devin-petersohn: I think we should do both. It doesn't hurt to always ensure positive number here in case other checks go wrong.

anmyachev: ok, I see. @vnlitvinov what do you think?

vnlitvinov: dunno, I would rather set an assert early on checking if variables have sensible values, rather than on anything weird default to some edge case

anmyachev: I added the check in `MinPartitionSize.put`. 

Although we have the ability to pass the erroneous value through `default_block_size` argument, it is not used or tested anywhere. Let's remove it. @vnlitvinov @devin-petersohn thoughts?

devin-petersohn: I think `default_block_size` is not a good name, but it is something we can plan to use in the future with better partitioning during the query planning phase. 

I see two options: (Option 1) is an `assert` here, (Option 2) is `max(1, ...)`. I don't care either way, but I don't want an unchecked and untested value here.

anmyachev: `assert` is added.

anmyachev:

devin-petersohn (approved):
LGTM!

modin/config/envvars.py
```diff
@@ -415,6 +415,20 @@ class MinPartitionSize(EnvironmentVariable, type=int):
...
     default = 32
 
+    @classmethod
+    def put(cls, value):
+        """
+        Set ``MinPartitionSize`` with extra checks.
+
+        Parameters
+        ----------
+        value : bool
```
YarShev, line 425:
```suggestion
        value : int
```

modin/core/storage_formats/pandas/utils.py
```diff
@@ -18,73 +18,38 @@
...
-
-    Returns
-    -------
-    int
-        Computed chunksize.
-    """
-    return (
-        length // num_splits if length % num_splits == 0 else length // num_splits + 1
-    )
+    Compute the number of elemenents (rows/columns) to include in each partition.
```
YarShev, line 23:
```suggestion
    Compute the number of elements (rows/columns) to include in each partition.
```

modin/core/storage_formats/pandas/utils.py
```diff
@@ -18,73 +18,38 @@
...
-    """
-    Compute the number of rows and/or columns to include in each partition.
+    Chunksize is defined the same for the axes.
 
     Parameters
     ----------
-    df : pandas.DataFrame
-        DataFrame to split.
+    axis_len : int
+        Element count in a axe.
```
YarShev, line 30:
```suggestion
        Element count in an axis.
```

modin/core/storage_formats/pandas/utils.py
```diff
@@ -18,73 +18,38 @@
...
+    if default_block_size is None:
+        default_block_size = MinPartitionSize.get()
 
-    return row_chunksize, col_chunksize
+    assert default_block_size > 0, "`default_block_size` should be > 0"
+
+    chunksize = axis_len // num_splits
+    if axis_len % num_splits:
+        chunksize += 1
+    # chunksize shouldn't be less that `default_block_size` to avoid a
```
YarShev, line 50:
```suggestion
    # chunksize shouldn't be less than `default_block_size` to avoid a
```

modin/core/storage_formats/pandas/utils.py
```diff
@@ -18,73 +18,38 @@
...
-            return row_chunksize
-    # We always execute this because we can only get here if axis is 1 or None.
-    col_chunksize = get_default_chunksize(len(df.columns), num_splits)
-    # Take the min of the default and the memory-usage chunksize first to avoid a
-    # large amount of small partitions.
-    col_chunksize = max(1, col_chunksize, default_block_size)
-    if axis == 1:
-        return col_chunksize
+    if default_block_size is None:
+        default_block_size = MinPartitionSize.get()
```
YarShev, line 43: Maybe we should put the assert that is below to `MinPartitionSize.get()` to avoid multiple places where we would check it.

modin/core/storage_formats/pandas/utils.py
```diff
@@ -18,73 +18,38 @@
 import pandas
 
 
-def get_default_chunksize(length, num_splits):
+def compute_chunksize(axis_len, num_splits, default_block_size=None):
```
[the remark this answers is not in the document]
anmyachev: I don’t remember in what other context `chunksize` is mentioned. If, except for the axes, this is not applied in Modin, then there is no need to clarify.

modin/core/storage_formats/pandas/utils.py
```diff
@@ -18,73 +18,38 @@
...
-
-def compute_chunksize(
-    df,
-    num_splits,
-    default_block_size=None,
-    axis=None,
-):
-    """
-    Compute the number of rows and/or columns to include in each partition.
+    Chunksize is defined the same for the axes.
```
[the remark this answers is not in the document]
anmyachev: So it's better to use next form, I suppose:
```suggestion
    Chunksize is defined the same for both axes.
```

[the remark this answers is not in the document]
YarShev: Correct

modin/core/storage_formats/pandas/utils.py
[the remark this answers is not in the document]
YarShev: ok

modin/core/storage_formats/pandas/utils.py
anmyachev: `default_block_size` can be passed into function directly. The assert check also need for the case.

YarShev: There might be a case when a user sets `MODIN_MIN_PARTITION_SIZE` to 0. That's why I suggest adding the assert to `MinPartitionSize.get()`.

modin/config/envvars.py
```diff
@@ -415,6 +415,20 @@ class MinPartitionSize(EnvironmentVariable, type=int):
...
+        """
+        Set ``MinPartitionSize`` with extra checks.
+
+        Parameters
+        ----------
+        value : int
+            Config value to set.
+        """
+        if value <= 0:
+            raise ValueError(f"Min partition size should be > 0, passed value {value}")
```
anmyachev, line 429: @YarShev We have check here for the case you talking about.

modin/core/storage_formats/pandas/utils.py
anmyachev: Check for the case here: https://github.com/modin-project/modin/pull/3769/files#r781197300

anmyachev: We need a check to set the value, not get it.

modin/config/envvars.py
YarShev: I am talking about other places where we call `MinPartitionSize.get()`. If a user sets a value via the env var (not via `MinPartitionSize.put(value)`, the check you specified in `MinPartitionSize.put()` won't be called.

anmyachev: @YarShev got it, I was relying on the idea that the only intended way to set values is `put` function, however this can happens in `get` function under different conditions.

@vnlitvinov it seems to me that it's better to rewrite these classes so that the setting of values goes only through `put` function. What do you think about?

vnlitvinov: This seems like a different problem. You're talking about verification of the correctness of the input.

There is already a mechanism which is specified per-type only that seems to be doing part of what you want, but for incoming values (like strings which we then cast to ints).
Seems like we should add similar mechanism to the values right before setting the internal value.

I propose to do it in a separate PR for all the values which should have such restrictions (like CPU count being >=1, etc.)

modin/core/storage_formats/pandas/utils.py
```diff
@@ -18,73 +18,38 @@
...
-    # Take the min of the default and the memory-usage chunksize first to avoid a
-    # large amount of small partitions.
-    col_chunksize = max(1, col_chunksize, default_block_size)
-    if axis == 1:
-        return col_chunksize
+    if default_block_size is None:
+        default_block_size = MinPartitionSize.get()
 
-    return row_chunksize, col_chunksize
+    assert default_block_size > 0, "`default_block_size` should be > 0"
```
vnlitvinov, line 45: It feels like the variable name is now wrong, it is no longer a `default` block size, but probably `min_block_size`.
I propose to rename the variable and update the docstring accordingly.

modin/config/envvars.py
YarShev: Yes, adding such a mechanism looks as a separate issue.

modin/core/storage_formats/pandas/utils.py
YarShev: I saw @devin-petersohn has a concern on `default_block_size` name too. `min_block_size` is okay to me.

modin/config/envvars.py
YarShev: For now, I would suggest adding the assert to `get` and remove it if necessary when that mechanism is added.

anmyachev: `assert` is added for `get` method. Ready for review.

YarShev (approved):
LGTM

modin/config/envvars.py
YarShev: @anmyachev, please create an issue for adding that mechanism.

PR #3770, opened 2021-12-02
FIX-#3767: Fix tests for Modin XGB Dmatrix

### LINKED ISSUE #3767: Fix dmatrix testing failures in CI
Starting from 4d7f6d4cfd2dc99d9bfd21bb62c5a74daa3ab256 `test_dmatrix` sporadically fails in CI testing. Possible reason is in strict check of equality between XGBoost/Modin XGBoost result:
https://github.com/modin-project/modin/blob/cc96f5d2c08b03f4bd5d3d261a6871b292d9230d/modin/experimental/xgboost/test/test_dmatrix.py#L129

prutskov:

## What do these changes do?

This PR fixes periodic failures in CI on `experimental/xgboost/test/test_dmatrix.py`. The fix is in changing of random generated data for training/prediction purposes to existing known dataset.

- [x] Resolves #3767  
- [x] tests added and passing

### DIFF
--- a/modin/experimental/xgboost/test/test_dmatrix.py
+++ b/modin/experimental/xgboost/test/test_dmatrix.py
@@ -13,6 +13,7 @@
 
 import numpy as np
 import pytest
+import pandas
 from sklearn.metrics import accuracy_score
 from sklearn.datasets import load_breast_cancer
 import xgboost as xgb
@@ -103,7 +104,7 @@ def test_feature_names():
     predictions = booster.predict(dmatrix)
     modin_predictions = md_booster.predict(md_dmatrix)
 
-    preds = pd.DataFrame(predictions).apply(np.round, axis=0)
+    preds = pandas.DataFrame(predictions).apply(np.round, axis=0)
     modin_preds = modin_predictions.apply(np.round, axis=0)
 
     accuracy = accuracy_score(y, preds)
--- a/modin/experimental/xgboost/test/test_dmatrix.py
+++ b/modin/experimental/xgboost/test/test_dmatrix.py
@@ -103,8 +103,8 @@ def test_feature_names():
     predictions = booster.predict(dmatrix)
     modin_predictions = md_booster.predict(md_dmatrix)
 
-    preds = pd.DataFrame(predictions).apply(lambda x: round(x))
-    modin_preds = modin_predictions.apply(lambda x: round(x))
+    preds = pd.DataFrame(predictions).apply(np.round, axis=0)
+    modin_preds = modin_predictions.apply(np.round, axis=0)
 
     accuracy = accuracy_score(y, preds)
     md_accuracy = accuracy_score(y, modin_preds)
--- a/modin/experimental/xgboost/test/test_dmatrix.py
+++ b/modin/experimental/xgboost/test/test_dmatrix.py
@@ -14,6 +14,7 @@
 import numpy as np
 import pytest
 from sklearn.metrics import accuracy_score
+from sklearn.datasets import load_breast_cancer
 import xgboost as xgb
 
 import modin.pandas as pd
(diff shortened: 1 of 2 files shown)


modin/experimental/xgboost/test/test_dmatrix.py
```diff
@@ -74,67 +75,49 @@ def test_dmatrix_feature_names_and_feature_types(data, feature_names, feature_ty
...
+    md_booster = mxgb.train(params, md_dmatrix, num_boost_round=10)
 
-    predictions = bst.predict(dm)
-    modin_predictions = md_bst.predict(md_dm)
+    predictions = booster.predict(dmatrix)
+    modin_predictions = md_booster.predict(md_dmatrix)
 
-    preds = np.asarray([np.argmax(line) for line in predictions])
-    md_preds = np.asarray([np.argmax(line) for line in modin_predictions.to_numpy()])
+    preds = pd.DataFrame(predictions).apply(lambda x: round(x))
```
YarShev, line 106:
```suggestion
    preds = pandas.DataFrame(predictions).apply(lambda x: round(x))
```
I would prefer to be explicit so let's specify `axis` parameter for `apply`. Why do we need `round`?

modin/experimental/xgboost/test/test_dmatrix.py
```diff
@@ -74,67 +75,49 @@ def test_dmatrix_feature_names_and_feature_types(data, feature_names, feature_ty
...
+    modin_preds = modin_predictions.apply(lambda x: round(x))
```
YarShev, line 107: same comments as above

modin/experimental/xgboost/test/test_dmatrix.py
```diff
@@ -74,67 +75,49 @@ def test_dmatrix_feature_names_and_feature_types(data, feature_names, feature_ty
...
-    md_dm = mxgb.DMatrix(pd.DataFrame(dummy), feature_names=list("abcde"))
+    # Different feature_names (default) must raise error in this case
+    dm = xgb.DMatrix(X)
+    md_dm = mxgb.DMatrix(pd.DataFrame(X))
     with pytest.raises(ValueError):
-        bst.predict(dm)
+        booster.predict(dm)
     with pytest.raises(ValueError):
-        md_bst.predict(md_dm)
+        repr(md_booster.predict(md_dm))
```
YarShev, line 120: Why do we need `repr` now?

modin/experimental/xgboost/test/test_dmatrix.py
```diff
@@ -74,67 +75,49 @@ def test_dmatrix_feature_names_and_feature_types(data, feature_names, feature_ty
...
-    md_dm = mxgb.DMatrix(
-        pd.DataFrame(data), label=pd.Series(label), feature_names=feature_names
+    dmatrix = xgb.DMatrix(X, label=y, feature_names=feature_names)
+    md_dmatrix = mxgb.DMatrix(
+        pd.DataFrame(X), label=pd.Series(y), feature_names=feature_names
     )
 
     params = {
-        "objective": "multi:softprob",
+        "objective": "binary:logistic",
```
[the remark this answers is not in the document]
prutskov: Because dataset with binary labels is used now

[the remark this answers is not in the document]
YarShev: Yes, I see. Why do we want to have a dataset for binary classification now?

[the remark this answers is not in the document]
prutskov: Previous data was incorrect. Binary classification is simply and fast.

modin/experimental/xgboost/test/test_dmatrix.py
prutskov: Modin XGB `predict` is async func. To get thrown exception here we should finish execution

modin/experimental/xgboost/test/test_dmatrix.py
vnlitvinov: also why don't you directly `.apply(round)` or `.apply(np.round)`?

modin/experimental/xgboost/test/test_dmatrix.py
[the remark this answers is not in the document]
YarShev: ok

modin/experimental/xgboost/test/test_dmatrix.py
YarShev: ok

modin/experimental/xgboost/test/test_dmatrix.py
prutskov: @YarShev, `axis` providing is added (I don't know if this makes sense, because we make `apply` for 1 column dataframe using non-reduce callable). `round` is need to determine class label (0 or 1) for each prediction value from column.
@vnlitvinov, I agree, `np.round` is used instead of lambda.

YarShev: I would also use pandas DataFrame for non Modin data here as I suggested earlier.

prutskov: No see the need in this, because in this test we don't compare pandas/modin dataframes functionality, we just use dataframe to apply `round`. But I don't mind, applied

prutskov:
Blocked by #3772.

modin/experimental/xgboost/test/test_dmatrix.py
YarShev: If we had a bug in `apply`, we might have tests passed and wouldn't see an actual problem (even though we test `apply` separately).

YarShev:
Merged #3773.

YarShev (approved):
LGTM assuming tests pass.

YarShev (approved):

PR #3826, opened 2021-12-09
FIX-#3818: Ignore files from /private/tmp/ray/ when detecting file leaks

### LINKED ISSUE #3818: read_excel leaves ray logs open
### Describe the problem
After running read_excel, there are open file handles for ray session logs:

in ipython:
Screen Shot 2021-12-08 at 12 51 56 PM

in test_io.py:
Screen Shot 2021-12-08 at 2 46 08 PM

Interestingly for test_io.py::TestExcel, only one of the test cases fails with this assertion if you run them together, but if you run them individually, they all fail.

Is related to #2559 but the root cause is probably different. To check open files, use

```
import psutil
p = psutil.Process()
# run code
print(p.open_files())
```

jeffreykennethli:

## What do these changes do?

[test_io.p::TestExcel](https://github.com/modin-project/modin/blob/7c847581f8ce8b61307690c87884775eba8d034b/modin/pandas/test/test_io.py#L1482) open file handle checks currently catches our [check_file_leaks decorator](https://github.com/modin-project/modin/blob/7c847581f8ce8b61307690c87884775eba8d034b/modin/pandas/test/utils.py#L1083) check due to Ray session driver logs that are still open after read_excel(). check_file_leaks compares open files before and after a test case, therefore only the first instance where a Ray session is initiated will catch check_file_leaks.

The session logs still being open is the [intended Ray behavior](https://github.com/modin-project/modin/issues/3818#issuecomment-989332461), so our tests should ignore these open files in check_file_leaks. This change adds this exception to check_file_leaks.

- [ ] Resolves #3818? 
- [ ] tests added and passing

### DIFF
--- a/modin/pandas/test/utils.py
+++ b/modin/pandas/test/utils.py
@@ -1105,9 +1105,9 @@ def check_file_leaks(func):
                     # modin reading any data (and this is what we care about).
                     if item[0].startswith("/proc/"):
                         continue
-                    # Ignore files in /private/tmp/ray/session/ (ray session logs)
-                    # because Ray intends to keep these logs open even after work
-                    # has been done.
+                    # Ignore files in /private/tmp/ray/ (ray session logs)
+                    # because Ray intends to keep these logs open even after
+                    # work has been done.
                     if item[0].startswith("/private/tmp/ray/"):
                         continue
                     leaks.append(item)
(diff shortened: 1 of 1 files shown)


devin-petersohn (commented):
Are the logs guaranteed to be here?

jeffreykennethli:
> Are the logs guaranteed to be here?

The logs structure is /tmp/ray/session_*/logs per the docs: https://docs.ray.io/en/master/ray-logging.html#logging-directory-structure

Modified the check to more explicitly check for this format.

devin-petersohn (approved):
LGTM

PR #3854, opened 2021-12-14
REFACTOR-#3853: interacting with Dask interface through 'DaskWrapper' class

### LINKED ISSUE #3853: REFACTOR: move work with `distributed.client.default_client` function into the single place
This change will allow us to concentrate the work with the third-party interface in one place (for example, `dask/common`), which will eliminate code duplication and make it more maintainable.

anmyachev:

## What do these changes do?

1. Changed the signature for all `deploy` functions.
2. Instead of explicitly calling `client.submit` function, `DaskTask.deploy` function is called.
3. `client.materialize` instead of `client.gather`
4. `client.scatter` (maybe rename to `put`) instead of `client.scatter`

- [x] Resolves #3853 


YarShev:
@anmyachev, that is what we were discussing earlier when restructuring, great changes! What about Ray?

anmyachev:
> @anmyachev, that is what we were discussing earlier when restructuring, great changes! What about Ray?

I plan to do the same for Ray, but on the separate PR. Is it ok for you? :)

YarShev:
> > @anmyachev, that is what we were discussing earlier when restructuring, great changes! What about Ray?
> 
> I plan to do the same for Ray, but on the separate PR. Is it ok for you? :)

Yes, that should be done in a separate PR.

modin/core/execution/ray/common/task_wrapper.py
```diff
@@ -53,16 +53,20 @@ def deploy(cls, func, num_returns, kwargs):
...
+            Additional keyword arguments to be passed in ``func``.
 
         Returns
         -------
         ray.ObjectRef or list
             Ray identifier of the result being put to Plasma store.
         """
-        return deploy_ray_func.options(num_returns=num_returns).remote(func, kwargs)
+        return deploy_ray_func.options(num_returns=num_returns).remote(
+            func, *args, kwargs
```
anmyachev, line 68: It looks strange, for this to be so `**kwargs`, `deploy_ray_func` signature should be changed.

modin/core/execution/dask/common/task_wrapper.py
```diff
@@ -62,3 +67,22 @@ def materialize(cls, future):
         """
         client = default_client()
         return client.gather(future)
+
+    @classmethod
+    def scatter(cls, data, **kwargs):
```
devin-petersohn, line 72: I prefer to call this `put`. It will be a little weird to say `DaskTask.put` or `DaskTask.scatter` because tasks aren't an object store.

modin/core/execution/ray/common/task_wrapper.py
devin-petersohn: Historically we have not been able to pass `**kwargs` and `*args` to a Ray task for implementation reasons. To get around this we pass `kwargs` as a dictionary.

modin/core/execution/dask/implementations/pandas_on_dask/partitioning/partition.py
```diff
@@ -137,19 +131,16 @@ def drain_call_queue(self):
         if len(self.call_queue) == 0:
             return
         call_queue = self.call_queue
-        client = default_client()
+
```
devin-petersohn, line 134:
```suggestion
```

devin-petersohn (commented):
Looks good overall. I wonder if we can align the two (ray and dask) implementations based on this interface.

modin/core/execution/dask/common/task_wrapper.py
anmyachev: What about using `DaskWrapper` class name instead of `DaskTask`?

anmyachev: Will change to `put`.

modin/core/execution/ray/common/task_wrapper.py
anmyachev: Do I understand correctly that such remote call did not work before?

```python
>>> def t2(*args, **kwargs):
...   print(args)  
...   print(kwargs) 
...                
>>> t2_rem = ray.remote(t2)
>>> t2_rem.remote("test_args", 2, 6, test_keyword="some value")
ObjectRef(4ee449587774c1f0ffffffffffffffffffffffff0100000001000000)
>>> (pid=28664) ('test_args', 2, 6)
(pid=28664) {'test_keyword': 'some value'}

```

modin/core/execution/dask/implementations/pandas_on_dask/partitioning/partition.py
```diff
@@ -137,19 +133,17 @@ def drain_call_queue(self):
...
             # We handle `len(call_queue) == 1` in a different way because
             # this improves performance a bit.
             func, args, kwargs = call_queue[0]
-            future = client.submit(apply_func, self.future, func, *args, **kwargs)
-        futures = [
-            client.submit(lambda l, i: l[i], future, i, pure=False) for i in range(2)
-        ]
+            futures = DaskWrapper.deploy(
+                apply_func, 2, self.future, func, *args, **kwargs
+            )
```
dchigarev, line 146: shouldn't we pass `pure=False` here as well?

modin/test/test_partition_api.py
```diff
@@ -33,10 +33,10 @@
     get_func = ray.get
     FutureType = ray.ObjectRef
 elif Engine.get() == "Dask":
-    from distributed.client import default_client
+    from modin.core.execution.dask.common.task_wrapper import DaskWrapper
     from distributed import Future
 
-    put_func = lambda x: default_client().scatter(x)  # noqa: E731
+    put_func = lambda x: DaskWrapper.put(x)  # noqa: E731
     get_func = lambda x: x.result()  # noqa: E731
```
dchigarev, line 40: shouldn't this be changed as well?
```suggestion
    put_func = lambda x: DaskWrapper.put(x)  # noqa: E731
    get_func = lambda x: DaskWrapper.materialize(x)  # noqa: E731
```

dchigarev:
> > @anmyachev, that is what we were discussing earlier when restructuring, great changes! What about Ray?
> 
> I plan to do the same for Ray, but on the separate PR. Is it ok for you? :)

Will this make sense to create a common interface for every `Task` class, defining a `put`, `materialize`, `deploy`, etc, so `DaskTask` and `RayTask` would implement it?

anmyachev:
> > > @anmyachev, that is what we were discussing earlier when restructuring, great changes! What about Ray?
> > 
> > 
> > I plan to do the same for Ray, but on the separate PR. Is it ok for you? :)
> 
> Will this make sense to create a common interface for every `Task` class, defining a `put`, `materialize`, `deploy`, etc, so `DaskTask` and `RayTask` would implement it?

This is the main goal of the refactoring, but it will only be done in the third step :) (After updating `DaskTask` and `RayTask` interfaces)

modin/core/execution/dask/implementations/pandas_on_dask/partitioning/partition.py
anmyachev: should

devin-petersohn (commented):
@anmyachev is there another PR that does something similar? or is this a different PR?

modin/core/execution/dask/common/task_wrapper.py
```diff
@@ -39,11 +41,14 @@ def deploy(cls, func, num_returns, kwargs):
...
         client = default_client()
-        remote_task_future = client.submit(func, **kwargs)
-        return [
-            client.submit(lambda l, i: l[i], remote_task_future, i)
-            for i in range(num_returns)
-        ]
+        remote_task_future = client.submit(func, *args, **kwargs)
+        pure = kwargs.get("pure", None)
+        if num_returns != 1:
+            remote_task_future = [
```
YarShev, line 47:
```suggestion
            return [
```

modin/core/execution/dask/implementations/pandas_on_dask/partitioning/partition.py
```diff
@@ -64,9 +62,9 @@ def get(self):
         """
         self.drain_call_queue()
         # blocking operation
-        if isinstance(self.future, pandas.DataFrame):
+        if not isinstance(self.future, Future):
```
YarShev, line 65: Why do we need to change this?

modin/core/execution/dask/implementations/pandas_on_dask/partitioning/partition.py
```diff
@@ -90,20 +88,18 @@ def apply(self, func, *args, **kwargs):
         -----
         The keyword arguments are sent as a dictionary.
         """
-        client = default_client()
         call_queue = self.call_queue + [[func, args, kwargs]]
         if len(call_queue) > 1:
-            future = client.submit(
-                apply_list_of_funcs, call_queue, self.future, pure=False
+            futures = DaskWrapper.deploy(
+                apply_list_of_funcs, 2, call_queue, self.future, pure=False
```
YarShev, line 94: We shouldn't use magic constants. Let's add a variable in line above like `num_returns` and put a comment nearby why the variable is equal to 2.

modin/core/execution/dask/implementations/pandas_on_dask/partitioning/partition.py
```diff
@@ -90,20 +88,18 @@ def apply(self, func, *args, **kwargs):
...
         else:
             # We handle `len(call_queue) == 1` in a different way because
             # this improves performance a bit.
             func, args, kwargs = call_queue[0]
-            future = client.submit(apply_func, self.future, func, *args, **kwargs)
-        futures = [
-            client.submit(lambda l, i: l[i], future, i, pure=False) for i in range(2)
-        ]
+            futures = DaskWrapper.deploy(
+                apply_func, 2, self.future, func, *args, pure=False, **kwargs
```
YarShev, line 101: Same comment on magic constants.

YarShev: address this where applicable

docs/img/pandas_on_dask_data_ingress.svg
```diff
@@ -1,4 +1,4 @@
...
..., 0)" stroke-miterlimit="10" pointer-events="all"/><path d="M 194.01 117 L 178.03 117 L 178.03 142.79 L 147.84 142.8" fill="none" stroke="rgb(0, 0, 0)" stroke-miterlimit="10" pointer-events="stroke"/><path d="M 209.01 117 L 201.51 121.41 L 194.01 117 L 201.51 112.59 Z" fill="none" stroke="rgb(0, 0, 0)" stroke-miterlimit="10" pointer-events="all"/><path d="M 159.72 136.29 L 146.72 142.8 L 159.72 149.29" fill="none" stroke="rgb(0, 0, 0)" stroke-miterlimit="10" pointer-events="all"/><path d="M 271 72.99 L 271 94.76" fill="none" stroke="rgb(0, 0, 0)" stroke-miterlimit="10" pointer-events="stroke"/><path d="M 271 57.99 L 275.41 65.49 L 271 72.99 L 266.59 65.49 Z" fill="none" stroke="rgb(0, 0, 0)" stroke-miterlimit="10" pointer-events="all"/><path d="M 264.5 82.88 L 271 95.88 L 277.5 82.88" fill="none" stroke="rgb(0, 0, 0)" stroke-miterlimit="10" pointer-events="all"/></g><switch><g requiredFeatures="http://www.w3.org/TR/SVG11/feature#Extensibility"/><a transform="translate(0,-5)" xlink:href="https://www.diagrams.net/doc/faq/svg-export-text-problems" target="_blank"><text text-anchor="middle" font-size="10px" x="50%" y="100%">Viewer does not support full SVG 1.1</text></a></switch></svg>
```
YarShev, line 4: What has been changed here?

modin/core/execution/dask/common/task_wrapper.py
```diff
@@ -16,11 +16,11 @@
 from distributed.client import default_client
 
 
-class DaskTask:
+class DaskWrapper:
     """The class responsible for execution of remote operations."""
 
     @classmethod
-    def deploy(cls, func, num_returns, kwargs):
+    def deploy(cls, func, num_returns, *args, **kwargs):
```
YarShev, line 23: We should probably rename this method to `deploy_func` or `deploy_task`.

modin/core/execution/dask/common/task_wrapper.py
```diff
@@ -39,11 +41,14 @@ def deploy(cls, func, num_returns, kwargs):
...
+        pure = kwargs.get("pure", None)
+        if num_returns != 1:
+            remote_task_future = [
+                client.submit(lambda l, i: l[i], remote_task_future, i, pure=pure)
+                for i in range(num_returns)
+            ]
+        return remote_task_future
 
     @classmethod
     def materialize(cls, future):
```
YarShev, line 54: I would probably prefer to rename this method to `get`.

docs/img/pandas_on_dask_data_ingress.svg
anmyachev: `DaskTask` -> `DaskWrapper`

modin/core/execution/dask/common/task_wrapper.py
anmyachev: I'm for it, only so far it's called that everywhere (including for Ray). Let's do it later and immediately for every engine.

modin/core/execution/dask/common/task_wrapper.py
anmyachev: I'm aligning on Ray for now.

modin/core/execution/dask/implementations/pandas_on_dask/partitioning/partition.py
anmyachev: Thus, the code looks clearer, we materialize when the object can be materialized.

modin/core/execution/dask/implementations/pandas_on_dask/partitioning/partition.py
anmyachev: The same problem is observed when using `deploy` function in other places.

The only way I can see to improve readability is to change the interface so that this option can be used in this way: `num_returns=2`.

In your case, there will be many places where these variables (`num_returns`) will be defined before `deploy` call with the same comments like "the number 2 is used here, because such and such a function returns two arguments"

modin/core/execution/dask/implementations/pandas_on_dask/partitioning/partition.py
anmyachev: I see, discussing above.

anmyachev:
@YarShev ready for review

modin/core/execution/dask/common/task_wrapper.py
YarShev: Ok, let's do that for all engines at once in a separate PR.

modin/core/execution/dask/common/task_wrapper.py
```diff
@@ -39,11 +41,14 @@ def deploy(cls, func, num_returns, kwargs):
             The result of ``func`` splitted into parts in accordance with ``num_returns``.
         """
         client = default_client()
-        remote_task_future = client.submit(func, **kwargs)
-        return [
-            client.submit(lambda l, i: l[i], remote_task_future, i)
-            for i in range(num_returns)
-        ]
+        remote_task_future = client.submit(func, *args, **kwargs)
+        pure = kwargs.get("pure", None)
```
YarShev, line 45: Why is a default value None? Given distributed's docs, pure is equal to True by default.

modin/core/execution/dask/common/task_wrapper.py
YarShev: ok

modin/core/execution/dask/implementations/pandas_on_dask/partitioning/partition.py
YarShev: The use of num_returns works for me.

modin/core/execution/dask/implementations/pandas_on_dask/partitioning/partition.py
YarShev: A partition should hold a pandas DataFrame. We may come to undefined behavior or errors in this case.

modin/core/execution/ray/common/task_wrapper.py
YarShev: @devin-petersohn, ping

modin/core/execution/dask/common/task_wrapper.py
YarShev: @devin-petersohn, ping

modin/core/execution/dask/implementations/pandas_on_dask/partitioning/partition.py
anmyachev: [Some code](https://github.com/modin-project/modin/pull/3991/files#diff-348036bdb519cf0606cdd3a08df8e5483c16fb9c39fe9f3d30983758fa391551R126) is written as if there could also be an object of `pandas.series` type.

YarShev:
Please fix conflicts.

modin/core/execution/dask/common/task_wrapper.py
anmyachev: I saw it in [the code](http://distributed.dask.org/en/stable/_modules/distributed/client.html#Client.submit).

I think it's beneficial in our case to make the default value - `False`, since that's what is used in most cases.

modin/core/execution/dask/implementations/pandas_on_dask/partitioning/partition.py
YarShev: Each partition holds a pandas DataFrame, but not a pandas Series. That is what we require and say in docs. So I am in favor of checking for a pandas DataFrame only, to be strongly restrict.

anmyachev: I think we are talking about different things. Checking for which objects can be stored in a partition should be as early as possible, for example in the constructor. Here it is the check whether it is possible to call `materialize` function on the stored object.

YarShev: You can't guarantee that a future in the constructor is exactly a pandas DataFrame. Otherwise, you'll break up asynchronous execution. Here we should check whether the object is a future to materialize or a pandas DataFrame to just return. We should probably add a check that materialized data is a pandas.DataFrame, and also, if a remote call is supposed to return a non pandas.DataFrame, we should execute the task via DaskWrapper in other places.

anmyachev: Why does `isinstance` function break asynchronous execution? Does it look at what lies inside the future?

anmyachev: How can a materialized object appear in a partition? At what point in time? I thought that only through the constructor. 

If the object is `future` type, then it is obviously not necessary to check what is inside until the moment of the necessary materialization.

modin/core/execution/dask/implementations/pandas_on_dask/partitioning/partition.py
anmyachev: added `num_returns` variable

modin/core/execution/dask/implementations/pandas_on_dask/partitioning/partition.py
YarShev: > How can a materialized object appear in a partition? At what point in time? I thought that only through the constructor.

Only a future (for Dask) and an object ref (for Ray) are allowed in the Partition constructor.

> If the object is future type, then it is obviously not necessary to check what is inside until the moment of the necessary materialization.

That is true.

modin/core/execution/dask/implementations/pandas_on_dask/partitioning/partition.py
```diff
@@ -89,20 +86,19 @@ def apply(self, func, *args, **kwargs):
         -----
         The keyword arguments are sent as a dictionary.
         """
-        client = default_client()
         call_queue = self.call_queue + [[func, args, kwargs]]
+        num_returns = 2
```
YarShev, line 90: To not introduce such a variable everywhere we should probably declare it a kw arg to deploy.

anmyachev: > The only way I can see to improve readability is to change the interface so that this option can be used in this way: num_returns=2.

Yeh, that's what I try to say above, but it also need to change the interface for Ray, and this already looks like a separate change.

modin/core/execution/dask/implementations/pandas_on_dask/partitioning/partition.py
anmyachev: It turns out that this check is not needed, since at the time of the `isinstance` call, there is always an object of `Future` type. Let's remove it.

As indirect evidence that this check is not needed, we can look at [the implementation](https://github.com/modin-project/modin/blob/b7027597985a442734a57949d72e3fa0625b4d43/modin/core/execution/ray/implementations/pandas_on_ray/partitioning/partition.py#L60) of this function for ray.

modin/core/execution/dask/implementations/pandas_on_dask/partitioning/partition.py
YarShev: Ok, let's do it as part of a separate PR.

modin/core/execution/dask/implementations/pandas_on_dask/partitioning/partition.py
YarShev: 1. We should add a check for a future passed in the constructor to detect if it is really Dask.Future (as for Ray).
2. I still tend to add a check in get if the materialized data is a pandas DataFrame.
3. We should remove isinstance for a pandas Series in to_pandas.

YarShev:
Overall, LGTM, except my concern in that discussion ([link](https://github.com/modin-project/modin/pull/3854#discussion_r807882936)).

modin/core/execution/dask/implementations/pandas_on_dask/partitioning/partition.py
anmyachev: https://github.com/modin-project/modin/pull/4207

YarShev: 2. I see, to_numpy is presumably the only case when we have a NumPy array in a partition. Okay, let's leave it as is but we should think of this.
3. We can remove isinstance for a pandas Series in to_pandas since we don't add that check in get.

modin/core/execution/dask/implementations/pandas_on_dask/partitioning/virtual_partition.py
```diff
@@ -81,9 +81,12 @@ def deploy_axis_func(
         list
             A list of distributed.Future.
         """
-        client = default_client()
-        axis_result = client.submit(
+        lengths = kwargs.get("_lengths", None)
+        result_num_splits = len(lengths) if lengths else num_splits
+        num_returns = result_num_splits * 4
+        axis_result = DaskWrapper.deploy(
```
YarShev, line 87:
```suggestion
        result = DaskWrapper.deploy(
```

modin/core/execution/dask/implementations/pandas_on_dask/partitioning/virtual_partition.py
```diff
@@ -93,16 +96,7 @@ def deploy_axis_func(
...
-        lengths = kwargs.get("_lengths", None)
-        result_num_splits = len(lengths) if lengths else num_splits
-
-        # We have to do this to split it back up. It is already split, but we need to
-        # get futures for each.
-        return [
-            client.submit(lambda l: l[i], axis_result, pure=False)
-            for i in range(result_num_splits * 4)
-        ]
+        return axis_result
```
YarShev, line 99:
```suggestion
        return result
```
We can probably return just DaskWrapper.deploy(...).

modin/core/execution/dask/implementations/pandas_on_dask/partitioning/virtual_partition.py
```diff
@@ -147,12 +142,7 @@ def deploy_func_between_two_axis_partitions(
             *partitions,
             pure=False,
         )
-        # We have to do this to split it back up. It is already split, but we need to
-        # get futures for each.
-        return [
-            client.submit(lambda l: l[i], axis_result, pure=False)
-            for i in range(num_splits * 4)
-        ]
+        return axis_result
```
YarShev, line 145: same

modin/core/execution/dask/implementations/pandas_on_dask/partitioning/partition.py
YarShev: I see the changes for ray in this PR. Are we able to put them in a separate PR?

modin/core/execution/dask/common/task_wrapper.py
```diff
@@ -16,11 +16,11 @@
 from distributed.client import default_client
```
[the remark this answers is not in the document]
anmyachev: Maybe just `wrapper.py` since the file in `dask` folder?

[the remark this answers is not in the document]
YarShev: I think that just a wrapper.py would be somewhat vague. dask_wrapper.py seems to me clearer with respect to DaskWrapper we have.

modin/core/execution/dask/implementations/pandas_on_dask/partitioning/partition.py
```diff
@@ -280,7 +267,7 @@ def ip(self):
         if self._ip_cache is None:
             self._ip_cache = self.apply(lambda df: df)._ip_cache
         if isinstance(self._ip_cache, Future):
-            self._ip_cache = self._ip_cache.result()
+            self._ip_cache = DaskWrapper.materialize(self._ip_cache)
```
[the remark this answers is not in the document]
anmyachev: Now I use `materialize` function everywhere. The approach to calculating these properties will be revised in another PR.

modin/core/execution/dask/common/task_wrapper.py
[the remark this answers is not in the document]
anmyachev: The import line will look something like this:
```python
from modin.core.execution.dask.common.dask_wrapper import DaskWrapper
```
Don't you think that the word `dask` is mentioned too many times here?

modin/core/execution/dask/implementations/pandas_on_dask/partitioning/partition.py
YarShev: ok

modin/core/execution/dask/common/task_wrapper.py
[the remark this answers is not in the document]
YarShev: I don't think `dask` is mentioned too many times here. I think we can either think up a different name for the class and the file or leave as is.

modin/core/execution/ray/common/task_wrapper.py
```diff
@@ -46,25 +46,29 @@ class RayTask:
...
-        num_returns : int
-            Amount of return values expected from `func`.
-        kwargs : dict
-            Keyword arguments to pass to remote instance of `func`.
+        *args : list
+            Additional positional arguments to be passed in ``func``.
+        num_returns : int, default: 1
+            Amount of return values expected from ``func``.
+        **kwargs : dict
+            Additional keyword arguments to be passed in ``func``.
```
YarShev, line 62:
```suggestion
            Additional keyword arguments to be passed in `func`.
```

modin/core/execution/dask/implementations/pandas_on_dask/partitioning/partition.py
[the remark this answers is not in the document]
YarShev: What is the purpose of revising?

anmyachev:
@YarShev ready for review. I answered all the comments except about the filename.

modin/core/execution/dask/implementations/pandas_on_dask/partitioning/partition.py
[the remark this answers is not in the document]
YarShev: ok

modin/core/execution/dask/common/task_wrapper.py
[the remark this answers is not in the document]
YarShev: api looks very common and a bit vague to me. I am wondering if we can leave DaskTask as is? I think Task word isn't meant to imply a dask remote function. It is just any dask task/job.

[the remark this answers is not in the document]
anmyachev: In terms of design patterns, `DaskWrapper` is a "Facade" - a unified interface to a subsystem. I would like the name to emphasize the purpose of the class.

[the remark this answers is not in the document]
YarShev: Let's keep the following then
```python
from modin.core.execution.dask.common.dask_wrapper import DaskWrapper
```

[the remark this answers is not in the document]
YarShev: To keep names of the files for Ray and Dask identical we could rename ray/dask_wrapper.py to engine_wrapper.py, where we declare Ray/DaskWrapper. What do you think on this?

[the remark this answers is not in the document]
anmyachev: In general, a unified name is convenient. However, will there be a conflict of terms: `execution` and `engine`?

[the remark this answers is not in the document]
YarShev: `execution`/`execution backend`/`execution system` consists of a `storage format` and an `engine` so there shouldn't be any conflict.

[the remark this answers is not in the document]
anmyachev: Ok, I would do it in a similar pull request for Ray. Do you mind?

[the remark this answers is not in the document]
YarShev: I think we should rename the file to `engine_wrapper.py` for Dask in this PR. As to Ray, that would be a separate PR.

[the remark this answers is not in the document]
anmyachev: ok

modin/core/execution/dask/implementations/pandas_on_dask/partitioning/partition.py
anmyachev: 3. This should not be done, because after calling some function on the `DataFrame` object, an object of `Series` type may be obtained, which will be stored in the partition (in that case, `get` function will return `Series`).

anmyachev:
@YarShev ready for review

docs/flow/modin/core/execution/dask/implementations/pandas_on_dask/index.rst
```diff
@@ -69,7 +69,7 @@ When reading data from a CSV file, for example, the :py:class:`~modin.core.execu
 the user query to the :meth:`~modin.core.io.text.CSVDispatcher._read` method of :py:class:`~modin.core.io.text.CSVDispatcher`, where the query's parameters are preprocessed
 to check if they are supported by the execution (defaulting to pandas if they are not) and computes some metadata
 common for all partitions to be read. Then, the file is split into row chunks, and this data is used to launch remote tasks on the Dask workers
-via the :meth:`~modin.core.execution.dask.common.task_wrapper.DaskTask.deploy` method of :py:class:`~modin.core.execution.dask.common.task_wrapper.DaskTask`.
+via the :meth:`~modin.core.execution.dask.common.dask_wrapper.DaskWrapper.deploy` method of :py:class:`~modin.core.execution.dask.common.dask_wrapper.DaskWrapper`.
```
YarShev, line 72:
```suggestion
via the :meth:`~modin.core.execution.dask.common.engine_wrapper.DaskWrapper.deploy` method of :py:class:`~modin.core.execution.dask.common.engine_wrapper.DaskWrapper`.
```

YarShev (commented):
Left the last comment. LGTM! Let's make it possible to put similar changes for Ray as well in 0.14.

YarShev:
Also, do not forget to add a release note.

anmyachev:
@YarShev the error with `readthedocs` can be ignored

PR #3876, opened 2021-12-23
FEAT-#3875: Optimize append time for omnisci engine

### LINKED ISSUE #3875: Dataframe.append works slow on omnisci engine with default arguments
with `ignore_index=True` `append` bypasses omnisci execution and concatenates arrow tables, but it is not set by default, so we need to optimize the default case as well.

<details>
<summary>Reproducer</summary>

```python
import os
import modin.config
modin.config.IsExperimental.put(True)
modin.config.StorageFormat.put("omnisci")

import pandas as pd
import modin.pandas as mpd
import numpy as np

def generate_data(nrows, ncols):
    data = {
        f"col{i}": np.random.randint(0, 1000000000, size=nrows)
        for i in range(ncols)
    }
    return data

NROWS = 10_000_000
NCOLS = 10
df = pd.DataFrame(generate_data(NROWS, NCOLS))
df2 = pd.DataFrame(generate_data(NROWS, NCOLS))

mdf = mpd.DataFrame(generate_data(NROWS, NCOLS))
mdf2 = mpd.DataFrame(generate_data(NROWS, NCOLS))

_ = mdf.shape 
_ = mdf2.shape

print("Start pandas")
for i in range(10):
    elapsed_time = timer()
    copy = df
    copy = copy.append(df2, ignore_index=True)
    
    elapsed_time = timer() - elapsed_time
    print("Pandas time:", elapsed_time)
print("Start modin")

for i in range(10):
    elapsed_time = timer()
    copy = mdf
    copy = copy.append(mdf2)
    
    copy = copy.shape
    
    elapsed_time = timer()- elapsed_time
    print("Modin time:", elapsed_time)
```

</details>

fexolm:
## What do these changes do?
Optimize `append` with `ignore_index=True` by bypassing omnisci engine and doing all operations with arrow table.

- [ ] Resolves #3875 
- [ ] tests added and passing

### DIFF
--- a/modin/experimental/core/execution/native/implementations/omnisci_on_native/dataframe/dataframe.py
+++ b/modin/experimental/core/execution/native/implementations/omnisci_on_native/dataframe/dataframe.py
@@ -1689,9 +1689,7 @@ class OmnisciOnNativeDataframe(PandasDataframe):
             )
         elif isinstance(self._op, TransformNode):
             return (
-                not self._uses_rowid
-                and self._op.is_simple_select()
-                and self._op.input[0]._can_execute_arrow()
+                self._op.is_simple_select() and self._op.input[0]._can_execute_arrow()
             )
         elif isinstance(self._op, UnionNode):
             return all(frame._can_execute_arrow() for frame in self._op.input)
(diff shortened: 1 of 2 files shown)


modin/experimental/core/execution/native/implementations/omnisci_on_native/dataframe/dataframe.py
```diff
@@ -1737,11 +1720,18 @@ def _arrow_select(self, exprs):
             The resulting table.
         """
         table = self._execute_arrow()
-        schema = table.schema
 
         new_fields = []
         new_columns = []
 
+        if "__index__" in exprs:
```
ienkovich, line 1727: This is an incorrect way to find the required columns. It doesn't have to be projected into `__index__` column, it might have any name (index can be renamed or moved to a data column). Also, there might be multiple columns referencing rowid. You should check referenced column name with the special name `__rowid__` to find required references.

ienkovich (changes requested):
Please, add tests to check the new feature.

dchigarev (commented):
LGTM besides unrelated code style changes. I assume that `black` did this automatically on your machine. If so, check that you're using the same version of black that is used in CI (`21.12b0`).

ienkovich (approved):

PR #3888, opened 2021-12-30
FEAT-#3887: added Dask diagnostics dependency

### LINKED ISSUE #3887: [Feature request] Install Dask diagnostics tools
Just like [Ray Dashboard](https://docs.ray.io/en/latest/ray-dashboard.html), Dask also has [diagnostics dashboard tool](https://docs.dask.org/en/stable/diagnostics-distributed.html). 

It's useful for debugging when the user submit a job. However, when the user run `pip install modin[dask]`, it wouldn't install the Dask diagnostic dependences. It would be convenience if install the diagnostic tools at same time.

Is it possible install not only `dask` but also `diagnostics tools`, when we run `pip install modin[dask]`? If it make sense, I could work on it.

orcahmlee:
Signed-off-by: Andrew Li <orcahmlee@gmail.com>

## What do these changes do?
Ray and Dask both have diagnostic dashboard. It's useful for debugging when the user submit a job. However, when the user run `pip install modin[dask]`, it wouldn't install the Dask diagnostic dependences. It would be convenience if install the diagnostic tools at same time.

- [x] Resolves #3887
- [ ] tests added and passing

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -4,7 +4,7 @@ import versioneer
 with open("README.md", "r", encoding="utf-8") as fh:
     long_description = fh.read()
 
-dask_deps = ["dask>=2.22.0", "distributed>=2.22.0"]
+dask_deps = ["dask[diagnostics]>=2.22.0", "distributed>=2.22.0"]
 ray_deps = ["ray[default]>=1.4.0", "pyarrow>=1.0"]
 remote_deps = ["rpyc==4.1.5", "cloudpickle", "boto3"]
 spreadsheet_deps = ["modin-spreadsheet>=0.1.0"]


RehanSD:
Hi Andrew! Thank you so much for your contribution! Could you add some more context (either in the PR description or by opening and linking an issue) so reviewers can understand why this change is necessary and what it's solving?

orcahmlee:
> Hi Andrew! Thank you so much for your contribution! Could you add some more context (either in the PR description or by opening and linking an issue) so reviewers can understand why this change is necessary and what it's solving?

Hi Rehan, Thank for the quick reply.
I already updated the PR description, hope it would be more clear.

RehanSD:
> > Hi Andrew! Thank you so much for your contribution! Could you add some more context (either in the PR description or by opening and linking an issue) so reviewers can understand why this change is necessary and what it's solving?
> 
> 
> 
> Hi Rehan, Thank for the quick reply.
> 
> I already updated the PR description, hope it would be more clear.

Hi Andrew! I see where you're coming from - but my concern is that introducing additional, optional, dependencies may make Modin bloated - and if we don't mention we're installing these packages, users may not know that they can be used. Instead of adding it to the pip install list, what do you think about adding a note to the installation sections for Dask and Ray that mention that they offer these debugging packages, and that users can install them for diagnostics on their cluster?

orcahmlee:
> > > Hi Andrew! Thank you so much for your contribution! Could you add some more context (either in the PR description or by opening and linking an issue) so reviewers can understand why this change is necessary and what it's solving?
> > 
> > 
> > Hi Rehan, Thank for the quick reply.
> > I already updated the PR description, hope it would be more clear.
> 
> Hi Andrew! I see where you're coming from - but my concern is that introducing additional, optional, dependencies may make Modin bloated - and if we don't mention we're installing these packages, users may not know that they can be used. Instead of adding it to the pip install list, what do you think about adding a note to the installation sections for Dask and Ray that mention that they offer these debugging packages, and that users can install them for diagnostics on their cluster?

Hi Rehan,
I totally understand what your concerns. I think that adding a note to the installation section on README.md would be better. Now, should I close this PR or I could just modify the title and commit to this branch?

RehanSD:
@orcahmlee I think creating a new PR and referencing this one in it and explaining the conclusion we came to is probably easier just in terms of having documentation of the process. (Otherwise people may see this PR with a totally different title and have no context for the discussion we had via comments)! Thank you so much for your contribution!

orcahmlee:
> @orcahmlee I think creating a new PR and referencing this one in it and explaining the conclusion we came to is probably easier just in terms of having documentation of the process. (Otherwise people may see this PR with a totally different title and have no context for the discussion we had via comments)! Thank you so much for your contribution!

@RehanSD I am not an experienced open source developer. Thank for your kindly reply.

PR #3894, opened 2022-01-05
FIX-#3860: Fix single row Series assignment.

### LINKED ISSUE #3860: Values are getting set to NaN upon using boolean indexes
### System information
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: Mac OS Monterey
- **Python version**: 3.7.6

```python3
import modin.pandas as mpd  # ray
series = mpd.Series(["123456", None])
null_mask = series.isnull()
print(series[~null_mask].astype(int).astype(str))  # prints 123456 correctly
series[~null_mask]=series[~null_mask].astype(int).astype(str)
print(series)  # Values in series: NaN, None
```

### Describe the problem
Values getting set to NaN upon using boolean indexes

mvashishtha:
## What do these changes do?

Fix https://github.com/modin-project/modin/issues/3860 by setting the correct `axis` argument to the query compiler's `setitem` method. In the case where a single row and column are to be replaced, we are assigning both an entire row and an entire column, so `axis` should be `None` instead of `1`, so we don't transpose the new value [here](https://github.com/modin-project/modin/blob/d8963a691958d34627e503c7a437adade4935aa8/modin/core/storage_formats/pandas/query_compiler.py#L2230).

- [x] Resolves #3860
- [x] tests added and passing

### DIFF
--- a/modin/pandas/indexing.py
+++ b/modin/pandas/indexing.py
@@ -532,7 +532,9 @@ class _LocationIndexerBase(object):
             for i, lookup in enumerate([row_lookup, col_lookup])
         ]
 
-        if (
+        if col_lookup_len == 1 and row_lookup_len == 1:
+            axis = None
+        elif (
             row_lookup_len == len(self.qc.index)
             and col_lookup_len == 1
             and isinstance(self.df, DataFrame)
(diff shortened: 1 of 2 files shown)


devin-petersohn (approved):
Thanks @mvashishtha, LGTM!

PR #3979, opened 2022-01-14
FIX-#3896: Change Series.values to default to to_numpy().

### LINKED ISSUE #3896: IndexError when calling modin.pandas.DataFrame.values.all()
### System information
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: Microsoft Windows 10 Entreprise 10.0.18363 N/A build 18363
- **Modin version** (`modin.__version__`): 0.12.1
- **Python version**: Python 3.8.5 (tags/v3.8.5:580fbb0, Jul 20 2020, 15:57:54)

```python
from pyreadline import Readline
readline = Readline()
import numpy as np
import ray
import modin.pandas as pd

ray.init()

df_movment_subset = pd.DataFrame([], columns=['Date transaction', 'Reference','Statut','MSISDN','Num Correspondant', 'Type', 'Movments.final.movmentAmount'])

if df_movment_subset['Type'].values.all() == 'Depot':
    print("Depot")
elif df_movment_subset['Type'].values.all() == 'Retrait':
    print("Retrait")
else:
    print("None")
```

### Describe the problem

An IndexError: tuple index out of range is thrown
The same code with native pandas library will return 'None'

### Source code / logs

```
UserWarning: `DataFrame.__getitem__` for empty DataFrame defaulting to pandas implementation.
UserWarning: Distributing <class 'pandas.core.series.Series'> object. This may take some time.
  File "C:\Users\TNTZ2660\AppData\Local\Programs\Python\Python38\lib\site-packages\modin\pandas\base.py", line 3076, in __getattribute__
    attr = super().__getattribute__(item)
  File "C:\Users\TNTZ2660\AppData\Local\Programs\Python\Python38\lib\site-packages\modin\pandas\series.py", line 546, in values
    return super(Series, self).to_numpy().flatten()
  File "C:\Users\TNTZ2660\AppData\Local\Programs\Python\Python38\lib\site-packages\modin\pandas\base.py", line 2694, in to_numpy
    return self._query_compiler.to_numpy(
  File "C:\Users\TNTZ2660\AppData\Local\Programs\Python\Python38\lib\site-packages\modin\core\storage_formats\pandas\query_compiler.py", line 338, in to_numpy
    arr = self._modin_frame.to_numpy(**kwargs)
  File "C:\Users\TNTZ2660\AppData\Local\Programs\Python\Python38\lib\site-packages\modin\core\dataframe\pandas\dataframe\dataframe.py", line 2282, in to_numpy
    return self._partition_mgr_cls.to_numpy(self._partitions, **kwargs)
  File "C:\Users\TNTZ2660\AppData\Local\Programs\Python\Python38\lib\site-packages\modin\core\execution\ray\generic\partitioning\partition_manager.py", line 51, in to_numpy
    n = partitions.shape[1]
IndexError: tuple index out of range

```

jeffreykennethli:

## What do these changes do?

Modify Series.values to default to Series.to_numpy(), which conveniently handles the empty dataframe/series case from #3896.

In [pandas/base.py](https://github.com/modin-project/modin/blob/master/modin/pandas/base.py#L3075-L3087), notice that if an attr is callable and self.empty, we default to pandas like in [Series.to_numpy()](https://github.com/modin-project/modin/blob/master/modin/pandas/series.py#L1863-L1877) which would fix our empty Series case from #3896. 

Because to_numpy() and .values are supposed to have identical behavior (see [pandas.Series.values](https://pandas.pydata.org/docs/reference/api/pandas.Series.values.html) and [this PR](https://github.com/pandas-dev/pandas/pull/23623)), we can safely make Series.values call Series.to_numpy().

- [ ] Resolves #3896 
- [ ] tests added and passing

### DIFF
--- a/modin/pandas/series.py
+++ b/modin/pandas/series.py
@@ -543,11 +543,7 @@ class Series(BasePandasDataset):
         """
         Return Series as ndarray or ndarray-like depending on the dtype.
         """
-
-        def values(df):
-            return df.values
-
-        return self._default_to_pandas(values)
+        return self.to_numpy()
 
     def add(self, other, level=None, fill_value=None, axis=0):  # noqa: PR01, RT01, D200
         """
--- a/modin/pandas/series.py
+++ b/modin/pandas/series.py
@@ -543,7 +543,11 @@ class Series(BasePandasDataset):
         """
         Return Series as ndarray or ndarray-like depending on the dtype.
         """
-        return super(Series, self).to_numpy().flatten()
+
+        def values(df):
+            return df.values
+
+        return self._default_to_pandas(values)
 
     def add(self, other, level=None, fill_value=None, axis=0):  # noqa: PR01, RT01, D200
         """
--- a/modin/pandas/test/test_series.py
+++ b/modin/pandas/test/test_series.py
@@ -3204,6 +3204,21 @@ def test_to_period():
 )
 def test_to_numpy(data):
     modin_series, pandas_series = create_test_series(data)
+    assert_array_equal(modin_series.to_numpy(), pandas_series.to_numpy())
+
+
+@pytest.mark.parametrize(
+    "data",
+    test_data_values + test_data_large_categorical_series_values,
+    ids=test_data_keys + test_data_large_categorical_series_keys,
+)
+def test_series_values(data):
+    modin_series, pandas_series = create_test_series(data)
+    assert_array_equal(modin_series.values, pandas_series.values)
+
+
+def test_series_empty_values():
+    modin_series = pandas_series = pd.Series([])
     assert_array_equal(modin_series.values, pandas_series.values)
 
 
--- a/modin/pandas/test/test_series.py
+++ b/modin/pandas/test/test_series.py
@@ -3218,7 +3218,7 @@ def test_series_values(data):
 
 
 def test_series_empty_values():
-    modin_series = pandas_series = pd.Series([])
+    modin_series, pandas_series = pd.Series(), pandas.Series()
     assert_array_equal(modin_series.values, pandas_series.values)
 
 


mvashishtha (approved):

RehanSD (commented):
In the PR description, you mention wanting to default to pandas only when the `Series` is empty - doesn't this default to pandas in all occasions?

modin/pandas/series.py
```diff
@@ -543,7 +543,11 @@ def values(self):  # noqa: RT01, D200
         """
         Return Series as ndarray or ndarray-like depending on the dtype.
         """
-        return super(Series, self).to_numpy().flatten()
+
+        def values(df):
+            return df.values
+
+        return self._default_to_pandas(values)
```
RehanSD, line 550:
```suggestion
        return self.to_numpy().flatten()
```

RehanSD (changes requested):
I did some digging, and it looks like the error comes from [this](https://github.com/modin-project/modin/blob/ed2a7a4adf973660eb15b9ea674cd5fb802f2424/modin/core/execution/ray/generic/partitioning/partition_manager.py#L51) line, which expects non-empty Dataframes.

If we remove the call to `super` the error is fixed, and we are defaulting to `to_numpy` like @jeffreykennethli suggested, which seems to be what pandas intends in the future as well.

modin/pandas/series.py
```diff
@@ -543,7 +543,7 @@ def values(self):  # noqa: RT01, D200
         """
         Return Series as ndarray or ndarray-like depending on the dtype.
         """
```
jeffreykennethli, line 545: was considering putting a:
```
UserWarning: 'Series.values` defaulting to to_numpy().
```

modin/pandas/series.py
jeffreykennethli: Done. I also removed the flatten() since presumably Series will already be 1-D, and flatten() is not necessary. Let me know if this is not a safe assumption.

modin/pandas/series.py
jeffreykennethli: This is not done for DataFrame.values.

Having this message for both DataFrame.values and Series.values is probably not super useful, except in the case where we're operating on an empty DataFrame/Series and we're already surfacing a 
```
UserWarning: 'DataFrame.to_numpy` for empty DataFrame defaulting to pandas implementation.
```
in which case it will be confusing why DataFrame/Series using to_numpy and defaulting to pandas when the user is calling .values? (and the "values defaulting to to_numpy()" message would clarify).

devin-petersohn (dismissed):
LGTM!

modin/pandas/test/test_series.py
```diff
@@ -3204,6 +3204,21 @@ def test_to_period():
...
+    test_data_values + test_data_large_categorical_series_values,
+    ids=test_data_keys + test_data_large_categorical_series_keys,
+)
+def test_series_values(data):
+    modin_series, pandas_series = create_test_series(data)
+    assert_array_equal(modin_series.values, pandas_series.values)
+
+
+def test_series_empty_values():
+    modin_series = pandas_series = pd.Series([])
```
RehanSD, line 3221: won't this make the `pandas_series` equal to `modin.pandas.Series`? I don't think this test actually tests that modin handles an empty series the same as pandas.

RehanSD (changes requested):
Quick thing about the tests, but once that's addressed LGTM!

modin/pandas/test/test_series.py
jeffreykennethli: done! ty for catching this.

RehanSD (approved):
LGTM!

devin-petersohn (approved):

PR #3986, opened 2022-01-17
FIX-#3985: Update pyarrow in environment files

### LINKED ISSUE #3985: Update pyarrow in environment files
When creating an env from enviroment.yml pyarrow 3.0.0 is installed. At the same time, when installing modin-ray and modin-omnisci in the same env pyarrow 5.0.0 is installed. So we should update pyarrow in requirement-dev.txt and enviroment.yml files to check correctness tests pass with a newer pyarrow version.

YarShev:

## What do these changes do?

- [x] Resolves #3985  
- [x] tests passing

### DIFF
--- a/environment-dev.yml
+++ b/environment-dev.yml
@@ -4,7 +4,7 @@ channels:
 dependencies:
   - pandas==1.3.5
   - numpy>=1.16.5
-  - pyarrow>=1.0.0
+  - pyarrow>=6.*
   - libthrift<0.14.1  # ref: modin gh-2840
   - dask[complete]>=2.22.0
   - distributed>=2.22.0
--- a/environment-dev.yml
+++ b/environment-dev.yml
@@ -4,7 +4,7 @@ channels:
 dependencies:
   - pandas==1.3.5
   - numpy>=1.16.5
-  - pyarrow>=6.0
+  - pyarrow>=4.0.1
   - dask[complete]>=2.22.0
   - distributed>=2.22.0
   - fsspec
--- a/environment-dev.yml
+++ b/environment-dev.yml
@@ -4,8 +4,7 @@ channels:
 dependencies:
   - pandas==1.3.5
   - numpy>=1.16.5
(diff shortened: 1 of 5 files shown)


environment-dev.yml
```diff
@@ -4,7 +4,7 @@ channels:
 dependencies:
   - pandas==1.3.5
   - numpy>=1.16.5
-  - pyarrow>=1.0.0
+  - pyarrow>=6.*
```
vnlitvinov, line 7: Don't use relative comparison and asterisk at the same time, conda gives a warning and discards the asterisk anyway.

Use `pyarrow>=6.0` to achieve the same effect.

YarShev:
Pyarrow should be updated in setup.py too.

environment-dev.yml
YarShev: fixed

YarShev:
libthrift<0.14.1 is a culprit of the conflict with a newer pyarrow so removed it to allow pyarrow itself to install libthrift.

setup.py
```diff
@@ -5,7 +5,7 @@
     long_description = fh.read()
 
 dask_deps = ["dask>=2.22.0", "distributed>=2.22.0"]
-ray_deps = ["ray[default]>=1.4.0", "pyarrow>=1.0"]
+ray_deps = ["ray[default]>=1.4.0", "pyarrow>=6.0"]
```
YarShev, line 8: An option for pinning pyarrow would be >= 4.0.1 like Ray requires it as extras ([link])(https://github.com/ray-project/ray/blob/master/python/setup.py#L197). However, our requirements on pyarrow version for modin-ray and modin-omnisci should be aligned. Any thoughts on what a better option would be?

Garra1980:
I'd say >= 4.0.1 is fine here

devin-petersohn (approved):
LGTM!

vnlitvinov (approved):

PR #3999, opened 2022-01-18
REFACTOR-#4000: move `progress_bar_wrapper` decorator in file with `PandasDataframePartitionManager` class

### LINKED ISSUE #4000: REFACTOR: remove code duplication for `PandasOnRayDataframePartitionManager`
In the current implementation, the following methods of `PandasOnRayDataframePartitionManager` class (child class of `PandasDataframePartitionManager`) are overridden just to apply the `progress_bar_wrapper` decorator:
- map_partitions
- lazy_map_partitions
- map_axis_partitions
- _apply_func_to_list_of_partitions
- apply_func_to_select_indices
- apply_func_to_select_indices_along_full_axis
- apply_func_to_indices_both_axis
- binary_operation

Moving the decorator to the base class will remove the duplication of functions.

anmyachev:

## What do these changes do?

- [x] Resolves #4000 
- [x] tests added and passing


modin/core/dataframe/pandas/partitioning/partition_manager.py
```diff
@@ -69,6 +71,68 @@ def wait(*args, **kwargs):
...
+    if Engine.get() != "Ray":
+        if ProgressBar.get():
+            raise NotImplementedError(
+                f"ProgressBar doesn't work with '{Engine.get()}' engine."
+            )
+        else:
+            return f
+
+    from functools import wraps
+    from modin.core.execution.ray.generic.modin_aqp import call_progress_bar
```
devin-petersohn, line 100: This progress bar increment is specific to the ray engine, are you sure this works?

anmyachev: Please note, that progress bar is enabled only for ray engine case.

That is, it has a specific implementation (at the moment), but its usage can always be the same, even when implementations for other engines appear. Therefore, the use of the wrapper itself can be moved to the base class.

modin/core/dataframe/pandas/partitioning/partition_manager.py
```diff
@@ -69,6 +71,68 @@ def wait(*args, **kwargs):
...
+    callable
+        Decorated version of `f` which reports progress.
+    """
+    if Engine.get() != "Ray":
+        if ProgressBar.get():
+            raise NotImplementedError(
+                f"ProgressBar doesn't work with '{Engine.get()}' engine."
+            )
+        else:
+            return f
```
anmyachev, line 97: These lines - the main changes that allow us to move progress bar usage into base class.

modin/core/execution/ray/implementations/pandas_on_ray/partitioning/partition_manager.py
```diff
@@ -26,64 +25,7 @@
     PandasOnRayDataframeRowPartition,
 )
 from .partition import PandasOnRayDataframePartition
-from modin.core.execution.ray.generic.modin_aqp import call_progress_bar
 from modin.error_message import ErrorMessage
-import pandas
```
RehanSD, line 31: This import is still necessary (it's used in `broadcast_apply`).

modin/core/dataframe/pandas/partitioning/partition_manager.py
```diff
@@ -806,6 +873,7 @@ def get_indices(cls, axis, partitions, index_func=None):
         return new_idx[0].append(new_idx[1:]) if len(new_idx) else new_idx
 
     @classmethod
+    @progress_bar_wrapper
     def _apply_func_to_list_of_partitions_broadcast(
```
RehanSD, line 877: This function doesn't have a progress bar wrapper in the original code - is there a reason why we've added one here?

RehanSD (changes requested):
I'm concerned about the performance implications of wrapping the functions at the pandas partitioning manager - its probably negligible, but we are adding an extra layer to the stack to the dask + python executions, when we don't really have any progress bar implemented for them yet.

anmyachev:
> I'm concerned about the performance implications of wrapping the functions at the pandas partitioning manager - its probably negligible, but we are adding an extra layer to the stack to the dask + python executions, when we don't really have any progress bar implemented for them yet.

@RehanSD an additional layer is created only for Ray (as it was). For all other engines, the additional cost will be only at the time of the first decorator call since the decorator will return the original function if there is no implementation we need.

modin/core/dataframe/pandas/partitioning/partition_manager.py
anmyachev: It's a mistake. Good catch!

modin/core/execution/ray/implementations/pandas_on_ray/partitioning/partition_manager.py
anmyachev: This import is there, I just moved it to line 17. There is a certain order of imports that most python developers support. https://www.python.org/dev/peps/pep-0008/#imports
>Imports should be grouped in the following order:
>
>Standard library imports.
>Related third party imports.
>Local application/library specific imports."

I'm fixing the imports a bit where I see.

anmyachev:
@RehanSD ready for review

YarShev:
Will #4167 be resolved with this PR?

anmyachev:
> Will #4167 be resolved with this PR?

No, this pull request is for refactoring only. For #4127 we need a PR with fixes/enhancements (depending on point of view).

modin/core/dataframe/pandas/partitioning/partition_manager.py
YarShev: I don't like we import something from ray directory to pandas, which is base for ray. We should probably have some base structure for the progress bar, which requires us do more complex refactor.

anmyachev: You are talking about the solution to which you need to strive, I agree. However, it is much more difficult to implement because we need to know the implementation details of the progress bar. This decision is the most balanced in relation to resources/result. I assume that the current version brings more benefits than problems.

devin-petersohn:
@RehanSD @anmyachev I do not think we should merge this. ProgressBar logic needs to be more abstract and this is moving code where I don't think it necessarily needs to be shared.

PR #4015, opened 2022-01-21
Exclude "scripts" from setup package

### LINKED ISSUE #4005: scripts/ is getting installed in site-packages
### System information
- **Python version**:  3.8

I believe the scripts/ folder is getting installed in site-packages with modin. I used ```pip install modin[dask]```. We ran into an import issue since we have a scripts module (conflict between using ours vs the site-packages installed.) but when we install modin we found that your scripts module is getting installed in the site-packages. I am not sure if this is a mistake or intentional.

vnlitvinov:
## What do these changes do?
Exclude `scripts` and its subdirectories from setup package.

- [x] Resolves #4005
- [x] tests added and passing

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -16,7 +16,7 @@ setup(
     version=versioneer.get_version(),
     cmdclass=versioneer.get_cmdclass(),
     description="Modin: Make your pandas code run faster by changing one line of code.",
-    packages=find_packages(),
+    packages=find_packages(exclude=["scripts", "scripts.*"]),
     include_package_data=True,
     license="Apache 2",
     url="https://github.com/modin-project/modin",


setup.py
```diff
@@ -16,7 +16,7 @@
     version=versioneer.get_version(),
     cmdclass=versioneer.get_cmdclass(),
     description="Modin: Make your pandas code run faster by changing one line of code.",
-    packages=find_packages(),
+    packages=find_packages(exclude=["scripts", "scripts.*"]),
```
YarShev, line 19: Even though our "scripts" folder is used for testing only, should we move it under "modin" directory so "scripts" is accessible via "modin.scripts"?

vnlitvinov: That is another option, I agree. But I don't see much value in that.

YarShev: We may put there something useful in future so that might be accessible when installing Modin package. Does that make sense?

vnlitvinov: In this case we'll move `scripts` when needed, and this exclude won't prevent it from inclusion 😉

devin-petersohn (approved):
LGTM, This is a good initial solution.

setup.py
devin-petersohn: Yes, this is a one line change rather than a refactor, so I am in favor of this change first

YarShev: Okay, makes sense.

YarShev (approved):

PR #4074, opened 2022-01-27
FIX-#4056: Allow an empty parse_date list in read_csv_glob.

### LINKED ISSUE #4056: `read_csv_glob` fails with empty parse_date list
### System information
Ray 1.9.2
Modin 0.12.0
Python 3.9.7

### Describe the problem

```
pd_experimental.read_csv_glob("somevalidpath", parse_dates=[])
----
...
File "##/python3.9/site-packages/modin/experimental/pandas/io.py", line 208, in _read
pd_obj = FactoryDispatcher.read_csv_glob(**kwargs)
File "##/python3.9/site-packages/modin/core/execution/dispatching/factories/dispatcher.py", line 185, in read_csv_glob
return cls.__factory._read_csv_glob(**kwargs)
File "##/python3.9/site-packages/modin/core/execution/dispatching/factories/factories.py", line 513, in _read_csv_glob
return cls.io_cls.read_csv_glob(**kwargs)
File "##/python3.9/site-packages/modin/core/io/text/csv_glob_dispatcher.py", line 238, in _read
if isinstance(parse_dates, list) and isinstance(parse_dates[0], list):
IndexError: list index out of range (worker#1713)
File "##/python3.9/site-packages/ray/worker.py", line 1713, in get
raise value.as_instanceof_cause()
```

it looks like code only checks that the type of `parse_dates` is a list, but doesn’t check for empty content, tries to access first element and throws an error

mvashishtha:

## What do these changes do?

Allow an empty parse_date list in read_csv_glob.

- [x] Resolves #4056
- [x] tests added and passing


modin/core/io/text/csv_glob_dispatcher.py
```diff
@@ -240,7 +240,9 @@ def _read(cls, filepath_or_buffer, **kwargs):
             # columns are on each partition.
             column_widths = None
             # Check if is list of lists
-            if isinstance(parse_dates, list) and isinstance(parse_dates[0], list):
```
RehanSD, line 243: Does it make sense to just short circuit the if statement here if we don't have a valid `parse_dates`, i.e.:
```python
if parse_dates is not None:
```
becomes
```
if parse_dates is not None and parse_dates != []
```
or something like that?

RehanSD (commented):
Left a comment, but besides that, lgtm!

modin/core/io/text/csv_glob_dispatcher.py
mvashishtha: That would work for this bug, but I prefer this version, because we should be checking that every element in the outer list is a list.

RehanSD: I agree with checking every element - can't we do both? Short cut the outside if and check that every element is a list in the inner one?

mvashishtha: I doubt that the short circuit would improve performance by any measurable amount. I'd rather keep the code simpler by not adding it.

RehanSD: Wfm!

modin/experimental/core/execution/native/implementations/omnisci_on_native/test/test_dataframe.py
```diff
@@ -283,7 +283,8 @@ def test_read_csv_datetime(
     ):
 
         parse_dates_unsupported = isinstance(parse_dates, dict) or (
-            isinstance(parse_dates, list) and isinstance(parse_dates[0], list)
+            isinstance(parse_dates, list)
+            and all(isinstance(date, list) for date in parse_dates)
```
RehanSD, line 287: I don't think the test needs to be added here right? Since we control the values for `parse_dates` and we don't pass an empty list to it.

RehanSD (commented):
Left a minor comment. after that, lgtm!

modin/experimental/core/execution/native/implementations/omnisci_on_native/test/test_dataframe.py
mvashishtha: Agreed. I reverted this change.

mvashishtha: We decided offline to test an empty list in omnisci, but that required re-added the check here, so I did that.

RehanSD:
@mvashishtha could you please add your PR Title + Number and your username to the release notes template go 0.14? You can find more info @ #4078!

docs/release_notes/release_notes-0.14.0.rst
```diff
@@ -4,7 +4,7 @@ Key Features and Updates
 ------------------------
 
 * Stability and Bugfixes
-  *
+  * Allow an empty parse_date list in `read_csv_glob` (f9ff6ac)
```
RehanSD, line 7:
```suggestion
  * FIX-#4056: Allow an empty parse_date list in `read_csv_glob` (#4074)
```

modin/experimental/core/execution/native/implementations/omnisci_on_native/io/io.py
```diff
@@ -466,7 +466,9 @@ def _read_csv_check_support(
             )
 
         parse_dates_unsupported = isinstance(parse_dates, dict) or (
-            isinstance(parse_dates, list) and isinstance(parse_dates[0], list)
+            isinstance(parse_dates, list)
+            and len(parse_dates) > 0
```
RehanSD, line 470: Does this mean that if `parse_dates` is an empty list, we'll just not parse dates silently? Is that ideal?

modin/experimental/core/execution/native/implementations/omnisci_on_native/io/io.py
```diff
@@ -466,7 +466,9 @@ def _read_csv_check_support(
...
+            and all(isinstance(date, list) for date in parse_dates)
```
RehanSD, line 471:
```suggestion
            and any(isinstance(date, list) for date in parse_dates)
```
I think this is maybe supposed to be `any`? Would `parse_dates=[['col1', 'col2'], 'col3']` be supported?

modin/experimental/core/execution/native/implementations/omnisci_on_native/test/test_dataframe.py
```diff
@@ -283,7 +284,8 @@ def test_read_csv_datetime(
...
```
RehanSD, line 288: Same comment as above!

modin/experimental/pandas/test/test_io_exp.py
```diff
@@ -119,6 +119,12 @@ def test_read_csv_without_glob(self):
             with pytest.raises(FileNotFoundError):
                 pd.read_csv_glob("s3://nyc-tlc/trip data/yellow_tripdata_2020-")
 
+    # This addresses https://github.com/modin-project/modin/issues/4056
+    def test_read_single_csv_with_invalid_parse_dates(self):
+        modin_df = pd.read_csv_glob(pytest.files[0], parse_dates=[])
+        pandas_df = pandas.read_csv(pytest.files[0], parse_dates=[])
```
RehanSD, line 125: Should we add the `nonexistent_column` test below here as well?

RehanSD (changes requested):
Left some comments, looking good though!

docs/release_notes/release_notes-0.14.0.rst
mvashishtha: I fixed this in the commit that merged master back in.

modin/experimental/core/execution/native/implementations/omnisci_on_native/test/test_dataframe.py
mvashishtha: As noted above, I changed to `any`.

modin/experimental/pandas/test/test_io_exp.py
mvashishtha: Done.

modin/core/io/text/csv_glob_dispatcher.py
```diff
@@ -239,12 +239,23 @@ def _read(cls, filepath_or_buffer, **kwargs):
...
+                    # Lists within the parse_dates list are sequences of
+                    # CSV columns that are parsed together as a single date
+                    # column. They can be a list of either string column names
+                    # or integer column indices. e.g. if parse_dates is
+                    # [[1, 2]] and columns at indices 1 and 2 are "b" and "c",
+                    # the output dataframe has the single date column "b_c". If
+                    # parse_dates is [["a", 1]] and the column at index 1 is
+                    # named "b", the output dataframe has the single date
+                    # column "a_b".
+                    if isinstance(date, list):
```
naren-ponder, line 253: Should there be an `else` here to correspond to this `if`, maybe logging some kind of specific error message?

mvashishtha: It's always ok if `date` is not a list. `parse_dates` can be a list of scalars, like , or a mix of scalars and lists of scalars, e.g.:

- `["timestamp"]`
- `["timestamp", ["year", "month", "date"]]`

from the values I defined in `parse_dates_values_by_id` in this PR.

`read_csv_glob` just needs to calculate the new column name from each list within `parse_dates`.

modin/experimental/core/execution/native/implementations/omnisci_on_native/io/io.py
```diff
@@ -466,13 +466,17 @@ def _read_csv_check_support(
             )
 
         parse_dates_unsupported = isinstance(parse_dates, dict) or (
-            isinstance(parse_dates, list) and isinstance(parse_dates[0], list)
+            isinstance(parse_dates, list)
+            and any(
+                isinstance(date, list) or not isinstance(date, str)
```
RehanSD, line 471: Does this mean that for OmniSci to work, all of the values in the list have to be strings?
If so:
```suggestion
                not isinstance(date, str)
```
isn't that equivalent?

modin/experimental/core/execution/native/implementations/omnisci_on_native/io/io.py
```diff
@@ -466,13 +466,17 @@ def _read_csv_check_support(
...
+                for date in parse_dates
+            )
         )
         if parse_dates_unsupported:
             return (
                 False,
                 "read_csv with 'arrow' engine supports only bool and "
-                "flattened lists 'parse_dates' parameter",
+                "flattened list of string column names 'parse_dates' parameter",
```
RehanSD, line 479:
```suggestion
                "flattened list of string column names for the 'parse_dates' parameter",
```

modin/experimental/core/execution/native/implementations/omnisci_on_native/test/test_dataframe.py
```diff
@@ -283,7 +286,11 @@ def test_read_csv_datetime(
     ):
 
         parse_dates_unsupported = isinstance(parse_dates, dict) or (
-            isinstance(parse_dates, list) and isinstance(parse_dates[0], list)
+            isinstance(parse_dates, list)
+            and any(
+                isinstance(date, list) or not isinstance(date, str)
```
RehanSD, line 291:
```suggestion
                not isinstance(date, str)
```

modin/experimental/pandas/test/test_io_exp.py
```diff
@@ -119,6 +120,27 @@ def test_read_csv_without_glob(self):
...
+    @pytest.mark.parametrize(
+        "parse_dates",
+        [pytest.param(value, id=id) for id, value in parse_dates_values_by_id.items()],
+    )
+    def test_read_single_csv_with_parse_dates(self, parse_dates):
+        try:
+            pandas_df = pandas.read_csv(time_parsing_csv_path, parse_dates=parse_dates)
+        except Exception as pandas_exception:
+            with pytest.raises(Exception) as modin_exception:
+                modin_df = pd.read_csv_glob(
```
RehanSD, line 132: if this is a single CSV, why are we using `read_csv_glob`?

modin/pandas/test/test_io.py
```diff
@@ -868,14 +870,29 @@ def test_read_csv_categories(self):
 
     @pytest.mark.parametrize("encoding", [None, "utf-8"])
     @pytest.mark.parametrize("encoding_errors", ["strict", "ignore"])
-    @pytest.mark.parametrize("parse_dates", [False, ["timestamp"]])
-    @pytest.mark.parametrize("index_col", [None, 0, 2])
+    @pytest.mark.parametrize(
+        "parse_dates",
+        [pytest.param(value, id=id) for id, value in parse_dates_values_by_id.items()],
+    )
+    @pytest.mark.parametrize("index_col", [None, 0, 5])
```
RehanSD, line 877: Why was this changed?

modin/pandas/test/utils.py
```diff
@@ -120,6 +120,20 @@
...
+parse_dates_values_by_id = {
+    "bool": False,
+    "list_of_single_int": [0],
+    "list_of_single_string": ["timestamp"],
+    "list_of_list_of_strings": [["year", "month", "date"]],
+    "list_of_list_of_ints": [[1, 2, 3]],
+    "list_of_list_of_strings_and_ints": [["year", 2, "date"]],
+    "list_of_string_and_list_of_strings": ["timestamp", ["year", "month", "date"]],
+    "empty_list": [],
+    "nonexistent_string_column": ["z"],
```
RehanSD, line 135: What about nonexistent int column?

RehanSD (changes requested):
Left some comments, great work!

modin/experimental/pandas/test/test_io_exp.py
mvashishtha: This CSV is checking the `parse_dates` functionality, not the "reading from a glob path" functionality. Generally I think it's better for each test case to [test a single thing](https://osherove.com/blog/2005/4/3/a-unit-test-should-test-only-one-thing.html). Here, I think we should test complex `parse_dates` behavior without also testing glob paths. We do the same thing in `test_read_csv_empty_frame` [here](https://github.com/modin-project/modin/blob/f1de99167be9b357a5710f71b320316642121e84/modin/experimental/pandas/test/test_io_exp.py#L112).

modin/pandas/test/test_io.py
mvashishtha: I changed the `parse_dates` line because I wanted to use the constant `parse_dates_values_by_id` in two different tests. I had to change the `index_col` line because I inserted "year", "month", and "date" as columns 1-3 in the CSV.

modin/pandas/test/utils.py
mvashishtha: Added.

docs/getting_started/troubleshooting.rst
```diff
@@ -209,13 +210,87 @@ This can happen when you use OmniSci engine along with ``pyarrow.gandiva``:
   cfg.IsExperimental.put(True)
   import modin.pandas as pd
   import pyarrow.gandiva as gandiva  # Error
-  CommandLine Error: Option 'enable-vfe' registered more than once!
-  LLVM ERROR: inconsistency in registered CommandLine options
-  Aborted (core dumped)
+  # CommandLine Error: Option 'enable-vfe' registered more than once!
+  # LLVM ERROR: inconsistency in registered CommandLine options
```
RehanSD, line 214: Is this PR supposed to touch this file?

RehanSD (commented):
Quick Q about diffs

docs/getting_started/troubleshooting.rst
mvashishtha: No. I think this is a GitHub bug in the PR page. If I pull from GitHub and diff against master, I don't see this diff. These lines are in master [here](https://github.com/modin-project/modin/blob/master/docs/getting_started/troubleshooting.rst).

mvashishtha: This has happened to me before.

dchigarev: it probably happened, because your branch introduces some commits from the master branch that are not part of any merge-commit:

<details><summary>screenshot</summary>

image

</details>

I would suggest to try to rebase on master

mvashishtha: Yes, that works. The spurious diffs are gone.

modin/experimental/core/execution/native/implementations/omnisci_on_native/io/io.py
```diff
@@ -466,13 +466,14 @@ def _read_csv_check_support(
             )
 
         parse_dates_unsupported = isinstance(parse_dates, dict) or (
-            isinstance(parse_dates, list) and isinstance(parse_dates[0], list)
+            isinstance(parse_dates, list)
+            and any(not isinstance(date, str) for date in parse_dates)
```
jeffreykennethli, line 470: I'm curious why only dates that are string column names are supported and not int

e.g. `parse_dates = [1, 2, 3]`

is this something specific to the arrow engine?

modin/pandas/test/utils.py
```diff
@@ -120,6 +120,21 @@
     #     "col2": True,
     # },
 }
+# The parse_dates param can take several different types and combinations of
+# types. Use the following values to test date parsing on a CSV created for
+# that purpose at `time_parsing_csv_path`
+parse_dates_values_by_id = {
```
jeffreykennethli, line 126: nit: could also be named `parse_dates_id_to_value`.

modin/core/io/text/csv_glob_dispatcher.py
```diff
@@ -239,12 +239,23 @@ def _read(cls, filepath_or_buffer, **kwargs):
...
+                        for i, part in enumerate(date):
+                            if not isinstance(part, str):
```
jeffreykennethli, line 255: instead of doing this conversion when `part` is not a `str`, would it be safer to only do this conversion if `part` is of a certain type, such as `int`?

modin/experimental/core/execution/native/implementations/omnisci_on_native/io/io.py
mvashishtha: It seems pyarrow doesn't support `parse_dates`, so we have to [manually convert](https://github.com/mvashishtha/modin/blob/ca8e7b5955527abbe0c7927febbe29d205d7daff/modin/experimental/engines/omnisci_on_ray/io.py#L147) to timestamps ourselves. We haven't implemented support for the full variety of parse date types on omnisci. I think supporting `parse_dates` in OmniSci is beyond the scope of this PR.

modin/pandas/test/utils.py
mvashishtha: There was once a Python style guide tip at Google voting for names like this as opposed to `parse_dates_id_to_value` 😄 . One reason was that they compose better, e.g. `parse_dates_values_by_id_by_name['id']['name']` is easier to read than `id_to_name_to_parse_dates['id']['name']` because in the former, the index names in the variable name appear in the same order as the index names in the brackets.

modin/core/io/text/csv_glob_dispatcher.py
mvashishtha: Yes, done.

vnlitvinov:
@mvashishtha @RehanSD what is the status of this PR?

RehanSD (dismissed):
LGTM!

modin/core/io/text/csv_glob_dispatcher.py
```diff
@@ -239,12 +239,23 @@ def _read(cls, filepath_or_buffer, **kwargs):
...
+                    # or integer column indices. e.g. if parse_dates is
+                    # [[1, 2]] and columns at indices 1 and 2 are "b" and "c",
+                    # the output dataframe has the single date column "b_c". If
+                    # parse_dates is [["a", 1]] and the column at index 1 is
+                    # named "b", the output dataframe has the single date
+                    # column "a_b".
+                    if isinstance(date, list):
+                        for i, part in enumerate(date):
+                            if isinstance(part, int):
+                                date[i] = column_names[part]
```
vnlitvinov, line 256: TBH this looks like an unrelated fix (i.e. a fix for a _different_ bug). I wonder if we should implement this in a separate PR...

vnlitvinov (commented):
LGTM but one question about "two bugfixes for a price of one" here (so not stamping approval yet).

modin/core/io/text/csv_glob_dispatcher.py
mvashishtha: It is an unrelated fix for the bug that `read_csv_glob` can't handle lists that contain things other than strings, e.g. `[[0, 1]]`, while pandas can. I prefer doing the wrong thing and not separating this fix because it's just 5 lines of code including tests and this PR has been open for over a month, but it's up to you.

mvashishtha: But I will be less impatient :) I removed the separate bug fix and filed #4310 for it.

RehanSD (approved):
LGTM!

vnlitvinov (approved):
LGTM! And thanks for bearing with me being so nit-picking 🙃

modin/experimental/pandas/test/test_io_exp.py
```diff
@@ -119,6 +120,29 @@ def test_read_csv_without_glob(self):
             with pytest.raises(FileNotFoundError):
                 pd.read_csv_glob("s3://nyc-tlc/trip data/yellow_tripdata_2020-")
 
+    @pytest.mark.parametrize(
+        "parse_dates",
+        [pytest.param(value, id=id) for id, value in parse_dates_values_by_id.items()],
+    )
+    def test_read_single_csv_with_parse_dates(self, parse_dates):
+        try:
```
YarShev, line 128: Why do we complicate the code? Why don't we use just eval_general or eval_io?

mvashishtha: pandas doesn't have `read_csv_glob`, so we can't use those functions, which call the same function name in both pandas and Modin.

PR #4148, opened 2022-02-03
FIX-#4071: Fix DataFrame and Series equals() for None values.

### LINKED ISSUE #4071: BUG: modin's df.equals() doesn't treat nans as equal like pandas
### System information
Modin Version: 0.12.0
Ray 1.7.1
python 3.9.7

### Describe the problem

```
>>> df = modin_pd.DataFrame({"d": [389.0, 24.0, 80.5, None]})
>>> df.equals(df)
False
```

Pandas returns true
```
>>> df = pandas_pd.DataFrame({"d": [389.0, 24.0, 80.5, None]})
>>> df.equals(df)
True
```

mvashishtha:

## What do these changes do?

Currently, we implement DataFrame.equals() by calling DataFrame.eq() on both
input frames, then checking that every element of the result is True. However, pandas does not define `equals` equivalently. `pandas.eq` does not consider `None` to be equal to `None`, while `pandas.equals` does:

```python
import pandas
assert pandas.DataFrame([[None]]).equals(pandas.DataFrame([[None]]))
assert not pandas.DataFrame([[None]]).eq(pandas.DataFrame([[None]])).any().any()
```

It's hard to tell what exactly `equals` means for pandas, so I reimplemented it as a binary operation that transforms a pair of dataframes to dataframes with a single element telling whether the two dataframes are `equals`, followed by a check that every element of the result is true.

- [x] Resolves #4071 
- [x] tests added and passing

### DIFF
--- a/modin/core/storage_formats/base/query_compiler.py
+++ b/modin/core/storage_formats/base/query_compiler.py
@@ -37,7 +37,7 @@ from pandas.core.dtypes.common import is_scalar
 import pandas.core.resample
 import pandas
 import numpy as np
-from typing import List, Hashable
+from typing import List, Hashable, Optional
 
 
 def _get_axis(axis):
@@ -482,6 +482,41 @@ class BaseQueryCompiler(abc.ABC):
     @doc_utils.doc_binary_method(operation="equality comparison", sign="==")
     def eq(self, other, **kwargs):  # noqa: PR02
         return BinaryDefault.register(pandas.DataFrame.eq)(self, other=other, **kwargs)
+
+    def equals(
+        self,
+        other,
+        new_index: Optional[pandas.Index] = None,
+        new_columns: Optional[pandas.Index] = None,
+    ):
+        """
+        Test whether two objects contain the same elements.
+
+        If axes are not equal, perform frames alignment first. Note that
+        because frame alignment includes filling in empty values with NaN,
+        the query compiler for DataFrame([[1], [np.NaN]]) is equal to the query
+        compiler for DataFrame([[1]]]). In that case, the latter dataframe's
+        partition is expanded to DataFrame([[1], [np.NaN]]]).
+
+        Parameters
+        ----------
+        other : BaseQueryCompiler
+            The other compiler to be compared with the first.
+        new_index : pandas.Index, optional
+            New index of the frame resulting from the binary operation.
+        new_columns : pandas.Index, optional
+            New columns of the frame reuslting from the binary operation.
+
+        Returns
+        -------
+        BaseQueryCompiler
+            Result where each.
+        """
+        return BinaryDefault.register(
+            lambda x, y: pandas.DataFrame(
+                pandas.DataFrame.equals(x, y), index=new_index, columns=new_columns
+            )
+        )(self, other=other)
 
     @doc_utils.doc_binary_method(operation="integer division", sign="//")
     def floordiv(self, other, **kwargs):  # noqa: PR02
--- a/modin/core/storage_formats/base/query_compiler.py
+++ b/modin/core/storage_formats/base/query_compiler.py
@@ -501,7 +501,7 @@ class BaseQueryCompiler(abc.ABC):
         Parameters
         ----------
         other : BaseQueryCompiler
-            The other compiler to be compared with the first.
+            The other compiler to be compared with this one.
         new_index : pandas.Index, optional
             New index of the frame resulting from the binary operation.
         new_columns : pandas.Index, optional
@@ -510,7 +510,9 @@ class BaseQueryCompiler(abc.ABC):
         Returns
         -------
         BaseQueryCompiler
-            Result where each.
+            Result where each partition has a single element telling whether
+            the two objects are pandas.DataFrame.equals() in their
+            corresponding partitions.
         """
         return BinaryDefault.register(
             lambda x, y: pandas.DataFrame(
--- a/modin/pandas/test/dataframe/test_binary.py
+++ b/modin/pandas/test/dataframe/test_binary.py
@@ -72,7 +72,7 @@ def test_math_functions(other, axis, op):
 
 @pytest.mark.parametrize(
     "other",
-    [lambda df: df[: -(2**4)], lambda df: df[df.columns[0]].reset_index(drop=True)],
+    [lambda df: df[: -(2 ** 4)], lambda df: df[df.columns[0]].reset_index(drop=True)],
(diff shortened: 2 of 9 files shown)


modin/core/storage_formats/base/query_compiler.py
```diff
@@ -430,6 +430,41 @@ def combine_first(self, other, **kwargs):  # noqa: PR02
...
+        If axes are not equal, perform frames alignment first. Note that
+        because frame alignment includes filling in empty values with NaN,
+        the query compiler for DataFrame([[1], [np.NaN]]) is equal to the query
+        compiler for DataFrame([[1]]]). In that case, the latter dataframe's
+        partition is expanded to DataFrame([[1], [np.NaN]]]).
+
+        Parameters
+        ----------
+        other : BaseQueryCompiler
+            The other compiler to be compared with the first.
```
jeffreykennethli, line 451: nit: is it more accurate to say that the other compiler is compared to self, as opposed to first?

modin/core/storage_formats/base/query_compiler.py
```diff
@@ -430,6 +430,41 @@ def combine_first(self, other, **kwargs):  # noqa: PR02
...
+        new_index : pandas.Index, optional
+            New index of the frame resulting from the binary operation.
+        new_columns : pandas.Index, optional
+            New columns of the frame reuslting from the binary operation.
+
+        Returns
+        -------
+        BaseQueryCompiler
+            Result where each.
```
jeffreykennethli, line 460: description appears to not be completed.

modin/pandas/test/dataframe/test_binary.py
```diff
@@ -167,28 +167,68 @@ def test_multi_level_comparison(data, op):
...
+
+    pandas_equals = pandas_df1.equals(pandas_df2)
+    assert pandas_equals == expected_pandas_equals, (
+        f"Test expected pandas to say the dataframes were"
+        f"{'' if expected_pandas_equals else ' not'} equal, but they were"
+        f"{' not' if expected_pandas_equals else ''} equal."
+    )
 
-    assert modin_df1.equals(modin_df2._query_compiler.to_pandas())
+    assert modin_df1.equals(modin_df2) == pandas_equals
```
jeffreykennethli, line 230: nit: would it be useful to have a test case for `modin_df1.equals(modin_df1)` given that the original bug is that a DF is not `.equals(` to itself.

mvashishtha: The author of the bug just used that to show that to show that there was definitely a bug in Modin, because any object should be equal to itself. The root cause was that our `equals` doesn't work correctly when the dataframes include `None`. I think it's clearer to have the tests compare dataframes that should be equivalent because they were constructed from the same data.

devin-petersohn (commented):
What about adding a `none_equals` parameter to the query_compiler's `eq` method and handling the `None` values conditionally in that way?

Right now, this PR is passing metadata state from the API layer all the way to the internal dataframe representation.

mvashishtha:
> What about adding a `none_equals` parameter to the query_compiler's `eq` method and handling the `None` values conditionally in that way?
> 
> Right now, this PR is passing metadata state from the API layer all the way to the internal dataframe representation.

@devin-petersohn How could I do handle the `none_equals=True` case? the pandas `equals` appears to check that all pairs of blocks [[1]](https://github.com/pandas-dev/pandas/blob/c0d746fae38be8f9204979323f3847705494cdfc/pandas/core/internals/managers.py#L1380) are array_equivalent [[2]](https://github.com/pandas-dev/pandas/blob/2cb628d2eb8750bd9f3d2b34eb98da16380c2625/pandas/core/dtypes/missing.py#L607), where `array_equivalent` checks [[3]](https://github.com/pandas-dev/pandas/blob/2cb628d2eb8750bd9f3d2b34eb98da16380c2625/pandas/core/dtypes/missing.py#L458) that the arrays have `None` in the same places. How could we port that logic to Modin?

modin/pandas/test/dataframe/test_binary.py
```diff
@@ -167,28 +167,68 @@ def test_multi_level_comparison(data, op):
...
-        df_equals(modin_df3, modin_df1)
-
-    with pytest.raises(AssertionError):
-        df_equals(modin_df3, modin_df2)
+@pytest.mark.parametrize(
+    "frame1_data,frame2_data,expected_pandas_equals",
+    [
+        pytest.param({}, {}, True, id="two_empty_dataframes"),
+        pytest.param([[1]], [[0]], False, id="single_unequal_values"),
+        pytest.param([[None]], [[None]], True, id="single_none_values"),
```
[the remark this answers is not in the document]
mvashishtha: Thank you for commenting to check. The fix should work for `np.NaN` too. I added a test case for that.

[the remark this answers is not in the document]
adrienpacifico: Thanks to you. Is the merge request blocked because of unsuccessful checks, or is it for other reasons?
If so, do you need help to solve the code coverage decrease and lake8 issues?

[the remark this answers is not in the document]
mvashishtha: While the solution in this PR is functionally correct (i.e. it fixes the bug), @devin-petersohn thinks it's violating some internal abstractions for Modin, so it shouldn't be merged as is. However, I don't know the proper way to make the fix. [This comment](https://github.com/modin-project/modin/pull/4148#issuecomment-1098564017) is the last one in the thread about that discussion.

If you just need a temporary fix for this issue while we get this PR merged, you could `pip install` modin from the git branch that this pull request is comparing to Modin, i.e.

`pip install git+https://github.com/mvashishtha/modin.git@4071-fix-equals-for-none`

mvashishtha:
@devin-petersohn please let me know what you think about [this question](https://github.com/modin-project/modin/pull/4148#issuecomment-1098564017). If we want to continue with the approach in this PR, maybe we could pass a new bool param `should_compute_axis_labels` to `binary_op` that tells whether to recompute axis labels because the binary op could change them? That way we at least won't be passing the new metadata down from the API layer.

jbrockmendel:
> It's hard to tell what exactly equals means for pandas

Roughly: `((x == y) | (isna(x) & isna(y))).all()`.   Similar to `np.array_equal(x, y, equal_nan=True)`.  For object-dtype we also check for matching-NAs, see, pd._libs.lib.array_equivalent_object

PR #4184, opened 2022-02-09
FIX-#4183: Make frame _siblings an instance variable.

### LINKED ISSUE #4183: DataFrame operations affect other DataFrames
### System information
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: MacOS 12.0.1
- **Modin version** (`modin.__version__`): 0.13.1
- **Python version**: 3.8.8
```
# import pandas as pd
import modin.pandas as pd

df = pd.DataFrame(data={"col1": [1, 2], "col2": [3, 4]})

a = pd.DataFrame(df.sum().reset_index())
b = pd.DataFrame(df.mean().reset_index())
b.columns = ["b_col1", "b_col2"]
b["b_col2"] = b["b_col2"] / 2 # This is the operation that causes the corruption.

print(a.columns) 
# Modin returns Index(['b_col1', 'b_col2'], dtype='object')
# pandas returns Index(['index', 0], dtype='object')
```

### Describe the problem
When two dataframes are created from the same DataFrame, their column names can become corrupted by an operation on the other. I haven't had time to dig into a simpler reproducible case that would help find the source of the bug, except that without the line `b["b_col2"] = b["b_col2"] / 2` this mistake doesn't happen.

mvashishtha:

## What do these changes do?

- [x] Resolves #4183 
- [N/A: I don't think we need to continuously test for this particular bug.] tests added and passing

### DIFF
--- a/docs/release_notes/release_notes-0.14.0.rst
+++ b/docs/release_notes/release_notes-0.14.0.rst
@@ -19,7 +19,7 @@ Key Features and Updates
   * FIX-#4208: Fix lazy metadata update for `PandasDataFrame.from_labels` (#4209)
 * Performance enhancements
   * FIX-#4138, FIX-#4009: remove redundant sorting in the internal '.mask()' flow (#4140)
-  * FIX-#4183: Make frame _siblings an instance variable (#4184)
+  * FIX-#4183: Stop shallow copies from creating global shared state. (#4184)
 * Benchmarking enhancements
   *
 * Refactor Codebase
--- a/docs/release_notes/release_notes-0.14.0.rst
+++ b/docs/release_notes/release_notes-0.14.0.rst
@@ -19,6 +19,7 @@ Key Features and Updates
   * FIX-#4208: Fix lazy metadata update for `PandasDataFrame.from_labels` (#4209)
 * Performance enhancements
   * FIX-#4138, FIX-#4009: remove redundant sorting in the internal '.mask()' flow (#4140)
+  * FIX-#4183: Make frame _siblings an instance variable (#4184)
 * Benchmarking enhancements
   *
 * Refactor Codebase
--- a/modin/pandas/dataframe.py
+++ b/modin/pandas/dataframe.py
@@ -2452,8 +2452,9 @@ class DataFrame(BasePandasDataset):
             Value to set.
         """
         # We have to check for this first because we have to be able to set
-        # _query_compiler before we check if the key is in self
-        if key in ["_query_compiler"] or key in self.__dict__:
+        # _query_compiler and _siblings before we check whether either key is
+        # in self.
+        if key in ["_query_compiler", "_siblings"] or key in self.__dict__:
             pass
         elif key in self and key not in dir(self):
             self.__setitem__(key, value)
--- a/modin/pandas/dataframe.py
+++ b/modin/pandas/dataframe.py
@@ -107,6 +107,9 @@ class DataFrame(BasePandasDataset):
         copy=None,
         query_compiler=None,
     ):
+        # Siblings are other dataframes that share the same query compiler. We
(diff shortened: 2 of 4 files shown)


modin/pandas/dataframe.py
```diff
@@ -2451,8 +2454,9 @@ def __setattr__(self, key, value):
             Value to set.
         """
         # We have to check for this first because we have to be able to set
-        # _query_compiler before we check if the key is in self
-        if key in ["_query_compiler"] or key in self.__dict__:
+        # _query_compiler and _siblings before we check whether either key is
+        # in self.
+        if key in ["_query_compiler", "_siblings"] or key in self.__dict__:
```
anmyachev, line 2459: How will the `_siblings` assignment take place in this case?

jeffreykennethli (commented):
Thanks for the fix! Are there any performance benefits/drawbacks to making `_siblings` an instance variable?

modin/pandas/dataframe.py
mvashishtha: In this case, we just set the `_siblings` variable as usual [here](https://github.com/modin-project/modin/blob/47b8a1ce82c7ff6d786da355edb034ffb1c21b4f/modin/pandas/dataframe.py#L2463). We do the same thing with `_query_compiler`. The point of the `pass` is to avoid trying to call `__setitem__` with these Modin instance variables.

mvashishtha:
> Thanks for the fix! Are there any performance benefits/drawbacks to making `_siblings` an instance variable?

@jeffreykennethli  I think fixing the wrong behavior here is so important that we shouldn't care about performance. This seems like a very serious bug, producing [spooky action at a distance](https://github.com/modin-project/modin/issues/4183#issuecomment-1034115612) across arbitrary dataframes.

jeffreykennethli:
@mvashishtha Agreed, fixing the behavior is highest priority!

anmyachev (approved):

modin/pandas/dataframe.py
RehanSD: Can we add a comment explaining this codepath please?

RehanSD (changes requested):
LGTM! Let's merge after comment!

modin/pandas/dataframe.py
mvashishtha: @RehanSD my explanation offline last week was incorrect. It turns out that if we remove the check for `_query_compiler` in master, or for either variable in this PR, we get an infinite recursion when we try to construct any dataframe. I added an explanation in a comment.

RehanSD (approved):
LGTM! Can you add this to release notes before merging?

RehanSD (approved):
LGTM!

docs/release_notes/release_notes-0.14.0.rst
```diff
@@ -19,6 +19,7 @@ Key Features and Updates
   * FIX-#4208: Fix lazy metadata update for `PandasDataFrame.from_labels` (#4209)
 * Performance enhancements
   * FIX-#4138, FIX-#4009: remove redundant sorting in the internal '.mask()' flow (#4140)
+  * FIX-#4183: Stop shallow copies from creating global shared state. (#4184)
```
YarShev, line 22: @mvashishtha, why are this release note and the PR description different? These should be aligned.

mvashishtha: @RehanSD asked me to make the release note describe the bug that this PR fixes. I've now made the PR description the same as the release note.

YarShev: That should've been done before merge as we now have `Make frame _siblings an instance variable.` in the commit history. Please keep in mind this in the future.

PR #4195, opened 2022-02-11
REFACTOR-#4194: create 'BaseDataframePartition' class

### LINKED ISSUE #4194: REFACTOR: create `BaseDataframePartition` class, which will define an interface
The same has already been done for `DataframeAxisPartition` hierarchy.

anmyachev:

## What do these changes do?

- [x] Resolves #4194 

### DIFF
--- a/modin/core/dataframe/base/partitioning/partition.py
+++ b/modin/core/dataframe/base/partitioning/partition.py
@@ -200,6 +200,7 @@ class BaseDataframePartition(ABC):  # pragma: no cover
         """
         pass
 
+    @abstractmethod
     def length(self):
         """
         Get the length of the object wrapped by this partition.
@@ -209,12 +210,9 @@ class BaseDataframePartition(ABC):  # pragma: no cover
         int
             The length of the object.
         """
-        if self._length_cache is None:
-            func = self._length_extraction_fn()
-            preprocessed_func = self.preprocess_func(func)
-            self._length_cache = self.apply(preprocessed_func)
-        return self._length_cache
+        pass
 
+    @abstractmethod
     def width(self):
         """
         Get the width of the object wrapped by the partition.
@@ -224,11 +222,7 @@ class BaseDataframePartition(ABC):  # pragma: no cover
         int
             The width of the object.
         """
-        if self._width_cache is None:
-            func = self._width_extraction_fn()
-            preprocessed_func = self.preprocess_func(func)
-            self._width_cache = self.apply(preprocessed_func)
-        return self._width_cache
+        pass
 
     @classmethod
     @abstractmethod
--- a/modin/core/dataframe/base/partitioning/partition.py
+++ b/modin/core/dataframe/base/partitioning/partition.py
@@ -1,0 +1,272 @@
+# Licensed to Modin Development Team under one or more contributor license agreements.
+# See the NOTICE file distributed with this work for additional information regarding
+# copyright ownership.  The Modin Development Team licenses this file to you under the
+# Apache License, Version 2.0 (the "License"); you may not use this file except in
+# compliance with the License.  You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software distributed under
+# the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific language
+# governing permissions and limitations under the License.
+
+"""The module defines base interface for a partition of a Modin DataFrame."""
+
+from abc import ABC
+
+
+class BaseDataframePartition(ABC):  # pragma: no cover
+    """
+    An abstract class that represents the parent class for any partition class.
+
+    Parameters
+    ----------
+    length : future-like or int, optional
+        Length or reference to it of wrapped DataFrame-like object.
+    width : future-like or int, optional
+        Width or reference to it of wrapped DataFrame-like object.
+    call_queue : list, optional
+        Call queue that needs to be executed on wrapped DataFrame-like object.
+    """
+
+    def __init__(self, length=None, width=None, call_queue=None):
+        self._length_cache = length
+        self._width_cache = width
+        self.call_queue = call_queue or []
+
+    def get(self):
+        """
+        Get the object wrapped by this partition.
+
+        Returns
+        -------
+        object
+            The object that was wrapped by this partition.
+
+        Notes
+        -----
+        This is the opposite of the classmethod `put`.
+        E.g. if you assign `x = BaseDataframePartition.put(1)`, `x.get()` should
+        always return 1.
+        """
+        pass
+
+    def apply(self, func, *args, **kwargs):
+        """
+        Apply a function to the object wrapped by this partition.
+
+        Parameters
+        ----------
+        func : callable
+            Function to apply.
+        *args : iterable
+            Additional positional arguments to be passed in `func`.
+        **kwargs : dict
+            Additional keyword arguments to be passed in `func`.
+
+        Returns
+        -------
+        BaseDataframePartition
+            New `BaseDataframePartition` object.
+
+        Notes
+        -----
+        It is up to the implementation how `kwargs` are handled. They are
+        an important part of many implementations. As of right now, they
+        are not serialized.
+        """
+        pass
+
+    def add_to_apply_calls(self, func, *args, **kwargs):
+        """
+        Add a function to the call queue.
+
+        Parameters
+        ----------
+        func : callable
+            Function to be added to the call queue.
+        *args : iterable
+            Additional positional arguments to be passed in `func`.
+        **kwargs : dict
+            Additional keyword arguments to be passed in `func`.
+
+        Returns
+        -------
+        BaseDataframePartition
+            New `BaseDataframePartition` object with the function added to the call queue.
+
+        Notes
+        -----
+        This function will be executed when `apply` is called. It will be executed
+        in the order inserted; apply's func operates the last and return.
+        """
+        pass
+
+    def drain_call_queue(self):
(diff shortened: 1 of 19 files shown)


modin/core/dataframe/pandas/partitioning/partition.py
```diff
@@ -263,38 +143,6 @@ def _width_extraction_fn(cls):
...
-        Returns
-        -------
-        int
-            The width of the object.
-        """
-        if self._width_cache is None:
-            cls = type(self)
-            func = cls._width_extraction_fn()
-            preprocessed_func = cls.preprocess_func(func)
-            self._width_cache = self.apply(preprocessed_func)
```
anmyachev, line 295: In the current implementation, `self._width_cache` variable stores an object of partition type. 

Modin's code works because each child class gets rid of this error by overriding the method.

modin/core/dataframe/pandas/partitioning/partition.py
```diff
@@ -13,102 +13,25 @@
...
 
 from modin.pandas.indexing import compute_sliced_len
 from modin.core.storage_formats.pandas.utils import length_fn_pandas, width_fn_pandas
+from modin.core.dataframe.base.partitioning.partition import BaseDataframePartition
 
 
-class PandasDataframePartition(ABC):  # pragma: no cover
+class PandasDataframePartition(
+    BaseDataframePartition
+):  # pragma: no cover; # noqa: PR01
```
anmyachev, line 28: `PR01` is a mute of the doc checks of  `__init__` method, which is inherited from the base class.

modin/core/dataframe/base/partitioning/partition.py
```diff
@@ -0,0 +1,272 @@
...
+    ----------
+    length : future-like or int, optional
+        Length or reference to it of wrapped DataFrame-like object.
+    width : future-like or int, optional
+        Width or reference to it of wrapped DataFrame-like object.
+    call_queue : list, optional
+        Call queue that needs to be executed on wrapped DataFrame-like object.
+    """
+
+    def __init__(self, length=None, width=None, call_queue=None):
```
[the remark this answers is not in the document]
anmyachev: Why not move part of the constructor to the base class if this part is repeated in all descendant' constructors?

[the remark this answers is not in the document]
YarShev: If this class was really a base class with some methods implemented, I would be okay with that. However, we should declare an interface as an abstract class to be implemented by child classes.

Note: there is no call_queue in omnisci engine. length and width might be anything that is beneficial for a concrete future engine, but not only future-like or int.

modin/core/dataframe/pandas/partitioning/partition.py
```diff
@@ -198,47 +121,6 @@ def try_recompute_cache(indices, previous_cache):
...
-        -----
-        This is a classmethod because the definition of how to preprocess
-        should be class-wide. Also, we may want to use this before we
-        deploy a preprocessed function to multiple `PandasDataframePartition`
-        objects.
-        """
-        pass
-
     @classmethod
     def _length_extraction_fn(cls):
```
YarShev, line 125: Should we really have this function?

modin/core/dataframe/base/partitioning/partition.py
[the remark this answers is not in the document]
devin-petersohn: Agree with @YarShev this is an ABC and should not have an `__init__`.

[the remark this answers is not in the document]
anmyachev: Moved in `PandasDataframePartition` class.

modin/core/dataframe/base/partitioning/partition.py
```diff
@@ -0,0 +1,224 @@
...
+
+        Notes
+        -----
+        This is the opposite of the classmethod `put`.
+        E.g. if you assign `x = BaseDataframePartition.put(1)`, `x.get()` should
+        always return 1.
+        """
+        pass
+
+    def apply(self, func, *args, **kwargs):
```
dchigarev, line 39: Do we imply that ANY partition implementation has to be able to apply arbitrary python functions? For instance, OmniSci can't do that, so wouldn't this make OmniSci a semi-backend since it won't fully implement the base partition's interface?

dchigarev: I mean, how general this base class should be? Do we really need all these methods in the base class?
- `apply`/`preprocess_func` - implies that any backend can execute callables again partition's content
- `add_to_apply_calls`/`drain_call_queue`/`wait`- implies that any implementation supports lazy execution

YarShev: I would prefer OmniSci is able to do that to follow Modin conception. But not sure how much efforts that would take to rewrite the code. For now I think we can omit these methods in the base class and will try to implement an OmniSci replacement backend, which will follow the conception.

YarShev: Or, we could remove the inheritance of the base class for OmniSci partition if it doesn't use anything from the base.

modin/core/dataframe/pandas/partitioning/partition.py
YarShev: Looks good. We could probably preprocess these functions at the class level. However, the perf should be checked.

anmyachev: Split the changes into https://github.com/modin-project/modin/pull/4271

modin/core/dataframe/base/partitioning/partition.py
anmyachev: > I mean, how general this base class should be? Do we really need all these methods in the base class?
> 
> * `apply`/`preprocess_func` - implies that any backend can execute callables again partition's content
> * `add_to_apply_calls`/`drain_call_queue`/`wait`- implies that any implementation supports lazy execution

I think these functions should be defined in the base class - thus we confirm that the inherited partition will have these properties. This is beneficial in terms of code reuse, since almost all implementations support this interface.

And yes, this interface is not convenient for OmniSci, simply because this implementation does not fit seamlessly into another parallelization mechanism. I think we can consider that partitions are an internal implementation detail, and not require for each implementation a one-to-one correspondence of interfaces.

So inheritance for OmniSci partition is removed.

modin/core/dataframe/base/partitioning/partition.py
```diff
@@ -0,0 +1,243 @@
...
+# the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific language
+# governing permissions and limitations under the License.
+
+"""The module defines base interface for a partition of a Modin DataFrame."""
+
+from abc import ABC, abstractmethod
+
+
+class BaseDataframePartition(ABC):  # pragma: no cover
```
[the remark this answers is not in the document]
anmyachev: removed

modin/core/dataframe/pandas/partitioning/partition.py
```diff
@@ -13,101 +13,36 @@
...
-        in the order inserted; apply's func operates the last and return.
-        """
-        pass
 
-    def drain_call_queue(self):
-        """Execute all operations stored in the call queue on the object wrapped by this partition."""
-        pass
+    Parameters
+    ----------
+    length : future-like or int, optional
```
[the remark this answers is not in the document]
anmyachev: In that case we need unify name of wrapped objects. (We have following names: `data`, `oid`, `futures`)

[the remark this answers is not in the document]
YarShev: I am ok with data.

anmyachev:
@YarShev ready for review

modin/core/dataframe/base/partitioning/partition.py
dchigarev: I think that the problem is you're trying to fit all the pandas storage-format-specific methods into a base class. We should not do that. The modin architecture shouldn't be limited by a single type of execution backend (the one that supports arbitrary lambdas/lazy execution/so on), the base class has to be general enough to be implemented by a various amount of execution systems, not only the one that's ray+pandas like.

> I think we can consider that partitions are an internal implementation detail, and not require for each implementation

The experience with integrating Omnisci into Modin shows, that even non-partition-based execution may exploit the partition concept by limiting the total amount of partitions to 1 and so being able to smoothly integrate into a partitioned-based execution system (Modin).

> So inheritance for OmniSci partition is removed.

Doing that makes a risk to end-up with monolith architecture where none of the layers can be replaced with each other, instead of the original concept, where it's implied that we can seamlessly replace one layer with another and everything should work properly (maybe not optimal, but properly). The refactoring activities should move us toward the goal of layers' independency, not throwing us back.

So, what's my suggestion? My suggestion is to carefully think about the base partition API, exclude all of the execution-specific methods, discuss the API with our consortium (apparently modin-core members), and define a decent base API that would be suitable for any Modin's implementation.

dchigarev: I think a good base for the discussion could be a partition class with excluded pandas-specific methods described in [this message](https://github.com/modin-project/modin/pull/4195/files#r810036924).

modin/experimental/core/execution/native/implementations/omnisci_on_native/partitioning/partition.py
```diff
@@ -15,12 +15,11 @@
 
 import pandas
 
-from modin.core.dataframe.pandas.partitioning.partition import PandasDataframePartition
 from ..omnisci_worker import OmnisciServer
 import pyarrow
 
 
-class OmnisciOnNativeDataframePartition(PandasDataframePartition):
+class OmnisciOnNativeDataframePartition:
```
dchigarev, line 22: [discussion](https://github.com/modin-project/modin/pull/4195#discussion_r811647081)

anmyachev:
Too hard to refresh.

PR #4203, opened 2022-02-15
TEST-#4201: Add read_parquet, explode, tail, and various arithmetic t…

### LINKED ISSUE #4201: TEST: Add more functions to asv_bench

We want to ensure that the most commonly used Pandas functions are a part of `asv_bench` and that they are running benchmarks on different dataset sizes across different numbers of CPUs comparing Modin vs. Pandas.

naren-ponder:

## What do these changes do?

Linked to #4201 

- [x] Resolves #4201 ? 
- [x] tests added and passing


asv_bench/benchmarks/benchmarks.py
```diff
@@ -314,6 +317,33 @@ def time_apply(self, shape, axis):
...
+    def time_mod(self, shape, axis):
+        execute(self.df.mod(2, axis=axis))
+
+    def time_abs(self, shape):
+        execute(self.df.abs())
+
+    def time_aggregate(self, shape, axis):
+        execute(self.df.aggregate(lambda df: df.sum(), axis=axis))
+
+    def value_counts(self, shape):
```
anmyachev, line 338: We already bench the function: https://github.com/modin-project/modin/blob/b7027597985a442734a57949d72e3fa0625b4d43/asv_bench/benchmarks/benchmarks.py#L477

asv_bench/benchmarks/benchmarks.py
```diff
@@ -314,6 +317,33 @@ def time_apply(self, shape, axis):
...
+        execute(self.df.value_counts())
+
+    def is_in(self, shape):
+        execute(self.df.isin([0, 2]))
+
+    def transpose(self, shape):
```
anmyachev, line 344:
```suggestion
    def time_transpose(self, shape):
```

naren-ponder:
Blocked for now by #4221.

asv_bench/benchmarks/benchmarks.py
```diff
@@ -376,6 +403,40 @@ def time_head(self, shape, head_count):
         execute(self.df.head(self.head_count))
 
 
+class TimeTail:
+    param_names = ["shape", "tail_count"]
+    params = [
+        get_benchmark_shapes("TimeTail"),
+        [5, 0.8],
```
jeffreykennethli, line 410: Is it worth adding other values for `tail_count` (and `head_count` for TimeHead) here? This could be a continuous range of percentages of the df to get head/tail for, e.g. `np.arange(0.1, 1, 0.1)`.

asv_bench/benchmarks/io/parquet.py
```diff
@@ -0,0 +1,58 @@
...
+    def setup(self, test_filenames, shape, *args, **kwargs):
+        # ray init
+        if ASV_USE_IMPL == "modin":
+            pd.DataFrame([])
+        self.shape_id = get_shape_id(shape)
+
+
+class TimeReadParquet(BaseReadParquet):
+    shapes = get_benchmark_shapes("TimeReadParquet")
+    data_type = "str_int"
```
jeffreykennethli, line 43: should data_type here include `{"int", "true_false_int"}` as well (from prepare_io_data_parquet).

asv_bench/benchmarks/utils/common.py
```diff
@@ -563,3 +563,32 @@ def prepare_io_data(test_filename: str, data_type: str, shapes: list):
...
+
+    Returns
+    -------
+    test_filenames : dict
+        Dictionary that maps dataset shape to the file on disk.
+    """
+    test_filenames = {}
+    for shape in shapes:
+        shape_id = get_shape_id(shape)
+        test_filenames[shape_id] = f"{test_filename}_{shape_id}.parquet"
```
jeffreykennethli, line 590: If multiple `prepare_io_data_parquet` are called with different `data_type`, the file names would conflict, so perhaps we could add `data_type` to the file name.

asv_bench/benchmarks/benchmarks.py
anmyachev: It seems to me that this change does not cover any new use case for the functions.

asv_bench/benchmarks/io/parquet.py
anmyachev: I don't think it's necessary to test the case where all data is numeric, since the `str_int` data type is more general - it includes both string columns and numeric. There is no specificity in our reading algorithms in these case.

As for `true_false_int` data type, we should write a separate benchmark (for example, like `TimeReadCsvTrueFalseValues`), but I would not add it until we stop falling into pandas when `true_values` or `false_values` parameters are specified.

asv_bench/benchmarks/utils/common.py
anmyachev: Good point - this also applies to `csv` format.

asv_bench/benchmarks/io/parquet.py
naren-ponder: With `read_csv` in Pandas there are parameters to specify `true_values` and `false_values` that we don't have with `read_sql`. Good point on the `data_type` comment with ensure consistency with `prepare_io_data_parquet`.

jeffreykennethli (dismissed):

anmyachev (dismissed):
LGTM!

asv_bench/benchmarks/io/parquet.py
```diff
@@ -0,0 +1,58 @@
...
+    ASV_USE_IMPL,
+    IMPL,
+    execute,
+    get_shape_id,
+    prepare_io_data_parquet,
+    get_benchmark_shapes,
+)
+
+
+class BaseReadParquet:
```
mvashishtha, line 26: I don't think we should add a base class until we have multiple subclasses.

mvashishtha (approved):
I left one comment. Apart from that, LGTM.

asv_bench/benchmarks/io/parquet.py
naren-ponder: We do this with read_csv, and we may add other things for parquet longer term so just wanted consistency.

naren-ponder: We do this with read_csv, so wanted some consistency in case in future we add things for parquet.

mvashishtha: I generally prefer in cases like this to assume [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it) and write the simpler code. However, I won't insist.

devin-petersohn: I agree with @mvashishtha, let's keep it simple.

mvashishtha (approved):

devin-petersohn (approved):
LGTM

PR #4262, opened 2022-02-24
FIX-#4185: Fix rolling across column partitions.

### LINKED ISSUE #4185: DataFrame.rolling() gives incorrect result when windows span two column partitions and axis == 1
### System information
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: MacOS Monterey 12.2
- **Modin version** (`modin.__version__`): 0.13.0+19.g62d73b78
- **Python version**: 3.9.9

```python
import modin.pandas as pd
df = pd.DataFrame([range(33)])
# result is nan
print(df.rolling(window=2, axis=1).min().iloc[0, -1])
import pandas
pdf = pandas.DataFrame([range(33)])
# result is 33
print(pdf.rolling(window=2, axis=1).min().iloc[0, -1])
```

### Describe the problem

When `axis=1`, we "fold" (really a better term is "map full axis partitions") along axis=0 instead of axis=1. The result is that each column partition gets results independently of other column partitions. We need to pass `axis` to the fold calls as [here](https://github.com/modin-project/modin/blob/62d73b7882fc58c4ef88f05bf596ece9f3cfe4fe/modin/core/storage_formats/pandas/query_compiler.py#L1076).

mvashishtha:
## What do these changes do?

Fix rolling across column partitions by mapping axis partitions that span the correct dimension.

- [x] Resolves #4185
- [x] tests added and passing

### DIFF
--- a/modin/core/dataframe/algebra/fold.py
+++ b/modin/core/dataframe/algebra/fold.py
@@ -36,7 +36,27 @@ class Fold(Operator):
         """
 
         def caller(query_compiler, fold_axis=None, *args, **kwargs):
-            """Execute Fold function against passed query compiler."""
+            """
+            Execute Fold function against passed query compiler.
+
+            Parameters
+            ----------
+            query_compiler : BaseQueryCompiler
+                The query compiler to execute the function on.
+            axis : int, optional
+                0 or None means apply across full column partitions. 1 means
+                apply across full row partitions.
+            *args : iterable
+                Additional arguments passed to fold_function.
+            **kwargs: dict
+                Additional keyword arguments passed to fold_function.
+
+            Returns
+            -------
+            BaseQueryCompiler
+                A new query compiler representing the result of executing the
+                function.
+            """
             return query_compiler.__constructor__(
                 query_compiler._modin_frame.fold(
                     cls.validate_axis(fold_axis),
--- a/modin/core/dataframe/algebra/fold.py
+++ b/modin/core/dataframe/algebra/fold.py
@@ -20,7 +20,7 @@ class Fold(Operator):
     """Builder class for Fold functions."""
 
     @classmethod
-    def call(cls, fold_function, axis=None):
+    def call(cls, fold_function):
         """
         Build Fold operator that will be performed across rows/columns.
 
@@ -28,8 +28,6 @@ class Fold(Operator):
         ----------
         fold_function : callable(pandas.DataFrame) -> pandas.DataFrame
             Function to apply across rows/columns.
-        axis : int, optional
-            Specifies axis to apply function along.
 
         Returns
         -------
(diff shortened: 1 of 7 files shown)


modin/core/storage_formats/base/doc_utils.py
```diff
@@ -567,10 +567,11 @@ def doc_window_method(
     callable
     """
     template = """
-        Create {win_type} and {action} for each window.
+        Create {win_type} and {action} for each window over the given axis.
 
         Parameters
         ----------
+        fold_axis : {{0, 1}}
```
naren-ponder, line 574: Perhaps I'm missing something, but don't see `fold_axis` as a parameter for `doc_window_method`.

mvashishtha: It's not a parameter to `doc_window_method`, but it's a parameter to each of the window methods in the query compiler :) `doc_window_method` generates the docstrings for those methods. The `fold_axis` line here is in a docstring template and will go into the docstring for each method.

modin/core/storage_formats/base/doc_utils.py
```diff
@@ -286,7 +286,7 @@ def doc_reduce_agg(method, refer_to, params=None, extra_params=None):
 
     Parameters
     ----------
-    axis : {{0, 1}}
+    fold_axis : {{0, 1}}
```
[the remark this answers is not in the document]
mvashishtha: The fold is basically a `map_axis_partitions`, which maps an operations across either entire rows or entire columns. Mapping across both at the same time would be equivalent to mapping across the entire dataframe, which the algebra has no support for.

modin/core/dataframe/algebra/fold.py
```diff
@@ -20,29 +20,26 @@ class Fold(Operator):
...
-            Specifies axis to apply function along.
 
         Returns
         -------
         callable
             Function that takes query compiler and executes Fold function.
         """
 
-        def caller(query_compiler, *args, **kwargs):
+        def caller(query_compiler, fold_axis=None, *args, **kwargs):
```
devin-petersohn, line 38: Let's add a doc for the new parameter.

devin-petersohn (commented):
Looks pretty good, left a comment.

modin/core/dataframe/algebra/fold.py
mvashishtha: This inner function didn't have a docstring at all, but I added it, and documented the new parameter.

mvashishtha: Done.

modin/core/dataframe/algebra/fold.py
```diff
@@ -20,29 +20,46 @@ class Fold(Operator):
...
-            _axis = kwargs.get("axis") if axis is None else axis
+        def caller(query_compiler, fold_axis=None, *args, **kwargs):
+            """
+            Execute Fold function against passed query compiler.
+
+            Parameters
+            ----------
+            query_compiler : BaseQueryCompiler
+                The query compiler to execute the function on.
+            axis : int, optional
```
naren-ponder, line 46: This should be `fold_axis` I think.

naren-ponder (dismissed):
Left one minor comment, but LGTM to me otherwise!

modin/core/dataframe/algebra/fold.py
mvashishtha: Thanks. Fixed.

naren-ponder (approved):
🚢

devin-petersohn (approved):
LGTM, thanks @mvashishtha !

PR #4324, opened 2022-03-16
FEAT-#3043: allow `pandas.DataFrame` to be passed into `merge` and `join` functions

### LINKED ISSUE #3043: Typeerror: can only merge series or dataframe objects a class 'pandas.core.frame.dataframe' was passed
File "D: \Python程序\因子入库0421\factor_ top_ pure_ czl_ 0421_ gtja191.py", line 323,
in BarraDataPreparelong

BarraMatrix long = pd . merge barraclass, industry，on=['code'，'date'], how= ' inner' )#在barra因子矩阵右侧追加行业哑变量矩阵

File "C: \Users \Administrator\AppData \Roaming \Python\Python38 \site - packages\modin\pandas \general.py", line 84,in merge
raise TypeError(
TypeError: Can only merge Series or DataFrame objects, a <class
pandas. core. frame .DataFrame' > was passed

anyone kowns how to deal with it? 0.0

anmyachev:

## What do these changes do?

- [x] Resolves #3043 
- [ ] tests added and passing

### DIFF
--- a/modin/pandas/dataframe.py
+++ b/modin/pandas/dataframe.py
@@ -1387,19 +1387,25 @@ class DataFrame(BasePandasDataset):
                 raise ValueError("Cannot merge a Series without a name")
             else:
                 right = right.to_frame()
-        if not isinstance(right, DataFrame):
+        if not isinstance(right, (DataFrame, pandas.DataFrame)):
             raise TypeError(
                 f"Can only merge Series or DataFrame objects, a {type(right)} was passed"
             )
 
         if left_index and right_index:
+            if not isinstance(right, DataFrame):
+                raise TypeError(
+                    f"Can only merge Series or DataFrame objects, a {type(right)} was passed"
+                )
             return self.join(
                 right, how=how, lsuffix=suffixes[0], rsuffix=suffixes[1], sort=sort
             )
 
         return self.__constructor__(
             query_compiler=self._query_compiler.merge(
-                right._query_compiler,
+                right._query_compiler
+                if not isinstance(right, pandas.DataFrame)
+                else right,
                 how=how,
                 on=on,
                 left_on=left_on,
--- a/modin/pandas/dataframe.py
+++ b/modin/pandas/dataframe.py
@@ -1258,7 +1258,9 @@ class DataFrame(BasePandasDataset):
         if on is not None:
             return self.__constructor__(
                 query_compiler=self._query_compiler.join(
-                    other._query_compiler,
+                    other._query_compiler
+                    if not isinstance(other, pandas.DataFrame)
+                    else other,
                     on=on,
                     how=how,
                     lsuffix=lsuffix,
@@ -1266,6 +1268,8 @@ class DataFrame(BasePandasDataset):
                     sort=sort,
                 )
             )
+        if isinstance(other, pandas.DataFrame):
+            other = DataFrame(other)
         if isinstance(other, DataFrame):
             # Joining the empty DataFrames with either index or columns is
             # fast. It gives us proper error checking for the edge cases that
--- a/modin/pandas/test/test_general.py
+++ b/modin/pandas/test/test_general.py
@@ -21,6 +21,7 @@ from modin.utils import get_current_execution, to_pandas
 from modin.test.test_utils import warns_that_defaulting_to_pandas
 
 from .utils import (
+    test_data,
     test_data_values,
     test_data_keys,
     df_equals,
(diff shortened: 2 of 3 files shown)


modin/pandas/dataframe.py
```diff
@@ -1387,19 +1391,25 @@ def merge(
                 raise ValueError("Cannot merge a Series without a name")
             else:
                 right = right.to_frame()
-        if not isinstance(right, DataFrame):
+        if not isinstance(right, (DataFrame, pandas.DataFrame)):
```
dchigarev, line 1394: we unconditionally converted `right` to pd.DataFrame above, why do we want to check for pandas.DataFrame here?

modin/pandas/dataframe.py
```diff
@@ -1387,19 +1391,25 @@ def merge(
...
             raise TypeError(
                 f"Can only merge Series or DataFrame objects, a {type(right)} was passed"
             )
 
         if left_index and right_index:
+            if not isinstance(right, DataFrame):
+                raise TypeError(
+                    f"Can only merge Series or DataFrame objects, a {type(right)} was passed"
+                )
```
dchigarev, line 1403: why the above check is not enough?

modin/pandas/test/test_general.py
```diff
@@ -179,6 +180,21 @@ def test_merge():
         pd.merge("Non-valid type", modin_df2)
 
 
+@pytest.mark.parametrize("data", test_data_values, ids=test_data_keys)
+def test_merge_pandas(data):
+    modin_df, pandas_df = pd.DataFrame(data), pandas.DataFrame(data)
```
dchigarev, line 185: this has to be imported from `modin.pandas.test.utils`
```suggestion
    modin_df, pandas_df = create_test_dfs(data)
```

modin/pandas/test/test_general.py
```diff
@@ -179,6 +180,21 @@ def test_merge():
...
+    df_equals(modin_df.merge(pandas_df), modin_df.merge(modin_df))
+
+
+def test_join_pandas():
+    data = test_data["int_data"]
+    modin_df, pandas_df = pd.DataFrame(data), pandas.DataFrame(data)
```
dchigarev, line 191:
```suggestion
    modin_df, pandas_df = create_test_dfs(data)
```

pyrito:
@anmyachev what's the status of this PR?

vnlitvinov:
I think @devin-petersohn was specifically against mixing Modin and pandas in that way, even though it's a "low hanging fruit" optimization in this case.

I feel that this should be better addressed by implementing https://github.com/modin-project/modin/issues/4605 and allowing the end user to easily "force" Modin to wrap certain pandas dataframe in this specific "small df" QueryCompiler.

anmyachev:
@pyrito for now, this most likely will not be merged, @vnlitvinov is right.

By the way, I have an interesting case, which may not be very relevant to this task, but still.

This call works for pandas, but fails for modin because modin tries to initialize ray on the worker (if I remember correctly).

In this case, the following change helps: `pd.Series...` -> `pandas.Series`. However, this is more like a short-term workaround than a general solution. What can we do here?

```python
import modin.pandas as pd

def agg_func(x):
    return int(pd.Series.count(x) > 0)

res = some_df.pivot_table(..., agg_func=agg_func)
```

RehanSD:
> @pyrito for now, this most likely will not be merged, @vnlitvinov is right.
> 
> 
> 
> By the way, I have an interesting case, which may not be very relevant to this task, but still.
> 
> 
> 
> This call works for pandas, but fails for modin because modin tries to initialize ray on the worker (if I remember correctly).
> 
> 
> 
> In this case, the following change helps: `pd.Series...` -> `pandas.Series`. However, this is more like a short-term workaround than a general solution. What can we do here?
> 
> 
> 
> ```python
> 
> import modin.pandas as pd
> 
> 
> 
> def agg_func(x):
> 
>     return int(pd.Series.count(x) > 0)
> 
> 
> 
> res = some_df.pivot_table(..., agg_func=agg_func)
> 
> ```

@anmyachev this fails since the function is running on a ray worker - maybe we can wrap deployed functions with a function that just imports pandas as pd, so if the user tries to use pd inside an apply function (for example) it's scope will have pandas instead of Modin, and it'll work?

anmyachev:
> maybe we can wrap deployed functions with a function that just imports pandas as pd, so if the user tries to use pd inside an apply function (for example) it's scope will have pandas instead of Modin, and it'll work?



> > @pyrito for now, this most likely will not be merged, @vnlitvinov is right.
> > By the way, I have an interesting case, which may not be very relevant to this task, but still.
> > This call works for pandas, but fails for modin because modin tries to initialize ray on the worker (if I remember correctly).
> > In this case, the following change helps: `pd.Series...` -> `pandas.Series`. However, this is more like a short-term workaround than a general solution. What can we do here?
> > ```python
> > import modin.pandas as pd
> > 
> > 
> > 
> > def agg_func(x):
> > 
> >     return int(pd.Series.count(x) > 0)
> > 
> > 
> > 
> > res = some_df.pivot_table(..., agg_func=agg_func)
> > ```
> 
> @anmyachev this fails since the function is running on a ray worker - maybe we can wrap deployed functions with a function that just imports pandas as pd, so if the user tries to use pd inside an apply function (for example) it's scope will have pandas instead of Modin, and it'll work?

The following example doesn't work.
```python
def agg_func(x):

    return int(pandas.Series.count(x) > 0)

def wrapper(x):
    import pandas as pd
    return agg_func(x)

res = some_df.pivot_table(..., agg_func=wrapper)
```
if I understand correctly, `pandas.Series` type is determined when `agg_func` is defined, and not when it is called, so it will not work to redefine the scope.

RehanSD:
@anmyachev We could modify the scope and replace modin.pandas with pandas by modifying the globals dict (although I'm not sure if we want to get into this level of meta-programming?)
image

anmyachev:
We need to first decide on an approach.

PR #4334, opened 2022-03-17
Release version 0.13.3 - pin pypi and docs

vnlitvinov:

This change is for _0.13.3 release_ branch, and its point is to make the link to documentation to always point to its documentation (not to the latest or stable or anything), so it keeps the same regardless of our further actions.

We also introduce the changes from PR #4114, which fixes issues #4113, #4116, #4115 to make CI green. The changes almost match the ones from the PR #4114 except pinning pip in `docs/requirements-doc.txt`. We unpin pip since we did that in master and the issue was temporary.


vnlitvinov:
Marking it as draft because I'm still waiting for CI to run for the release branch itself: https://github.com/modin-project/modin/runs/5589407262?check_suite_focus=true

YarShev (approved):

PR #4372, opened 2022-04-06
FEAT-#4371: Add logging to Modin

### LINKED ISSUE #4371: Low-Level Logging for Modin

Add logging for Modin.

naren-ponder:
## What do these changes do?

Add logging to Modin to trace API function calls.

- [x] Resolves #4371 
- [x] tests added and passing


modin/logging/config.py
```diff
@@ -0,0 +1,100 @@
...
+class MyFormatter(logging.Formatter):
+    converter = dt.datetime.fromtimestamp
+
+    def formatTime(self, record, datefmt=None):
+        ct = self.converter(record.created)
+        if datefmt:
+            s = ct.strftime(datefmt)
+        else:
+            t = ct.strftime("%Y-%m-%d %H:%M:%S")
+            s = "%s,%03d" % (t, record.msecs)
```
mvashishtha, line 24: It's hard to tell what these lines are doing. Could you add more descriptive names or comments?

modin/logging/config.py
```diff
@@ -0,0 +1,100 @@
...
+        return s
+
+
+def get_size(bytes, suffix="B"):
```
mvashishtha, line 28: A more descriptive name might be `bytes_int_to_str`

mvashishtha, line 28: Please add a docstring.

mvashishtha, line 28: `bytes` is a built-in type. Let's use another name.

modin/logging/config.py
```diff
@@ -0,0 +1,100 @@
...
+    """
+    Scale bytes to its proper format
+    e.g:
+        1253656 => '1.20MB'
+        1253656678 => '1.17GB'
+    """
+    factor = 1024
+    for unit in ["", "K", "M", "G", "T", "P"]:
```
mvashishtha, line 36: `get_size(1_000_000_000_000_000_000_000)` is `None` because it's past the last prefix.

mvashishtha, line 36: Can we please make these Ki, Mi, Gi, etc.?

modin/logging/config.py
```diff
@@ -0,0 +1,100 @@
...
+import psutil
+import pkg_resources
+import threading
+import time
+from modin.config import LogMode
+
+__LOGGER_CONFIGURED__: bool = False
+
+
+class MyFormatter(logging.Formatter):
```
mvashishtha, line 15: Maybe call this `ModinFormatter`?

mvashishtha (commented):
@naren-ponder thank you for your work on this! I have a lot of comments because there's a lot of new code.

Here's one batch of comments. I reviewed a few files completely and made it partway through `modin/logging/config.py`. I'll finish this round of review later.

modin/logging/logger_function.py
```diff
@@ -0,0 +1,28 @@
...
+            if LogMode.get() != "none":
+                logger = get_logger()
+                try:
+                    getattr(logger, log_level.lower())(
+                        f"START::{modin_layer.upper()}::{function_name}"
+                    )
+                except AttributeError:
+                    raise AttributeError(f"Invalid log_level: {log_level}")
+                result = f(*args, **kwargs)
+                getattr(logger, log_level.lower())(
```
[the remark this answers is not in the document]
naren-ponder: A bit confused on what this means here, do you mean something like:

```
logger = get_logger()
try:
    log_line = getattr(logger, log_level.lower())(
           f"START::{modin_layer.upper()}::{function_name}"
    )
except AttributeError:
    raise AttributeError(f"Invalid Log Statement: {log_line}")
result = f(*args, **kwargs)
getattr(logger, log_level.lower())(
       f"STOP::{modin_layer.upper()}::{function_name}"
)   
return result
```

modin/logging/logger_function.py
```diff
@@ -0,0 +1,28 @@
+from .config import get_logger
+from functools import wraps
+from modin.config import LogMode
+
+
+def logger_decorator(modin_layer: str, function_name: str, log_level: str):
+    def decorator(f):
+        @wraps(f)
+        def run_and_log(*args, **kwargs):
+            if LogMode.get() != "none":
```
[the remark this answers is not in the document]
naren-ponder: Even if the check is `== "none"`, I think we still need an `else` clause? But since `return f(*args, **kwargs)` is common code among both cases, I can just remove the `else` and the two lines of code in the current if case that does this. This does not avoid the negative check though.

modin/logging/config.py
naren-ponder: Added a case for Exabyte, but do you mean to show this example as just a case we should hardcode if above a certain value?

modin/logging/logger_function.py
```diff
@@ -0,0 +1,28 @@
...
+                    )
+                except AttributeError:
+                    raise AttributeError(f"Invalid log_level: {log_level}")
+                result = f(*args, **kwargs)
+                getattr(logger, log_level.lower())(
+                    f"STOP::{modin_layer.upper()}::{function_name}"
+                )
+                return result
+            else:
+                return f(*args, **kwargs)
```
mvashishtha, line 24: This is what I meant by my suggestion [here](https://github.com/modin-project/modin/pull/4372/files#r862258012).

```suggestion
            if LogMode.get() == "none":
                return f(*args, **kwargs)
            logger = get_logger()
            try:
                getattr(logger, log_level.lower())(
                   f"START::{modin_layer.upper()}::{function_name}"
                )
            except AttributeError:
                raise AttributeError(f"Invalid log_level: {log_level}")
            result = f(*args, **kwargs)
            getattr(logger, log_level.lower())(
                f"STOP::{modin_layer.upper()}::{function_name}"
            )
            return result
                
```

modin/logging/logger_function.py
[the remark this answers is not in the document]
mvashishtha: I added a suggested change in another comment to clarify what I meant.

modin/logging/logger_function.py
mvashishtha, line 19: This is what I was thinking of in my comment [here](https://github.com/modin-project/modin/pull/4372/files#r862258012).

```suggestion
                try:
                    logger_level = getattr(logger, log_level.lower())
                except AttributeError:
                    raise AttributeError(f"Invalid log_level: {log_level}")                    
                logger_level(f"START::{modin_layer.upper()}::{function_name}")
                result = f(*args, **kwargs)
                logger_level(
```

modin/logging/config.py
mvashishtha: Yes, I think we should hardcode beyond a certain value. I don't think we'll ever hit this bug in practice, but I prefer to fix it for tidiness.

modin/logging/logger_function.py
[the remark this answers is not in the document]
mvashishtha: I added a suggested change in another comment to clarify what I meant.

mvashishtha (commented):
I've gone through all the code changes now.

modin/core/execution/ray/implementations/pandas_on_ray/partitioning/partition.py
```diff
@@ -93,15 +107,20 @@ def apply(self, func, *args, **kwargs):
...
         if len(call_queue) > 1:
             result, length, width, ip = _apply_list_of_funcs.remote(call_queue, oid)
+            logger.debug(f"SUBMIT::_apply_list_of_funcs::{self._identity}")
         else:
             # We handle `len(call_queue) == 1` in a different way because
             # this dramatically improves performance.
             func, args, kwargs = call_queue[0]
             result, length, width, ip = _apply_func.remote(oid, func, *args, **kwargs)
+            logger.debug(f"SUBMIT::_apply_func::{self._identity}")
+            logger.debug(f"EXIT::Partition.apply::{self._identity}")
```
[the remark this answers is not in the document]
naren-ponder: Oops my mistake, I must've messed up the indentation when I did a rebase.

modin/pandas/io.py
```diff
@@ -703,6 +726,7 @@ def read_orc(
 
 
 @_inherit_docstrings(pandas.read_xml, apilink="pandas.read_xml")
+@logger_decorator("PANDAS-API", "read_orc", "info")
```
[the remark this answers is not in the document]
naren-ponder: Good catch!

modin/logging/config.py
```diff
@@ -0,0 +1,127 @@
...
+
+def get_logger():
+    if not __LOGGER_CONFIGURED__ and LogMode.get() != "none":
+        if LogMode.get() == "api_only":
+            configure_logging(logging.INFO)
+        else:
+            configure_logging(logging.DEBUG)
+
+        logger = logging.getLogger("modin.logger")
+        logger.info("OS Version: " + platform.platform())
```
jeffreykennethli, line 105: nit: use f-strings here to be consistent

modin/logging/config.py
```diff
@@ -0,0 +1,127 @@
...
+    logger.setLevel(level)
+
+    __LOGGER_CONFIGURED__ = True
+
+
+def memory_thread(logger):
+    while True:
+        svmem = psutil.virtual_memory()
+        logger.info(f"Memory Percentage: {svmem.percent}%")
+        time.sleep(5)
```
jeffreykennethli, line 94: nit: make the memory thread interval a parameter that defaults to 5 (seconds)

modin/logging/config.py
```diff
@@ -0,0 +1,127 @@
...
+        datefmt="%Y-%m-%d,%H:%M:%S.%f",
+    )
+    logfile.setFormatter(formatter)
+    logger.addHandler(logfile)
+    logger.setLevel(level)
+
+    __LOGGER_CONFIGURED__ = True
+
+
+def memory_thread(logger):
```
jeffreykennethli, line 90: question: given we may want to log the memory usage of distributed execution, how would it change the current design? would we have to use something like psutil processes?

jeffreykennethli (commented):
Looks good so far! I think there may be more design considerations to go over, so I will add another review.

modin/core/execution/ray/implementations/pandas_on_ray/partitioning/partition.py
```diff
@@ -56,6 +58,14 @@ def __init__(self, object_id, length=None, width=None, ip=None, call_queue=None)
         self._length_cache = length
         self._width_cache = width
         self._ip_cache = ip
+        self._identity = uuid.uuid4().hex[8:]
+
+        logger = get_logger()
+        logger.debug(
+            "Partition OID: {}, Height: {}, Width: {}".format(
```
jeffreykennethli, line 65: could you also add the object `_identity` here? It would allow us to reference the partition `_identity` logged in `get()`, `apply()`, etc.

jeffreykennethli: it may also be helpful to log the `_ip_cache`. I think @RehanSD had mentioned it might be useful, could you comment on this?

modin/core/execution/ray/implementations/pandas_on_ray/partitioning/partition.py
```diff
@@ -404,6 +438,8 @@ def deserialize(obj):
         else:
             return obj
 
+    logger = get_logger()
+    logger.debug(f"BEGIN::apply_list_of_funcs::{oid_hash}")
```
jeffreykennethli, line 442: I have tried running some sample scripts (with log level 'all' to get this line to log, but to no avail, despite being able to get the line `logger.debug(f"SUBMIT::_apply_list_of_funcs::{self._identity}")` on line partition.py::116 to log.

Do you have an example script that can log this line? This line would be nice to have because it could help us see start/end of remote computation.

modin/core/execution/ray/implementations/pandas_on_ray/partitioning/partition.py
```diff
@@ -133,27 +152,32 @@ def add_to_apply_calls(self, func, *args, **kwargs):
...
     def drain_call_queue(self):
         """Execute all operations stored in the call queue on the object wrapped by this partition."""
+        logger = get_logger()
+        logger.debug(f"ENTER::Partition.drain_call_queue::{self._identity}")
         if len(self.call_queue) == 0:
             return
         oid = self.oid
         call_queue = self.call_queue
         if len(call_queue) > 1:
+            logger.debug(f"SUBMIT::_apply_list_of_funcs::{self._identity}")
```
jeffreykennethli, line 162: nit: on line 116 of `apply()`, the line `logger.debug(f"SUBMIT::_apply_list_of_funcs::{self._identity}")` is called after `_apply_list_of_funcs`. Is there a reason for this, or should the order be changed to be consistent?

jeffreykennethli:
nit: with this PR checked out and pip installed, I am unable to run `jupyter lab` with an error: `AttributeError: partially initialized module 'logging' has no attribute 'getLogger' (most likely due to a circular import)`

I'm 90% sure it's because JupyterLab initializes with its own logger, which is conflicting with the logging module introduced in this PR. (something like https://stackoverflow.com/a/60677039/19027728). I then locally renamed the logging dir from `logging` -> `modin-logging` which fixed this. Can you consider renaming the `logging` dir to avoid this module name conflict?

modin/core/execution/ray/implementations/pandas_on_ray/partitioning/partition.py
jeffreykennethli: I dig some digging and found that the line `logger = get_logger()` returns a logger with level set to WARNING, even when LogMode is set to all.

This is because in [logging/config](https://github.com/modin-project/modin/pull/4372/files#diff-91f7f65eaa299999008d45ca833ad96f6ba57131b0cec49d3df441500617633aR98) logging is configured but LogMode is not set. I'm not sure why this is, since in my debugging script I set `LogMode.put("all")` and in the same script, `Partition.apply` is able to `get_logger` with the right level. I wonder if the envvars is not being picked up by Ray remote functions for some reason.

naren-ponder: I think the reason for this is that LogMode does not support the `all` flag. As it's currently written it only supports `enable`, `disable`, and `enable_api_only`. If you do `LogMode.put(enable)` you won't have this issue.

naren-ponder: But you bring up a good point that perhaps I should throw an error of some kind if any other flag is passed

modin/core/execution/ray/implementations/pandas_on_ray/partitioning/partition.py
naren-ponder: Yes, will log these as well. Object `_identity` will actually be better than the partition's `oid` so will change that.

modin/logging/config.py
naren-ponder: I don't think `psutil` would help in this case, but we have access to the Ray Object IDs, so we could asynchronously hit the Ray API and access some of the object's attributes. The challenge here is the Ray API may take some time to return so we should set a timeout or have this running in another separate thread that ModinVision parses out later from the logs.

modin/core/execution/ray/implementations/pandas_on_ray/partitioning/partition.py
naren-ponder: I think that you are correct w.r.t. to the ray remotes, will dig into this further.

modin/config/envvars.py
```diff
@@ -358,6 +358,40 @@ def put(cls, value):
...
+        """Disable logging feature."""
+        cls.put("none")
+
+    @classmethod
+    def enable_api_only(cls):
+        """Enable API level logging."""
+        cls.put("api_only")
+
+    @classmethod
+    def put(cls, value):
```
anmyachev, line 383: Why override this method?

modin/pandas/io.py
```diff
@@ -611,6 +631,7 @@ def read_spss(
 
 
 # Adding docstring since pandas docs don't have web section for this function.
+@logger_decorator("PANDAS-API", "read_to_pickle", "info")
```
anmyachev, line 634:
```suggestion
@logger_decorator("PANDAS-API", "to_pickle", "info")
```

modin/pandas/io.py
```diff
@@ -668,6 +689,7 @@ def to_pickle(
 
 
 @_inherit_docstrings(pandas.json_normalize, apilink="pandas.json_normalize")
+@logger_decorator("PANDAS-API", "read_json_normalize", "info")
```
anmyachev, line 692:
```suggestion
@logger_decorator("PANDAS-API", "json_normalize", "info")
```

modin/modin_logging/config.py
```diff
@@ -0,0 +1,129 @@
...
+def memory_thread(logger, sleep_time):
+    while True:
+        svmem = psutil.virtual_memory()
+        logger.info(f"Memory Percentage: {svmem.percent}%")
+        time.sleep(sleep_time)
+
+
+def get_logger(mem_sleep=5):
+    if not __LOGGER_CONFIGURED__ and LogMode.get() != "none":
+        if LogMode.get() == "api_only":
```
anmyachev, line 99:
```suggestion
        if LogMode.get() == "enable_api_only":
```

modin/modin_logging/config.py
```diff
@@ -0,0 +1,129 @@
...
+        logger.info(f"Modin Version: {modin_version}")
+        logger.info(f"Pandas Version: {pandas_version}")
+        logger.info(f"Physical Cores: {num_physical_cores}")
+        logger.info(f"Total Cores: {num_total_cores}")
+        logger.info(f"Memory Total: {bytes_int_to_str(svmem.total)}")
+        logger.info(f"Memory Available: {bytes_int_to_str(svmem.available)}")
+        logger.info(f"Memory Used: {bytes_int_to_str(svmem.used)}")
+        logger.info(f"Memory Percentage: {svmem.percent}%")
+
+        if LogMode.get() != "api_only":
```
anmyachev, line 121:
```suggestion
        if LogMode.get() != "enable_api_only":
```

modin/core/execution/ray/implementations/pandas_on_ray/partitioning/partition.py
```diff
@@ -56,6 +58,14 @@ def __init__(self, object_id, length=None, width=None, ip=None, call_queue=None)
         self._length_cache = length
         self._width_cache = width
         self._ip_cache = ip
+        self._identity = uuid.uuid4().hex[8:]
```
anmyachev, line 61: Why are the changes made only for Ray? What about Dask?

modin/core/io/text/text_file_dispatcher.py
```diff
@@ -48,6 +49,7 @@ class TextFileDispatcher(FileDispatcher):
     read_callback = None
 
     @classmethod
+    @logger_decorator("PANDAS-API", "TextFileDispatcher.get_path_or_buffer", "INFO")
```
anmyachev, line 52: Is it possible to use a class decorator instead of all these function-level decorators?

modin/core/execution/ray/implementations/pandas_on_ray/partitioning/partition.py
jeffreykennethli: @naren-ponder I think these changes could be left off this PR as well, as with the Ray remote logs.

modin/core/io/text/text_file_dispatcher.py
anmyachev: I agree, it just seems to me that all methods of `TextFileDispatcher` class are wrapped in decorators. Why not replace it with a class-level decorator in this case?

modin/modin_logging/config.py
```diff
@@ -0,0 +1,142 @@
...
+        datefmt="%Y-%m-%d,%H:%M:%S.%f",
+    )
+    logfile.setFormatter(formatter)
+    logger.addHandler(logfile)
+    logger.setLevel(level)
+
+    __LOGGER_CONFIGURED__ = True
+
+
+def memory_thread(logger, sleep_time):
```
mvashishtha, line 103: Can we please make sure that this doesn't leave a hanging thread so that we have to  cntl-c twice to exit ipython? @RehanSD mentioned offline that he had that problem.

docs/usage_guide/advanced_usage/index.rst
```diff
@@ -84,6 +85,14 @@ Modin provides an implementation of `distributed XGBoost`_ machine learning algo
 :doc:`Modin XGBoost architecture documentation </flow/modin/experimental/xgboost>` for information about implementation and
 internal execution flow.
 
+Logging and Tracing with Modin
+------------------------------
+
+Modin logging and tracing allows users to gain observability into their queries by logging internal Modin API calls, partition metadata,
```
RehanSD, line 91: nit:
```suggestion
Modin logging allows users to gain observability into their queries by logging internal Modin API calls, partition metadata, 
```
this feels more concise and is a better name imo

docs/usage_guide/advanced_usage/index.rst
```diff
@@ -84,6 +85,14 @@ Modin provides an implementation of `distributed XGBoost`_ machine learning algo
 :doc:`Modin XGBoost architecture documentation </flow/modin/experimental/xgboost>` for information about implementation and
 internal execution flow.
 
+Logging and Tracing with Modin
+------------------------------
+
+Modin logging and tracing allows users to gain observability into their queries by logging internal Modin API calls, partition metadata, 
+and profiling memory. When Modin logging is enabled (default disabled), log files are written to a local `.modin` directory at the same
+directory level as the notebook/script used to run Modin. See our :doc:`Logging and Tracing with Modin documentation <modin_logging>`
```
RehanSD, line 93: Same here - is there a reason we specify both Logging and Tracing?

docs/usage_guide/advanced_usage/modin_logging.rst
```diff
@@ -0,0 +1,41 @@
+Logging and Tracing with Modin
+==============================
+
+Modin logging and tracing allows users to gain observability into their queries by logging internal Modin API calls, partition metadata,
```
RehanSD, line 4: Isn't this the exact same paragraph as above? Maybe change the above one to:
Modin also includes the ability to log internal API calls and metadata, to offer users greater insight into their queries. Please see our docs ... for more.

docs/usage_guide/advanced_usage/modin_logging.rst
```diff
@@ -0,0 +1,41 @@
...
+
+.. code-block:: python
+
+  import modin.pandas as pd
+  from modin.config import LogMode
+  LogMode.enable_api_only()
+
+  # Your code goes here
+
+In this example, we enable logging for internal Modin API calls, partition metadata, and memory profiling.
```
RehanSD, line 21: This gets a little repetitive since its essentially the same thing.
```suggestion
In the next example, we add logging not only for internal API calls, but also for partition metadata and memory profiling.```

File: docs/usage_guide/advanced_usage/modin_logging.rst [hunk 28]
```diff
@@ -0,0 +1,41 @@
...
+We can set the granularity (in seconds) at which the system memory utilization is logged using `LogMemoryInterval`. 
+
+.. code-block:: python
+
+  import modin.pandas as pd
+  from modin.config import LogMode, LogMemoryInterval
+  LogMode.enable()
+  LogMemoryInterval.put(2) # Defaults to 5 seconds
```
RehanSD, line 29:
```suggestion
  LogMemoryInterval.put(2) # Default is 5 seconds - New Interval is 2 seconds
```
just to make it a little more clear what's happening here.

docs/usage_guide/advanced_usage/modin_logging.rst
```diff
@@ -0,0 +1,41 @@
...
+
+  # Your code goes here
+
+Modin logging is default disabled, but if currently enabled you can manually disable the logs as follows:
```
RehanSD, line 33:
```suggestion
Disable Modin logging like so:
```

modin/modin_logging/config.py
```diff
@@ -0,0 +1,143 @@
...
+import pkg_resources
+import threading
+import time
+from modin.config import LogMemoryInterval, LogMode
+
+__LOGGER_CONFIGURED__: bool = False
+
+
+class ModinFormatter(logging.Formatter):
+    def formatTime(self, record, datefmt=None):
```
RehanSD, line 29: From the docs for the Formatter class:
> formatTime(record, datefmt=None)
> This method should be called from [format()](https://docs.python.org/3/library/functions.html#format) by a formatter which wants to make use of a formatted time. This method can be overridden in formatters to provide for any specific requirement, but the basic behavior is as follows: if datefmt (a string) is specified, it is used with [time.strftime()](https://docs.python.org/3/library/time.html#time.strftime) to format the creation time of the record. Otherwise, the format ‘%Y-%m-%d %H:%M:%S,uuu’ is used, where the uuu part is a millisecond value and the other letters are as per the time.strftime() documentation. An example time in this format is 2003-01-23 00:29:50,411. The resulting string is returned.

Why are we redefining this method if it does the same thing as its parent method?

modin/modin_logging/config.py
```diff
@@ -0,0 +1,143 @@
...
+    suffix: str, default: "B"
+        Suffix to add to conversion of num_bytes.
+
+    Returns
+    -------
+    str
+        Human-readable string format.
+    """
+    factor = 1024
+    if num_bytes > 1000000000000000000:
```
RehanSD, line 74: Why do we do this? Is this like a sanity checking thing where if the number of bytes is too high we make this complaint? Should we clarify the ValueError? Its not super clear what this is catching.

modin/modin_logging/config.py
```diff
@@ -0,0 +1,143 @@
...
+        raise ValueError("System memory exceeds expectations")
+    for unit in ["", "Ki", "Mi", "Gi", "Ti", "Pi"]:
```
RehanSD, line 76: Why are we using the powers of 2 one instead of the powers of 10 ones? (i.e. Gigabyte vs. Gibibyte)?

modin/modin_logging/config.py
```diff
@@ -0,0 +1,143 @@
...
+        num_bytes /= factor
+
+
+def configure_logging(level):
+    global __LOGGER_CONFIGURED__
+    logger = logging.getLogger("modin.logger")
+    job_id = uuid.uuid4().hex
+    log_filename = f".modin/logs/job_{job_id}.log"
+
+    if not os.path.isdir(".modin/logs"):
```
RehanSD, line 88: Do we really need to special case this? Can't we just run this with `exist_ok=True`?

modin/modin_logging/config.py
```diff
@@ -0,0 +1,143 @@
...
+        os.makedirs(os.path.dirname(log_filename), exist_ok=False)
+
+    logfile = logging.FileHandler(log_filename, "a")
+    formatter = ModinFormatter(
+        fmt="%(process)d, %(thread)d, %(asctime)s, %(message)s",
+        datefmt="%Y-%m-%d,%H:%M:%S.%f",
```
RehanSD, line 94: Is this why we needed to create our own subclass? If we specify datefmt for the default formatter will it lop off the milliseconds?

modin/modin_logging/config.py
```diff
@@ -0,0 +1,143 @@
...
+    logfile.setFormatter(formatter)
+    logger.addHandler(logfile)
+    logger.setLevel(level)
+
+    __LOGGER_CONFIGURED__ = True
+
+
+def memory_thread(logger, sleep_time):
+    while True:
+        svmem = psutil.virtual_memory()
```
RehanSD, line 105: For my own understanding - if someone has multiple things running on this device, psutil will return memory used by all processes right? Not just the Modin processes? Also, doesn't this only return the memory used on the head node? If the user is running on a cluster this won't really be accurate right? And if they're running on a cluster that doesn't schedule on the head node isn't this going to be super wrong?

modin/modin_logging/logger_function.py
```diff
@@ -0,0 +1,39 @@
...
+# the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific language
+# governing permissions and limitations under the License.
+
+from .config import get_logger
+from functools import wraps
+from modin.config import LogMode
+
+
+def logger_decorator(modin_layer: str, function_name: str, log_level: str):
```
RehanSD, line 19: Could we add a docstring?

modin/modin_logging/config.py
```diff
@@ -0,0 +1,143 @@
...
+    factor = 1024
+    if num_bytes > 1000000000000000000:
+        raise ValueError("System memory exceeds expectations")
+    for unit in ["", "Ki", "Mi", "Gi", "Ti", "Pi"]:
+        if num_bytes < factor:
+            return f"{num_bytes:.2f}{unit}{suffix}"
+        num_bytes /= factor
+
+
+def configure_logging(level):
```
RehanSD, line 82: could we add a docstring?

modin/modin_logging/config.py
```diff
@@ -0,0 +1,143 @@
...
+        datefmt="%Y-%m-%d,%H:%M:%S.%f",
+    )
+    logfile.setFormatter(formatter)
+    logger.addHandler(logfile)
+    logger.setLevel(level)
+
+    __LOGGER_CONFIGURED__ = True
+
+
+def memory_thread(logger, sleep_time):
```
RehanSD, line 103: Could we add a docstring here?

modin/modin_logging/config.py
```diff
@@ -0,0 +1,143 @@
...
+    while True:
+        svmem = psutil.virtual_memory()
+        logger.info(f"Memory Percentage: {svmem.percent}%")
+        time.sleep(sleep_time)
+
+
+def get_logger():
```
RehanSD, line 110: Could we add a docstring here?

modin/modin_logging/logger_metaclass.py
```diff
@@ -0,0 +1,45 @@
...
+# ANY KIND, either express or implied. See the License for the specific language
+# governing permissions and limitations under the License.
+
+from functools import wraps
+from .config import get_logger
+from types import FunctionType, MethodType
+from modin.config import LogMode
+
+
+def logger_class_wrapper(classname, name, method):
```
RehanSD, line 20: Could we add a docstring?

modin/modin_logging/logger_metaclass.py
```diff
@@ -0,0 +1,45 @@
...
+        else:
+            return method(*args, **kwargs)
+
+    return log_wrap
+
+
+class LoggerMetaClass(type):
+    def __new__(mcs, classname, bases, class_dict):
+        new_class_dict = {}
+        for attribute_name, attribute in class_dict.items():
```
RehanSD, line 38: Don't you need to add `__getattribute__` to new_class_dict as well? Maybe this should be
```suggestion
        new_class_dict = {}
        new_class_dict.update(class_dict)
        for attribute_name, attribute in new_class_dict.items():
```

modin/modin_logging/logger_metaclass.py
```diff
@@ -0,0 +1,45 @@
...
+            if (
+                isinstance(attribute, (FunctionType, MethodType))
+                and attribute_name != "__getattribute__"
+            ):
+                attribute = logger_class_wrapper(classname, attribute_name, attribute)
+            new_class_dict[attribute_name] = attribute
+        return type.__new__(mcs, classname, bases, new_class_dict)
```
RehanSD, line 45:
```suggestion
        return super().__new__(mcs, classname, bases, new_class_dict)
```

modin/core/execution/ray/implementations/pandas_on_ray/partitioning/partition.py
RehanSD, line 61: Why are we skipping the first 8?

modin/core/execution/ray/implementations/pandas_on_ray/partitioning/partition.py
```diff
@@ -56,6 +58,14 @@ def __init__(self, object_id, length=None, width=None, ip=None, call_queue=None)
...
+
+        logger = get_logger()
+        logger.debug(
+            "Partition ID: {}, Height: {}, Width: {}, Node IP: {}".format(
```
RehanSD, line 65: Should we do a check if these are None, and print something more useful than `Width: None` e.g. if they are (maybe like width has not been initialized or something)?

modin/core/execution/ray/implementations/pandas_on_ray/partitioning/partition.py
```diff
@@ -66,9 +76,13 @@ def get(self):
         pandas.DataFrame
             The object from the Plasma store.
         """
+        logger = get_logger()
```
RehanSD, line 79: Shouldn't we be using the logger_function + logger decorator for these things?

modin/core/execution/ray/implementations/pandas_on_ray/partitioning/partition.py
```diff
@@ -415,7 +442,6 @@ def deserialize(obj):
         # we absolutely have to.
         except ValueError:
             partition = func(partition.copy(), *args, **kwargs)
-
```
RehanSD, line 418: Is there a reason for this deletion?

modin/core/storage_formats/pandas/parsers.py
```diff
@@ -136,7 +137,7 @@ def find_common_type_cat(types):
         return find_common_type(list(types))
 
 
-class PandasParser(object):
+class PandasParser(object, metaclass=LoggerMetaClass):
```
RehanSD, line 140: Given that the parser is more internal, should it really be logging with the loggermetaclass?

modin/pandas/indexing.py
```diff
@@ -267,7 +268,7 @@ def _compute_ndim(row_loc, col_loc):
     return ndim
 
 
-class _LocationIndexerBase(object):
+class _LocationIndexerBase(object, metaclass=LoggerMetaClass):
```
RehanSD, line 271: Again, this is more internal - should it be logging with the `PANDAS_API` tag?

RehanSD (changes requested):
Great work on this @naren-ponder and @jeffreykennethli ! I left some comments + questions, and had a few overarching questions:
* The LoggerMetaClass says its logging at the `PANDAS_API` level, but the `PandasDataFrame` along with the `QueryCompiler` inherit from it. The QueryCompiler is still fairly high level, so it makes sense for it to inherit from it, but the PandasDataFrame, etc. are very much Modin internals, and maybe shouldn't include `PANDAS_API` in their logs, since they don't 1:1 match up to the Pandas API?
* I think we should have some of the logging code moved down into `core` - so something like `modin/core/logging/*`, although maybe the logger meta class should stay fairly high level since its imported by `modin/pandas/*`? Maybe it should be in the experimental folder?

docs/usage_guide/advanced_usage/modin_logging.rst
naren-ponder: I think it's fine for the descriptions to be the same here. The descriptions are the same for XGBoost and Spreadsheet API as well. It's a fairly short description, and I think that each line in the summary is important (what logging logs, default disabled, and that it is a local file).

modin/core/execution/ray/implementations/pandas_on_ray/partitioning/partition.py
naren-ponder: For now, the logs are a fairly advanced feature and I think printing `None` is not totally necessary for the partition metadata. We will have more detailed Ray remote logging for the partitions in the future anyhow that @jeffreykennethli is working on.

modin/core/execution/ray/implementations/pandas_on_ray/partitioning/partition.py
naren-ponder: The function and decorator really only logs the timestamp for when the top-level `get` function starts and ends, but we want to log at a finer level of granularity here inside of the function.

modin/modin_logging/config.py
naren-ponder: The parent method is returning a millisecond value, this is being overridden to record microsecond timestamps.

modin/core/storage_formats/pandas/parsers.py
naren-ponder: I guess I've been logging every API call with the `LoggerMetaClass`. I don't think there is a need to distinguish between the pandas queries and the Modin internals. If folks want to just look at the pandas queries they are using, they have the script/notebook they are running them in and don't need Modin logging. Would you suggest that I rename `START::PANDAS-API::{classname}.{name}"` to `START::MODIN-API::{classname}.{name}"` in the LoggerMetaClass?

modin/modin_logging/config.py
naren-ponder: Randomly picked, changed it.

modin/modin_logging/config.py
naren-ponder: Yeah this is sanity checking that they have for than a Petabyte of system memory (we are likely never going to hit this, but it is good to have these sanity checks for completeness). What would you suggest changing the ValueError to?

modin/modin_logging/config.py
naren-ponder: Addressed in a previous comment, we want to enable microsecond level timestamps which the default formatter for the python logging module does not allow.

modin/modin_logging/config.py
naren-ponder: This is more intended as a general profiling tool (not one specific to Modin, we essentially will use this to identify all memory spikes). Correct, this is going to be useful in most user cases but could be off for some (scheduling on the head node is fairly standard, this will total system memory utilization). To get more specific memory utilization data about the Modin processes, we would have to log inside of the Ray workers which will be done in a future PR by @jeffreykennethli.

modin/pandas/indexing.py
naren-ponder: Addressed in previous comment.

modin/core/io/text/text_file_dispatcher.py
devin-petersohn: @anmyachev it was not possible in this case due to the way that `TextFileDispatcher` is architected (it is a mix-in rather than a directly created/called class). We tried multiple ways, and this was the only way we could get it to work.

modin/core/storage_formats/pandas/parsers.py
devin-petersohn: I think it's fine the way it is.

modin/modin_logging/config.py
```diff
@@ -0,0 +1,173 @@
...
+    """
+
+    factor = 1000
+    if num_bytes > 1000000000000000000:
+        raise ValueError("System memory exceeds expectations")
+    for unit in ["", "K", "M", "G", "T", "P"]:
+        if num_bytes < factor:
+            return f"{num_bytes:.2f}{unit}{suffix}"
+        num_bytes /= factor
+
```
devin-petersohn, line 82:
```suggestion
    return f"{num_bytes:.2f}{unit}{suffix}"

```

modin/modin_logging/config.py
devin-petersohn: I think we can remove this and add the suggestion below. 1000+PB is fine to log.

modin/modin_logging/config.py
devin-petersohn: 👍 @naren-ponder was this resolved?

modin/modin_logging/config.py
```diff
@@ -0,0 +1,173 @@
...
+    Configures Modin logger based on Modin config and returns the logger.
+
+    Returns
+    -------
+    logging.Logger
+        The Modin logger.
+    """
+
+    if not __LOGGER_CONFIGURED__ and LogMode.get() != "disable":
+        if LogMode.get() == "enable_api_only":
```
devin-petersohn, line 142: I prefer to factor this out into a separate function and call it separately since it should only be executed at the start.

modin/modin_logging/logger_metaclass.py
devin-petersohn: @RehanSD no we do not need it, it will be inherited.

modin/modin_logging/logger_metaclass.py
devin-petersohn: @RehanSD I prefer to keep it as it is.

devin-petersohn (dismissed):
LGTM once the tests pass 😄

naren-ponder:
@anmyachev Could you re-run the TeamCity tests? Seems like the failures are around a connection endpoint timing out and not related to this PR?

naren-ponder:
I see that `ci / test cloud` is running into some metaclass type errors related to `class Wrapper(override, origin_cls, metaclass=ProxyMeta):` which I haven't touched in this PR but will investigate that now. Otherwise, seems that all tests are passing.

modin/modin_logging/logger_metaclass.py
```diff
@@ -0,0 +1,93 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software distributed under
+# the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific language
+# governing permissions and limitations under the License.
+
+from functools import wraps
+from .config import get_logger
+from types import FunctionType, MethodType
+from modin.config import LogMode
```
YarShev, line 17:
```suggestion
from functools import wraps
from types import FunctionType, MethodType

from .config import get_logger
from modin.config import LogMode
```
Third party imports go first and Modin ones go next.

modin/modin_logging/logger_metaclass.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+def logger_class_wrapper(classname, name, method):
+    """
+    Helper method for Modin logging used by the LoggerMetaClass
+    to log internal Modin functions.
+
+    Parameters
+    ----------
+    classname: str
```
YarShev, line 27:
```suggestion
    classname : str
```

YarShev: wherever applicable

modin/modin_logging/__init__.py
```diff
@@ -0,0 +1,20 @@
...
+# ANY KIND, either express or implied. See the License for the specific language
+# governing permissions and limitations under the License.
+
+from .logger_metaclass import (  # noqa: F401
+    LoggerMetaClass,
+    logger_class_wrapper,
+    metaclass_resolver,
+)
+from .config import get_logger  # noqa: F401
+from .logger_function import logger_decorator  # noqa: F401
```
YarShev, line 20: What about __all__ = [...]?

modin/core/storage_formats/base/query_compiler.py
```diff
@@ -32,6 +32,7 @@
 )
 from modin.error_message import ErrorMessage
 import modin.core.storage_formats.base.doc_utils as doc_utils
+from modin.modin_logging import LoggerMetaClass
```
YarShev, line 35:
```suggestion
from modin.logging import LoggerMetaClass
```
Do you think this would be better?

docs/release_notes/release_notes-0.15.0.rst
```diff
@@ -21,7 +21,7 @@ Key Features and Updates
 * Performance enhancements
   * FEAT-#4320: Add connectorx as an alternative engine for read_sql (#4346)
 * Benchmarking enhancements
-  *
+  * FEAT-#4371: Add logging to Modin (#4372)
```
YarShev, line 24: We should add all the new code to linting jobs in GH actions.

modin/modin_logging/config.py
YarShev: ping

modin/core/storage_formats/base/query_compiler.py
naren-ponder: This was the initial name, but when running jupyter notebooks the name conflicts with JupyterLab and breaks which was the reason for renaming it.

docs/usage_guide/advanced_usage/index.rst
```diff
@@ -84,6 +85,14 @@ Modin provides an implementation of `distributed XGBoost`_ machine learning algo
 :doc:`Modin XGBoost architecture documentation </flow/modin/experimental/xgboost>` for information about implementation and
 internal execution flow.
 
+Logging with Modin
+------------------
+
+Modin logging offers users greater insight into their queries by logging internal Modin API calls, partition metadata, 
+and profiling system memory. When Modin logging is enabled (default disabled), log files are written to a local `.modin` directory at the same
+directory level as the notebook/script used to run Modin. See our :doc:`Logging with Modin documentation <modin_logging>` 
+for usage information.
```
mvashishtha, line 94:
```suggestion
Modin logging offers users greater insight into their queries by logging internal Modin API calls, partition metadata, 
and system memory. Logging is disabled by default, but when it is enabled, log files are written to a local `.modin` directory at the same directory level as the notebook/script used to run Modin. See our :doc:`Logging with Modin documentation <modin_logging>` for usage information.
```

modin/pandas/dataframe.py
```diff
@@ -807,8 +808,8 @@ def _update_var_dicts_in_kwargs(self, expr, kwargs):
             return
         frame = sys._getframe()
         try:
-            f_locals = frame.f_back.f_back.f_locals
-            f_globals = frame.f_back.f_back.f_globals
+            f_locals = frame.f_back.f_back.f_back.f_locals
```
mvashishtha, line 811: Can we add `query` and `eval` back now that we have this fix? I think I was able to get them to work with two `f_back`s.

mvashishtha (commented):
I have two more minor comments.

modin/core/storage_formats/base/query_compiler.py
naren-ponder: @jeffreykennethli Could you describe the error you ended up getting here?

modin/core/execution/ray/implementations/pandas_on_ray/partitioning/partition.py
```diff
@@ -207,6 +233,7 @@ def mask(self, row_labels, col_labels):
             new_obj._width_cache = compute_sliced_len.remote(
                 col_labels, self._width_cache
             )
+        logger.debug(f"EXIT::Partition.mask::{self._identity}")
```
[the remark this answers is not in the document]
naren-ponder: Some of these functions execute with the ray remote decorator, so the logger is not compatible with ray workers. There will be a follow-up PR that @jeffreykennethli is working on that will make logging fully compatible with ray and dask.

modin/modin_logging/__init__.py
naren-ponder: What do you mean by this? Not familiar with this syntax so can you elaborate?

modin/core/storage_formats/base/query_compiler.py
jeffreykennethli: Here was the original comment: https://github.com/modin-project/modin/pull/4372#issuecomment-1117825144

I can confirm the above issue with JupyterLab still exists (I tested by renaming the module locally on the latest commit). Here is the debug output, where it appears JupterLab tries to import modin logging and running `getLogger` on a file import instead of the python logging module.

https://gist.github.com/jeffreykennethli/371e70b2d938143ec591e8caf3c4ae84

devin-petersohn: @jeffreykennethli is this because you were working in the same directory as the `modin.logging` directory? Does this same issue occur elsewhere?

Calling this `modin_logging` is really clunky and I prefer `logging`

jeffreykennethli: I hadn't thought of that, but yes, I can confirm this only happens when working from the repo's base dir and `modin`, but not any other dir, not even `modin.logging`.

I do agree that `modin_logging` is clunky. Perhaps we can use `logging` and have this edge case, or `logger`?

naren-ponder: @devin-petersohn Thoughts on this? I think `logger` is fine.

modin/modin_logging/__init__.py
YarShev: An example of this https://github.com/modin-project/modin/blob/60518ca737bf13cc37c8f1269985dcbb68dad4c7/modin/pandas/__init__.py#L235

modin/modin_logging/logger_metaclass.py
```diff
@@ -0,0 +1,133 @@
...
+
+def logger_class_wrapper(classname, name, method):
+    """
+    Execute Modin functions with logging if enabled.
+
+    Parameters
+    ----------
+    classname : str
+        The name of the class the LoggerMetaClass is being applied to.
+    name : str
```
YarShev, line 35:
```suggestion
    method_name : str
```

modin/modin_logging/logger_metaclass.py
```diff
@@ -0,0 +1,133 @@
...
+from modin.config import LogMode
+
+
+def logger_class_wrapper(classname, name, method):
+    """
+    Execute Modin functions with logging if enabled.
+
+    Parameters
+    ----------
+    classname : str
```
YarShev, line 33:
```suggestion
    class_name : str
```

modin/modin_logging/logger_metaclass.py
```diff
@@ -0,0 +1,133 @@
...
+        The name of the class the LoggerMetaClass is being applied to.
+    name : str
+        The name of the Modin function within the class.
+    method : str
```
YarShev, line 37:
```suggestion
    method : callable
```

modin/modin_logging/logger_metaclass.py
```diff
@@ -0,0 +1,133 @@
...
+class LoggerMetaClass(type):  # noqa: PR01
+    """Log Metaclass to attach to class definitions."""
+
+    def __new__(mcs, classname, bases, class_dict):
+        """
+        Complete class instance creation with metaclass.
+
+        Parameters
+        ----------
+        mcs : class
```
YarShev, line 105:
```suggestion
        mcls : class
```

modin/modin_logging/logger_function.py
```diff
@@ -0,0 +1,79 @@
...
+
+"""
+Module contains ``logger_decorator`` function.
+
+``logger_decorator`` is used for decorating individual Modin functions.
+"""
+
+from .config import get_logger
+from functools import wraps
+from modin.config import LogMode
```
YarShev, line 22:
```suggestion
from functools import wraps

from .config import get_logger
from modin.config import LogMode
```

modin/core/execution/ray/implementations/pandas_on_ray/partitioning/partition.py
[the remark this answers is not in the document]
YarShev: Maybe we should leave out the changes for engine-related functionality and introduce them as part of that follow-up PR?

modin/modin_logging/__init__.py
naren-ponder: This will change how we imports those particular functions/modules though right? As in, `from modin.modin_logging import metaclass_resolver` will no longer work?

modin/modin_logging/logger_metaclass.py
naren-ponder: Are you suggesting to change the function signature as well and not just the doc string?

modin/modin_logging/__init__.py
YarShev: Everything should work correctly with this. `from modin.modin_logging import metaclass_resolver` will work. You likely want to read something on the use of `__all__` (e.g. https://docs.python.org/3/tutorial/modules.html#importing-from-a-package).

modin/core/execution/ray/implementations/pandas_on_ray/partitioning/partition.py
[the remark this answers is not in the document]
YarShev: I do not have strong preferences to leave out this but I would prefer to add this as part of that follow-up PR.

modin/modin_logging/logger_metaclass.py
YarShev: Yep.

.github/workflows/ci.yml
```diff
@@ -116,6 +116,7 @@ jobs:
             modin/experimental/core/execution/native/implementations/omnisci_on_native/omnisci_worker.py \
       - run: python scripts/doc_checker.py modin/experimental/core/storage_formats/omnisci
       - run: python scripts/doc_checker.py modin/experimental/core/execution/native/implementations/omnisci_on_native/exchange/dataframe_protocol
+      - run: python scripts/doc_checker.py modin/modin_logging
```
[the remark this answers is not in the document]
naren-ponder: I think this code is already tested on black and flake8. At least, there have been errors via black and flake8 caught in these modules that I have fixed.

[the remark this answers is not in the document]
YarShev: Ah, I see.

modin/core/execution/ray/implementations/pandas_on_ray/partitioning/partition.py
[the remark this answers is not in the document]
devin-petersohn: In the interest of merging, I think we can leave this in.

devin-petersohn (approved):
Thanks @naren-ponder LGTM

Issue #4059, opened 2022-01-26
`add` fails with empty DataFrame argument


c3-cjazra:
### System information
Modin: 0.12.0
Ray 1.7.1
python 3.9.7

### Describe the problem
```
import numpy as np
df_empty = modin_pd.DataFrame([])
df1 = modin_pd.DataFrame([1, 2, 3], columns=["col"])
# df_empty.add(df1) <-- OK
df1.add(df_empty) <-- FAILS
```

error:
```
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
/var/folders/9l/q_y0w5yx2_1_h65cprqgczyc0000gq/T/ipykernel_10949/3077187932.py in <module>
----> 6 df1.add(df)

/python3.9/site-packages/modin/pandas/dataframe.py in add(self, other, axis, level, fill_value)
    548         Get addition of ``DataFrame`` and `other`, element-wise (binary operator `add`).
    549         """
--> 550         return self._binary_op(
    551             "add",
    552             other,

/python3.9/site-packages/modin/pandas/base.py in _binary_op(self, op, other, **kwargs)
    425         if op in exclude_list:
    426             kwargs.pop("axis")
--> 427         new_query_compiler = getattr(self._query_compiler, op)(other, **kwargs)
    428         return self._create_or_update_from_compiler(new_query_compiler)
    429 

/python3.9/site-packages/modin/core/dataframe/algebra/binary.py in caller(query_compiler, other, broadcast, *args, **kwargs)
     90                 else:
     91                     return query_compiler.__constructor__(
---> 92                         query_compiler._modin_frame.binary_op(
     93                             lambda x, y: func(x, y, *args, **kwargs),
     94                             other._modin_frame,

/python3.9/site-packages/modin/core/dataframe/pandas/dataframe/dataframe.py in binary_op(self, op, right_frame, join_type)
   2030         # unwrap list returned by `copartition`.
   2031         right_parts = right_parts[0]
-> 2032         new_frame = self._partition_mgr_cls.binary_operation(
   2033             1, left_parts, lambda l, r: op(l, r), right_parts
   2034         )

/python3.9/site-packages/modin/core/execution/ray/implementations/pandas_on_ray/partitioning/partition_manager.py in magic(*args, **kwargs)
     55     @wraps(f)
     56     def magic(*args, **kwargs):
---> 57         result_parts = f(*args, **kwargs)
     58         if ProgressBar.get():
     59             current_frame = inspect.currentframe()

[...]

PR #4391, opened 2022-04-15
PERF-#4182, FIX-#4059: Add cell-wise execution for binary ops, fix bin ops for empty dataframes

### LINKED ISSUE #4182: Binary operations in Modin is slower than in pandas in case operands are Modin structures
<details><summary>Benchmark code</summary>

```python
import time

import numpy as np
import modin.pandas as pd
# import pandas as pd
import modin.config as cfg

cfg.BenchmarkMode.put(True)
# cfg.NPartitions.put(1)

def generate_data(path, nrows=5000, n_random_cols=100):
    labels = ["red", "blue", "black"]
    data = {
        "Int": np.random.randint(0, 100, nrows),
        "Float": np.random.rand(nrows),
        "Color": [labels[np.random.randint(0, len(labels), 1)[0]] for _ in range(nrows)],
    }
    data_random = {f"col{i}":  np.random.rand(nrows) for i in range(n_random_cols)}
    data.update(data_random)
    df = pd.DataFrame(data)

    df.to_csv(path, index=False)

if __name__ == "__main__":
    path = "data.csv"

    # generate_data(path, nrows=5000000, n_random_cols=100)

    print("Reading data...")
    df = pd.read_csv(path)

    # df = df._to_pandas()
    print(f"Original shape: {df.shape}")

    print("Filters...")

    t1 = time.time()
    filter_series1 = (df.Int < 50)
    print(f'Time binary op #1 (scalar): {time.time() - t1} s')

    t2 = time.time()
    filter_series2 = (df.Color == "red")
    print(f'Time binary op #2 (scalar): {time.time() - t2} s')

    t3 = time.time()
    filter_series3 = (df.Color != "blue")
    print(f'Time binary op #3 (scalar): {time.time() - t3} s')

    t4 = time.time()
    filter_series4 = (df.Float < 0.5)
    print(f'Time binary op #4 (scalar): {time.time() - t4} s')

    t5 = time.time()
    filter_series = filter_series1 & filter_series2 & filter_series3 & filter_series4
    print(f'Time binary op #5 (query compilers): {time.time() - t5} s')

    t6 = time.time()
    df = df[filter_series] # Default to pandas in case indexer is query_compiler
    print(f'Time getting by mask: {time.time() - t6} s')

    print(f"Result shape: {df.shape}")
    print(f'Time filters: {time.time() - t1} s')

```

</details>

Table below shows number from binary operation №5 (from the code above) (`&` operation between Modin Series).
| Shape | (10m, 103) | (10m, 4) | (5m, 103) |
| ----------- | ----------- | ----------- | ----------- |
| modin      | 4.996 | 5.33 | 4.98 |
| modin (NPartitions==1)| 0.184| 0.17| 0.11 |
| pandas | 0.025 | 0.024 | 0.008 |

Decreasing of number of partitions improves Modin's time. Possible, there is some gaps in parallelization.

prutskov:

## What do these changes do?

PR adds cell-wise execution for binary operations as attempt to improve performance.

- [x] Resolves #4182, #4059 
- [x] tests added and passing

<details><summary>Performance results</summary>

Hardware info: OS: Windows11 21H2, CPU: AMD Ryzen 7 5800HS 3.2 GHz 8 cores (16 threads) RAM: 16 GB DDR4 3200 MHz

1. Benchmarks results for binary opeartions. Command `asv continuous -f 1.01 master HEAD -b TimeBinaryOp.time_binary_op --launch-method=spawn` is used with different `MODIN_NPARTITIONS`.

`MODIN_NPARTITIONS=16` (default)

| shape (rows x cols left, rows x cols right), ms| 250x250, 250x250|10k x 20, 20k 10| 5k x 250, 5k x 250|5k x 5k, 5k x 5k|500k x 20, 1m x 10| 1m x 100, 1m x 100|500 x 10k, 500 x 10k|
| ----------- | :-----------: | :-----------: | :-----------: | :-----------: | :-----------: |:-----------: | :-----------: |
| modin #4391| 27     |181|56|208|916|228|138|
| modin master   | 28 |60  |54|275|787|711|150|

`MODIN_NPARTITIONS=4` (optimal number of partitions for cell-wise execution)

| shape (rows x cols left, rows x cols right), ms| 250x250, 250x250|10k x 20, 20k 10| 5k x 250, 5k x 250|5k x 5k, 5k x 5k|500k x 20, 1m x 10| 1m x 100, 1m x 100|500 x 10k, 500 x 10k|
| ----------- | :-----------: | :-----------: | :-----------: | :-----------: | :-----------: |:-----------: | :-----------: |
| modin #4391| 9    |60|18   |62  |744|185|30|
| modin master | 12 |40  |21|166|695|681|78|

2. Benchmarks for `fillna` function where `binary_op` is used as internal implementation (asv bench `TimeFillnaSeries.time_fillna`). Parameters are `self=Series`, `value` has `Series` type:

`MODIN_NPARTITIONS=16` (default)

| shape rows x cols, ms| 10k x 1|100k x 1| 1m x 100 | 10m x 10| 
| ----------- | :-----------: | :-----------: | :-----------: | :-----------: | 
| modin #4391| 24.5     | 26.1|5.1 s| 13.4 s|
| modin master   | 54.6 | 46.9| 5.3 s| 14.7 s| 

`MODIN_NPARTITIONS=4` (optimal number of partitions for cell-wise execution)

| shape rows x cols, ms| 10k x 1|100k x 1| 1m x 100 | 10m x 10| 
| ----------- | :-----------: | :-----------: | :-----------: | :-----------: | 
| modin #4391| 8.1     | 9.5|1.5 s| 6.82 s|
| modin master   | 9.43 | 10.8| 1.6 s| 6.81 s| 

3. Benchmark for `df.where` where `binary_op` is used as internal impl.

<details><summary>Benchmark code for `df.where` testing</summary>

```python
import time

import ray
ray.init(num_cpus=16)

import numpy as np
import modin.pandas as pd
import modin.config as cfg

# cfg.NPartitions.put(4)
cfg.BenchmarkMode.put(True)

def generate_data(nrows=5000, ncols=100):
    data = {f"col{i}":  np.random.rand(nrows) for i in range(ncols)}
    return pd.DataFrame(data), pd.DataFrame(data)

def get_time(op):
    start = time.time()
    result = op()
    return time.time() - start


if __name__ == "__main__":
    print("Generating data...")
    df1, df2 = generate_data(nrows=10_000, ncols=10)

    print(f"Original shape: {df1.shape}")

    def operation():
        return df1.where(df1 < 0.5, -df2)

    n_runs = 3
    all_time = 0
    get_time(operation)

    for i in range(n_runs):
        t = get_time(operation)
        print(f"Iteration #{i}: {t} s")
        all_time += t

    print(f'Time where op: {all_time / n_runs} s')

```

</details>

Results:
`MODIN_NPARTITIONS=16` (default)

| shape rows x cols, s| 10k x 10| 10k x 100|100k x 10|1m x 100|
| ----------- | :-----------: | :-----------: |:-----------: | :-----------: |
| modin #4391| 0.05     |0.38|0.21|23.45|
| modin master   | 0.12 |0.39|0.3|27.86|

`MODIN_NPARTITIONS=4` (optimal number of partitions for cell-wise execution)

| shape rows x cols, s| 10k x 10| 10k x 100|100k x 10|1m x 100|
| ----------- | :-----------: | :-----------: |:-----------: | :-----------: |
| modin #4391| 0.045     |0.21|0.23|24.8|
| modin master   | 0.053 |0.37|0.25|31.3|

4. Benchmark for `df.update` where `binary_op` is used as internal impl.

<details><summary>Benchmark code for `df.update` testing</summary>

```python
import time

import ray
ray.init(num_cpus=16)

import numpy as np
import modin.pandas as pd
import modin.config as cfg

# cfg.NPartitions.put(4)
cfg.BenchmarkMode.put(True)

def generate_data(nrows=5000, ncols=100):
    data1 = {f"col{i}":  np.random.rand(nrows) for i in range(ncols)}
    data2 = {f"col{i}":  np.random.rand(nrows) for i in range(ncols)}
    return pd.DataFrame(data1), pd.DataFrame(data2)

def get_time(op):
    start = time.time()
    result = op()
    return time.time() - start


if __name__ == "__main__":
    print("Generating data...")
    df1, df2 = generate_data(nrows=1000_000, ncols=100)

    print(f"Original shape: {df1.shape}")

    def operation():
        df1.update(df2)
        return df1

    n_runs = 3
    all_time = 0
    get_time(operation)

    for i in range(n_runs):
        t = get_time(operation)
        print(f"Iteration #{i}: {t} s")
        all_time += t

    print(f'Time where op: {all_time / n_runs} s')
```

</details>

Results:
`MODIN_NPARTITIONS=16` (default)

| shape rows x cols, s| 10k x 10| 10k x 100|100k x 100|1m x 100|
| ----------- | :-----------: | :-----------: |:-----------: | :-----------: |
| modin #4391| 0.019     |0.14|0.5|6.9|
| modin master   | 0.057 |0.12|1.61|15.43|

`MODIN_NPARTITIONS=4` (optimal number of partitions for cell-wise execution)

| shape rows x cols, s| 10k x 10| 10k x 100|100k x 100|1m x 100|
| ----------- | :-----------: | :-----------: |:-----------: | :-----------: |
| modin #4391| 0.007     |0.045|0.65|5.73|
| modin master   | 0.016 |0.12|1.3|14.52|

</details>

### DIFF
--- a/modin/core/dataframe/pandas/dataframe/dataframe.py
+++ b/modin/core/dataframe/pandas/dataframe/dataframe.py
@@ -2528,7 +2528,7 @@ class PandasDataframe(object, metaclass=LoggerMetaClass):
         # unwrap list returned by `copartition`.
         right_parts = right_parts[0]
         new_frame = self._partition_mgr_cls.binary_operation(
-            left_parts, lambda l, r: op(l, r), right_parts
+            left_parts, op, right_parts
         )
 
         return self.__constructor__(
--- a/modin/core/dataframe/pandas/dataframe/dataframe.py
+++ b/modin/core/dataframe/pandas/dataframe/dataframe.py
@@ -2525,10 +2525,12 @@ class PandasDataframe(object, metaclass=LoggerMetaClass):
             column_widths,
         ) = new_left_frame._copartition(1, new_right_frame, join_type, sort=True)
 
-        # unwrap list returned by `copartition`.
-        right_parts = right_parts[0]
-        new_frame = self._partition_mgr_cls.binary_operation(
-            left_parts, op, right_parts
+        new_frame = (
+            np.array([])
+            if len(left_parts) == 0 or len(right_parts[0]) == 0
+            else self._partition_mgr_cls.binary_operation(
+                left_parts, op, right_parts[0]
+            )
         )
 
         return self.__constructor__(
--- a/modin/core/dataframe/pandas/dataframe/dataframe.py
+++ b/modin/core/dataframe/pandas/dataframe/dataframe.py
@@ -2507,18 +2507,36 @@ class PandasDataframe(object, metaclass=LoggerMetaClass):
         left_parts, right_parts, joined_index, row_lengths = self._copartition(
             0, right_frame, join_type, sort=True
         )
+        new_left_frame = self.__constructor__(
+            left_parts, joined_index, self.columns, row_lengths, self._column_widths
+        )
+        new_right_frame = self.__constructor__(
+            right_parts[0],
+            joined_index,
+            right_frame.columns,
+            row_lengths,
+            right_frame._column_widths,
+        )
+
+        (
+            left_parts,
+            right_parts,
+            joined_columns,
+            column_widths,
+        ) = new_left_frame._copartition(1, new_right_frame, join_type, sort=True)
+
         # unwrap list returned by `copartition`.
         right_parts = right_parts[0]
         new_frame = self._partition_mgr_cls.binary_operation(
             left_parts, lambda l, r: op(l, r), right_parts, axis=1
         )
-        new_columns = self.columns.join(right_frame.columns, how=join_type)
+
         return self.__constructor__(
             new_frame,
             joined_index,
-            new_columns,
+            joined_columns,
             row_lengths,
-            column_widths=self._column_widths_cache,
+            column_widths,
         )
 
     @lazy_metadata_decorator(apply_axis="both")
--- a/modin/core/dataframe/pandas/dataframe/dataframe.py
+++ b/modin/core/dataframe/pandas/dataframe/dataframe.py
@@ -2510,7 +2510,7 @@ class PandasDataframe(object, metaclass=LoggerMetaClass):
         # unwrap list returned by `copartition`.
         right_parts = right_parts[0]
         new_frame = self._partition_mgr_cls.binary_operation(
-            1, left_parts, lambda l, r: op(l, r), right_parts
+            left_parts, lambda l, r: op(l, r), right_parts
         )
         new_columns = self.columns.join(right_frame.columns, how=join_type)
         return self.__constructor__(
--- a/modin/core/dataframe/pandas/dataframe/dataframe.py
+++ b/modin/core/dataframe/pandas/dataframe/dataframe.py
@@ -2510,7 +2510,7 @@ class PandasDataframe(object, metaclass=LoggerMetaClass):
         # unwrap list returned by `copartition`.
         right_parts = right_parts[0]
         new_frame = self._partition_mgr_cls.binary_operation(
-            left_parts, lambda l, r: op(l, r), right_parts
+            left_parts, lambda l, r: op(l, r), right_parts, axis=1
         )
         new_columns = self.columns.join(right_frame.columns, how=join_type)
         return self.__constructor__(
--- a/modin/core/dataframe/pandas/dataframe/dataframe.py
+++ b/modin/core/dataframe/pandas/dataframe/dataframe.py
@@ -2528,7 +2528,7 @@ class PandasDataframe(object, metaclass=LoggerMetaClass):
         # unwrap list returned by `copartition`.
         right_parts = right_parts[0]
         new_frame = self._partition_mgr_cls.binary_operation(
-            left_parts, lambda l, r: op(l, r), right_parts, axis=1
+            left_parts, lambda l, r: op(l, r), right_parts
         )
 
         return self.__constructor__(
--- a/modin/core/dataframe/pandas/partitioning/partition_manager.py
+++ b/modin/core/dataframe/pandas/partitioning/partition_manager.py
@@ -1242,34 +1242,44 @@ class PandasDataframePartitionManager(ABC):
 
     @classmethod
     @wait_computations_if_benchmark_mode
-    def binary_operation(cls, left, func, right):
+    def binary_operation(cls, left, func, right, axis=1):
         """
-        Apply a function that requires two PandasDataframe objects.
+        Apply a function that requires two ``PandasDataframe`` objects.
 
         Parameters
         ----------
         left : np.ndarray
-            The partitions of left PandasDataframe.
+            The partitions of left ``PandasDataframe``.
         func : callable
             The function to apply.
         right : np.ndarray
-            The partitions of right PandasDataframe.
+            The partitions of right ``PandasDataframe``.
+        axis : {0, 1}, default: 1
+            The axis to apply the function over (0 - rows, 1 - columns).
 
         Returns
         -------
         np.ndarray
             A NumPy array with new partitions.
         """
+        if axis:
+            left_partitions = cls.row_partitions(left)
+            right_partitions = cls.row_partitions(right)
+        else:
+            left_partitions = cls.column_partitions(left)
+            right_partitions = cls.column_partitions(right)
         func = cls.preprocess_func(func)
-        return np.array(
+        result = np.array(
             [
-                [
-                    part.apply(func, right[row_idx][col_idx].oid)
-                    for col_idx, part in enumerate(left[row_idx])
-                ]
-                for row_idx in range(len(left))
+                left_partitions[i].apply(
+                    func,
+                    num_splits=NPartitions.get(),
+                    other_axis_partition=right_partitions[i],
+                )
+                for i in range(len(left_partitions))
             ]
         )
+        return result if axis else result.T
 
     @classmethod
     @wait_computations_if_benchmark_mode
--- a/modin/core/dataframe/pandas/partitioning/partition_manager.py
+++ b/modin/core/dataframe/pandas/partitioning/partition_manager.py
@@ -1260,6 +1260,7 @@ class PandasDataframePartitionManager(ABC):
         np.ndarray
             A NumPy array with new partitions.
         """
+        [part.drain_call_queue() for part in right.flatten()]
 
         def op_with_empty_check(x, y, *args, **kwargs):
             y = pandas.DataFrame(index=x.index, columns=x.columns) if y is None else y
@@ -1272,7 +1273,7 @@ class PandasDataframePartitionManager(ABC):
                 [
                     part.apply(
                         op_with_empty_check,
-                        right[row_idx][col_idx].physical_data if len(right) else None,
+                        right[row_idx][col_idx]._data if len(right) else None,
                     )
                     for col_idx, part in enumerate(left[row_idx])
                 ]
--- a/modin/core/dataframe/pandas/partitioning/partition_manager.py
+++ b/modin/core/dataframe/pandas/partitioning/partition_manager.py
@@ -1262,18 +1262,13 @@ class PandasDataframePartitionManager(ABC):
         """
         [part.drain_call_queue() for part in right.flatten()]
 
-        def op_with_empty_check(x, y, *args, **kwargs):
-            y = pandas.DataFrame(index=x.index, columns=x.columns) if y is None else y
-
-            return func(x, y, *args, **kwargs)
-
-        op_with_empty_check = cls.preprocess_func(op_with_empty_check)
+        func = cls.preprocess_func(func)
         return np.array(
             [
                 [
                     part.apply(
-                        op_with_empty_check,
-                        right[row_idx][col_idx]._data if len(right) else None,
+                        func,
+                        right[row_idx][col_idx]._data,
                     )
                     for col_idx, part in enumerate(left[row_idx])
                 ]
--- a/modin/core/dataframe/pandas/partitioning/partition_manager.py
+++ b/modin/core/dataframe/pandas/partitioning/partition_manager.py
@@ -1242,14 +1242,12 @@ class PandasDataframePartitionManager(ABC):
 
     @classmethod
     @wait_computations_if_benchmark_mode
-    def binary_operation(cls, axis, left, func, right):
+    def binary_operation(cls, left, func, right):
         """
         Apply a function that requires two PandasDataframe objects.
 
         Parameters
         ----------
-        axis : {0, 1}
-            The axis to apply the function over (0 - rows, 1 - columns).
         left : np.ndarray
             The partitions of left PandasDataframe.
         func : callable
@@ -1262,24 +1260,16 @@ class PandasDataframePartitionManager(ABC):
         np.ndarray
             A NumPy array with new partitions.
         """
-        if axis:
-            left_partitions = cls.row_partitions(left)
-            right_partitions = cls.row_partitions(right)
-        else:
-            left_partitions = cls.column_partitions(left)
-            right_partitions = cls.column_partitions(right)
         func = cls.preprocess_func(func)
-        result = np.array(
+        return np.array(
             [
-                left_partitions[i].apply(
-                    func,
-                    num_splits=NPartitions.get(),
-                    other_axis_partition=right_partitions[i],
-                )
-                for i in range(len(left_partitions))
+                [
+                    part.apply(func, right[row_idx][col_idx].oid)
+                    for col_idx, part in enumerate(left[row_idx])
+                ]
+                for row_idx in range(len(left))
             ]
         )
-        return result if axis else result.T
 
     @classmethod
     @wait_computations_if_benchmark_mode
--- a/modin/core/dataframe/pandas/partitioning/partition_manager.py
+++ b/modin/core/dataframe/pandas/partitioning/partition_manager.py
@@ -1242,7 +1242,7 @@ class PandasDataframePartitionManager(ABC):
 
     @classmethod
     @wait_computations_if_benchmark_mode
-    def binary_operation(cls, left, func, right, axis=1):
+    def binary_operation(cls, left, func, right):
         """
         Apply a function that requires two ``PandasDataframe`` objects.
 
@@ -1254,32 +1254,31 @@ class PandasDataframePartitionManager(ABC):
             The function to apply.
         right : np.ndarray
             The partitions of right ``PandasDataframe``.
-        axis : {0, 1}, default: 1
-            The axis to apply the function over (0 - rows, 1 - columns).
 
         Returns
         -------
         np.ndarray
             A NumPy array with new partitions.
         """
-        if axis:
-            left_partitions = cls.row_partitions(left)
-            right_partitions = cls.row_partitions(right)
-        else:
-            left_partitions = cls.column_partitions(left)
-            right_partitions = cls.column_partitions(right)
-        func = cls.preprocess_func(func)
-        result = np.array(
+
+        def op_with_empty_check(x, y, *args, **kwargs):
+            y = pandas.DataFrame(index=x.index, columns=x.columns) if y is None else y
+
+            return func(x, y, *args, **kwargs)
+
+        op_with_empty_check = cls.preprocess_func(op_with_empty_check)
+        return np.array(
             [
-                left_partitions[i].apply(
-                    func,
-                    num_splits=NPartitions.get(),
-                    other_axis_partition=right_partitions[i],
-                )
-                for i in range(len(left_partitions))
+                [
+                    part.apply(
+                        op_with_empty_check,
+                        right[row_idx][col_idx].physical_data if len(right) else None,
+                    )
+                    for col_idx, part in enumerate(left[row_idx])
+                ]
+                for row_idx in range(len(left))
             ]
         )
-        return result if axis else result.T
 
     @classmethod
     @wait_computations_if_benchmark_mode
--- a/modin/core/execution/ray/implementations/pandas_on_ray/partitioning/partition_manager.py
+++ b/modin/core/execution/ray/implementations/pandas_on_ray/partitioning/partition_manager.py
@@ -18,6 +18,7 @@ import threading
 
 import numpy as np
 import ray
+import pandas
 
 from modin.config import ProgressBar, NPartitions
 from modin.core.execution.ray.generic.partitioning import (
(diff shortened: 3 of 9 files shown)


anmyachev:
@prutskov any updates here?

prutskov:
> @prutskov any updates here?

This PR is in a draft state, work is on hold.

modin/core/execution/ray/implementations/pandas_on_ray/partitioning/partition_manager.py
```diff
@@ -525,26 +525,34 @@ def apply_func_to_indices_both_axis(
...
-        return super(PandasOnRayDataframePartitionManager, cls).binary_operation(
-            axis, left, func, right
+        func = cls.preprocess_func(func)
+        [part.drain_call_queue() for row in right for part in row]
+        return np.array(
+            [
+                [
+                    part.apply(func, right[row_idx][col_idx].oid)
+                    for col_idx, part in enumerate(left[row_idx])
+                ]
```
anmyachev, line 555: What if the operand indexes don't match?

anmyachev, line 555: What if the operand indexes don't match?

prutskov:
Cell-wise execution requires an align partitioning by rows and cols at the same time that's expensive and sometimes isn't possible (in the case when left and right operands have significantly different shapes). Possible, that's reason why this is implemented as row-wise processing (https://github.com/modin-project/modin/issues/3100#issuecomment-852116896, @dchigarev).

After additional investigation it was found, that Ray uses the same workers for execution for several axis-partitions. In this case at least one of the remote tasks waits until worker finished previous tasks. This can be observed for any Modin/Ray setup. Example:

```python
import time

import ray

ray.init(num_cpus=16) # We have 16 workers to distribute execution of remote tasks
import numpy as np
# import pandas as pd
import modin.pandas as pd
import modin.config as cfg
from modin.logging import get_logger

cfg.NPartitions.put(2) # We have 2 axis partitions, which should be processed in different workers to be parallel processed
cfg.BenchmarkMode.put(True)
cfg.LogMode.enable()

logger = get_logger()

def generate_data(nrows=5000, ncols=100):
    data = {f"col{i}":  np.random.rand(nrows) for i in range(ncols)}
    return pd.DataFrame(data), pd.DataFrame(data)


if __name__ == "__main__":
    print("Generating data...")
    df1, df2 = generate_data(nrows=2_000_000, ncols=50)

    print(f"Original shape: {df1.shape}")

    t = time.time()
    logger.info("Start")
    result = df1 + df2             ---------> # Internally we have 2 row-partitions, which are processed in the same worker (print `pid` from remote function call to see).
    logger.info("End")
    print(f'Time binary op: {time.time() - t} s')
```

In case of correct scheduling the time of axis-partitions processing could be reduced at least to 2 times.

Does anyone faced with this scheduling problems?

CC @modin-project/modin-core, @modin-project/modin-contributors

mvashishtha:
@prutskov I am taking a look at this and have made some progress! I'll hopefully have an update within a few hours.

mvashishtha:
@prutskov I can reproduce that serial execution on my mac. I added a print statement to the remote function [here](https://github.com/modin-project/modin/blob/3d4404e9d9a9b2a3327f8aee664a8e71ac1f18b8/modin/core/execution/ray/implementations/pandas_on_ray/partitioning/virtual_partition.py#L506) and saw a single task executing both remote functions. However, on an Ubuntu EC2 instance (specs below) with more RAM, I got parallel execution.  I also get parallel execution on my mac when I decrease `nrows` to 10_000. I tracked the serial execution down to what looks like a ray bug: https://github.com/ray-project/ray/issues/25448

What are the specs of the machine you're testing on?


My mac:

macOS Monterey 12.2.1
MacBook Pro (16-inch, 2019) with 2.3 GHz 8-Core Intel Core i9
Memory: 16 GB 2667 MHz DDR4
Ray: 1.12.1
Python: 3.9.12

The Ubuntu EC2 instance:

AMI name: ubuntu/images/hvm-ssd/ubuntu-focal-20.04-amd64-server-20211129
instance type: t2.2xlarge
Ubuntu 20.04.3 LTS
8 physical CPUs
Memory: 32 GiB
Ray: 1.12.0
Python: 3.9.7

prutskov:
> @prutskov I can reproduce that serial execution on my mac. I added a print statement to the remote function [here](https://github.com/modin-project/modin/blob/3d4404e9d9a9b2a3327f8aee664a8e71ac1f18b8/modin/core/execution/ray/implementations/pandas_on_ray/partitioning/virtual_partition.py#L506) and saw a single task executing both remote functions. However, on an Ubuntu EC2 instance (specs below) with more RAM, I got parallel execution. I also get parallel execution on my mac when I decrease `nrows` to 10_000. I tracked the serial execution down to what looks like a ray bug: [ray-project/ray#25448](https://github.com/ray-project/ray/issues/25448)
> 
> What are the specs of the machine you're testing on?
> 
> My mac:
> 
> macOS Monterey 12.2.1 MacBook Pro (16-inch, 2019) with 2.3 GHz 8-Core Intel Core i9 Memory: 16 GB 2667 MHz DDR4 Ray: 1.12.1 Python: 3.9.12
> 
> The Ubuntu EC2 instance:
> 
> AMI name: ubuntu/images/hvm-ssd/ubuntu-focal-20.04-amd64-server-20211129 instance type: t2.2xlarge Ubuntu 20.04.3 LTS 8 physical CPUs Memory: 32 GiB Ray: 1.12.0 Python: 3.9.7

Thank you @mvashishtha! Really, reducing of data size makes execution parallel. Thank you a lot for investigating! I will continue investigating with taking this in mind

My laptop:

OS: Linux WSL2 5.10.16.3-microsoft-standard-WSL2
CPU: AMD Ryzen 7 5800HS 3.2 GHz 8 cores (16 threads)
RAM: 16 GB DDR4 3200 MHz 
Ray: 1.12.1
Python: 3.8.13

mvashishtha:
I think we have a good explanation of the ray serial execution here: https://github.com/ray-project/ray/issues/25448#issuecomment-1151639035

@Alexey Prutskov i think you saw the serial execution in your script [here](https://github.com/modin-project/modin/pull/4391#issuecomment-1145824540) because the default ray object store from `ray.init()` is smaller than the size modin should be using. on my mac, ray artificially limits object store size even lower to 2 GB so there was no way it could handle all of the data. You can confirm by looking for statements like `Dispatching task 9d172512a06e513526abaf75f5a71b612dd74bb601000000 would put this node over the max memory allowed for arguments of executing tasks (1503238528). Waiting to dispatch task until other tasks complete` in `/tmp/ray/session_latest/logs/` if you run your script with `RAY_BACKEND_LOG_LEVEL=debug`. 

In practice, do we expect the total active arguments to hit the limit of 70% of plasma store? If so, we should look into how we can get away with using less plasma.

prutskov:
@mvashishtha, thank you for finding of explanation of such behavior case!

> In practice, do we expect the total active arguments to hit the limit of 70% of plasma store? If so, we should look into how we can get away with using less plasma.

It's difficult question. I guess we can to hit the limit of 70% of plasma store for both cell/axis-wise executions type in case we have "large" `ray.ObjectRef`(s) per one remote call, and also we should have several remote calls. It will be more visibly for operations with several Modin DataFrames (binary ops, concat, merge etc). In this case we will have a number of `ray.ObjectRef` with partition data per one remote call.

For axis-wise operations, waiting/execution of remaining operations will take more time that for the same cell-wise (all depends from size of one remote task, which is bigger for axis-wise opertions).

For now I have no ideas how we can reduce the usage of plasma except increasing the size of plasma store during ray initialization.

For binary operations cell-wise execution will be used, because in that case we have a smaller impact of possible serial execution in case we reach the limit of 70% of plasma store.

mvashishtha:
> For binary operations cell-wise execution will be used, because in that case we have a smaller impact of possible serial execution in case we reach the limit of 70% of plasma store.

As I understand the 70% limit, making binary operations cell-wise could help, but in a subtle way. The limit seems to be on the **total** size of arguments to **all** executing tasks. Right now for your serial script above we have 2 remote functions that each take 4 block partitions (a full column partition for each 2x2 partitioned frame). The total size of the arguments when we kick off the remote tasks are 4 block partitions. If instead we had 4 remote functions that each took 1 pair of blocks, we'd have 4 tasks but the total size of all arguments would be the same. However, when each task's argument size is smaller, ray can pack more tasks in at once. For example, say your object store size is 10 GB and your total argument cap is 7 GB, and each of the 8 blocks is 1 GB. In the axis-partition case, we can only schedule one task at once, because each task requires 4 GB (4 partitions). In the block-partition case, we can schedule 7 of the tasks at once. The smaller tasks can be "packed" tighter and closer to the limit.

prutskov:
> > For binary operations cell-wise execution will be used, because in that case we have a smaller impact of possible serial execution in case we reach the limit of 70% of plasma store.
> 
> As I understand the 70% limit, making binary operations cell-wise could help, but in a subtle way. The limit seems to be on the **total** size of arguments to **all** executing tasks. Right now for your serial script above we have 2 remote functions that each take 4 block partitions (a full column partition for each 2x2 partitioned frame). The total size of the arguments when we kick off the remote tasks are 4 block partitions. If instead we had 4 remote functions that each took 1 pair of blocks, we'd have 4 tasks but the total size of all arguments would be the same. However, when each task's argument size is smaller, ray can pack more tasks in at once. For example, say your object store size is 10 GB and your total argument cap is 7 GB, and each of the 8 blocks is 1 GB. In the axis-partition case, we can only schedule one task at once, because each task requires 4 GB (4 partitions) and both tasks would be 8 GB, over the limit. In the block-partition case, we can schedule 3/4 of the tasks at once since each requires 2 GB, and 3 tasks require 6 GB, which is under the limit. The smaller tasks can be "packed" tighter and closer to the limit, with 3x parallelism for most of the execution instead of no parallelism.

I thought about the same example:) 

> edit: but again, I don't know whether this argument limit is something we hit in practice with the size of object store that modin sets by default. If not, we shouldn't take the limit into consideration in our design.

I don't think that we should take the limit into consideration for our implementations. The warning message from ray about serial execution (which is mentioned in https://github.com/ray-project/ray/issues/25448#issuecomment-1151731616 ) will be enough.
> 
> edit: made some math errors :)
> 
> edit: Also, in practice for us, the object store for us probably has many things besides arguments. how does that affect this problem 🤔 ? Does that mean we are even tighter on space for arguments?

I think many other objects in object storage can exacerbate the problem because in this case we will have less space for active arguments and sequential execution will happen more often.

mvashishtha (commented):
The performance looks great!

modin/core/dataframe/pandas/partitioning/partition_manager.py
```diff
@@ -1242,44 +1242,43 @@ def compute_part_size(indexer, remote_part, part_idx, axis):
...
-        if axis:
-            left_partitions = cls.row_partitions(left)
-            right_partitions = cls.row_partitions(right)
-        else:
-            left_partitions = cls.column_partitions(left)
-            right_partitions = cls.column_partitions(right)
-        func = cls.preprocess_func(func)
-        result = np.array(
+
+        def op_with_empty_check(x, y, *args, **kwargs):
```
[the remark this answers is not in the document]
YarShev: The algebra operators should be generic as possible. I think this corner case should be handled in `_copartition`.

YarShev:
I am not sure the performance looks great and comprehensive. We should check all/most corner cases such as applying a function to df and s, s and df, for example, as well as at least 2 or 3 general scenarios.

modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -2507,18 +2507,36 @@ def binary_op(self, op, right_frame, join_type="outer"):
...
+            right_parts,
+            joined_columns,
+            column_widths,
+        ) = new_left_frame._copartition(1, new_right_frame, join_type, sort=True)
+
         # unwrap list returned by `copartition`.
         right_parts = right_parts[0]
         new_frame = self._partition_mgr_cls.binary_operation(
-            1, left_parts, lambda l, r: op(l, r), right_parts
+            left_parts, lambda l, r: op(l, r), right_parts
```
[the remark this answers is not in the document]
prutskov: This is same. It's strange why it was done by this way before. Fixed

dchigarev:
@YarShev 

> We should check all/most corner cases such as applying a function to df and s, s and df

The PR only modifies the [`modin_frame.binary_op`](https://github.com/modin-project/modin/blob/4ec7f6347903f9133c65ebc5b6e0e15553b98577/modin/core/dataframe/pandas/dataframe/dataframe.py#L2489) behavior. The thing is that the `binary_op` is designed for `df + df` operations only. The handling of mixin a frame and a series has to be done differently via broadcasting a series to every column of the frame instead of attempting to align the shapes of two operands. We already have the [broadcasting logic inside the `Binary` operator](https://github.com/modin-project/modin/blob/4ec7f6347903f9133c65ebc5b6e0e15553b98577/modin/core/dataframe/algebra/binary.py#L71-L89), the flow goes to the `modin_frame.broadcast_apply` which wasn't touched in this PR.

We would potentially try the same cell-wise execution approach on the Series+DF operations, but it seems like a different task.

YarShev:
@dchigarev, thanks for pointing this out! I didn't notice that Binary operator and the broadcast logic are untouched. Still, I suggest checking the performance for [fillna](https://github.com/modin-project/modin/blob/4ec7f6347903f9133c65ebc5b6e0e15553b98577/modin/core/storage_formats/pandas/query_compiler.py#L1843) and [where](https://github.com/modin-project/modin/blob/4ec7f6347903f9133c65ebc5b6e0e15553b98577/modin/core/storage_formats/pandas/query_compiler.py#L406), which call the [binary_op](https://github.com/modin-project/modin/blob/4ec7f6347903f9133c65ebc5b6e0e15553b98577/modin/core/dataframe/pandas/dataframe/dataframe.py#L2489) at the QC layer, as well as for [series_update](https://github.com/modin-project/modin/blob/4ec7f6347903f9133c65ebc5b6e0e15553b98577/modin/core/storage_formats/pandas/query_compiler.py#L398) and one of the binary operations. We have the numbers for the latter (`mul` as I see) in the PR description.

modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -2507,18 +2507,36 @@ def binary_op(self, op, right_frame, join_type="outer"):
...
+            row_lengths,
+            right_frame._column_widths,
+        )
+
+        (
+            left_parts,
+            right_parts,
+            joined_columns,
+            column_widths,
+        ) = new_left_frame._copartition(1, new_right_frame, join_type, sort=True)
```
YarShev, line 2526: Can we avoid double copartitioning?

modin/core/dataframe/pandas/partitioning/partition_manager.py
```diff
@@ -1242,44 +1242,44 @@ def compute_part_size(indexer, remote_part, part_idx, axis):
...
         """
-        if axis:
-            left_partitions = cls.row_partitions(left)
-            right_partitions = cls.row_partitions(right)
-        else:
-            left_partitions = cls.column_partitions(left)
-            right_partitions = cls.column_partitions(right)
-        func = cls.preprocess_func(func)
-        result = np.array(
+        [part.drain_call_queue() for part in right.flatten()]
```
YarShev, line 1263: This is the case when we can pass the right partitions and the respective call queues in the remote calls.

modin/core/dataframe/pandas/partitioning/partition_manager.py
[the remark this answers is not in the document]
prutskov: Implementation of this corner case was changed in 9e90c5e9fca72c98182a13929a895d95cf22b1a9. For now, the implementation doesn't do a remote call in case the left or right operands are empty and just return an empty Modin DataFrame with correct index/columns as pandas does.

I don't see the reason to create an empty dataframes of right partitions in `_copartition` because it will just add superfluous execution time.

modin/core/dataframe/pandas/partitioning/partition_manager.py
prutskov: We can but I don't see any benefit from applying of the respective call queue of right partition in the remote call here. This will be useful in case we have lazy execution (but even that is questionable, described below), but here we use `partition.apply` which runs an execution of remote calls right now.

In this row we run an early execution of the call queue in background for each `right_partition` and immediately actualize `right_partition._data`. After, during serial logic of `left_partition.apply` call for each partition we will have already finished remote calls from the call queue of the each `right_partition` with some probability.

If we will use your way we can than draining of call queue for `right_partition` will happen only after draining of call queue for `left_partition` in remote call.

Also, by providing the call queue of right partition together with `right_partition._data` in `left_partition.apply` we don't drain `right_partition.call_queue` and don't actualize the `right_partition._data`. If we will work with the right dataframe further in the code we will again run `drain_call_queue` for each partition of right dataframe. In the result we have double work of the same calls but in different places.

What benefits do you see from your suggestion? How we will resolve the issue which I described in this case?

mvashishtha (approved):

modin/core/dataframe/pandas/partitioning/partition_manager.py
YarShev: These are by far valid points but on the other hand there are others that can affect the performance. For instance, it we are considering Ray, the following factors but probably not limited to can affect the performance: 1. how many partitions are contained in the right array. 2. how many physical (materialized) data of the right partitions will be saved into in-process memory of the driver. 3. how many physical (materialized) data of the right partitions will be saved into the plasma.

We can experiment this case as part of a separate issue if we see good times without passing the call queues of the right partitions in to the remote calls.

modin/core/dataframe/pandas/partitioning/partition_manager.py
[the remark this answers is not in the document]
YarShev: Makes sense to me.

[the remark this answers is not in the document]
YarShev: Since #4059 seems to be resolved in this PR, please include it in the PR description and in release notes.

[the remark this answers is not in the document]
prutskov: #4059 was added in PR title/commit message/release notes

prutskov:
The actual benchmark results for all affected funcionality(binary ops, `fillna`, `where`, `update`) were added in the PR description. In the most cases performance was increased or not changed for some cases of usage.

modin/core/dataframe/pandas/partitioning/partition_manager.py
prutskov: I still believe that the main issue with implementation you suggest will in duplicating of processing operations for right partition(the first one is draining of call queue in remote call from the left partition, the second one in right partition directly in other part of code).

I mean the next flow in the sentences above:

```python
df1 = pd.DataFrame(...) 
df2 = pd.DataFrame(...).T # Add tranpose operation in call queue of parts in df2

result = df1 + df2    # Perform cell-wise execution. Each remote call for left partition also runs transpose operation for right part. The parts of df2 isn't updated.
print(df2) # Drain call queue for parts of df2 (transpose operation)
```

So, I suggest to experiment with this new execution flow in a separate issue, because it, possible, requires a lot of architectural changes.

YarShev (approved):
@prutskov, thanks for the changes and measurements, LGTM!

PR #4420, opened 2022-04-25
FEAT-#4419: Extend virtual partitioning API to pandas on Dask

### LINKED ISSUE #4419: Extend Virtual Partitioning API to Dask
We should extend the virtual partitioning API to pandas on Dask.

RehanSD:

## What do these changes do?

- [x] Resolves #4419 ? 
- [ ] tests added and passing


devin-petersohn (commented):
Should we add tests for this?

modin/core/execution/dask/implementations/pandas_on_dask/partitioning/virtual_partition.py
```diff
@@ -36,23 +38,97 @@ class PandasOnDaskDataframeAxisPartition(PandasDataframeAxisPartition):
...
+        self.full_axis = full_axis
+        if isinstance(list_of_blocks, PandasOnDaskDataframePartition):
+            list_of_blocks = [list_of_blocks]
+        if not any(
+            isinstance(o, PandasOnDaskDataframeVirtualPartition) for o in list_of_blocks
+        ):
+            self.list_of_partitions_to_combine = list_of_blocks
+            return
+
+        assert (
```
mvashishtha, line 58: Can we the ray virtual partition constructor comments, e.g. [this](https://github.com/modin-project/modin/blob/0ab04e104a60419c464fc1833a13f7f2b92e7983/modin/core/execution/ray/implementations/pandas_on_ray/partitioning/virtual_partition.py#L61), explaining some of the non-trivial code?

modin/core/execution/dask/implementations/pandas_on_dask/partitioning/virtual_partition.py
```diff
@@ -36,23 +38,97 @@ class PandasOnDaskDataframeAxisPartition(PandasDataframeAxisPartition):
...
+                if isinstance(o, PandasOnDaskDataframeVirtualPartition)
+            ).axis
+            == self.axis
+        ):
+            new_list_of_blocks = []
+            for o in list_of_blocks:
+                new_list_of_blocks.extend(
+                    o.list_of_partitions_to_combine
+                ) if isinstance(
+                    0, PandasOnDaskDataframeVirtualPartition
```
mvashishtha, line 81:
```suggestion
                    o, PandasOnDaskDataframeVirtualPartition
```

mvashishtha (commented):
I have some minor comments.

We can't test much of this code until we add something like repeated partition rebalancing, which can result in creating virtual partitions out of other virtual partitions.

modin/core/execution/dask/implementations/pandas_on_dask/partitioning/virtual_partition.py
RehanSD: Sure, that makes sense to me!

modin/core/execution/dask/implementations/pandas_on_dask/partitioning/virtual_partition.py
RehanSD: Committed, thank you!

mvashishtha (approved):

anmyachev:
@RehanSD we need to implement `rebalance_partitions` method here, right? As it was for Ray: https://github.com/modin-project/modin/blob/830e8dcea0f05118e4a2b66d66ed42dce9b84d5a/modin/core/execution/ray/implementations/pandas_on_ray/partitioning/partition_manager.py#L148

anmyachev:
@RehanSD we need to implement `rebalance_partitions` method here, right? As it was for Ray: https://github.com/modin-project/modin/blob/830e8dcea0f05118e4a2b66d66ed42dce9b84d5a/modin/core/execution/ray/implementations/pandas_on_ray/partitioning/partition_manager.py#L148

modin/core/execution/dask/implementations/pandas_on_dask/partitioning/virtual_partition.py
```diff
@@ -36,23 +38,97 @@ class PandasOnDaskDataframeAxisPartition(PandasDataframeAxisPartition):
...
+            self.list_of_partitions_to_combine = list_of_blocks
+            return
+
+        assert (
+            len(
+                set(
+                    o.axis
+                    for o in list_of_blocks
+                    if isinstance(o, PandasOnDaskDataframeVirtualPartition)
+                )
```
naren-ponder, line 64: Nit: Instead of having two different loops here, can we combine them into one if/else case?

naren-ponder (commented):
LGTM overall!

devin-petersohn (commented):
@RehanSD please rebase this on current master

YarShev:
> @RehanSD we need to implement `rebalance_partitions` method here, right? As it was for Ray:
> 
> https://github.com/modin-project/modin/blob/830e8dcea0f05118e4a2b66d66ed42dce9b84d5a/modin/core/execution/ray/implementations/pandas_on_ray/partitioning/partition_manager.py#L148

cc @RehanSD

modin/core/execution/dask/implementations/pandas_on_dask/partitioning/partition_manager.py
```diff
@@ -48,3 +52,150 @@ def get_objects_from_partitions(cls, partitions):
...
+        result = super(PandasOnDaskDataframePartitionManager, cls).concat(
+            axis, left_parts, right_parts
+        )
+        if axis == 0:
+            return cls.rebalance_partitions(result)
+        else:
+            return result
+
+    @classmethod
+    def rebalance_partitions(cls, partitions):
```
mvashishtha, line 90: I think this method can be moved to the parent class.  Its logic is very complex and we shouldn't duplicate it.

modin/core/execution/dask/implementations/pandas_on_dask/partitioning/partition_manager.py
```diff
@@ -48,3 +52,150 @@ def get_objects_from_partitions(cls, partitions):
             The objects wrapped by `partitions`.
         """
         return DaskWrapper.materialize([partition._data for partition in partitions])
+
+    @classmethod
+    def concat(cls, axis, left_parts, right_parts):
```
mvashishtha, line 57: Can this be moved to the parent class instead, too?

mvashishtha (commented):
@RehanSD I left two comments about trying to move some code that might be shared between Dask and Ray. I can review in more detail later.

modin/test/storage_formats/pandas/test_internals.py
```diff
@@ -110,3 +110,38 @@ def func_to_apply(partition, row_internal_indices, col_internal_indices, item):
...
+def test_rebalance_partitions():
+    small_dfs = [
+        pd.DataFrame(
+            [[i + j for j in range(0, 100)]],
+            columns=[f"col{j}" for j in range(1, 101)],
+            index=pd.Index([i - 1]),
+        )
+        for i in range(1, 10001, 100)
+    ]
+    large_df = pd.concat(small_dfs)
```
mvashishtha, line 128: Could you also add a case that forces us to split a partition? e.g. concat a DF with 10_000 rows to 5  dfs with 1 row each? e.g. this produces splitting with `NPartitions` of 16:

```
pd.concat([pd.DataFrame([[i] for i in range(1000)])] + [pd.DataFrame([[i]]) for i in range(16)])
```

you need enough partitions in the first partition for the partitioning to be imbalanced enough to trigger repartitioning.

modin/test/storage_formats/pandas/test_internals.py
```diff
@@ -110,3 +110,38 @@ def func_to_apply(partition, row_internal_indices, col_internal_indices, item):
...
+    large_modin_frame = large_df._query_compiler._modin_frame
+    assert large_modin_frame._partitions.shape == (
+        4,
+        4,
+    ), "Partitions were not rebalanced after concat."
+    assert all(
+        isinstance(ptn, large_modin_frame._partition_mgr_cls._column_partitions_class)
+        for ptn in large_modin_frame._partitions.flatten()
+    )
+    large_df = large_df.apply(lambda x: x + 1)
```
mvashishtha, line 138: Could you please add a comment about why you're testing what the apply does? One reason I'm thinking of is that it requires building full-axis virtual partitions out of non-full-axis virtual partitions.

modin/test/storage_formats/pandas/test_internals.py
RehanSD: That makes a lot of sense - will go ahead and add!

modin/test/storage_formats/pandas/test_internals.py
mvashishtha: Isn't this a full-axis apply? I thought we would create full-axis virtual partitions out the virtual partitions, apply the UDF to the full-axis partitions, then split the virtual partitions into block partitions?

RehanSD: You are correct - my bad!

modin/test/storage_formats/pandas/test_internals.py
```diff
@@ -110,3 +110,79 @@ def func_to_apply(partition, row_internal_indices, col_internal_indices, item):
...
+@pytest.mark.skipif(
+    Engine.get() not in ("Dask", "Ray"),
+    reason="Rebalancing partitions is only supported for Dask and Ray engines",
+)
+@pytest.mark.parametrize(
+    "large_df,col_length",
+    [
+        (pd.concat(small_dfs), 100),
+        (pd.concat([pd.concat(small_dfs)] + small_dfs[:3]), 103),
+        (pd.concat([large_df] + small_dfs[:3]), 103),
```
mvashishtha, line 139: I don't think we should call the Modin function under test within the pytest params. If the Modin `concat` throws an unexpected exception (not that unlikely when someone is modifying modin), I think that pytest will fail while gathering the tests, and the developer will probably be confused. I understand that there is some repeated work in these test cases, and you want to test rebalancing different kinds of partitions, but could you make it more succinct in some other way? What about using labels like `many_small_dfs`, `rebalance_twice`, and `large_partitions_with_smaller_partitions`, and doing the concat that corresponds to the label?

modin/test/storage_formats/pandas/test_internals.py
```diff
@@ -110,3 +110,79 @@ def func_to_apply(partition, row_internal_indices, col_internal_indices, item):
...
+
+small_dfs = [
+    pd.DataFrame(
+        [[i + j for j in range(0, 1000)]],
+        columns=[f"col{j}" for j in range(1, 1001)],
+        index=pd.Index([i - 1]),
+    )
+    for i in range(1, 100001, 1000)
+]
+large_df = pd.DataFrame(
```
mvashishtha, line 123: If you move the concats out of the params, I think you should move these definitions inside the test body.

modin/test/storage_formats/pandas/test_internals.py
```diff
@@ -110,3 +110,79 @@ def func_to_apply(partition, row_internal_indices, col_internal_indices, item):
...
+    ), "Partitions were not rebalanced after concat."
+    assert all(
+        isinstance(ptn, large_modin_frame._partition_mgr_cls._column_partitions_class)
+        for ptn in large_modin_frame._partitions.flatten()
+    )
+    # The following check tests that we can correctly form full-axis virtual partitions
+    # over the orthogonal axis from non-full-axis virtual partitions.
+
+    def col_apply_func(col):
+        assert len(col) == col_length, "Partial axis partition detected."
```
mvashishtha, line 156: Why do we need this `assert`? Why isn't it enough to check that the column-wise apply does what pandas would do?

modin/test/storage_formats/pandas/test_internals.py
```diff
@@ -110,3 +110,79 @@ def func_to_apply(partition, row_internal_indices, col_internal_indices, item):
...
+        for ptn in new_large_modin_frame._partitions.flatten()
+    ), "Partitions are not block partitioned after apply."
+    large_df = pd.DataFrame(
+        query_compiler=large_df._query_compiler.__constructor__(large_modin_frame)
+    )
+    # The following check tests that we can correctly form full-axis virtual partitions
+    # over the same axis from non-full-axis virtual partitions.
+
+    def row_apply_func(row):
+        assert len(row) == 1000, "Partial axis partition detected."
```
mvashishtha, line 176: ditto about the assert

modin/test/storage_formats/pandas/test_internals.py
```diff
@@ -110,3 +110,79 @@ def func_to_apply(partition, row_internal_indices, col_internal_indices, item):
...
+
+@pytest.mark.skipif(
+    Engine.get() not in ("Dask", "Ray"),
+    reason="Rebalancing partitions is only supported for Dask and Ray engines",
+)
+@pytest.mark.parametrize(
+    "large_df,col_length",
+    [
+        (pd.concat(small_dfs), 100),
+        (pd.concat([pd.concat(small_dfs)] + small_dfs[:3]), 103),
```
mvashishtha, line 138: Does this really trigger two rebalances?

modin/test/storage_formats/pandas/test_internals.py
```diff
@@ -110,3 +110,79 @@ def func_to_apply(partition, row_internal_indices, col_internal_indices, item):
...
+
+
+small_dfs = [
+    pd.DataFrame(
+        [[i + j for j in range(0, 1000)]],
+        columns=[f"col{j}" for j in range(1, 1001)],
+        index=pd.Index([i - 1]),
+    )
+    for i in range(1, 100001, 1000)
+]
```
mvashishtha, line 122:
```suggestion
small_dfs = [
    pd.DataFrame([[i + j for j in range(0, 1000)]]).add_prefix('col')
    for i in range(100 * 1000, 1000)
]
```

I prefer to have the values start from 0. Then we can create the dataframe like this. Let me know if you prefer the current way.

modin/test/storage_formats/pandas/test_internals.py
```diff
@@ -110,3 +110,79 @@ def func_to_apply(partition, row_internal_indices, col_internal_indices, item):
...
+large_df = pd.DataFrame(
+    [[i + j for j in range(1, 1000)] for i in range(0, 100000, 1000)],
+    columns=[f"col{j}" for j in range(1, 1000)],
+    index=pd.Index(list(range(0, 100000, 1000))),
+)
```
mvashishtha, line 127:
```suggestion
large_df = pd.DataFrame(
    [[i + j for j in range(1000)] for i in range(0, 100 * 1000, 1000)]]
).add_prefix('col')
```

Unless you see a good reason for the 1-indexing and a custom index, I prefer this.

modin/core/execution/dask/implementations/pandas_on_dask/dataframe/dataframe.py
```diff
@@ -41,22 +42,63 @@ class PandasOnDaskDataframe(PandasDataframe):
 
     _partition_mgr_cls = PandasOnDaskDataframePartitionManager
 
+    def _get_partition_size_along_axis(self, partition, axis=0):
```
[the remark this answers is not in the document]
mvashishtha: Ah, I'm wrong about this function blocking on inner partitions, because it doesn't call `length` or `width`. So I think it's correct. Still, this part seems outside the scope of this PR. I'd rather worry about this fix in a separate PR for #4494.

modin/core/execution/dask/implementations/pandas_on_dask/partitioning/virtual_partition.py
```diff
@@ -165,8 +251,209 @@ def _wrap_partitions(self, partitions):
...
+            self.drain_call_queue()
+        kwargs["args"] = args
+        result = super(PandasOnDaskDataframeVirtualPartition, self).apply(
+            func, num_splits, other_axis_partition, maintain_partitioning, **kwargs
+        )
+        if self.full_axis:
+            return result
+        else:
+            # If this is a full axis partition, just take out the single split in the result.
+            return result[0]
```
pyrito, line 306:
```suggestion
        return result if self.full_axis else result[0]
```
Maybe something like this?

modin/core/dataframe/pandas/partitioning/partition_manager.py
```diff
@@ -1302,6 +1314,103 @@ def rebalance_partitions(cls, partitions):
...
+            # put into the current new partition.
+            start = 0
+            total_rows = sum(part.length() for part in partitions[:, 0])
+            ideal_partition_size = compute_chunksize(
+                total_rows, ideal_num_new_partitions, min_block_size=1
+            )
+            for _ in range(ideal_num_new_partitions):
+                # We might pick up old partitions too quickly and exhaust all of them.
+                if start >= len(partitions):
+                    break
```
pyrito, line 1374: I don't understand when this would happen?

pyrito (commented):
Looks pretty good to me - mainly just style nits and questions to clarify my understanding

docs/flow/modin/core/execution/dask/implementations/pandas_on_dask/partitioning/axis_partition.rst
```diff
@@ -1,14 +1,14 @@
-PandasOnDaskDataframeAxisPartition
-""""""""""""""""""""""""""""""""""
+PandasOnDaskDataframeVirtualPartition
```
YarShev, line 1: This file should be renamed to virtual_partition.rst.

modin/core/dataframe/pandas/partitioning/partition_manager.py
```diff
@@ -1292,7 +1296,15 @@ def finalize(cls, partitions):
     @classmethod
     def rebalance_partitions(cls, partitions):
         """
-        Return the provided array of partitions without rebalancing it.
+        Rebalance a 2-d array of partitions if we are using PandasOnRay or PandasOnDask executions.
```
YarShev, line 1299:
```suggestion
        Rebalance a 2-d array of partitions if we are using ``PandasOnRay`` or ``PandasOnDask`` executions.
```

modin/core/execution/dask/implementations/pandas_on_dask/dataframe/dataframe.py
```diff
@@ -41,22 +42,63 @@ class PandasOnDaskDataframe(PandasDataframe):
...
         list
             A list of row partitions lengths.
         """
         if self._row_lengths_cache is None:
-            self._row_lengths_cache = (
-                self._partition_mgr_cls.get_objects_from_partitions(
-                    [obj.apply(lambda df: len(df)) for obj in self._partitions.T[0]]
-                )
+            row_lengths_list = DaskWrapper.materialize(
+                [
```
YarShev, line 96: Probably, there is no need for [] as self._get_partition_size_along_axis returns a list.

modin/core/execution/dask/implementations/pandas_on_dask/dataframe/dataframe.py
```diff
@@ -70,12 +112,13 @@ def _column_widths(self):
...
         """
         if self._column_widths_cache is None:
-            self._column_widths_cache = (
-                self._partition_mgr_cls.get_objects_from_partitions(
-                    [
-                        obj.apply(lambda df: len(df.columns))
-                        for obj in self._partitions[0]
-                    ]
-                )
+            col_widths_list = DaskWrapper.materialize(
```
YarShev, line 115: same

modin/core/execution/dask/implementations/pandas_on_dask/dataframe/dataframe.py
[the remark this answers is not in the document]
RehanSD: I think it makes sense to include here, since the original code to get length + width in parallel breaks when applied to axis partitions so I'll need to fix that code anyways, and if I'm doing that, I may as well fix the code to just do it all in parallel right?

modin/test/storage_formats/pandas/test_internals.py
RehanSD: We want to ensure that this is a full-axis partition, and not just a partial axis partition - the function as I've written it will run correctly even if it is a partial axis partition, so I just thought this assert would be a good way to keep the function simple, but still test that it is full axis.

modin/core/execution/dask/implementations/pandas_on_dask/dataframe/dataframe.py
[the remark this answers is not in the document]
mvashishtha: If the original code was broken, I think it's okay to include the partial fix for #4494 here.

modin/test/storage_formats/pandas/test_internals.py
mvashishtha: I think this test should check for functional correctness. We already check that we get the `apply` result correct when the starting partitions are non-full-axis virtual partitions. I don't think it's right to check for internal implementation details here.

RehanSD: Hmm - I'm not sure that I agree with that - this file is called `test_internals` which would imply its testing internal details, and it makes sense to ensure that the internal partitioning code didn't incorrectly make a full axis partition.

modin/core/dataframe/pandas/partitioning/partition_manager.py
```diff
@@ -1302,6 +1314,103 @@ def rebalance_partitions(cls, partitions):
...
+            engine and storage format.
         """
+        if Engine.get() in ["Ray", "Dask"] and StorageFormat.get() == "Pandas":
+            # Rebalancing partitions is currently only implemented for PandasOnRay and PandasOnDask.
+            # We rebalance when the ratio of the number of existing partitions to
+            # the ideal number of partitions is larger than this threshold. The
+            # threshold is a heuristic that may need to be tuned for performance.
+            max_excess_of_num_partitions = 1.5
+            num_existing_partitions = partitions.shape[0]
+            ideal_num_new_partitions = NPartitions.get()
```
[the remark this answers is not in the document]
RehanSD: You can override `NPartitions.get()` to do so. It just asks Ray how many cores there are, and returns that, so as long as your cluster is fully initialized when modin is initialized, and you set `IsRayCluster` correctly, it should.

modin/core/dataframe/pandas/partitioning/partition_manager.py
```diff
@@ -1302,6 +1314,103 @@ def rebalance_partitions(cls, partitions):
...
+                total_rows, ideal_num_new_partitions, min_block_size=1
+            )
+            for _ in range(ideal_num_new_partitions):
+                # We might pick up old partitions too quickly and exhaust all of them.
+                if start >= len(partitions):
+                    break
+                # `stop` is the index of the last existing partition so far that we
+                # want to put into the current new partition.
+                stop = start
+                partition_size = partitions[start][0].length()
```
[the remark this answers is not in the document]
RehanSD: We try to cache it, but it will block if it is not cached. If it is blocking, it submits a remote task that gets the length of the object, and only materializes the length in memory on the main node.

modin/core/execution/dask/implementations/pandas_on_dask/partitioning/virtual_partition.py
```diff
@@ -24,35 +26,119 @@
...
+                for obj in list_of_blocks
+            ]
 
     partition_type = PandasOnDaskDataframePartition
     instance_type = Future
 
+    @property
+    def list_of_blocks(self):
+        """
+        Get the list of physical partition objects that compose this partition.
```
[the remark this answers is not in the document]
RehanSD: Yup - a virtual partition is just one or more block partitions (physical partitions) and an axis that they are aligned along.

modin/core/execution/dask/implementations/pandas_on_dask/partitioning/virtual_partition.py
```diff
@@ -165,8 +251,209 @@ def _wrap_partitions(self, partitions):
...
+            Additional keywords arguments to be passed in `func`.
+
+        Returns
+        -------
+        list
+            A list of `PandasOnDaskDataframeVirtualPartition` objects.
+        """
+        if not self.full_axis:
+            # If this is not a full axis partition, it already contains a subset of
+            # the full axis, so we shouldn't split the result further.
```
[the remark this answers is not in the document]
RehanSD: After a rebalance for example. If we have 20 very small partitions, we may chunk up some of them into non-full axis virtual partitions, so maybe we have 4 virtual partitions of 5 block partitions each. After we apply a function to the virtual partitions, we want to materialize the result as 1 block partition, instead of as 5 small block partitions (as it was before) in order to complete the rebalance, which would trigger this if statement.

modin/core/execution/dask/implementations/pandas_on_dask/partitioning/virtual_partition.py
```diff
@@ -165,8 +251,209 @@ def _wrap_partitions(self, partitions):
...
+
+        Returns
+        -------
+        pandas DataFrame.
+        """
+        return self.force_materialization().list_of_partitions_to_combine[0].to_pandas()
+
+    _length_cache = None
+
+    def length(self):
```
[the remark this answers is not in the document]
RehanSD: We don't really have any unit testing at the partition level. This PR is actually the first to add any testing thats supposed to cover that level. We could look into adding more testing, but I think that would belong in a separate PR.

modin/core/execution/dask/implementations/pandas_on_dask/partitioning/virtual_partition.py
```diff
@@ -165,8 +251,209 @@ def _wrap_partitions(self, partitions):
...
+        -------
+        list
+            A list of `PandasOnDaskDataframeVirtualPartition` objects.
+        """
+        if not self.full_axis:
+            # If this is not a full axis partition, it already contains a subset of
+            # the full axis, so we shouldn't split the result further.
+            num_splits = 1
+        if len(self.call_queue) > 0:
+            self.drain_call_queue()
```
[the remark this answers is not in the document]
RehanSD: Good question - tbh I don't think there's really any huge reason to do so. It does save an extra call on the stack, but thats an extremely minuscule effect (so def not why). I think it just boils down to the idea that the `drain_call_queue` method (conceptually) should only be called when we have a call queue?

modin/test/storage_formats/pandas/test_internals.py
mvashishtha: I don't think it's appropriate to test any internal detail here just because it's `test_internals`. Why not do something like have this function return `col[::-1]`? That way we can be pretty confident that modin is really applying the function across the full axis. I think testing the intermediate representations during the function call is outside the scope of this particular test case.

modin/core/execution/dask/implementations/pandas_on_dask/partitioning/virtual_partition.py
RehanSD: It is more condensed, but I think the comment is helpful from a code clarity perspective.

modin/core/dataframe/pandas/partitioning/partition_manager.py
RehanSD: The comment above kind of explains, but basically, if we have very, very small partitions, we may want to coalesce more than we actually have - e.g. if we have rebalanced all but the last three partitions, and those three alone are not enough to make a new partition, we would run off the end of the list before we've satisfied the min constraint for the length of the new partition, and hit this case.

modin/core/execution/dask/implementations/pandas_on_dask/dataframe/dataframe.py
RehanSD: The brackets are for a list comprehension.

mvashishtha (dismissed):
LGTM!

mvashishtha (approved):

modin/test/storage_formats/pandas/test_internals.py
RehanSD: I could do that, but then I would have to go into each partition to determine what the last row is and then double check the output, so this feels "less intrusive" open to other suggestions though! Perhaps the solution may be to add unit testing for partition layer stuff into the codebase like @pyrito suggested, and move this code to there, so that `test_internals` just tests the overall internals as its supposed to, and implementation details can be tested at that level?

modin/core/execution/dask/implementations/pandas_on_dask/dataframe/dataframe.py
```diff
@@ -41,22 +42,63 @@ class PandasOnDaskDataframe(PandasDataframe):
...
+            partition.list_of_partitions_to_combine[0]
+            .apply(lambda df: len(df) if not axis else (len(df.columns)))
+            ._data
+        ]
+
     @property
     def _row_lengths(self):
         """
-        Compute the row partitions lengths if they are not cached.
+        Compute ther row partitions lengths if they are not cached.
```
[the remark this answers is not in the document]
YarShev: not fixed

modin/core/dataframe/pandas/partitioning/partition_manager.py
```diff
@@ -1302,6 +1314,103 @@ def rebalance_partitions(cls, partitions):
...
+            # to avoid materializing any unmaterialized lengths.
+            if any(
+                partition._length_cache is None
+                for row in partitions
+                for partition in row
+            ):
+                # We need each partition to go into an axis partition, but the
+                # number of axis partitions may not evenly divide the number of
+                # partitions.
+                chunk_size = compute_chunksize(
```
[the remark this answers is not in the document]
RehanSD: Chunk size tells us how many of the old partitions go into one new partition - its basically `num_existing_partitions/ideal_num_new_partitions `

modin/core/dataframe/pandas/partitioning/partition_manager.py
```diff
@@ -1302,6 +1314,103 @@ def rebalance_partitions(cls, partitions):
...
+                        partition_size += partitions[stop][0].length()
+                # If the new partition is larger than we want, split the last
+                # current partition that it contains into two partitions, where
+                # the first partition has just enough rows to make the current
+                # new partition have length `ideal_partition_size`, and the second
+                # partition has the remainder.
+                if partition_size > ideal_partition_size * max_excess_of_num_partitions:
+                    new_last_partition_size = ideal_partition_size - sum(
+                        row[0].length() for row in partitions[start:stop]
+                    )
```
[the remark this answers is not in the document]
RehanSD: That would be a negative number - basically, we only enter this if case if the sum of the lengths of the partitions in partitions[start:stop] is less than the ideal partition size we want - so we're trying to see how many rows/cols we want to slice from this new partition to add to our partitions[start:stop] to get the sum of the lengths of partitions[start:stop] + slice of current partition equal to ideal_partition_size.

modin/core/execution/dask/implementations/pandas_on_dask/partitioning/virtual_partition.py
```diff
@@ -24,35 +26,120 @@
...
+        else:
+            self.list_of_partitions_to_combine = [
+                obj.force_materialization().list_of_partitions_to_combine[0]
+                if isinstance(obj, PandasOnDaskDataframeVirtualPartition)
+                else obj
+                for obj in list_of_blocks
+            ]
 
     partition_type = PandasOnDaskDataframePartition
     instance_type = Future
```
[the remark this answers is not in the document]
YarShev: Either way of these is fine to me

PR #4427, opened 2022-04-26
PERF-#4711: avoid `to_pandas` call when calling left join

### LINKED ISSUE #4711: PERF: avoid `to_pandas` call when calling left join

anmyachev:

## What do these changes do?

Key idea: the current implementation for left join can be replaced with a combination of `right.reindex` + `concat` functions.
This approach can be extended to all types of join.

<details>
<summary>
Benchmark (>2x gain on my machine with 8 cores, compared with master, both for Ray and Dask):
</summary>

```python
import modin.pandas as pd
#import pandas as pd
import pandas
import numpy as np
from time import time

count_rows = 1_000_000
count_cols = 64

arrays = [
    np.array(["bar"]*count_rows),
    np.array([f"ind{x}" for x in range(count_rows)]),
]
index1 = pandas.MultiIndex.from_arrays(arrays, names=["name1", "name2"])

data = random_state.randn(count_rows, count_cols)
df1 = pd.DataFrame(data, index=index1)

random_state.shuffle(arrays[1])

index2 = pandas.MultiIndex.from_arrays(arrays, names=["name1", "name2"])
df2 = pd.DataFrame(data, index=index2, columns=[f"col{x}" for x in range(count_cols)])

df2.drop(f"ind{count_rows-1}", axis=0, level=1, inplace=True)

for _ in range(3):
    start = time()
    res = df1.join(df2, on=["name1", "name2"])
    # partitions = res._query_compiler._modin_frame._partitions
    # all(map(lambda partition: partition.wait() or True, partitions.flatten()))
    print(f"join time: {time()-start}\n")
```
</details>

- [x] Resolves #4711 
- [x] tests added and passing


modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -931,7 +931,19 @@ def _reorder_labels(self, row_positions=None, col_positions=None):
...
+            row_lengths = self._row_lengths
+        if col_positions is None or len(self.columns) == len(col_positions):
+            column_widths = self._column_widths
+        return self.__constructor__(
+            ordered_cols,
+            row_idx,
+            col_idx,
+            row_lengths=row_lengths,
+            column_widths=column_widths,
+        )
```
anmyachev, line 946: Allows us to speed up `getitem_array` function (`->mask->_reoder_labels`).

modin/core/storage_formats/pandas/query_compiler.py
```diff
@@ -512,11 +564,24 @@ def map_func(left, right=right, kwargs=kwargs):
...
+        actual_index = None
+        if "_reset_index" in kwargs:
+            actual_index = kwargs.pop("_reset_index")
+
+        def _reindex(df):
+            df = df.reindex(labels=labels, axis=axis, **kwargs)
+            if actual_index is not None:
+                df.index = actual_index
+            return df
+
```
anmyachev, line 576: Reindexing for operation `join` needs to be done in a special way. First, reindexing is performed by the index obtained from the names in `on` variable, and then it needs to immediately update the internal index of right operand to the one used for the left operand (this will help to avoid expensive index synchronization)

modin/core/storage_formats/pandas/query_compiler.py
```diff
@@ -493,15 +493,97 @@ def join(self, right, **kwargs):
...
+                    if indexer is not None:
+                        # `take` doesn't work with indexer, where `-1` value can be
+                        new_index = self.index.take(indexer)
+                    else:
+                        new_index = self.index
+                    new_self = self.reindex(axis=0, labels=new_index)
+                    right.index = new_self.index
+                    new_self = new_self.concat(1, right, join="left")
+            else:
+                right = right.to_pandas()
```
jbrockmendel, line 579: is there any way to avoid this to_pandas?

modin/core/storage_formats/pandas/query_compiler.py
```diff
@@ -493,15 +493,97 @@ def join(self, right, **kwargs):
...
+                if len(on_in_index) == len(on) == len(index.names):
+                    # fast path
+                    new_index = index
+                else:
+                    if not on_in_index.empty:
+                        frame1 = index.to_frame()[on_in_index]
+                    if not on_in_columns.empty:
+                        frame2 = query_compiler.getitem_array(on_in_columns).to_pandas()
+                    if frame1 is not None and frame2 is not None:
+                        frame = pandas.concat([frame1, frame2], axis=1)
```
jbrockmendel, line 512: might save some with copy=False here

modin/core/storage_formats/pandas/query_compiler.py
```diff
@@ -512,11 +594,24 @@ def map_func(left, right=right, kwargs=kwargs):
     def reindex(self, axis, labels, **kwargs):
         new_index = self.index if axis else labels
         new_columns = labels if axis else self.columns
+        actual_index = None
+        if "_reset_index" in kwargs:
+            actual_index = kwargs.pop("_reset_index")
+
+        def _reindex(df):
+            df = df.reindex(labels=labels, axis=axis, **kwargs)
```
jbrockmendel, line 602: pandas.DataFrame.reindex can take copy=False, can we do that here?

modin/core/storage_formats/pandas/query_compiler.py
anmyachev: I suppose that yes, but there is out of scope of the PR.

modin/core/storage_formats/pandas/query_compiler.py
anmyachev: let's try

modin/core/storage_formats/pandas/query_compiler.py
```diff
@@ -514,11 +566,25 @@ def map_func(left, right=right, kwargs=kwargs):
     def reindex(self, axis, labels, **kwargs):
         new_index = self.index if axis else labels
         new_columns = labels if axis else self.columns
+        actual_index = None
+        if "_reset_index" in kwargs:
```
jbrockmendel, line 570: how does this happen?

modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -1893,6 +1894,12 @@ def apply_full_axis(
             The data types of the result. This is an optimization
             because there are functions that always result in a particular data
             type, and allows us to avoid (re)computing it.
+        sync_axes : bool, default: True
```
jbrockmendel, line 1897: maybe check_axes_sync?  make it obvious that setting to False means skipping a check?

modin/core/storage_formats/pandas/query_compiler.py
```diff
@@ -514,11 +566,25 @@ def map_func(left, right=right, kwargs=kwargs):
...
         new_modin_frame = self._modin_frame.apply_full_axis(
             axis,
-            lambda df: df.reindex(labels=labels, axis=axis, **kwargs),
-            new_index=new_index,
+            _reindex,
+            new_index=actual_index
+            if actual_index is not None and axis == 0
+            else new_index,
             new_columns=new_columns,
+            sync_axes=False,
```
anmyachev, line 587:
```suggestion
            check_axes_sync=False,
```

modin/core/dataframe/pandas/dataframe/dataframe.py
anmyachev, line 1897:
```suggestion
        check_axes_sync : bool, default: True
```

modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -1873,6 +1873,7 @@ def apply_full_axis(
         new_index=None,
         new_columns=None,
         dtypes=None,
+        sync_axes=True,
```
anmyachev, line 1876:
```suggestion
        check_axes_sync=True,
```

modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -2310,6 +2319,12 @@ def broadcast_apply_full_axis(
             Data types of the result. This is an optimization
             because there are functions that always result in a particular data
             type, and allows us to avoid (re)computing it.
+        sync_axes : bool, default: True
```
anmyachev, line 2322:
```suggestion
        check_axes_sync : bool, default: True
```

modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -2356,9 +2371,10 @@ def broadcast_apply_full_axis(
             None,
             dtypes,
         )
-        if new_index is not None:
+        # There is can be case where synchronization is not needed
+        if sync_axes and new_index is not None:
```
anmyachev, line 2375:
```suggestion
        if check_axes_sync and new_index is not None:
```

modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -2356,9 +2371,10 @@ def broadcast_apply_full_axis(
...
             result.synchronize_labels(axis=0)
-        if new_columns is not None:
+        if sync_axes and new_columns is not None:
```
anmyachev, line 2377:
```suggestion
        if check_axes_sync and new_columns is not None:
```

YarShev:
> `>2x gain on my machine with 8 cores, compared with master, both for Ray and Dask`

@anmyachev, are we still slower that pandas on that bench?

anmyachev:
> > `>2x gain on my machine with 8 cores, compared with master, both for Ray and Dask`
> 
> @anmyachev, are we still slower than pandas on that bench?

Yes, more than 3x slower

modin/core/storage_formats/pandas/query_compiler.py
anmyachev: We have to force the index of the right operand to be overwritten by the index of the left operand in order for the dataframe to be correct after the `concat` operation.

modin/core/storage_formats/pandas/query_compiler.py
```diff
@@ -514,11 +573,25 @@ def map_func(left, right=right, kwargs=kwargs):
     def reindex(self, axis, labels, **kwargs):
         new_index = self.index if axis else labels
         new_columns = labels if axis else self.columns
+        actual_index = None
+        if "_reset_index" in kwargs:
+            actual_index = kwargs.pop("_reset_index")
+
+        def _reindex(df):
+            kwargs["copy"] = False
```
jbrockmendel, line 581: this changes kwargs, which is probably benign but not good practice. can you just pass copy to df.reindex

jbrockmendel:
Looking at this, isnt The Right Way to handle this to copy pandas' _MergeOperation code and adapt it?

anmyachev:
> Looking at this, isnt The Right Way to handle this to copy pandas' _MergeOperation code and adapt it?

think so, im just not familiar with this code. will look

vnlitvinov:
haven't read the diff yet, but the bench seems to _not_ be waiting for the computations to finish - are you running it under benchmark mode?..

modin/pandas/test/dataframe/test_join_sort.py
```diff
@@ -158,6 +158,100 @@ def test_join(test_data, test_data2):
...
+        eval_general(
+            modin_df1,
+            pandas_df1,
+            lambda df: df.join(
+                modin_df2.set_index(right_columns[:2])
+                if isinstance(df, pd.DataFrame)
+                else pandas_df2.set_index(right_columns[:2]),
+                on=on,
+            ),
+        )
```
vnlitvinov, line 183: maybe something like
```python
def _get_other(df):
    return modin_df2 if isinstance(df, pd.DataFrame) else pandas_df2

eval_general(
    modin_df1,
    pandas_df1,
    lambda df, other: df.join(other, on=on),
    other=_get_other,
)
```
(and use this `other=_get_other` in all `eval_general()`?)

jbrockmendel:
>> Looking at this, isnt The Right Way to handle this to copy pandas' _MergeOperation code and adapt it?

> think so, im just not familiar with this code. will look

I'll get started on this tomorrow.  I'm getting nowhere on making the lower-level stuff not-painfully-slow.

anmyachev:
Too outdated

Issue #3804, opened 2021-12-07
`df_equals` testing util does not check that dtypes are equal


RehanSD:
### System information
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**:
- **Modin version** (`modin.__version__`):
- **Python version**:
- **Code we can use to reproduce**:

### Describe the problem

The `df_equals` testing utility which compares two dataframes to ensure equality does not check dtypes - allowing cases like the following to succeed, when they should error out:
```python
>>> pandas_df
    A  B
0  1  2
1  3  4
>>> modin_df
     A    B
0  1.0  2.0
1  3.0  4.0
>>> df_equals(modin_df, pandas_df)
# Should raise assertion error, but passes instead
```

devin-petersohn:
@Garra1980 I think this needs to be high priority. We will need a lot of efforts to correct some of these.

Garra1980:
@devin-petersohn  agree, let's try to cleanup everything in the upcoming release

Garra1980:
we can create or convert this issue into epic and track all problems there

devin-petersohn:
Let's make this an epic

PR #4437, opened 2022-05-03
FIX-#4436: Fix to_pydatetime dtype for timezone None.

### LINKED ISSUE #4436: call to_pydatetime get datetime64[ns] dtype
### System information
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: Win10 x64
- **Modin version** (`modin.__version__`):   modin- 0.11.3 or modin-0.14.0+27.g830e8dc
- **Python version**: 3.8.8

### Describe the problem

call a datetime column dt attribute to_pydatetime should get python datetime but  return datetime64[ns] dtype data.
to_pydatetime only worked on DateTimeIndex

### Source code / logs

```
import modin.pandas as pd

df = pd.DataFrame(pd.date_range('2022-01-01', '2022-01-10', freq='1h'), columns=['date'])

pydates = df.date.dt.to_pydatetime()

pydates.dtype
```

the result gives dtype('<M8[ns]') instead of dtype('O').

while DatetimeIndex worked as expected
```dateindx = pd.date_range('2022-01-01', '2022-01-10', freq='1h')

pydates2 = dateindx.to_pydatetime()

pydates2.dtype

```
The result is dtype('O')

mvashishtha:

## What do these changes do?

Fix to_pydatetime dtype for timezone None.

- [x] Resolves #4436 
- [x] tests added and passing

### DIFF
--- a/modin/core/dataframe/algebra/default2pandas/default.py
+++ b/modin/core/dataframe/algebra/default2pandas/default.py
@@ -78,17 +78,26 @@ class DefaultMethod(Operator):
             """
             df = cls.frame_wrapper(df)
             result = fn(df, *args, **kwargs)
-
             if (
                 not isinstance(result, pandas.Series)
                 and not isinstance(result, pandas.DataFrame)
                 and func != "to_numpy"
                 and func != pandas.DataFrame.to_numpy
             ):
+                # When applying a DatetimeProperties function, if we don't
+                # specify the dtype for the DataFrame, the frame might get the
+                # wrong dtype, e.g. for to_pydatetime in
+                # https://github.com/modin-project/modin/issues/4436
+                astype_kwargs = {}
+                dtype = getattr(result, "dtype", None)
+                if dtype and isinstance(
+                    df, pandas.core.indexes.accessors.DatetimeProperties
+                ):
+                    astype_kwargs["dtype"] = dtype
                 result = (
-                    pandas.DataFrame(result)
+                    pandas.DataFrame(result, **astype_kwargs)
                     if is_list_like(result)
-                    else pandas.DataFrame([result])
+                    else pandas.DataFrame([result], **astype_kwargs)
                 )
             if isinstance(result, pandas.Series):
                 if result.name is None:
--- a/modin/core/dataframe/algebra/default2pandas/default.py
+++ b/modin/core/dataframe/algebra/default2pandas/default.py
@@ -78,6 +78,7 @@ class DefaultMethod(Operator):
             """
             df = cls.frame_wrapper(df)
             result = fn(df, *args, **kwargs)
+
             if (
                 not isinstance(result, pandas.Series)
                 and not isinstance(result, pandas.DataFrame)
--- a/modin/pandas/test/test_series.py
+++ b/modin/pandas/test/test_series.py
@@ -1586,8 +1586,14 @@ def test_dtype(data):
     df_equals(modin_series.dtype, pandas_series.dtypes)
 
 
-def test_dt():
-    data = pd.date_range("2016-12-31", periods=128, freq="D", tz="Europe/Berlin")
+# Bug https://github.com/modin-project/modin/issues/4436 in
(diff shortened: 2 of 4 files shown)


mvashishtha:
@modin-project/modin-contributors why is flake8 failing with all these prints I didn't add? https://github.com/modin-project/modin/runs/6278631418?check_suite_focus=true

RehanSD (changes requested):
LGTM! The only thing is I would explicitly compare the dtypes, since `df_equals` doesn't actually do that. (https://github.com/modin-project/modin/issues/3804). Should we also add tests to ensure that the date time conversion on index doesn't regress?

mvashishtha:
@RehanSD I added another fix for the Python engine. Please take another look.

modin/pandas/test/test_series.py
```diff
@@ -1627,10 +1633,11 @@ def test_dt():
         modin_series.dt.tz_localize(None),
         pandas_series.dt.tz_localize(None),
     )
-    df_equals(
-        modin_series.dt.tz_convert(tz="Europe/Berlin"),
-        pandas_series.dt.tz_convert(tz="Europe/Berlin"),
-    )
+    if timezone:
+        df_equals(
+            modin_series.dt.tz_convert(tz="Europe/Berlin"),
```
RehanSD, line 1638: I don't think `df_equals` checks dtypes - you may want to manually double check that the dtypes are equal.

RehanSD (commented):
left some quick questions

modin/core/dataframe/algebra/default2pandas/default.py
```diff
@@ -78,17 +78,26 @@ def applyier(df, *args, **kwargs):
             """
             df = cls.frame_wrapper(df)
             result = fn(df, *args, **kwargs)
-
```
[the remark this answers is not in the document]
mvashishtha: No, I reverted it.

modin/pandas/test/test_series.py
mvashishtha: There's no bug with dtypes in `tz_convert`, so we don't need to check the dtypes here. I added a `timezone` param with the new value `None` to the entire unit test case and `tz_convert` doesn't work unless `timezone` is something other than `None`. I had to change this part of the test case so it only checked `tz_convert` when `timezone` was not None.

modin/pandas/test/test_series.py
```diff
@@ -1620,8 +1620,14 @@ def test_dtype(data):
...
-    data = pd.date_range("2016-12-31", periods=128, freq="D", tz="Europe/Berlin")
+# Bug https://github.com/modin-project/modin/issues/4436 in
+# Series.dt.to_pydatetime is only reproducible when the date range out of which
+# the frame is created has timezone None, so that its dtype is datetime64[ns]
+# as opposed to, e.g. datetime64[ns, Europe/Berlin]. To reproduce that bug, we
+# use timezones None and Europe/Berlin.
+@pytest.mark.parametrize("timezone", [None, "Europe/Berlin"])
+def test_dt(timezone):
+    data = pd.date_range("2016-12-31", periods=128, freq="D", tz=timezone)
     modin_series = pd.Series(data)
```
RehanSD, line 1631: I think the tests need to be modified to actually ensure that the dtypes are correct. As I mentioned before, `df_equals` doesn't check dtypes by default, so even if we have a clause in the test that should be triggering the bug, we don't actually confirm that the bugfix works since we don't check the output dtypes.

RehanSD (changes requested):

modin/pandas/test/test_series.py
RehanSD: That makes sense, thank you!

RehanSD (approved):
LGTM!

RehanSD:
This PR relates slightly to issue #3804 - we should potentially revisit this PR in that context in the future.

PR #4451, opened 2022-05-10
FIX-#4450: Ensure Modin successfully initializes when Ray cluster has no resources

### LINKED ISSUE #4450: Modin is unable to handle ray cluster scaling from 0
### System information
- **Python version**:

### Describe the problem

When we have a ray cluster that autoscales from 0, the `ray.cluster_resources()` dictionary does not initially include any CPU resources, and so Modin errors out. We should probably catch this case, use a default, and warn the user to update `NPartitions` appropriately.

RehanSD:

## What do these changes do?
This patch ensures that Modin can initialize successfully even when the ray cluster has no resources (because it is autoscaling from 0).

- [] passes `flake8 modin/ asv_bench/benchmarks scripts/doc_checker.py`
- [ ] Resolves #4450 ? 
- [ ] tests added and passing

### DIFF
--- a/modin/core/execution/ray/common/utils.py
+++ b/modin/core/execution/ray/common/utils.py
@@ -217,7 +217,13 @@ def initialize_ray(
         _move_stdlib_ahead_of_site_packages
     )
     ray.worker.global_worker.run_function_on_all_workers(_import_pandas)
-    num_cpus = int(ray.cluster_resources()["CPU"])
+    num_cpus = ray.cluster_resources().get("CPU", None)
+    if num_cpus is None:
+        import warnings
+        warnings.warn("The current Ray cluster does not have any CPU Resources.\nModin uses the number of CPUs to determine how many partitions to create.\nNumber of partitions defaulting to 4. To update, run the following python code:\n\tfrom modin.config import NPartitions\n\tNPartitions.put(desired_num_cpus)")
+        num_cpus = 4
+    else:
+        num_cpus = int(num_cpus)
     num_gpus = int(ray.cluster_resources().get("GPU", 0))
     if StorageFormat.get() == "Cudf":
         NPartitions._put(num_gpus)


RehanSD:
While this patch does fix the issue, I think it would be helpful to have a discussion around how we want Modin to behave in such scenarios (i.e. whether it should just use some number, say 16, for the number of partitions until the user specifies how many partitions they want, whether it should ask the user to set the number of partitions manually, and not set num partitions, or whether it should launch ray tasks to cause the autoscaler to spin up workers, and then get the number of CPUs).

modin/core/execution/ray/common/utils.py
```diff
@@ -217,7 +217,17 @@ def initialize_ray(
...
-    num_cpus = int(ray.cluster_resources()["CPU"])
+    num_cpus = ray.cluster_resources().get("CPU", None)
+    if num_cpus is None:
+        warnings.warn(
+            "The current Ray cluster does not have any CPU Resources.\nModin uses the number of "
+            + "CPUs to determine how many partitions to create.\nNumber of partitions defaulting to"
+            + " 4. To update, run the following python code:\n\tfrom modin.config import "
+            + "NPartitions\n\tNPartitions.put(desired_num_cpus)"
+        )
+        num_cpus = 4
```
YarShev, line 228: I think we should add an inline comment why we set `num_cpus=4` by default. I remember there was a question why we did that and whether we should have changed it. Also, it would be good to reflect this in the docs somewhere.

YarShev: Btw, is such kind of a scenario applicable to Dask?

modin/core/execution/ray/common/utils.py
```diff
@@ -217,7 +217,17 @@ def initialize_ray(
         _move_stdlib_ahead_of_site_packages
     )
     ray.worker.global_worker.run_function_on_all_workers(_import_pandas)
-    num_cpus = int(ray.cluster_resources()["CPU"])
+    num_cpus = ray.cluster_resources().get("CPU", None)
```
devin-petersohn, line 220:
```suggestion
    num_cpus = ray.cluster_resources().get("CPU", None) or 4
```

modin/core/execution/ray/common/utils.py
devin-petersohn, line 228:
```suggestion
```

modin/core/execution/ray/common/utils.py
```diff
@@ -217,7 +217,17 @@ def initialize_ray(
...
+    if num_cpus is None:
+        warnings.warn(
+            "The current Ray cluster does not have any CPU Resources.\nModin uses the number of "
+            + "CPUs to determine how many partitions to create.\nNumber of partitions defaulting to"
+            + " 4. To update, run the following python code:\n\tfrom modin.config import "
+            + "NPartitions\n\tNPartitions.put(desired_num_cpus)"
+        )
+        num_cpus = 4
+    else:
```
devin-petersohn, line 229:
```suggestion
```

modin/core/execution/ray/common/utils.py
```diff
@@ -217,7 +217,17 @@ def initialize_ray(
...
+        num_cpus = int(num_cpus)
```
devin-petersohn, line 230:
```suggestion
    num_cpus = int(num_cpus)
```

modin/core/execution/ray/common/utils.py
devin-petersohn: +1 we need documentation.

I think this might also be applicable to Dask, but we'd need to test it.

RehanSD: I'm happy to add a comment here! Before that: Is 4 a good heuristic? I picked it bc most personal laptops have 4 cores/it seemed like a good number, but I'd prefer to have a better reason behind how we treat this case, so I'd appreciate y'all's insight into how we can better handle this case!

modin/core/execution/ray/common/utils.py
RehanSD: The reason I did it like this is because I wanted to warn users that we were unable to determine how many `CPU`s there were and were relying on an internal heuristic - should we keep it this way, or change it to the more streamlined version you propose?

devin-petersohn (commented):
I'm not sure setting the value to 4 is actually right here either. Should we throw an exception asking the user to specify the `NPartitions` value?

RehanSD:
I double checked dask, and it uses `CpuCount.get` which should work even if its an autoscaling cluster!

mvashishtha (approved):
LGTM!

modin/core/execution/ray/common/utils.py
```diff
@@ -217,7 +217,17 @@ def initialize_ray(
         _move_stdlib_ahead_of_site_packages
     )
     ray.worker.global_worker.run_function_on_all_workers(_import_pandas)
-    num_cpus = int(ray.cluster_resources()["CPU"])
+    num_cpus = ray.cluster_resources().get("CPU", None)
+    if num_cpus is None:
+        warnings.warn(
+            "The current Ray cluster does not have any CPU Resources.\nModin uses the number of "
+            + "CPUs to determine how many partitions to create.\nNumber of partitions defaulting to"
+            + " number of CPUs on head node. To update, run the following python code:\n\tfrom "
```
jeffreykennethli, line 225: nit: could you instead instantiate `num_cpus` above this warning, and add the actual value of `num_cpus` to the warning message? such as, with an f-string.

jeffreykennethli:
> While this patch does fix the issue, I think it would be helpful to have a discussion around how we want Modin to behave in such scenarios (i.e. whether it should just use some number, say 16, for the number of partitions until the user specifies how many partitions they want, whether it should ask the user to set the number of partitions manually, and not set num partitions, or whether it should launch ray tasks to cause the autoscaler to spin up workers, and then get the number of CPUs).

Is there another (more complicated but better) solution, which is to wait until ray tasks launch, then get the number of CPUs, update the number of partitions, and propagate the change? I think you mentioned in the Discuss forum some challenges with this approach since we don't currently repartition when `NumPartitions` is updated. However, could this be a longer term approach?

In the nearer term, I think the current approach of showing a warning and defaulting to 4 (or some other number) is best for this edge case.

jeffreykennethli:
Another option, in contrast to surfacing a warning and setting a default, is to raise an Error with an error message that gives instructions to bypass the error with a config variable to ignore the error, or to set the default `NPartitions` manually. 

The reason this might be better is that users in batch mode won't see a warning, but they will see an error (because the script will fail), and we don't want users to have mysterious behavior with Modin without knowing about it.

PR #4460, opened 2022-05-13
FEAT-#4445: Stop recomputing both indices for axis-wide applies.

### LINKED ISSUE #4445: PERF: Stop recomputing both indices for user-defined and dict-like axis-wide applies
### Describe the problem
We recompute both column and row indices for column-wise and row-wise applies of user-defined callables and dict-like functions. [Here](https://github.com/modin-project/modin/blob/49f47b1ed66583d4b90d3c06faf3e1d457cf4df8/modin/core/storage_formats/pandas/query_compiler.py#L2394) for dict applies and [here](https://github.com/modin-project/modin/blob/49f47b1ed66583d4b90d3c06faf3e1d457cf4df8/modin/core/storage_formats/pandas/query_compiler.py#L2465) for callable applies, we specify neither `new_index` nor `new_columns` so we end up [recomputing](https://github.com/modin-project/modin/blob/685f1b4eb60c3dd6c9127bee93b6c138f890f700/modin/core/dataframe/pandas/dataframe/dataframe.py#L2325) both axes, so we always block on both the first row of partitions and the first column of partitions. You can observe this unnecessary blocking here:

```python
import modin.pandas as pd
import numpy as np
import time
from modin.config import MinPartitionSize

num_columns = MinPartitionSize.get() + 1

# 3 rows where each row has the numbers from 0 through num_columns 
# exclusive in sequential order
# so the entire frame has a single row of two partitions,
# where the first has num_columns - 1 columns and the second has one.
df = pd.DataFrame(np.tile(np.arange(num_columns), (3, 1)))

# This takes hours for the the column containing 32,
# but should return much faster for the other columns.
def col_func(col):
    if col[0] == num_columns - 1:
        time.sleep(10000)
    return col * 2

# This blocks on the last column to recompute
# the column index, so it blocks 10000 seconds.
# instead, it should finish immediately while the last 
# apply completes asynchronously.
print('starting apply...')
result = df.apply(col_func)
```

mvashishtha:

## What do these changes do?

Stop recomputing both indices for axis-wide applies. These changes make the slow computation in the snippet from https://github.com/modin-project/modin/issues/4445#issue-1228486253 asynchronous.

- [x] Resolves #4445
- [ ] tests added and passing


modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -1864,6 +1864,7 @@ def apply_full_axis(
         new_index=None,
         new_columns=None,
         dtypes=None,
+        func_may_change_complementary_index_size=True,
```
YarShev, line 1867: What if we would pass new row lengths and new column widths from the QC?

YarShev: The QC probably shouldn't know anything about this thought. However, would do you thing on this?

cc @mvashishtha, @devin-petersohn

devin-petersohn (commented):
@mvashishtha I like the idea of splitting the functionality like this, but maybe we need a new operator instead of handling it like this?

mvashishtha:
> @mvashishtha I like the idea of splitting the functionality like this, but maybe we need a new operator instead of handling it like this?

The difference between the new operator and the current `apply_full_axis` seems small. I have a slight preference for making the change this way. The extra code is only 7 lines. @YarShev what do you think?

YarShev:
If we do not want to pass lengths from the QC to the core dataframe, I would suggest renaming `func_may_change_complementary_index_size` to `preserve_opposite_lengths` and passing it `apply_full_axis` to `broadcast_apply_full_axis`, where we would handle it.

devin-petersohn:
@mvashishtha For the algebra, typically there are two things that qualify an operator:
* Unique handling of metadata
* Unique handling of data

We typically want each operator to have certain guarantees without requiring too much information from upper layers. This is why I believe it should be a different operator, since both metadata and data can change at the same time, which isn't something the current operators explicitly support.

mvashishtha:
@devin-petersohn The new param I introduced to `apply_full_axis` in this PR seems nearly identical in purpose to `preserve_labels` in `broadcast_apply`. Should `broadcast_apply` also be split into two methods, one of which assumes labels change, and the other of which assumes they do not? https://github.com/modin-project/modin/blob/cca9468648521e9317de1cb69cf8e6b1d5292d21/modin/core/dataframe/pandas/dataframe/dataframe.py#L2101-L2102

devin-petersohn:
@mvashishtha for now, let's keep this fix local to the axis-wide apply. we should revisit after.

PR #4468, opened 2022-05-18
FIX-#4467: 'drop_duplicates' no longer removes items based on index value

### LINKED ISSUE #4467: drop_duplicates after pd.concat not working as expected
### System information
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: MacOS 12.3.1
- **Modin version** (`modin.__version__`): 0.14.1
- **Python version**: Python 3.9.12

```python
import pandas as pd
import modin.pandas as mpd
import numpy as np

# Pandas
np.random.seed(0)
df1 = pd.DataFrame(np.random.randint(0, 1000, size=1000), columns=["A"])
np.random.seed(1)
df2 = pd.DataFrame(np.random.randint(0, 1000, size=1000), columns=["A"])

df = pd.concat([df1, df2])
print(len(df))  # 2000
print(len(df.drop_duplicates()))  # 874

# Modin
np.random.seed(0)
df1 = mpd.DataFrame(np.random.randint(0, 1000, size=1000), columns=["A"])
np.random.seed(1)
df2 = mpd.DataFrame(np.random.randint(0, 1000, size=1000), columns=["A"])

df = mpd.concat([df1, df2])
print(len(df))  # 2000
print(len(df.drop_duplicates()))  # 322 ?!?

# Issue caused by index? Use ignore_index=True to fix
df = mpd.concat([df1, df2], ignore_index=True)
print(len(df))  # 2000
print(len(df.drop_duplicates()))  # 874
```

### Describe the problem

After combining DataFrames using `mpd.concat()`, dropping duplicate values removes WAY too many items. I believe this is related to the indexing of the combined DataFrames, as using `ignore_index=True` in the `concat()` call seems to fix the issue.

In regular pandas, the result is the same whether or not `ignore_index=True` is used.

The above example is a toy one. I noticed this when combining large CSV files. Happy to share those, but this takes much less time to run!

jrsacher:
Signed-off-by: Joshua Sacher <joshuasacher@gmail.com>

## What do these changes do?

Fix a case where, if index values were not unique, multiple values could be incorrectly removed when using `drop_duplicates`

- [x] Resolves #4467  
- [ ] tests added and passing

### DIFF
--- a/modin/pandas/base.py
+++ b/modin/pandas/base.py
@@ -54,6 +54,7 @@ from modin.error_message import ErrorMessage
 import modin.pandas as pd
 from modin.pandas.utils import is_scalar
 from modin.config import IsExperimental
+from modin.logging import LoggerMetaClass
 
 # Similar to pandas, sentinel value to use as kwarg in place of None when None has
 # special meaning and needs to be distinguished from a user explicitly passing None.
@@ -99,7 +100,7 @@ _doc_binary_op_kwargs = {"returns": "BasePandasDataset", "left": "BasePandasData
 
 
 @_inherit_docstrings(pandas.DataFrame, apilink=["pandas.DataFrame", "pandas.Series"])
-class BasePandasDataset(object):
+class BasePandasDataset(object, metaclass=LoggerMetaClass):
     """
     Implement most of the common code that exists in DataFrame/Series.
 
--- a/modin/pandas/base.py
+++ b/modin/pandas/base.py
@@ -1400,7 +1400,7 @@ class BasePandasDataset(object):
 
         result = self[-duplicates]
         if ignore_index:
-            result.index = list(range(len(result)))
+            result.index = pandas.RangeIndex(stop=len(result))
         if inplace:
             self._update_inplace(result)
             return None
--- a/modin/pandas/base.py
+++ b/modin/pandas/base.py
@@ -1397,13 +1397,11 @@ class BasePandasDataset(object):
             duplicates = self.duplicated(keep=keep, subset=subset)
         else:
             duplicates = self.duplicated(keep=keep)
-
-        result = self[-duplicates]
+        result = self[~duplicates]
         if ignore_index:
             result.index = pandas.RangeIndex(stop=len(result))
         if inplace:
             self._update_inplace(result)
-            return None
         else:
             return result
 
--- a/modin/pandas/base.py
+++ b/modin/pandas/base.py
@@ -1401,7 +1401,7 @@ class BasePandasDataset(object):
         if ignore_index:
             result.index = pandas.RangeIndex(stop=len(result))
         if inplace:
-            self._update_inplace(result)
+            self._update_inplace(result._query_compiler)
         else:
             return result
 
--- a/modin/pandas/base.py
+++ b/modin/pandas/base.py
@@ -1387,6 +1387,7 @@ class BasePandasDataset(object):
         """
         inplace = validate_bool_kwarg(inplace, "inplace")
         subset = kwargs.get("subset", None)
+        ignore_index = kwargs.get("ignore_index", False)
         if subset is not None:
             if is_list_like(subset):
                 if not isinstance(subset, list):
@@ -1396,8 +1397,15 @@ class BasePandasDataset(object):
             duplicates = self.duplicated(keep=keep, subset=subset)
         else:
             duplicates = self.duplicated(keep=keep)
-        indices = duplicates.values.nonzero()[0]
-        return self.drop(index=self.index[indices], inplace=inplace)
+
+        result = self[-duplicates]
+        if ignore_index:
+            result.index = list(range(len(result)))
+        if inplace:
+            self._update_inplace(result)
+            return None
+        else:
+            return result
 
     def eq(self, other, axis="columns", level=None):  # noqa: PR01, RT01, D200
         """
--- a/modin/pandas/test/dataframe/test_map_metadata.py
+++ b/modin/pandas/test/dataframe/test_map_metadata.py
@@ -746,30 +746,43 @@ def test_droplevel():
     [None, "col1", "name", ("col1", "col3"), ["col1", "col3", "col7"]],
     ids=["None", "string", "name", "tuple", "list"],
 )
-def test_drop_duplicates(data, keep, subset):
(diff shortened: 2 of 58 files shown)


jrsacher:
A test could (and should?) be added, but I'm not familiar with how to write them. If I have time to do some research, I can try to come up with something.

modin/pandas/base.py
```diff
@@ -1396,8 +1397,15 @@ def drop_duplicates(
             duplicates = self.duplicated(keep=keep, subset=subset)
         else:
             duplicates = self.duplicated(keep=keep)
-        indices = duplicates.values.nonzero()[0]
-        return self.drop(index=self.index[indices], inplace=inplace)
+
+        result = self[-duplicates]
+        if ignore_index:
+            result.index = list(range(len(result)))
```
jrsacher, line 1403: `list(range(len(results)))` doesn't feel very elegant.
Maybe `pandas.RangeIndex(stop=len(result))` instead? This may better match what Pandas does.

modin/pandas/base.py
```diff
@@ -1387,6 +1387,7 @@ def drop_duplicates(
         """
         inplace = validate_bool_kwarg(inplace, "inplace")
         subset = kwargs.get("subset", None)
+        ignore_index = kwargs.get("ignore_index", False)
```
jrsacher, line 1390: is it preferable to pull from `kwargs` or to add an argument and default value to the function?

modin/pandas/base.py
jrsacher: yeah, I decided I liked that better.

devin-petersohn (commented):
Hi @jrsacher, thanks for the PR, this was a great catch! 

A couple of high level points:
- DataFrame isn't passing in `ignore_index`, which is another piece of the bug. see: https://github.com/modin-project/modin/blob/cf0eb393daa41abdd8cdf32b52ceee938cdcbe1a/modin/pandas/dataframe.py#L285
- The test can be added to the existing test by adding a `pytest.mark.parametrize` for `ignore_index` here: https://github.com/modin-project/modin/blob/cf0eb393daa41abdd8cdf32b52ceee938cdcbe1a/modin/pandas/test/dataframe/test_map_metadata.py#L738-L773

This is awesome, thanks again for the PR, and happy to help if needed!

modin/pandas/test/dataframe/test_map_metadata.py
```diff
@@ -746,30 +746,43 @@ def test_droplevel():
     [None, "col1", "name", ("col1", "col3"), ["col1", "col3", "col7"]],
     ids=["None", "string", "name", "tuple", "list"],
 )
-def test_drop_duplicates(data, keep, subset):
+@pytest.mark.parametrize("ignore_index", [True, False], ids=["True", "False"])
+def test_drop_duplicates(data, keep, subset, ignore_index):
```
mvashishtha, line 750: While the issue that this PR is fixing reveals a bug with dropping duplicates from a dataframe with a repeated index, none of the frames in `test_data_with_duplicates` has a repeated index. We should test that we are dropping the correct duplicates when the index has duplicates. For Modin right now we have:

```python
import modin.pandas as pd
import pandas

df = pd.DataFrame([[1], [2], [1]], index=[0, 0, 0])
# This gives an empty frame!
print(df.drop_duplicates())

pdf = pandas.DataFrame([[1], [2], [1]], index=[0, 0, 0])
# This drops just the second 1
print(pdf.drop_duplicates())
```

`test_data_with_duplicates_values` has just the data with no columns, so we can't add custom columns there. Could you please add another test case with a small reproducer as above for just this issue, and add a link to the github issue it's testing?

mvashishtha (commented):
@jrsacher thank you for the fix! I am leaving one comment about the tests.

modin/pandas/test/dataframe/test_map_metadata.py
```diff
@@ -866,6 +879,13 @@ def test_drop_duplicates_after_sort():
     df_equals(modin_result, pandas_result)
 
 
+def test_drop_duplicates_with_repeated_index_values():
```
mvashishtha, line 882: Could you please add a link to the GitHub issue?

modin/pandas/test/dataframe/test_map_metadata.py
```diff
@@ -866,6 +879,13 @@ def test_drop_duplicates_after_sort():
...
+    data = {"data": [[0], [1], [0]], "index": [0, 0, 0]}
+    pandas_df = pandas.DataFrame(data["data"], index=data["index"])
+    modin_df = pd.DataFrame(data["data"], index=data["index"])
+    df_equals(modin_df.drop_duplicates(), pandas_df.drop_duplicates())
```
mvashishtha, line 886:
```suggestion
def test_drop_duplicates_with_repeated_index_values():
    data = [[0], [1], [0]]
    index = [0, 0, 0]
    modin_df, pandas_df = create_test_dfs(data, index=index)
    eval_general(modin_df, pandas_df, lambda df: df.drop_duplicates())
```

Applying some modin testing utilities like this should make the test clearer.

mvashishtha (commented):
@jrsacher  Thank you for adding the test case! I have two more minor comments about the test.

modin/pandas/test/dataframe/test_map_metadata.py
jrsacher: I pushed a new commit with the link as a comment. Is that what you're referring to or is there some better way to link the issue?

mvashishtha: That's what I was thinking of. Thank you!

mvashishtha (approved):
LGTM! Thank you for the contribution.

devin-petersohn (commented):
@jrsacher this looks really close to being ready to merge! The test are failing for the Python debugger engine: 
image

@mvashishtha - do you know why that would happen with these changes?

mvashishtha:
> @jrsacher this looks really close to being ready to merge! The test are failing for the Python debugger engine: image
> 
> @mvashishtha - do you know why that would happen with these changes?

@devin-petersohn The cause is that there is a bug in `BaseOnPython` execution for calling `__getitem__` with a Modin series. I'll make a fix for #4482.

devin-petersohn (dismissed):
Awesome work @jrsacher , LGTM!

devin-petersohn (approved):

PR #4472, opened 2022-05-19
FIX-#4449 Drain the call queue before waiting on result in benchmark mode.

### LINKED ISSUE #4449: Benchmark Mode blocks on computation so all computation occurs on one node
### System information
- **Python version**:

### Describe the problem

Since Benchmark Mode calls `wait` on each partition sequentially, computation that can be done in parallel ends up happening sequentially on a single node. We need to instead collect all the OID's of all partitions, and wait on all of them together - that way each of the tasks is kicked off in parallel, and computation happens in parallel.

mvashishtha:
## What do these changes do?

- [x] Resolves #4449 
- [ ] tests added and passing

### DIFF
--- a/modin/core/dataframe/pandas/partitioning/partition_manager.py
+++ b/modin/core/dataframe/pandas/partitioning/partition_manager.py
@@ -60,6 +60,13 @@ def wait_computations_if_benchmark_mode(func):
                 partitions = result[0]
             else:
                 partitions = result
+            # When partitions have a deferred call queue, calling
+            # parittion.wait() on each partition serially will serially kick
+            # off each deferred computation and wait for each partition to
+            # finish before kicking off the next one. Instead, we want to
+            # serially kick off all the deferred computations so that they can
+            # all run aynschronously, then serially wait on all the results.
+            [part.drain_call_queue() for row in partitions for part in row]
             # need to go through all the values of the map iterator
             # since `wait` does not return anything, we need to explicitly add
             # the return `True` value from the lambda
--- a/modin/core/dataframe/pandas/partitioning/partition_manager.py
+++ b/modin/core/dataframe/pandas/partitioning/partition_manager.py
@@ -65,7 +65,7 @@ def wait_computations_if_benchmark_mode(func):
             # off each deferred computation and wait for each partition to
             # finish before kicking off the next one. Instead, we want to
             # serially kick off all the deferred computations so that they can
-            # all run aynschronously, then serially wait on all the results.
+            # all run asynchronously, then wait on all the results.
             [part.drain_call_queue() for part in partitions.flatten()]
             # need to go through all the values of the map iterator
             # since `wait` does not return anything, we need to explicitly add
--- a/modin/core/dataframe/pandas/partitioning/partition_manager.py
+++ b/modin/core/dataframe/pandas/partitioning/partition_manager.py
@@ -70,6 +70,8 @@ def wait_computations_if_benchmark_mode(func):
             # need to go through all the values of the map iterator
             # since `wait` does not return anything, we need to explicitly add
             # the return `True` value from the lambda
+            # TODO(https://github.com/modin-project/modin/issues/4491): Wait
+            # for all the partitions in parallel.
             all(map(lambda partition: partition.wait() or True, partitions.flatten()))
             return result
 
--- a/docs/release_notes/release_notes-0.15.0.rst
+++ b/docs/release_notes/release_notes-0.15.0.rst
@@ -18,6 +18,7 @@ Key Features and Updates
   * FIX-#4394: Fix issue with multiindex metadata desync (#4395)
   * FIX-#4425: Add parameters to groupby pct_change (#4429)
   * FIX-#4461: Fix S3 CSV data path (#4462)
+  * FIX-#4449: Drain the call queue before waiting on result in benchmark mode (#4472)
 * Performance enhancements
   * FEAT-#4320: Add connectorx as an alternative engine for read_sql (#4346)
 * Benchmarking enhancements


modin/core/dataframe/pandas/partitioning/partition_manager.py
```diff
@@ -60,6 +60,13 @@ def wait(*args, **kwargs):
                 partitions = result[0]
             else:
                 partitions = result
+            # When partitions have a deferred call queue, calling
+            # parittion.wait() on each partition serially will serially kick
```
anmyachev, line 64:
```suggestion
            # partition.wait() on each partition serially will serially kick
```

modin/core/dataframe/pandas/partitioning/partition_manager.py
```diff
@@ -60,6 +60,13 @@ def wait(*args, **kwargs):
                 partitions = result[0]
             else:
                 partitions = result
+            # When partitions have a deferred call queue, calling
+            # partition.wait() on each partition serially will serially kick
+            # off each deferred computation and wait for each partition to
+            # finish before kicking off the next one. Instead, we want to
+            # serially kick off all the deferred computations so that they can
+            # all run aynschronously, then serially wait on all the results.
+            [part.drain_call_queue() for row in partitions for part in row]
```
anmyachev, line 69:
```suggestion
            [part.drain_call_queue() for part in partitions.flatten()]
```

anmyachev (dismissed):

anmyachev (approved):

YarShev:
It seems to me that @RehanSD meant something like this when creating the issue.
```python
ray.wait([part.oid for part in partitions], num_returns=len(partitions))
```
@RehanSD, please correct me if I am wrong.

mvashishtha:
> It seems to me that @RehanSD meant something like this when creating the issue.
> 
> ```python
> ray.wait([part.oid for part in partitions], num_returns=len(partitions))
> ```
> 
> @RehanSD, please correct me if I am wrong.

@YarShev that won't fix the bug. e.g. if we have a transpose on the call queue, `ray.wait` will return immediately for each partition's `oid`. We need to drain the call queue before waiting.

YarShev:
@mvashishtha, yes, I understand. I am saying about waiting on partition ids after we drain the call queue.

mvashishtha:
> @mvashishtha, yes, I understand. I am saying about waiting on partition ids after we drain the call queue.

@YarShev what's wrong with waiting the way we are doing now? We are not waiting in parallel, but any remote functions should be executing in parallel anyway. Doing a `ray wait` would let us wait for everything at once, but I don't expect it to save any significant time because the cost of iterating through the partition is so little.

YarShev:
@mvashishtha, I would like to make sure if there is a difference between waiting on each partition separately and on all partitions at once both in a single node and in a cluster. @modin-project/modin-ray, could you shed a light on this?

mvashishtha:
@YarShev Isn't that question separate from the problem this PR is addressing? Draining the call queue is something we have to do serially anyway.

YarShev:
@mvashishtha, I think my question may be related to the problem this PR is addressing. I am wondering if there is a parallel wait on object references in case of a cluster?

wuisawesome:
If I'm understanding this conversation correctly, `ray.wait(refs)` waits on all the refs in parallel. (Note `num_returns=1` by default so we have to wait for all of them to know which one finishes first).

YarShev:
@wuisawesome, that is what I expected. Could you tell us about how the parallel wait works? Does the parallelization happen per workers or per nodes (plasma)?

devin-petersohn (approved):
@mvashishtha Makes sense, I thought this bug was an issue with synchronous execution.

LGTM

YarShev:
Still, I think we should use a parallel wait from Ray (and probably from Dask) to check the performance correctly. We likely want to wait for partitions when all computations are complete in parallel. I guess we could add a method named `wait` to the partition manager where we wait for all partitions to be ready.

mvashishtha:
> Still, I think we should use a parallel wait from Ray (and probably from Dask) to check the performance correctly. We likely want to wait for partitions when all computations are complete in parallel. I guess we could add a method named `wait` to the partition manager where we wait for all partitions to be ready.

@YarShev I still think this is a separate issue that is independent of this one. I have filed #4491 for it. This issue is about deferred tasks that are executed in sequence in benchmark mode, whereas #4491 is about more precisely waiting for all async computations in benchmark mode in general. Let's please keep this PR focused on #4449.

YarShev:
> > Still, I think we should use a parallel wait from Ray (and probably from Dask) to check the performance correctly. We likely want to wait for partitions when all computations are complete in parallel. I guess we could add a method named `wait` to the partition manager where we wait for all partitions to be ready.
> 
> @YarShev I still think this is a separate issue that is independent of this one. I have filed #4491 for it. This issue is about deferred tasks that are executed in sequence in benchmark mode, whereas #4491 is about more precisely waiting for all async computations in benchmark mode in general. Let's please keep this PR focused on #4449.

@mvashishtha, ok

modin/core/dataframe/pandas/partitioning/partition_manager.py
```diff
@@ -60,9 +60,18 @@ def wait(*args, **kwargs):
                 partitions = result[0]
             else:
                 partitions = result
+            # When partitions have a deferred call queue, calling
```
YarShev, line 63: I think this comment should be rephrased taking into account these discussions - [link1](https://github.com/modin-project/modin/pull/4472#issuecomment-1136042874) and [link2](https://github.com/modin-project/modin/pull/4472#issuecomment-1136219274).

mvashishtha: What would you change it to? It seems like as is, it accurately explains why the `drain_call_queue` is there.

YarShev: I read again this and probably would change the last part a bit. Left a suggestion.

modin/core/dataframe/pandas/partitioning/partition_manager.py
```diff
@@ -60,9 +60,18 @@ def wait(*args, **kwargs):
...
+            # partition.wait() on each partition serially will serially kick
+            # off each deferred computation and wait for each partition to
+            # finish before kicking off the next one. Instead, we want to
+            # serially kick off all the deferred computations so that they can
+            # all run aynschronously, then serially wait on all the results.
```
YarShev, line 68:
```suggestion
            # all run asynchronously, then wait on all the results in parallel.
```

mvashishtha: I wanted to clarify that while the computations are happening in parallel, we wait on them in serial (for now). What about this?

```suggestion
            # all run asynchronously in parallel, then wait serially on all the results.
```

YarShev: The sentence begins with `we want to ...` while actuaclly we do not want to wait serially and will change this in the future. Let's just not use `in parallel` here. Corrected my suggestion.

YarShev (approved):

PR #4473, opened 2022-05-19
FIX-#4464: Fix virtual partition handling args improperly

### LINKED ISSUE #4464: groupby.count fails with virtual partitions
### System information
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: MacOS 12.0.1
- **Modin version** (`modin.__version__`): 0.14
- **Python version**: 3.9.9
```
import modin.pandas as pd
import numpy as np
import ray

ray.init()
data = np.random.randint(0, 100, size=(2**10, 2**4))
df = pd.DataFrame(data) # if we add add_prefix("col") here, the partition types change and this script succeeds
# df.reset_index(inplace=True) also makes this script succeed by changing the partition type.
big_df = pd.concat([df for _ in range(5)])
big_df.groupby(1).count()
```

### Describe the problem
When a DataFrame is created by concatenating multiple DataFrames column-wise, the resulting DataFrame is composed of partitions that are of type `PandasOnRayDataframeColumnPartition`, and `groupby.count` fails. When the DataFrames partitions are of type `PandasOnRayDataframePartition`, the `groupby.count` succeeds. You can inspect this by running the two versions of the above script and checking `big_df._query_compiler._modin_frame._partitions`.

Even more confusingly, this doesn't fail with methods like `groupby.mean`, `groupby.sum`, and when the internal method for groupby.count (`_wrap_aggregation`) is called with `numeric_only=True` instead of `False`, it succeeds. See https://github.com/modin-project/modin/blob/master/modin/pandas/groupby.py#L798-L802.

Opening the issue because I don't understand groupby behavior well enough to debug this.

jeffreykennethli:
## What do these changes do?
**TDLR**: refactor code to better support passing a `func`, `*args` and `**kwargs` because the current way was confusing and led to an error where we were passing in wrong arguments to Virtual Partitions.

Fixes a bug where partitions are called with `*args` from https://github.com/modin-project/modin/blob/c17dde71ae8114723a13279e36bf5532dcba328a/modin/core/dataframe/pandas/partitioning/partition_manager.py#L364-L379 but in the case of `PandasDataframeAxisPartition`s (concretely, `PandasOnRayDataframeColumnPartition`s), the `apply` was being handled incorrectly. Due to how positional arguments work, it was being interpreted as the argument `num_splits`.

This led to the `map_func` of `PandasDataframePartitionManager.broadcast_apply` not correctly being applied to a list of blocks, given that `*args` that `func` is applied to wasn't available.

*note: this was a rare bug that only showed up in the strange case of doing `broadcast_apply` on virtual partitions. This doesn't happen with `groupby.sum`, but does with `groupby.count` and `groupby.min` and possibly others. Has not been exhaustively tested.

- [ ] Resolves #4464? 
- [ ] tests added and passing


modin/core/dataframe/pandas/partitioning/axis_partition.py
```diff
@@ -65,7 +66,7 @@ def apply(
         """
         if num_splits is None:
             num_splits = len(self.list_of_blocks)
-
+        func_call = (func, args, kwargs)
```
mvashishtha, line 69: Elsewhere in Modin we are using `NamedTuple` for objects like this, e.g. [here](https://github.com/modin-project/modin/blob/9b94cb27bd7ea0b28ce6e6a163c056c3fa38d609/modin/config/pubsub.py#L20). I suggest switching to that.

mvashishtha (commented):
I like this approach. I left two minor comments.

modin/core/dataframe/pandas/partitioning/axis_partition.py
jeffreykennethli: will switch to this after getting the dask virtual partition changed + passing tests.

modin/core/base.py
```diff
@@ -0,0 +1,20 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software distributed under
+# the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific language
+# governing permissions and limitations under the License.
+
+from typing import Callable, Dict, NamedTuple, Tuple
+
+
+class FuncCall(NamedTuple):
```
jeffreykennethli, line 17: open to suggestions on naming here. some other options I've considered: `CallableSet` `FunctionSet` `Applyable`...

PR #4502, opened 2022-05-26
FEAT-#4501: Add RSS Memory Profiling to Modin Logging

### LINKED ISSUE #4501: Add RSS Memory to Modin Logging

naren-ponder:

## What do these changes do?

Add RSS Memory Profiling at `LogMemoryInterval` interval to Modin logging. This helps with out-of-core measurements.

- [x] Resolves #4501? 
- [x] tests added and passing

### DIFF
--- a/docs/usage_guide/advanced_usage/modin_logging.rst
+++ b/docs/usage_guide/advanced_usage/modin_logging.rst
@@ -10,6 +10,9 @@ repository or in the `modin/modin` directory may result in circular dependency i
 `modin/logging` directory and the Python `logging` module, which may be used as a default in such environments. As a resolution, one can simply
 run Modin logging from a different directory or manually manage the name conflicts.
 
+**User Warning:** In the interactive case (using modin logging inside of an ipython process), it may require quitting ipython several times
+(cntl-c) if you want to exit ipython when the memory profiling is enabled. 
+
 Usage example
 -------------
 
--- a/docs/usage_guide/advanced_usage/modin_logging.rst
+++ b/docs/usage_guide/advanced_usage/modin_logging.rst
@@ -10,8 +10,6 @@ repository or in the `modin/modin` directory may result in circular dependency i
 `modin/logging` directory and the Python `logging` module, which may be used as a default in such environments. As a resolution, one can simply
 run Modin logging from a different directory or manually manage the name conflicts.
 
-**User Warning:** In the interactive case (using modin logging inside of an ipython process), it may require quitting ipython several times
-(cntl-c) if you want to exit ipython when the memory profiling is enabled. 
 
 Usage example
 -------------
--- a/docs/usage_guide/advanced_usage/modin_logging.rst
+++ b/docs/usage_guide/advanced_usage/modin_logging.rst
@@ -10,7 +10,6 @@ repository or in the `modin/modin` directory may result in circular dependency i
 `modin/logging` directory and the Python `logging` module, which may be used as a default in such environments. As a resolution, one can simply
 run Modin logging from a different directory or manually manage the name conflicts.
 
-
 Usage example
 -------------
 
--- a/modin/error_message.py
+++ b/modin/error_message.py
@@ -86,7 +86,11 @@ class ErrorMessage(object):
     @classmethod
     def missmatch_with_pandas(cls, operation, message):
         logger = get_logger()
-        logger.debug("Modin Warning: {} mismatch with pandas, Message: {}".format(operation, message))
+        logger.debug(
+            "Modin Warning: {} mismatch with pandas, Message: {}".format(
+                operation, message
+            )
+        )
         cls.single_warning(
             f"`{operation}` implementation has mismatches with pandas:\n{message}."
         )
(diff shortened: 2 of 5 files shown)


modin/logging/config.py
```diff
@@ -144,8 +144,10 @@ def memory_thread(logger, sleep_time):
         The interval at which to profile system memory.
     """
     while True:
+        rss_mem = bytes_int_to_str(psutil.Process().memory_info().rss)
         svmem = psutil.virtual_memory()
-        logger.info(f"Memory Percentage: {svmem.percent}%")
+        logger.info(f"CPU Utilization: {svmem.percent}%")
```
mvashishtha, line 149:
```suggestion
        logger.info(f"Memory Percentage: {svmem.percent}%")
```

mvashishtha (dismissed):
LGTM with one fix.

modin/logging/config.py
naren-ponder: Oop good catch!

modin/logging/config.py
```diff
@@ -144,8 +144,10 @@ def memory_thread(logger, sleep_time):
         The interval at which to profile system memory.
     """
     while True:
+        rss_mem = bytes_int_to_str(psutil.Process().memory_info().rss / 1000000)
```
devin-petersohn:
```suggestion
        rss_mem = bytes_int_to_str(psutil.Process().memory_info().rss / 1_000_000)
```

naren-ponder: I think this got updated, so we use `bytes_int_to_str` instead of dividing by any number here.

jeffreykennethli (dismissed):

docs/usage_guide/advanced_usage/modin_logging.rst
```diff
@@ -10,6 +10,9 @@ repository or in the `modin/modin` directory may result in circular dependency i
 `modin/logging` directory and the Python `logging` module, which may be used as a default in such environments. As a resolution, one can simply
 run Modin logging from a different directory or manually manage the name conflicts.
 
+**User Warning:** In the interactive case (using modin logging inside of an ipython process), it may require quitting ipython several times
```
mvashishtha, line 13: I think fix #4515 using `atexit` should work. Let's discuss there.

modin/logging/config.py
```diff
@@ -144,8 +144,10 @@ def memory_thread(logger, sleep_time):
         The interval at which to profile system memory.
     """
     while True:
+        rss_mem = bytes_int_to_str(psutil.Process().memory_info().rss)
         svmem = psutil.virtual_memory()
         logger.info(f"Memory Percentage: {svmem.percent}%")
+        logger.info(f"RSS Memory: {rss_mem}")
```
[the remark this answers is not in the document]
naren-ponder: This addition is for out-of-core measurement, so the memory percent here is not that useful given we have the total system memory and the RSS already logged.

modin/error_message.py
```diff
@@ -21,10 +21,11 @@ class ErrorMessage(object):
     printed_warnings = set()
 
     @classmethod
-    @logger_decorator("MODIN-ERROR", "ErrorMessage.not_implemented", "debug")
```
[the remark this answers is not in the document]
naren-ponder: Moved to #4519.

modin/logging/config.py
```diff
@@ -96,7 +97,12 @@ def configure_logging():
 
     os.makedirs(os.path.dirname(log_filename), exist_ok=True)
 
-    logfile = logging.FileHandler(log_filename, "a")
+    logfile = RotatingFileHandler(
+        filename=log_filename,
+        mode="a",
+        maxBytes=100000000,
```
RehanSD, line 103: Why do we have `maxBytes` set so high?

RehanSD (commented):
left a quick question, but looks good to me otherwise! I think instead of writing over the file maybe it would be better to truncate the middle part? So that way users have start and end?

modin/logging/config.py
RehanSD: Where is the 10GB max?

RehanSD: Sounds good to me!

RehanSD (dismissed):
LGTM!

modin/logging/config.py
mvashishtha: This looks like 100 MB (10^8)? Is that what you meant to put? I think something like `int(1e8)` would be more readable.

devin-petersohn: I think it might be better to make this configurable instead of hard coded. 

I think setting a default max of 10GB of logs is way too high. This is 10GB per Modin job, not total. I prefer to cap it at 100MB per job by default. If more is needed, they can set a configuration.

devin-petersohn (changes requested):

modin/logging/config.py
naren-ponder: So 10MB per file, and a max of 10 files, for a total of 100MB per job?

devin-petersohn (approved):
This LGTM, @naren-ponder thanks!

I want to separate the different kinds of logs in a separate PR. Will open an issue for that.

PR #4515, opened 2022-05-31
FIX-#4503: Stop the memory logging thread after session exit.

### LINKED ISSUE #4503: Logging mode leaves a hanging thread in interactive python session
### System information
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: MacOS Monterey
- **Modin version** (`modin.__version__`): latest source
- **Python version**: 3.9.12

In ipython, run 

```python
import modin.pandas as pd

LogMode.enable()
df = pd.DataFrame([[1]])
```

and try to exit. You have to control-c or control-d twice.

This bug came up in the discussion on #4372 but it wasn't fixed in that PR.

The problematic thread is started [here](https://github.com/modin-project/modin/blob/49fc2cf3733f20ac6cf8a7c61e42ef7aa5cf4b03/modin/logging/config.py#L127).

mvashishtha:

## What do these changes do?

Make the memory thread [daemonic](https://docs.python.org/3/library/threading.html#thread-objects) so that "the entire Python program exits when only daemon threads are left."

- [x] Resolves #4503
- [x] tests added and passing

### DIFF
--- a/modin/logging/config.py
+++ b/modin/logging/config.py
@@ -21,8 +21,6 @@ import logging
 import datetime as dt
 import os
 import uuid
-import atexit
-
 import platform
 import psutil
 import pkg_resources
@@ -31,15 +29,6 @@ import time
 from modin.config import LogMemoryInterval, LogMode
 
 __LOGGER_CONFIGURED__: bool = False
-
-
-def _turn_off_logging():
-    """Turn off __LOGGER_CONFIGURED__."""
-    global __LOGGER_CONFIGURED__
-    __LOGGER_CONFIGURED__ = False
-
-
-atexit.register(_turn_off_logging)
 
 
 class ModinFormatter(logging.Formatter):  # noqa: PR01
@@ -137,7 +126,9 @@ def configure_logging():
 
     if LogMode.get() != "enable_api_only":
         mem_sleep = LogMemoryInterval.get()
-        mem = threading.Thread(target=memory_thread, args=[logger, mem_sleep])
+        mem = threading.Thread(
+            target=memory_thread, args=[logger, mem_sleep], daemon=True
+        )
         mem.start()
 
     __LOGGER_CONFIGURED__ = True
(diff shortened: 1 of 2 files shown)


naren-ponder (approved):
This looks great!

devin-petersohn (approved):
Looks great, thanks @mvashishtha!

devin-petersohn (commented):
I think there might be a docstyle failure @mvashishtha

mvashishtha:
> I think there might be a docstyle failure @mvashishtha

@devin-petersohn yes, I added a fix.

modin/logging/config.py
```diff
@@ -31,6 +33,15 @@
 __LOGGER_CONFIGURED__: bool = False
 
 
+def _turn_off_logging():
+    """Turn off __LOGGER_CONFIGURED__."""
+    global __LOGGER_CONFIGURED__
+    __LOGGER_CONFIGURED__ = False
+
+
+atexit.register(_turn_off_logging)
```
prutskov, line 42: Shouldn't we add  processing of additional signals (ctrl+c, for example), as we made in unidist?:
https://github.com/modin-project/unidist/blob/81a355ac9a308aeb9b746a7e62859c4a097427eb/unidist/core/backends/mpi/core/executor.py#L879-L881

mvashishtha: I was able to fix the bug by making the thread daemonic instead of by registering a custom `atexit`. It turns out that that solution wasn't working, anyway.

naren-ponder (approved):
Just tested and is working correctly. Thanks @mvashishtha!

mvashishtha:
> Just tested and is working correctly. Thanks @mvashishtha!

@naren-ponder did you use the latest commit? The fix has changed quite a bit since the first commit.

naren-ponder:
> > Just tested and is working correctly. Thanks @mvashishtha!
> 
> @naren-ponder did you use the latest commit? The fix has changed quite a bit since the first commit.

Yes used the one that made the thread daemonic.

PR #4529, opened 2022-06-06
FIX-#4478: Support level param in query and eval.

### LINKED ISSUE #4478: Remove hardcoding of local and global scopes for df.query and df.eval

For the pandas API functions `df.query` and `df.eval`, Modin hardcodes the function depth for local and global variable scopes. This is done in `modin/pandas/dataframe.py` in the `_update_var_dicts_in_kwargs` function:

```
f_locals = frame.f_back.f_back.f_back.f_locals
f_globals = frame.f_back.f_back.f_back.f_globals
```

Instead of this approach, Modin should support the `levels` parameter in `pandas.eval` (called by `df.query` and `df.eval`): https://pandas.pydata.org/docs/reference/api/pandas.eval.html#pandas.eval to create the local and global scopes.

mvashishtha:

## What do these changes do?

Support level param in query and eval.

- [x] Resolves #4478 
- [x] tests added and passing

### DIFF
--- a/modin/pandas/dataframe.py
+++ b/modin/pandas/dataframe.py
@@ -795,25 +795,40 @@ class DataFrame(metaclass_resolver(BasePandasDataset)):
             and self.eq(other).all().all()
         )
 
-    def _update_var_dicts_in_kwargs(self, expr, kwargs):
+    def _update_var_dicts_in_kwargs(self, expr, level: int, kwargs):
         """
         Copy variables with "@" prefix in `local_dict` and `global_dict` keys of kwargs.
+
+        This function exists because we neeed to infer local and variables
+        ourselves here in the main node instead of having the remote functions
+        infer them on their own. The remote functions take place in a different
+        thread with their own stacks that normally do not match the stack that
+        leads to the Modin eval or query call.
 
         Parameters
         ----------
         expr : str
             The expression string to search variables with "@" prefix.
+        level : int
+            The number of prior stack frames back to look for local variables.
+            level=0 means look just one frame back.
         kwargs : dict
             See the documentation for eval() for complete details on the keyword arguments accepted by query().
         """
         if "@" not in expr:
             return
-        frame = sys._getframe()
-        try:
-            f_locals = frame.f_back.f_back.f_back.f_back.f_locals
-            f_globals = frame.f_back.f_back.f_back.f_back.f_globals
-        finally:
-            del frame
+
+        # Bump level up one because this function is wrapped in the Modin
+        # logging function wrapper. The alternative is for the wrapper to
+        # give the Modin functions that deal with `level` the special
+        # treatment of bumping `level` up by one, but that's not so nice
+        # either.
+        level += 1
+
+        frame = sys._getframe(level + 1)
+        f_locals = frame.f_locals
+        f_globals = frame.f_globals
+        del frame
         local_names = set(re.findall(r"@([\w]+)", expr))
         local_dict = {}
         global_dict = {}
@@ -836,14 +851,27 @@ class DataFrame(metaclass_resolver(BasePandasDataset)):
         """
         Evaluate a string describing operations on ``DataFrame`` columns.
         """
+        # Remove `level` from the kwargs if it's already there. "level" doesn't
+        # make sense within the remote execution context, where the remote
+        # functions have a stack which is different from the stack on the
+        # driver node that ends in the modin eval call.
+        level = kwargs.pop("level", 0)
+
+        # Bump level up one because this function is wrapped in the Modin
+        # logging function wrapper. The alternative is for the wrapper to
+        # give the Modin functions that deal with `level` the special
+        # treatment of bumping `level` up by one, but that's not so nice
+        # either.
+        level += 1
         self._validate_eval_query(expr, **kwargs)
         inplace = validate_bool_kwarg(inplace, "inplace")
-        self._update_var_dicts_in_kwargs(expr, kwargs)
+        self._update_var_dicts_in_kwargs(expr, level + 1, kwargs)
+        # Make sure to not pass level to query compiler
         new_query_compiler = self._query_compiler.eval(expr, **kwargs)
         return_type = type(
             pandas.DataFrame(columns=self.columns)
             .astype(self.dtypes)
-            .eval(expr, **kwargs)
+            .eval(expr, level=level + 1, **kwargs)
         ).__name__
         if return_type == type(self).__name__:
             return self._create_or_update_from_compiler(new_query_compiler, inplace)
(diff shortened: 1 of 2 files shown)


modin/pandas/dataframe.py
```diff
@@ -795,25 +795,40 @@ def equals(self, other):  # noqa: PR01, RT01, D200
...
-            del frame
+
+        # Bump level up one because this function is wrapped in the Modin
+        # logging function wrapper. The alternative is for the wrapper to
+        # give the Modin functions that deal with `level` the special
+        # treatment of bumping `level` up by one, but that's not so nice
+        # either.
+        level += 1
+
+        frame = sys._getframe(level + 1)
```
naren-ponder, line 828: Why are we bumping level by one again here?

modin/pandas/dataframe.py
```diff
@@ -795,25 +795,40 @@ def equals(self, other):  # noqa: PR01, RT01, D200
...
+        f_locals = frame.f_locals
```
naren-ponder, line 829: Is there a reason you removed the `try` `finally` statements?

naren-ponder (commented):
Great work on this @mvashishtha. Just have a few questions, but overall looks good to me.

modin/pandas/dataframe.py
vnlitvinov: Yes, please retain the `try..finally`, it's here to protect against creating a reference cycle if any exception fires during getting the variables. Maybe we should extend this `finally` until `for name in local_names` loop is over and remove `f_locals` and `f_globals` as well for the same refcycle reason...

modin/pandas/dataframe.py
```diff
@@ -795,25 +795,40 @@ def equals(self, other):  # noqa: PR01, RT01, D200
...
-            f_globals = frame.f_back.f_back.f_back.f_back.f_globals
-        finally:
-            del frame
+
+        # Bump level up one because this function is wrapped in the Modin
+        # logging function wrapper. The alternative is for the wrapper to
+        # give the Modin functions that deal with `level` the special
+        # treatment of bumping `level` up by one, but that's not so nice
+        # either.
+        level += 1
```
vnlitvinov, line 826: I wonder if we should just discard any `modin`-owned frames (or at least `logging`-owned) instead of calculating this offset manually

modin/pandas/dataframe.py
```diff
@@ -836,14 +851,27 @@ def eval(self, expr, inplace=False, **kwargs):  # noqa: PR01, RT01, D200
...
         self._validate_eval_query(expr, **kwargs)
         inplace = validate_bool_kwarg(inplace, "inplace")
-        self._update_var_dicts_in_kwargs(expr, kwargs)
+        self._update_var_dicts_in_kwargs(expr, level + 1, kwargs)
```
vnlitvinov, line 868: you had already increased `level` above, why again `+1`?

modin/pandas/dataframe.py
```diff
@@ -836,14 +851,27 @@ def eval(self, expr, inplace=False, **kwargs):  # noqa: PR01, RT01, D200
...
+        # Make sure to not pass level to query compiler
```
vnlitvinov, line 869: I don't think this comment is needed

modin/pandas/dataframe.py
```diff
@@ -836,14 +851,27 @@ def eval(self, expr, inplace=False, **kwargs):  # noqa: PR01, RT01, D200
...
         new_query_compiler = self._query_compiler.eval(expr, **kwargs)
         return_type = type(
             pandas.DataFrame(columns=self.columns)
             .astype(self.dtypes)
-            .eval(expr, **kwargs)
+            .eval(expr, level=level + 1, **kwargs)
```
vnlitvinov, line 874: again, why `+1`?

modin/pandas/dataframe.py
```diff
@@ -1692,9 +1720,22 @@ def query(self, expr, inplace=False, **kwargs):  # noqa: PR01, RT01, D200
...
+        # functions have a stack which is different from the stack on the
+        # driver node that ends in the modin eval call.
+        level = kwargs.pop("level", 0)
+
+        # Bump level up one because this function is wrapped in the Modin
+        # logging function wrapper. The alternative is for the wrapper to
+        # give the Modin functions that deal with `level` the special
+        # treatment of bumping `level` up by one, but that's not so nice
+        # eitiher.
+        level += 1
```
vnlitvinov, line 1734: given I see exact same wall of comments three times already, maybe we should turn this into a helper function like

```python
def _extract_eval_level(kwargs):
    """
    Extract `level` from eval or query `kwargs`.

    Remove `level` from the kwargs if it's already there. "level" doesn't
    make sense within the remote execution context, where the remote
    functions have a stack which is different from the stack on the
    driver node that ends in the modin eval call.

    Bump level up one because eval/query are wrapped in the Modin
    logging function wrapper.
    """
    level = kwargs.pop("level", 0) + 1
    return level, kwargs
```
and then use it here like

```python
def query(self, expr, inplace=False, **kwargs):
    # docstring
    level, kwargs = _extract_eval_level(kwargs)
    # do the rest of the stuff
```

modin/pandas/test/dataframe/test_udf.py
```diff
@@ -285,12 +285,29 @@ def test_eval_df_arithmetic_subexpression():
...
+        # stack to access this variable
+        return getattr(df, method)(*args, **kwargs)
+
     modin_df, pandas_df = pd.DataFrame(data), pandas.DataFrame(data)
     op = "+" if method == "eval" else "<"
     for expr in (f"col1 {op} @local_var", f"col1 {op} @TEST_VAR"):
-        df_equals(getattr(modin_df, method)(expr), getattr(pandas_df, method)(expr))
+        df_equals(
+            df_method_with_local_var(modin_df, method, expr, level=level),
+            df_method_with_local_var(pandas_df, method, expr, level=level),
```
vnlitvinov, line 309: I think we should also test the case of omitting the `level` argument (to ensure we and pandas stay on the same page about default value of `level`).

modin/pandas/dataframe.py
```diff
@@ -795,25 +795,40 @@ def equals(self, other):  # noqa: PR01, RT01, D200
             and self.eq(other).all().all()
         )
 
-    def _update_var_dicts_in_kwargs(self, expr, kwargs):
+    def _update_var_dicts_in_kwargs(self, expr, level: int, kwargs):
         """
         Copy variables with "@" prefix in `local_dict` and `global_dict` keys of kwargs.
 
+        This function exists because we neeed to infer local and variables
```
jeffreykennethli, line 802: should this say `local and *global* variables` here? it feels like it's missing a word.

modin/pandas/dataframe.py
```diff
@@ -795,25 +795,40 @@ def equals(self, other):  # noqa: PR01, RT01, D200
...
+        ourselves here in the main node instead of having the remote functions
+        infer them on their own. The remote functions take place in a different
+        thread with their own stacks that normally do not match the stack that
+        leads to the Modin eval or query call.
+
         Parameters
         ----------
         expr : str
             The expression string to search variables with "@" prefix.
```
RehanSD, line 811:
```suggestion
            The expression string to search for variables with "@" prefix.
```

RehanSD (changes requested):
Left a quick comment.

Issue #4508, opened 2022-05-27
TEST: modin/test/test_partition_api.py is very flaky


mvashishtha:
### System information
- **Modin version** (`modin.__version__`): ce8f18765dd9201039769b022f92735792d63454

### Describe the problem

modin/test/test_partition_api.py seems to be failing more than 50% of the time.

### Source code / logs
Here's a run with a more informative stack trace: https://github.com/modin-project/modin/runs/6513946096?check_suite_focus=true

<details>

<summary>Show stack trace</summary>

```
[gw0] linux -- Python 3.8.13 /usr/share/miniconda/envs/modin/bin/python
worker 'gw0' crashed while running 'modin/test/test_partition_api.py::test_from_partitions[0-index-columns-row_lengths-None]'
__________ test_from_partitions[None-index-columns-row_lengths-None] ___________
[gw1] linux -- Python 3.8.13 /usr/share/miniconda/envs/modin/bin/python

axis = None, index = RangeIndex(start=0, stop=65536, step=1)
columns = Int64Index([  0,   1,   2,   3,   4,   5,   6,   7,   8,   9,
            ...
            246, 247, 248, 249, 250, 251, 252, 253, 254, 255],
           dtype='int64', length=512)
row_lengths = [65536], column_widths = None

    @pytest.mark.parametrize("column_widths", [None, "column_widths"])
    @pytest.mark.parametrize("row_lengths", [None, "row_lengths"])
    @pytest.mark.parametrize("columns", [None, "columns"])
    @pytest.mark.parametrize("index", [None, "index"])
    @pytest.mark.parametrize("axis", [None, 0, 1])
    def test_from_partitions(axis, index, columns, row_lengths, column_widths):
        num_rows = 2**16
        num_cols = 2**8
        data = np.random.randint(0, 100, size=(num_rows, num_cols))
        df1, df2 = pandas.DataFrame(data), pandas.DataFrame(data)
        expected_df = pandas.concat([df1, df2], axis=1 if axis is None else axis)
    
        index = expected_df.index if index == "index" else None
        columns = expected_df.columns if columns == "columns" else None
        row_lengths = (
            None
            if row_lengths is None
            else [num_rows, num_rows]
            if axis == 0
            else [num_rows]
        )
        column_widths = (
            None
            if column_widths is None
            else [num_cols]
            if axis == 0
            else [num_cols, num_cols]
        )
    
        if Engine.get() == "Ray":
            if axis is None:
[...]

PR #4551, opened 2022-06-08
TEST-#4508: Reduce test_partition_api pytest threads to deflake it.

mvashishtha:
Signed-off-by: mvashishtha <mahesh@ponder.io>

## What do these changes do?

Reduce test_partition_api pytest threads to deflake it.

I was able to reproduce the failure in 3/3 CI runs on a branch where the CI in push.yml only ran this test. I reduced the number of pytest cores to the default of 1 and 4/4 runs passed. 

- [x] commit message follows format outlined [here](https://modin.readthedocs.io/en/latest/development/contributing.html#commit-message-formatting)
- [x] passes `flake8 modin/ asv_bench/benchmarks scripts/doc_checker.py`
- [x] passes `black --check modin/ asv_bench/benchmarks scripts/doc_checker.py`
- [x] signed commit with `git commit -s` 
- [x] Resolves #4508
- [ ] tests added and passing
- [x] module layout described at `docs/development/architecture.rst` is up-to-date 
- [x] added (Issue Number: PR title (PR Number)) and github username to release notes for next major release

PR #4554, opened 2022-06-08
TEST-#4550: Use much less data in test_partition_api

### LINKED ISSUE #4550: TEST: Figure out why modin/test/test_partition_api.py with 2 pytest cores is so flaky.
The original issue is #4508. When I ran this test on an ubuntu machine I saw object store memory going up and some spilling to disk. Maybe that's related to why setting 2 pytest cores is flaky. The test takes a while with 1 core, so it would be good to re-enable 2 cores again. For now in the fix for #4508, I will move the test to 1 core.

mvashishtha:
## What do these changes do?

TEST-#4550: Use much less data in test_partition_api.

We don't need to test the partition API with such huge data. The large data made the test slow, which is why we originally why we had 2 cores. The 2 ray clusters started in the 2 cores probably could not handle all the data, resulting in #4508. We should just test with enough data to fill all the partitions. `test_data['int_data']` from the test utils should be good enough.

- [x] Resolves #4550
- [x] tests added and passing

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -257,7 +257,10 @@ jobs:
       - run: python -m pytest modin/test/storage_formats/base/test_internals.py
       - run: python -m pytest modin/test/storage_formats/pandas/test_internals.py
       - run: python -m pytest modin/test/test_envvar_npartitions.py
-      - run: python -m pytest -n 2 modin/test/test_partition_api.py
+      # TODO(https://github.com/modin-project/modin/issues/4550): this test is
+      # slow (~140 sec), but using 2 cores usually causes ray worker crashes.
+      # Fix the crashes and increase cores to 2.
+      - run: python -m pytest modin/test/test_partition_api.py
       - run: python -m pytest modin/test/test_utils.py
       - run: python -m pytest asv_bench/test/test_utils.py
       - run: python -m pytest modin/test/exchange/dataframe_protocol/base
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -257,9 +257,6 @@ jobs:
       - run: python -m pytest modin/test/storage_formats/base/test_internals.py
       - run: python -m pytest modin/test/storage_formats/pandas/test_internals.py
       - run: python -m pytest modin/test/test_envvar_npartitions.py
-      # TODO(https://github.com/modin-project/modin/issues/4550): this test is
-      # slow (~140 sec), but using 2 cores usually causes ray worker crashes.
-      # Fix the crashes and increase cores to 2.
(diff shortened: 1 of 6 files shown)


mvashishtha:
I need to wait for #4551 to be merged  so I can delete the comment about understanding the ncores bug. Converted this PR to draft.

prutskov (dismissed):
LGTM!

.github/workflows/ci.yml
```diff
@@ -257,9 +257,6 @@ jobs:
       - run: python -m pytest modin/test/storage_formats/base/test_internals.py
       - run: python -m pytest modin/test/storage_formats/pandas/test_internals.py
       - run: python -m pytest modin/test/test_envvar_npartitions.py
-      # TODO(https://github.com/modin-project/modin/issues/4550): this test is
-      # slow (~140 sec), but using 2 cores usually causes ray worker crashes.
-      # Fix the crashes and increase cores to 2.
       - run: python -m pytest modin/test/test_partition_api.py
```
YarShev, line 260: Should we return back pytest -n 2 for modin/test/test_partition_api.py?

mvashishtha: The test is so fast now that `-n 2` seems to actually slow the test down. on my mac, I get 24.07 sec with `-n 2`, and 13.3 sec with the default of 1. I'm on 

- macOSMonterey 12.2.1
- 2.3 GHz 8-Core Intel Core i9
- 16 GB 2667 MHz DDR4

YarShev: Ok, but your machine is different from those machines that are used in GH actions. It seems to me those have 2 CPUs each. I would suggest re-checking the need of `pytest -n 2 usage` for tests to understand if we really see the benefit. Of course, as part of a different issue.

YarShev (approved):
@mvashishtha, LGTM, thanks!

docs/release_notes/release_notes-0.16.0.rst
```diff
@@ -23,6 +23,7 @@ Key Features and Updates
   *
 * Update testing suite
   * TEST-#4508: Reduce test_partition_api pytest threads to deflake it (#4551)
+  * TEST-#4550: Use much less data in test_partition_api (#4554).
```
YarShev, line 26:
```suggestion
  * TEST-#4550: Use much less data in test_partition_api (#4554)
```

YarShev (approved):

.github/workflows/ci.yml
prutskov: I agree with @YarShev that it's need to definne the need of using 2 parallel jobs with Ray/Dask clusters on 2 cores GH servers. I was also faced with issues when `-n 2` only increase execution time.

Issue #4491, opened 2022-05-24
BUG: benchmark mode should wait for all partitions in parallel


mvashishtha:
We are waiting for all partitions in sequence, but it's more accurate to wait for them all at once, e.g with `ray.wait(partitions, num_returns=len(partitions))` for Ray. The performance difference is probably negligible because iterating across all the partitions should be really fast.

https://github.com/modin-project/modin/blob/c17dde71ae8114723a13279e36bf5532dcba328a/modin/core/dataframe/pandas/partitioning/partition_manager.py#L66

PR #4563, opened 2022-06-09
REFACTOR-#4530: Standardize access to physical data in partitions

### LINKED ISSUE #4530: Standardize the `oid/future/data/...` property of the `<StorageFormat>On<Engine>DataframePartition` class

Currently, we have a different APIs for property of physical data stored by specific partition class.
https://github.com/modin-project/modin/blob/dcee13d57ebf9a006460deedb734c15791acae7a/modin/core/execution/dask/implementations/pandas_on_dask/partitioning/partition.py#L45
https://github.com/modin-project/modin/blob/dcee13d57ebf9a006460deedb734c15791acae7a/modin/core/execution/python/implementations/pandas_on_python/partitioning/partition.py#L45
https://github.com/modin-project/modin/blob/dcee13d57ebf9a006460deedb734c15791acae7a/modin/core/execution/ray/implementations/cudf_on_ray/partitioning/partition.py#L55
https://github.com/modin-project/modin/blob/dcee13d57ebf9a006460deedb734c15791acae7a/modin/core/execution/ray/implementations/pandas_on_ray/partitioning/partition.py#L54
https://github.com/modin-project/modin/blob/dcee13d57ebf9a006460deedb734c15791acae7a/modin/experimental/core/execution/native/implementations/omnisci_on_native/partitioning/partition.py#L61-L63

We have a number of possible problems:
 1. Different name of property doesn't allow to create a unified implementation over all types of partitions. For example, case when we should provide other partition to args of `apply` call. Current logic of partition class doesn't expect partition type in args/kwargs. Therefore, we should provide physical data to args:
```python
class PandasDataframePartitionManager:
   ...
  def some_unified_functionality(cls, parts_left, parts_right):
      ...
      [lpart.apply(rpart.<oid/future/data/...>)
         for lpart, rpart in zip(parts_left, parts_right)]  # --> We should implement this logic for every engine, that introduced a lot of code duplication
```
 2.  Physical data can be corrupted because the property is public and can be changed outside after `__init__` during life of partition object (property setter issue).
 3. Physical data should be provided to outside only after `self.drain_call_queue` call. Because without executing of all operations in call queue we will get incorrect physical data (property getter issue).

The possible solutions:

1. API solutions:
- a) Standardize physical data property. Need to have the single API for acces to physical data for any partition type.
OR
- b) Add logic of expecting of partition type in `apply/add_to_apply_calls`. The soultion, possible, isn't good because in that case we should check every argument of `apply, add_to_apply_calls` on `isinstance(arg, partition_type)`, it will introduce performance gaps.
2. Setter issue solution:
- Physical data property should be readonly to exclude unexpected changing outside. 
3. Getter issue solution:
- Physical data property should include `drain_call_queue` logic before providing physical data to outside to be up to date

The concept from ideas above:
```python
class <StorageFormat>On<Engine>DataFramePartition:
   def __init__(self, physical_data, ...):
      self._physical_data = physical_data
   
   @property
   def physical_data(self):
      self.drain_call_queue()
      return self._physical_data 
```

@modin-project/modin-core, @modin-project/modin-contributors,  what're you thoughts?

prutskov:

## What do these changes do?

This PR unifies the API for accessing to the internal data in partition.

- [x] Resolves #4530  
- [x] tests added and passing

### DIFF
--- a/asv_bench/benchmarks/utils/common.py
+++ b/asv_bench/benchmarks/utils/common.py
@@ -513,11 +513,11 @@ def execute(
         if ASV_USE_ENGINE == "ray":
             from ray import wait
 
-            all(map(lambda partition: wait([partition.oid]), partitions))
+            all(map(lambda partition: wait([partition.physical_data]), partitions))
         elif ASV_USE_ENGINE == "dask":
             from dask.distributed import wait
 
-            all(map(lambda partition: wait(partition.future), partitions))
+            all(map(lambda partition: wait(partition.physical_data), partitions))
         elif ASV_USE_ENGINE == "python":
             pass
 
--- a/asv_bench/benchmarks/utils/common.py
+++ b/asv_bench/benchmarks/utils/common.py
@@ -513,11 +513,11 @@ def execute(
         if ASV_USE_ENGINE == "ray":
             from ray import wait
 
-            all(map(lambda partition: wait([partition.physical_data]), partitions))
+            all(map(lambda partition: wait([partition._data]), partitions))
         elif ASV_USE_ENGINE == "dask":
             from dask.distributed import wait
 
-            all(map(lambda partition: wait(partition.physical_data), partitions))
+            all(map(lambda partition: wait(partition._data), partitions))
         elif ASV_USE_ENGINE == "python":
             pass
 
--- a/modin/core/dataframe/pandas/partitioning/partition.py
+++ b/modin/core/dataframe/pandas/partitioning/partition.py
@@ -32,6 +32,20 @@ class PandasDataframePartition(ABC):  # pragma: no cover
 
     _length_cache = None
     _width_cache = None
+    _physical_data = None
+
+    @property
+    def physical_data(self):
+        """
+        Flush the `call_queue` and get the physical data wrapped by this partition.
+
+        Returns
+        -------
+        object
+            The physical data wrapped by this partition.
+        """
+        self.drain_call_queue()
+        return self._physical_data
 
     def get(self):
         """
--- a/modin/core/dataframe/pandas/partitioning/partition.py
+++ b/modin/core/dataframe/pandas/partitioning/partition.py
@@ -37,7 +37,7 @@ class PandasDataframePartition(ABC):  # pragma: no cover
     @property
     def physical_data(self):
         """
-        Flush the `call_queue` and get the physical data wrapped by this partition.
+        Drain the `call_queue` and get the physical data wrapped by this partition.
 
         Returns
         -------
--- a/modin/core/dataframe/pandas/partitioning/partition.py
+++ b/modin/core/dataframe/pandas/partitioning/partition.py
@@ -32,20 +32,7 @@ class PandasDataframePartition(ABC):  # pragma: no cover
 
     _length_cache = None
     _width_cache = None
-    _physical_data = None
-
-    @property
-    def physical_data(self):
-        """
-        Drain the `call_queue` and get the physical data wrapped by this partition.
-
-        Returns
-        -------
-        object
-            The physical data wrapped by this partition.
-        """
-        self.drain_call_queue()
-        return self._physical_data
+    _data = None
 
     def get(self):
         """
--- a/modin/core/execution/dask/implementations/pandas_on_dask/partitioning/partition_manager.py
+++ b/modin/core/execution/dask/implementations/pandas_on_dask/partitioning/partition_manager.py
@@ -47,6 +47,4 @@ class PandasOnDaskDataframePartitionManager(PandasDataframePartitionManager):
         list
             The objects wrapped by `partitions`.
         """
-        return DaskWrapper.materialize(
-            [partition.physical_data for partition in partitions]
-        )
+        return DaskWrapper.materialize([partition._data for partition in partitions])
--- a/modin/core/execution/dask/implementations/pandas_on_dask/partitioning/partition_manager.py
+++ b/modin/core/execution/dask/implementations/pandas_on_dask/partitioning/partition_manager.py
@@ -47,4 +47,6 @@ class PandasOnDaskDataframePartitionManager(PandasDataframePartitionManager):
         list
             The objects wrapped by `partitions`.
         """
-        return DaskWrapper.materialize([partition.future for partition in partitions])
+        return DaskWrapper.materialize(
+            [partition.physical_data for partition in partitions]
+        )
--- a/modin/core/execution/python/implementations/pandas_on_python/partitioning/partition.py
+++ b/modin/core/execution/python/implementations/pandas_on_python/partitioning/partition.py
@@ -52,7 +52,7 @@ class PandasOnPythonDataframePartition(PandasDataframePartition):
     @property
     def physical_data(self):
         """
-        Flush the `call_queue` and get the physical data wrapped by this partition.
+        Drain the `call_queue` and get the physical data wrapped by this partition.
 
         Returns
         -------
--- a/modin/core/execution/python/implementations/pandas_on_python/partitioning/partition.py
+++ b/modin/core/execution/python/implementations/pandas_on_python/partitioning/partition.py
@@ -25,7 +25,7 @@ class PandasOnPythonDataframePartition(PandasDataframePartition):
 
     Parameters
     ----------
-    physical_data : pandas.DataFrame
+    data : pandas.DataFrame
         ``pandas.DataFrame`` that should be wrapped with this class.
     length : int, optional
         Length of `data` (number of rows in the input dataframe).
@@ -41,30 +41,17 @@ class PandasOnPythonDataframePartition(PandasDataframePartition):
     subclasses. There is no logic for updating in-place.
     """
 
-    def __init__(self, physical_data, length=None, width=None, call_queue=None):
-        self._physical_data = physical_data
+    def __init__(self, data, length=None, width=None, call_queue=None):
+        self._data = data
         if call_queue is None:
             call_queue = []
         self.call_queue = call_queue
         self._length_cache = length
         self._width_cache = width
 
-    @property
-    def physical_data(self):
-        """
-        Drain the `call_queue` and get the physical data wrapped by this partition.
-
-        Returns
-        -------
-        pandas.DataFrame
-            Copy of pandas DataFrame wrapped by this partition.
-        """
-        self.drain_call_queue()
-        return self._physical_data.copy()
-
     def get(self):
         """
-        Flush the `call_queue` and return copy of the physical data.
+        Flush the `call_queue` and return copy of the data.
 
         Returns
         -------
@@ -76,7 +63,7 @@ class PandasOnPythonDataframePartition(PandasDataframePartition):
         Since this object is a simple wrapper, just return the copy of data.
         """
         self.drain_call_queue()
-        return self._physical_data.copy()
+        return self._data.copy()
 
     def apply(self, func, *args, **kwargs):
         """
@@ -121,10 +108,10 @@ class PandasOnPythonDataframePartition(PandasDataframePartition):
                     raise e
             return result
 
-        self._physical_data = call_queue_closure(self._physical_data, self.call_queue)
+        self._data = call_queue_closure(self._data, self.call_queue)
         self.call_queue = []
         return PandasOnPythonDataframePartition(
-            func(self._physical_data.copy(), *args, **kwargs)
+            func(self._data.copy(), *args, **kwargs)
         )
 
     def add_to_apply_calls(self, func, *args, **kwargs):
@@ -146,7 +133,7 @@ class PandasOnPythonDataframePartition(PandasDataframePartition):
             New ``PandasOnPythonDataframePartition`` object with extended call queue.
         """
         return PandasOnPythonDataframePartition(
-            self._physical_data.copy(),
+            self._data.copy(),
             call_queue=self.call_queue + [(func, args, kwargs)],
         )
 
@@ -213,7 +200,7 @@ class PandasOnPythonDataframePartition(PandasDataframePartition):
             The length of the object.
         """
         if self._length_cache is None:
-            self._length_cache = self.apply(self._length_extraction_fn())._physical_data
+            self._length_cache = self.apply(self._length_extraction_fn())._data
         return self._length_cache
 
     def width(self):
@@ -226,5 +213,5 @@ class PandasOnPythonDataframePartition(PandasDataframePartition):
             The width of the object.
         """
         if self._width_cache is None:
-            self._width_cache = self.apply(self._width_extraction_fn())._physical_data
+            self._width_cache = self.apply(self._width_extraction_fn())._data
         return self._width_cache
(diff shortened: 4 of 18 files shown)


modin/core/execution/python/implementations/pandas_on_python/partitioning/partition.py
```diff
@@ -41,17 +41,30 @@ class PandasOnPythonDataframePartition(PandasDataframePartition):
     subclasses. There is no logic for updating in-place.
     """
 
-    def __init__(self, data, length=None, width=None, call_queue=None):
-        self.data = data
+    def __init__(self, physical_data, length=None, width=None, call_queue=None):
+        self._physical_data = physical_data
```
YarShev, line 45:
```suggestion
        self._data = data
```
I think `data` should be good enough.

modin/core/execution/ray/implementations/pandas_on_ray/partitioning/virtual_partition.py
```diff
@@ -111,13 +111,12 @@ def list_of_blocks(self):
...
-        # Defer draining call queue until we get the partitions
-        # TODO Look into draining call queue at the same time as the task
-        result = [None] * len(self.list_of_partitions_to_combine)
-        for idx, partition in enumerate(self.list_of_partitions_to_combine):
-            partition.drain_call_queue()
-            result[idx] = partition.oid
-        return result
+        # Draining call queue of partition is happening during access to physical
+        # data of partition. Defer draining call queue until we get the partitions.
+        # TODO: Look into draining call queue at the same time as the task.
```
mvashishtha, line 116: @devin-petersohn I think you wrote this TODO. How would we do that? Should we file an issue for it?

mvashishtha (commented):
Thanks for the change! I have minor comments.

modin/core/execution/python/implementations/pandas_on_python/partitioning/partition.py
```diff
@@ -41,17 +41,30 @@ class PandasOnPythonDataframePartition(PandasDataframePartition):
...
+    def physical_data(self):
+        """
+        Flush the `call_queue` and get the physical data wrapped by this partition.
+
+        Returns
+        -------
+        pandas.DataFrame
+            Copy of pandas DataFrame wrapped by this partition.
+        """
+        self.drain_call_queue()
```
YarShev, line 62: I don't think we want to drain the call queue every time when accessing the data. We likely want to drain the call queue in a worker.

asv_bench/benchmarks/utils/common.py
```diff
@@ -513,11 +513,11 @@ def execute(
         if ASV_USE_ENGINE == "ray":
             from ray import wait
 
-            all(map(lambda partition: wait([partition.oid]), partitions))
+            all(map(lambda partition: wait([partition.physical_data]), partitions))
```
vnlitvinov, line 516: side note - `map()` is an iterator, so using `all()` would stop on _any_ of the `.wait()` returning something false-ish, we should probably use `tuple()` or `list()` if we don't want to use the results of `map()` but want all of its iterations to finish

modin/core/execution/python/implementations/pandas_on_python/partitioning/partition.py
vnlitvinov: I second this. It would cause a lot of draining, is this what we really want?

asv_bench/benchmarks/utils/common.py
mvashishtha: See my comment [here](https://github.com/modin-project/modin/pull/4404#discussion_r877465281): I don't understand why this is different from `[partition.wait() for partition in partitions]`

modin/core/execution/dask/implementations/pandas_on_dask/partitioning/partition_manager.py
```diff
@@ -47,4 +47,6 @@ def get_objects_from_partitions(cls, partitions):
         list
             The objects wrapped by `partitions`.
         """
-        return DaskWrapper.materialize([partition.future for partition in partitions])
+        return DaskWrapper.materialize(
+            [partition.physical_data for partition in partitions]
```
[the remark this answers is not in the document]
mvashishtha: Strictly speaking, `get_objects_from_partitions` should get the "latest" physical data, after all the deferred operations on the call queue are applied. However, I can't think of a Modin-level bug where we use `get_objects_from_partitions` to get partitions that have deferred operations on their call queue. dask uses `get_objects_from_partitions` like [here](https://github.com/modin-project/modin/blob/97769988a6f19e4b76f34238c97bf159ee7626a5/modin/core/execution/dask/implementations/pandas_on_dask/dataframe/dataframe.py#L56) but the `apply` drains the call queue. `get_indices` [here](https://github.com/modin-project/modin/blob/57e29bc5d82348006c5170ef9ac0a9eedcd9acf9/modin/core/dataframe/pandas/partitioning/partition_manager.py#L810) uses `get_objects_from_partitions`, but I can't think of a case where we have something on the call queue at the point where we are getting the new index from the partitions.

modin/core/execution/python/implementations/pandas_on_python/partitioning/partition.py
prutskov: In case of accessing to physical data holded by partition the getter property should return actual data after all deffered operations. For example, in a case of creating the new partition from another one using physical data the correct state of source-partition will be lost in case not calling the call queue.

In the current code, before all accesses to physical data using `oid/future` `drain_call_queue` is called for the each partition.
If `call_queue` is empty, the call `drain_call_queue` doesn't include any additional computations.

> We likely want to drain the call queue in a worker.

@YarShev, what do you mean here?

modin/core/execution/python/implementations/pandas_on_python/partitioning/partition.py
prutskov: @devin-petersohn proposed `physical_data` naming. Devin, what do you think about `_data`/`data`?

modin/distributed/dataframe/pandas/partitions.py
```diff
@@ -52,28 +52,29 @@ def unwrap_partitions(api_layer_object, axis=None, get_ip=False):
...
-        elif actual_engine in ("PandasOnDaskDataframePartition",):
-            return _unwrap_partitions("future")
+        if actual_engine in (
+            "PandasOnRayDataframePartition",
+            "PandasOnDaskDataframePartition",
+        ):
+            return _unwrap_partitions()
         raise ValueError(
             f"Do not know how to unwrap '{actual_engine}' underlying partitions"
         )
```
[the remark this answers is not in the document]
prutskov: We need this check, because behavior was not changed: in case `actual_engine` is not Ray/Dask we should continue raising `ValueError`

modin/core/execution/python/implementations/pandas_on_python/partitioning/partition.py
```diff
@@ -41,17 +41,30 @@ class PandasOnPythonDataframePartition(PandasDataframePartition):
...
         if call_queue is None:
             call_queue = []
         self.call_queue = call_queue
         self._length_cache = length
         self._width_cache = width
 
+    @property
+    def physical_data(self):
+        """
+        Flush the `call_queue` and get the physical data wrapped by this partition.
```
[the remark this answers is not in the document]
prutskov: applied

modin/core/dataframe/pandas/partitioning/partition.py
```diff
@@ -32,6 +32,20 @@ class PandasDataframePartition(ABC):  # pragma: no cover
 
     _length_cache = None
     _width_cache = None
+    _physical_data = None
+
+    @property
+    def physical_data(self):
+        """
+        Flush the `call_queue` and get the physical data wrapped by this partition.
```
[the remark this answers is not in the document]
prutskov: applied

asv_bench/benchmarks/utils/common.py
prutskov: I agree with @vnlitvinov, that implementation looks incorrect because we might not wait of all of partitions computations to finish.

Implementation `[partition.wait() for partition in partitions]` is applicable, but in this case we will stop on the each iteration to wait the each partition. I believe, that something like that will be worked a bit faster:

```python
wait([partition.physical_data for partition in partitions], num_returns=len(partitions))
```

I think, that this kind of changes should be implemented in another PR.

mvashishtha: @prutskov that's correct. I have filed #4491 for that already.

mvashishtha: > side note - map() is an iterator, so using all() would stop on any of the .wait() returning something false-ish

@vnlitvinov yes, that's a bug , that's why we wait on `partition.wait() or True` elsewhere, e.g.: https://github.com/modin-project/modin/blob/e67db1dd3ee448eec23a05ffdb3ca38a2b2b709e/asv_bench/benchmarks/utils/common.py#L509

modin/core/execution/python/implementations/pandas_on_python/partitioning/partition.py
YarShev: > > We likely want to drain the call queue in a worker.
> 
> @YarShev, what do you mean here?

Imagine a case when the partition has the call queue not drained and we want to pass it as an input in a remote call. It might be beneficial (I think it is rather beneficial) to pass the call queue along with the partition and drain the call queue of the partition in the remote call rather than perform that before.

I understand your concern on the actual physical data held by a partition. We have two options here.
1. Not drain the call queue in the getter and hand on this responsibility to the partition manager that handles partitions.
2. Drain the call queue in the getter and hand on the responsibility of accessing private `_data` field and draining the call queue to the partition manager in case we want to achieve the behavior which I said about above.

Do you think which option makes much more sense?

prutskov: > Imagine a case when the partition has the call queue not drained and we want to pass it as an input in a remote call. It might be beneficial (I think it is rather beneficial) to pass the call queue along with the partition and drain the call queue of the partition in the remote call rather than perform that before.

Do you mean case when we want pass some other `partition1` together with their `partiiton1.call_queue` in a remote call of `partition0`? Something like that:

```python
# partition manager impl
def some_operation(cls, foo, left_parts, other_parts):
   ...
   def func(lpart, rpart, rpart_call_queue, *args,  **kwargs):
      actual_rpart = rpart_call_queue(rpart) # drain call queue of other partition in remote call
      return foo(lpart, rpart)

   [lpart.apply(func, rpart._data, rpart.call_queue) for lpart, rpart in zip(left_parts, other_parts)] #--> We have access to private `partition._data`.

```
If you mean that I've described above, that it will be better the 3rd option, in my opinion:

3. Direct access to `_physical_data/_data` (naming discussion is in other thread) must happen only in partition class to resolve issues described in #4530. But instead of property for accessing to `_physical_data/_data` need to use getter func with parameter `is_drain_call_queue` with default value `True`. This approach allows you to transfer responsibility of draining call queue to partition manager directly with expected behavior for such cases. Also you will always have the actual physical data from default call `actual_data = prtition.get_data()`. 

```python
# partition class

def get_data(self, is_drain_call_queue=True):
   if is_drain_call_queue:
      self.drain_call_queue
   return self._data
   
```
What do you think @YarShev?

cc @vnlitvinov, @mvashishtha @modin-project/modin-core

YarShev: Yes, I mean that that you linked above. I think this option makes sense too with the getter named just `data()`. However, other question has arisen in my head. If we go with this option, do we need `drain_call_queue()` to be a public method? It seems to me we should make it private in that case.

YarShev: Also, there might be a confusion between `data()` and `get()` methods.

YarShev: I would probably narrow the scope of the issue just to standardizing access to physical data of partitions (not introducing the getter) since we also use internal members such as `._width_cache` and `._length_cache` outside of a partition class. Accessing the fields that might be corrupted is rather a broader problem and should probably be resolved as part of a separate issue.

prutskov: > Yes, I mean that that you linked above. I think this option makes sense too with the getter named just `data()`. However, other question has arisen in my head. If we go with this option, do we need `drain_call_queue()` to be a public method? It seems to me we should make it private in that case.

`data()` is also makes sense to me though I prefer to use verbs in functions/methods to highlight "action".

> Also, there might be a confusion between data() and get() methods.

Probably, it will be need to change name of `get` to something other with additional highlighting of data materialization.

> I would probably narrow the scope of the issue just to standardizing access to physical data of partitions (not introducing the getter) since we also use internal members such as ._width_cache and ._length_cache outside of a partition class. Accessing the fields that might be corrupted is rather a broader problem and should probably be resolved as part of a separate issue.

Agree, I'll change the scope of this PR to cover only 1st problem described in #4530. It will unblock PR #4391.
I'm not sure that examples of usage `._width_cache`/`._length_cache` outside of partition class is a good approach. Possible, for these properties it will be need also check/fix publicity.

modin/core/execution/python/implementations/pandas_on_python/partitioning/partition.py
prutskov: `_data` name is used for private property

modin/core/execution/python/implementations/pandas_on_python/partitioning/partition.py
prutskov: PR scope was changed to cover only 1st problem described in #4530 (renaming `oid/future/data -> _data`). The public getter will be added in an other PR as resolution of 2nd and 3rd problems from #4530.

modin/core/execution/dask/implementations/pandas_on_dask/partitioning/partition_manager.py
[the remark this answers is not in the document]
prutskov: Thread is out of date. Therefore I resolve conversation.

PR scope was changed to cover only 1st problem described in #4530 (renaming `oid/future/data -> _data`). The public getter will be added in an other PR as resolution of 2nd and 3rd problems from #4530.

modin/core/execution/ray/implementations/pandas_on_ray/partitioning/virtual_partition.py
prutskov: Thread is out of date. Therefore I resolve conversation. The comment was reverted

PR scope was changed to cover only 1st problem described in https://github.com/modin-project/modin/issues/4530 (renaming oid/future/data -> _data). The public getter will be added in an other PR as resolution of 2nd and 3rd problems from https://github.com/modin-project/modin/issues/4530.

mvashishtha (approved):

YarShev (approved):
@prutskov, LGTM, thanks!

PR #4571, opened 2022-06-11
FIX-#4570: Replace np.bool -> np.bool_

### LINKED ISSUE #4570: BUG: DeprecationWarning from using np.bool
I get:
```
../../../Library/Application Support/hatch/env/virtual/mismo-5Vu_DQpX/mismo/lib/python3.10/site-packages/modin/core/storage_formats/pandas/query_compiler.py:1350: 1 warning
../../../Library/Application Support/hatch/env/virtual/mismo-5Vu_DQpX/mismo/lib/python3.10/site-packages/modin/core/storage_formats/pandas/query_compiler.py:1351: 1 warning
../../../Library/Application Support/hatch/env/virtual/mismo-5Vu_DQpX/mismo/lib/python3.10/site-packages/modin/core/storage_formats/pandas/query_compiler.py:1356: 1 warning
../../../Library/Application Support/hatch/env/virtual/mismo-5Vu_DQpX/mismo/lib/python3.10/site-packages/modin/core/storage_formats/pandas/query_compiler.py:1376: 1 warning
../../../Library/Application Support/hatch/env/virtual/mismo-5Vu_DQpX/mismo/lib/python3.10/site-packages/modin/core/storage_formats/pandas/query_compiler.py:1378: 1 warning
../../../Library/Application Support/hatch/env/virtual/mismo-5Vu_DQpX/mismo/lib/python3.10/site-packages/modin/core/storage_formats/pandas/query_compiler.py:1383: 1 warning
../../../Library/Application Support/hatch/env/virtual/mismo-5Vu_DQpX/mismo/lib/python3.10/site-packages/modin/core/storage_formats/pandas/query_compiler.py:1384: 1 warning
../../../Library/Application Support/hatch/env/virtual/mismo-5Vu_DQpX/mismo/lib/python3.10/site-packages/modin/core/storage_formats/pandas/query_compiler.py:1385: 1 warning
../../../Library/Application Support/hatch/env/virtual/mismo-5Vu_DQpX/mismo/lib/python3.10/site-packages/modin/core/storage_formats/pandas/query_compiler.py:1386: 1 warning
../../../Library/Application Support/hatch/env/virtual/mismo-5Vu_DQpX/mismo/lib/python3.10/site-packages/modin/core/storage_formats/pandas/query_compiler.py:1387: 1 warning
../../../Library/Application Support/hatch/env/virtual/mismo-5Vu_DQpX/mismo/lib/python3.10/site-packages/modin/core/storage_formats/pandas/query_compiler.py:1388: 1 warning
../../../Library/Application Support/hatch/env/virtual/mismo-5Vu_DQpX/mismo/lib/python3.10/site-packages/modin/core/storage_formats/pandas/query_compiler.py:1389: 1 warning
../../../Library/Application Support/hatch/env/virtual/mismo-5Vu_DQpX/mismo/lib/python3.10/site-packages/modin/core/storage_formats/pandas/query_compiler.py:1390: 1 warning
../../../Library/Application Support/hatch/env/virtual/mismo-5Vu_DQpX/mismo/lib/python3.10/site-packages/modin/core/storage_formats/pandas/query_compiler.py:1391: 1 warning
../../../Library/Application Support/hatch/env/virtual/mismo-5Vu_DQpX/mismo/lib/python3.10/site-packages/modin/core/storage_formats/pandas/query_compiler.py:1412: 1 warning
  DeprecationWarning: `np.bool` is a deprecated alias for the builtin `bool`. To silence this warning, use `bool` by itself. Doing this will not modify any behavior and is safe. If you specifically wanted the numpy scalar type, use `np.bool_` here.
  Deprecated in NumPy 1.20; for more details and guidance: https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations
```

Simple follow up patch to follow.

NickCrews:
I was getting:
```
  DeprecationWarning: `np.bool` is a deprecated alias for the builtin `bool`. To silence this warning, use `bool` by itself. Doing this will not modify any behavior and is safe. If you specifically wanted the numpy scalar type, use `np.bool_` here.
  Deprecated in NumPy 1.20; for more details and guidance: https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations
```

I assume that this is the correct substitution, instead of changing to a simple `bool`, but I don't understand what this code does here, so I may be wrong.

If this looks good I can do the final TODOs in this checklist

## What do these changes do?

- [x] Resolves #4570 
- [x] tests added and passing


devin-petersohn (commented):
Hi @NickCrews, thanks for the contribution! The changes look great!

CI will fail on the commit message, it needs to have a signed off message (`git commit -s`) and the first line of the message needs to follow this: https://modin.readthedocs.io/en/latest/development/contributing.html#commit-message-formatting . Setting the first line of the commit message to the PR title will fix that. 

We use commit messages to generate the release notes. Thanks, and let me know if I can help in any way!

NickCrews:
@devin-petersohn Thanks, I think I updated the branch correctly. I didn't do the task:

> added (Issue Number: PR title (PR Number)) and github username to release notes for next major release

because you are saying that is autogenerated? Or I need to do that?

YarShev:
@NickCrews, we should probably replace all entries of `np.bool` in the codebase (search for `np.bool` -> replace all to `np.bool_`). Would you be able to handle that?

NickCrews:
@YarShev That followup commit replaced all the other instances of np.bool. But I think in those cases they should get replaced with just bool. Double check me.

devin-petersohn:
>because you are saying that is autogenerated? Or I need to do that?

Not exactly, sorry for not being clear. We are working toward that, but for now the message also needs to be manually added here: https://github.com/modin-project/modin/blame/master/docs/release_notes/release_notes-0.16.0.rst#L9

NickCrews:
Added entry to release notes! That last CI run failed, though it looked unrelated to this PR? IDK we will see this new run again.

I will note, as a new contributor the number of hoops I have to jump through, as compared to contributing to say pandas, is annoying. Autogenerated release notes, and looser requirements on how to format commit messages and issue numbers etc would probably make me more likely to contribute in the future. Just my two cents, I'm sure there are reasons for the structure!

RehanSD:
> Added entry to release notes! That last CI run failed, though it looked unrelated to this PR? IDK we will see this new run again.
> 
> 
> 
> I will note, as a new contributor the number of hoops I have to jump through, as compared to contributing to say pandas, is annoying. Autogenerated release notes, and looser requirements on how to format commit messages and issue numbers etc would probably make me more likely to contribute in the future. Just my two cents, I'm sure there are reasons for the structure!

Hi @NickCrews! That makes a lot of sense, thank you - we'll be sure to keep that in mind moving forward! One of the reasons we like to lint commits is so that the commit history on the main branch remains clear + descriptive!

I've approved the CI to run, and look forward to merging when everything is green!

NickCrews:
@RehanSD I just had to force-push again to fix a error in the docs build, thanks!

RehanSD:
> @YarShev That followup commit replaced all the other instances of np.bool. But I think in those cases they should get replaced with just bool. Double check me.

Hi @NickCrews quick question - why are we replacing those instances with `bool` instead of `np.bool_`? I did some checking, and it seems that `bool` and `np.bool_` behave similarly when doing dtypes checking (i.e. `df.dtypes == bool` and `df.dtypes == np.bool_` return the same result), but `np.bool_ != bool` - so unless there's a compelling reason to swap to `bool`, I'd prefer sticking with `np.bool_`, in case there are some edge cases we're missing!

(I've also approved the run again)!

YarShev:
@RehanSD, thanks, that is a great question! Some info to be considered - https://stackoverflow.com/questions/55905690/how-exactly-does-the-behavior-of-python-bool-and-numpy-bool-differ.

NickCrews:
I made the choice between going to bool vs np.bool_ based on:
- where possible, use bool. I figured anywhere we could be basic python and not doing stuff with numpy the better. numpy recommends using bool, and I figured it would be the one users are more likely to use. These dtype checks seemed like prime use cases where someone would use bool instead of np.bool_. Per your tests you did above (nice, I didn't even bother doing that explicitly, thanks!), I think that is even more evidence this is an OK thing to do. Also, I read [numpy's release notes](https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations) as saying that going to bool in these cases is fine in cases where we are only interacting with the public API of numpy.
- In the cases where I kept np.bool_ , it was because it seemed like we were getting into internals, doing some meta-programming to create new types and methods etc. I thought the distinction was more important there so I wanted to let sleeping dogs lie.

Hopefully that reasoning makes sense? Other things I'm missing? I will change everything to np.bool_ if you tell me, but I think I still prefer the way the PR is now.

CI failures still look unrelated? These same workflows are failing on `main`.

RehanSD:
I think that makes sense to me - but that begets the broader question, why use `np.bool_` at all? I'm fairly certain that we could just use `bool` everywhere - the only benefit from using `np.bool_` (as far as I can tell) is that it is more memory efficient than `bool`. If we do decide to use `np.bool_` I think it would be good to document why we made that decision (e.g. a comment that goes something like `The following APIs use np.bool_ to maintain backwards compatibility/be more memory efficient/etc.`

Thank you for pointing out the CI errors - I'll take a closer look at those later today/tomorrow!

RehanSD:
Just checked and the errors are due to a dependency bug we are currently trying to resolve. #4568 is one particular solution - once we've got something merged for that, you'll be able to pull from master and have CI actually run!

NickCrews:
Yeah that's true, and I think in a nice simple world we DO only use bool. I'm not sure what the implications are. I just started diving into the internals of what `isin = Map.register(pandas.DataFrame.isin, dtypes=np.bool_)` actually does, but then I stopped. Too complicated for me to put that much effort in, someone who is more familiar will have to speak to that.

I agree a comment would be good. At my current level of understanding I think I would say something like `Perhaps we could just use bool here?`, I dont want to pretend I know more than I do, or it might keep the next maintainer from actually making a change that's good.

RehanSD:
Just ran a quick test, which supports the idea that we could probably just swap `np.bool_` with `bool` everywhere - the following code snippet seems to indicate that pandas treats `np.bool_` as `bool` in the backend:
```python
In [1]: import pandas as pd; import numpy as np

In [2]: df = pd.DataFrame([[True, False, True]]*1000, columns=["A", "B", "C"], dtype=np.bool_)

In [3]: df.dtypes
Out[3]:
A    bool
B    bool
C    bool
dtype: object

In [4]: df.info()
<class 'pandas.core.frame.DataFrame'>
RangeIndex: 1000 entries, 0 to 999
Data columns (total 3 columns):
 #   Column  Non-Null Count  Dtype
---  ------  --------------  -----
 0   A       1000 non-null   bool
 1   B       1000 non-null   bool
 2   C       1000 non-null   bool
dtypes: bool(3)
memory usage: 3.1 KB

In [5]: bool_df = pd.DataFrame([[True, False, True]]*1000, columns=["A", "B", "C"])

In [6]: bool_df.dtypes
Out[6]:
A    bool
B    bool
C    bool
dtype: object

In [7]: bool_df.info()
<class 'pandas.core.frame.DataFrame'>
RangeIndex: 1000 entries, 0 to 999
Data columns (total 3 columns):
 #   Column  Non-Null Count  Dtype
---  ------  --------------  -----
 0   A       1000 non-null   bool
 1   B       1000 non-null   bool
 2   C       1000 non-null   bool
dtypes: bool(3)
memory usage: 3.1 KB

In [8]: np_bool_df = pd.DataFrame([[np.bool_(True), np.bool_(False), np.bool_(True)]]*1000, columns=["A", "B", "C"])

In [9]: np_bool_df.dtypes
Out[9]:
A    bool
B    bool
C    bool
dtype: object

In [10]: np_bool_df.info()
<class 'pandas.core.frame.DataFrame'>
RangeIndex: 1000 entries, 0 to 999
Data columns (total 3 columns):
 #   Column  Non-Null Count  Dtype
---  ------  --------------  -----
 0   A       1000 non-null   bool
 1   B       1000 non-null   bool
 2   C       1000 non-null   bool
dtypes: bool(3)
memory usage: 3.1 KB
```

YarShev:
> > @YarShev That followup commit replaced all the other instances of np.bool. But I think in those cases they should get replaced with just bool. Double check me.
> 
> Hi @NickCrews quick question - why are we replacing those instances with `bool` instead of `np.bool_`? I did some checking, and it seems that `bool` and `np.bool_` behave similarly when doing dtypes checking (i.e. `df.dtypes == bool` and `df.dtypes == np.bool_` return the same result), but `np.bool_ != bool` - so unless there's a compelling reason to swap to `bool`, I'd prefer sticking with `np.bool_`, in case there are some edge cases we're missing!
> 
> (I've also approved the run again)!

@RehanSD, I am not sure that is the fair check you did because when comparing `df.dtypes == bool` or `df.dtypes == np.bool_`, as well as `np.bool_ != bool` it seems we are doing different operations.

```python
import pandas as pd; import numpy as np

df = pd.DataFrame([[True, False, True]]*1000, columns=["A", "B", "C"], dtype=np.bool_)
df.dtypes == bool
A    True
B    True
C    True
dtype: bool

df.dtypes == np.bool
DeprecationWarning: `np.bool` is a deprecated alias for the builtin `bool`. To silence this warning, use `bool` by itself. Doing this will not modify any behavior and is safe. If you specifically wanted the numpy scalar type, use `np.bool_` here.
Deprecated in NumPy 1.20; for more details and guidance: https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations
  df.dtypes == np.bool
A    True
B    True
C    True
dtype: bool

df.dtypes == np.bool_
A    True
B    True
C    True
dtype: bool

np.bool_ == bool
False
```

devin-petersohn:
>I will note, as a new contributor the number of hoops I have to jump through, as compared to contributing to say pandas, is annoying. Autogenerated release notes, and looser requirements on how to format commit messages and issue numbers etc would probably make me more likely to contribute in the future. Just my two cents, I'm sure there are reasons for the structure!

Thanks @NickCrews this is good feedback. I think the contribution requirements list has grown and a lot of things have been added to solve one problem or another, but I agree that we should definitely try to soften as many of these requirements. This checklist has grown to be a bit too long: 

image

Most of these requirements were reminders for myself/other maintainers to enforce some good discipline, but don't necessarily need to be so strict.

devin-petersohn:
@RehanSD @YarShev I think the conversation is getting a bit too in the weeds. I think the way @NickCrews has structured this is actually fine: `np.bool_` is used within the query compiler and `bool` is used at the higher layers in Modin (where code should be more readable). I don't think we need to necessarily go so deep. 

I am fine with the changes as-is.

YarShev:
> @RehanSD @YarShev I think the conversation is getting a bit too in the weeds. I think the way @NickCrews has structured this is actually fine: `np.bool_` is used within the query compiler and `bool` is used at the higher layers in Modin (where code should be more readable). I don't think we need to necessarily go so deep.
> 
> I am fine with the changes as-is.

@devin-petersohn, a Modin developer might be confused with this behavior in future and he will have a question why `np.bool_` is used in some places, but Python `bool` is used in others. I am rather for the consistent behavior everywhere if there is no any implication or side effect of using a concrete boolean dtype.

RehanSD:
To be honest, I think @devin-petersohn is right - I'm happy to merge this PR as is and open a new PR to do a refactor.

@YarShev I don't know if `np.bool_` is actually more memory efficient when used with pandas. The `np.bool_` type seems to take up about 3 bits less than `bool`, so in the dfs I created above, we'd expect the `np.bool_` ones to be about a KB less in memory, but both seem to be the same amount of memory. Obviously this could be due to some caching that Python/pandas are doing (since my tests aren't very rigorous, and the whole df seems to take less memory than I'd expect), but it seems to me that pandas is just converting `np.bool_` to `bool` under the hood, negating any memory savings we would expect.

modin/pandas/base.py
```diff
@@ -739,7 +739,7 @@ def all(
                             type(self).__name__, "all"
                         )
                     )
-                data_for_compute = self[self.columns[self.dtypes == np.bool]]
+                data_for_compute = self[self.columns[self.dtypes == bool]]
```
devin-petersohn, line 742:
```suggestion
                data_for_compute = self[self.columns[self.dtypes == np.bool_]]
```

modin/pandas/base.py
```diff
@@ -802,7 +802,7 @@ def any(
...
```
devin-petersohn, line 805:
```suggestion
                data_for_compute = self[self.columns[self.dtypes == np.bool_]]
```

modin/pandas/base.py
```diff
@@ -1105,7 +1105,7 @@ def count(self, axis=0, level=None, numeric_only=False):  # noqa: PR01, RT01, D2
         Count non-NA cells for `BasePandasDataset`.
         """
         axis = self._get_axis_number(axis)
-        frame = self.select_dtypes([np.number, np.bool]) if numeric_only else self
+        frame = self.select_dtypes([np.number, bool]) if numeric_only else self
```
devin-petersohn, line 1108:
```suggestion
        frame = self.select_dtypes([np.number, np.bool_]) if numeric_only else self
```

modin/pandas/series.py
```diff
@@ -2441,7 +2441,7 @@ def _getitem(self, key):
             Series with retrieved data.
         """
         key = apply_if_callable(key, self)
-        if isinstance(key, Series) and key.dtype == np.bool:
+        if isinstance(key, Series) and key.dtype == bool:
```
devin-petersohn, line 2444:
```suggestion
        if isinstance(key, Series) and key.dtype == np.bool_:
```

modin/pandas/test/dataframe/test_reduce.py
```diff
@@ -186,7 +186,7 @@ def test_2195(datetime_is_numeric, has_numeric_column):
     [
         ([np.float64], None),
         (np.float64, None),
-        (None, [np.timedelta64, np.datetime64, np.object, np.bool]),
+        (None, [np.timedelta64, np.datetime64, np.object, bool]),
```
devin-petersohn, line 189:
```suggestion
        (None, [np.timedelta64, np.datetime64, np.object, np.bool_]),
```

modin/pandas/test/test_series.py
```diff
@@ -1393,16 +1393,16 @@ def test_describe(data):
 
     try:
         pandas_result = pandas_series.describe(
-            include=[np.timedelta64, np.datetime64, np.object, np.bool]
+            include=[np.timedelta64, np.datetime64, np.object, bool]
```
devin-petersohn, line 1396:
```suggestion
            include=[np.timedelta64, np.datetime64, np.object, np.bool_]
```

modin/pandas/test/test_series.py
```diff
@@ -1393,16 +1393,16 @@ def test_describe(data):
...
         )
     except Exception as e:
         with pytest.raises(type(e)):
             modin_series.describe(
-                include=[np.timedelta64, np.datetime64, np.object, np.bool]
+                include=[np.timedelta64, np.datetime64, np.object, bool]
```
devin-petersohn, line 1401:
```suggestion
                include=[np.timedelta64, np.datetime64, np.object, np.bool_]
```

modin/pandas/test/test_series.py
```diff
@@ -1393,16 +1393,16 @@ def test_describe(data):
...
             )
     else:
         modin_result = modin_series.describe(
-            include=[np.timedelta64, np.datetime64, np.object, np.bool]
+            include=[np.timedelta64, np.datetime64, np.object, bool]
```
devin-petersohn, line 1405:
```suggestion
            include=[np.timedelta64, np.datetime64, np.object, np.bool_]
```

devin-petersohn (commented):
@YarShev I don't necessarily agree that a Modin developer would get confused, especially given the `git blame` would link back to this discussion, but I have made it easy for @NickCrews to make it all consistent.

YarShev:
@devin-petersohn, makes sense to me, thanks! @NickCrews, please rebase on current master to fix failures in CI.

NickCrews:
Suggestions applied. Thank you all! That was easy to apply those suggestions. I haven't used that feature before but that was slick!

YarShev (dismissed):
@NickCrews, thanks for the changes! Will look forward to seeing further contributions from your side.

NickCrews:
I'm nervous that adding a oneline fixer of `or (inc in [bool, np.bool_] and d in [bool, np.bool_])` is a weird exceptional case that separates bool/np.bool_ from all the other dtypes. Seems like a code smell, and we should solve this in a more generalized way.

Looking at how pandas does this, their `describe()` uses `select_dtypes()` so that those two methods are guaranteed to have the same behavior and always be in sync. Do we want to / can we do the same thing? Or somehow merge that logic together so they both use the same helper functions?

IDK, could just do this tweak, merge, and then fixup later, but then it runs the risk of being a TODO that never is fixed 😆

RehanSD:
That makes sense - but I think there may be an easier way to handle this - we can just use `pandas.core.dtypes.common.pandas_dtype` which converts a type to a pandas or numpy dtype. The result of calling this on `np.bool_` is `bool`, which works well, so we can just map this across `include` and `exclude` in both `select_dtypes` and `describe`.

NickCrews:
@RehanSD OK that seems like a nice simple alternative. Do you think you could actually implement it though? I am getting blocked from being able to work on this locaclly by https://github.com/modin-project/modin/issues/4579

RehanSD:
Of course - happy to do so (unless you'd prefer to do it, since it seems that Mahesh may have unblocked you)?

NickCrews:
@RehanSD thank you, that commit looks great! @YarShev I think this is ready for review again.

YarShev:
Thank you guys, LGTM!

RehanSD:
Thank you @NickCrews! I've gone ahead and merged this!

PR #4583, opened 2022-06-17
FIX-#4582: Inherit custom log layer

### LINKED ISSUE #4582: Modin custom logging layer is not inherited
### System information
- **Modin version** (`modin.__version__`): 81648e711f7525d87dbc58f53e67a2fd22650ae7

### Describe the problem
It's possible to override default logging layer (`PANDAS-API`) when constructing a class, like
```python
class Foo(ClassLogger, modin_layer="CUSTOM"):
    def method1(self):
        pass
```
but then it won't get inherited, i.e.
```python
class Bar(Foo):
    def method2(self):
        pass
```
would get default logging layer again, which feels wrong.

vnlitvinov:

## What do these changes do?
Fix inheritance of custom log layer. It's still possible to use `PANDAS-API` log layer in subclasses, but to do so one has to set it explicitly.

- [x] Resolves #4582
- [x] tests added and passing

### DIFF
--- a/modin/logging/class_logger.py
+++ b/modin/logging/class_logger.py
@@ -31,10 +31,12 @@ class ClassLogger:
     This mixin must go first in class bases declaration to have the desired effect.
     """
 
+    _modin_logging_layer = "PANDAS-API"
+
     @classmethod
     def __init_subclass__(
         cls,
-        modin_layer: str = "PANDAS-API",
+        modin_layer: Optional[str] = None,
         class_name: Optional[str] = None,
         log_level: Optional[str] = "info",
         **kwargs,
@@ -53,5 +55,7 @@ class ClassLogger:
             The log level (INFO, DEBUG, WARNING, etc.).
         **kwargs : dict
         """
+        modin_layer = modin_layer or cls._modin_logging_layer
         super().__init_subclass__(**kwargs)
         enable_logging(modin_layer, class_name, log_level)(cls)
+        cls._modin_logging_layer = modin_layer
--- a/docs/release_notes/release_notes-0.16.0.rst
+++ b/docs/release_notes/release_notes-0.16.0.rst
@@ -9,6 +9,7 @@ Key Features and Updates
   * FIX-#4570: Replace ``np.bool`` -> ``np.bool_`` (#4571)
   * FIX-#4543: Fix `read_csv` in case skiprows=<0, []> (#4544)
   * FIX-#4059: Add cell-wise execution for binary ops, fix bin ops for empty dataframes (#4391)
+  * FIX-#4582: Inherit custom log layer (#4583)
 * Performance enhancements
   * PERF-#4182: Add cell-wise execution for binary ops, fix bin ops for empty dataframes (#4391)
 * Benchmarking enhancements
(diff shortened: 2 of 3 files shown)


modin/logging/class_logger.py
```diff
@@ -31,10 +31,12 @@ class ClassLogger:
     This mixin must go first in class bases declaration to have the desired effect.
     """
 
+    _modin_logging_layer = "PANDAS-API"
+
     @classmethod
     def __init_subclass__(
         cls,
-        modin_layer: str = "PANDAS-API",
+        modin_layer: Optional[str] = None,
```
naren-ponder, line 39: Why do we need the change here on Line 39 and Line 58? Maybe I am missing something, but does just adding  `cls._modin_logging_layer = modin_layer` after defining `_modin_logging_layer` fix this issue?

If not, can we also update the function comment to reflect the `Optional[str] = None`?

devin-petersohn: @naren-ponder I think the Pythonic way here is to default to `None`, I don't think what you describe will work because `cls` is not in scope in the signature.

devin-petersohn (approved):
LGTM, but will wait for @naren-ponder to agree.

modin/logging/class_logger.py
naren-ponder: @devin-petersohn Makes sense.

naren-ponder (approved):
LGTM 🚢

modin/logging/class_logger.py
```diff
@@ -31,10 +31,12 @@ class ClassLogger:
     This mixin must go first in class bases declaration to have the desired effect.
     """
 
+    _modin_logging_layer = "PANDAS-API"
```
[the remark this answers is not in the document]
vnlitvinov: If you mean declare the string as a constant outside of the class (and re-use it in this line to set the variable) then yes, sure, it will work. But why?..

Unless we want to re-use the same constant in the `@enable_logging`...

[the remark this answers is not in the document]
devin-petersohn: Sorry, was responding to the comment here: https://github.com/modin-project/modin/pull/4583#discussion_r902463682

I thought you were saying this didn't work as we wanted it to, my mistake. Is this ok to merge?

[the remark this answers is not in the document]
vnlitvinov: Unless you want me to extract this constant (which I don't see much value in, TBH), it should be good to go.

vnlitvinov:
@devin-petersohn @naren-ponder I had to merge to solve the conflict, please re-approve and let's merge the thing if no objections.

modin/logging/class_logger.py
```diff
@@ -53,5 +55,7 @@ def __init_subclass__(
             The log level (INFO, DEBUG, WARNING, etc.).
         **kwargs : dict
         """
+        modin_layer = modin_layer or cls._modin_logging_layer
         super().__init_subclass__(**kwargs)
         enable_logging(modin_layer, class_name, log_level)(cls)
+        cls._modin_logging_layer = modin_layer
```
jeffreykennethli, line 61: Does this change the `_modin_logging_layer` for all children of `ClassLogger` or just the children of `ClassLogger` and class `Foo(ClassLogger)`? For example, if there are classes:

`Foo(ClassLogger, modin_layer="Custom")`, `Bar(Foo)` and `Other(ClassLogger)`, is `Other` guaranteed to have `modin_layer="PANDAS-API"`?

RehanSD (approved):
LGTM!

devin-petersohn (approved):

PR #4585, opened 2022-06-17
FIX-#4584: Enable pdb debug when running cloud tests

### LINKED ISSUE #4584: Facilitate pdb usage in Modin-in-the-Cloud
Currently it's impossible to use `pdb` when debugging some failure on remote side of the RPyC tunnel, even if tests are running locally, because `pdb` requires stdin/stdout/stderr to be user-controllable, and current implementation of `LocalConnection` hides those either in `subprocess.PIPE` or in `subprocess.DEVNULL`.

vnlitvinov:

## What do these changes do?
Enable using `pdb.set_trace()` and stuff like that on remote side of cloud connection, to facilitate debugging cloud-related issues.

- [x] Resolves #4584 
- [x] tests added and passing

### DIFF
--- a/modin/conftest.py
+++ b/modin/conftest.py
@@ -129,6 +129,18 @@ def simulate_cloud(request):
     if mode == "off":
         yield
         return
+    if (
+        request.config.getoption("usepdb")
+        and request.config.getoption("capture") != "no"
+    ):
+        with request.config.pluginmanager.getplugin(
+            "capturemanager"
+        ).global_and_fixture_disabled():
+            print(
+                "WARNING! You're running tests in simulate-cloud mode. "
+                + "To enable pdb in remote side please disable output capturing "
+                + "by passing '-s' or '--capture=no' to pytest command line"
+            )
 
     if mode not in ("normal", "experimental"):
         raise ValueError(f"Unsupported --simulate-cloud mode: {mode}")
--- a/modin/conftest.py
+++ b/modin/conftest.py
@@ -136,10 +136,10 @@ def simulate_cloud(request):
         with request.config.pluginmanager.getplugin(
             "capturemanager"
         ).global_and_fixture_disabled():
-            print(
+            sys.stderr.write(
                 "WARNING! You're running tests in simulate-cloud mode. "
                 + "To enable pdb in remote side please disable output capturing "
-                + "by passing '-s' or '--capture=no' to pytest command line"
+                + "by passing '-s' or '--capture=no' to pytest command line\n"
             )
 
     if mode not in ("normal", "experimental"):
--- a/modin/experimental/cloud/local_cluster.py
+++ b/modin/experimental/cloud/local_cluster.py
@@ -22,6 +22,10 @@ from .rpyc_proxy import WrappingConnection, WrappingService
 from .tracing.tracing_connection import TracingWrappingConnection
 from modin.config import DoTraceRpyc
 
+_IS_PYTEST_DEBUG = (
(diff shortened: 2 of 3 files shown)


modin/experimental/cloud/local_cluster.py
```diff
@@ -57,9 +61,9 @@ def _run(self, sshcmd: list, cmd: list, capture_out: bool = True):
             redirect.write(f"Running: {cmd}\n")
         return subprocess.Popen(
             cmd,
-            stdin=subprocess.DEVNULL,
-            stdout=redirect,
-            stderr=redirect,
+            stdin=None if _IS_PYTEST_DEBUG and not capture_out else subprocess.DEVNULL,
```
jeffreykennethli, line 64: I'm not familiar with these test fixtures, so could you explain these lines? If you're running `pdb` (and `not capture_out`) why should `stdin`, `stdout`, and `stderr` be None?

devin-petersohn (dismissed):
LGTM, will wait to merge until @jeffreykennethli questions are answered

modin/experimental/cloud/local_cluster.py
vnlitvinov: These are not fixtures, it's a normal `subprocess.Popen()` being called here.
And the reason for those to be `None` is so the "remote" end actually can be interacted with using std streams.

vnlitvinov:
@jeffreykennethli @devin-petersohn can we approve and merge this?

jeffreykennethli:
> @jeffreykennethli @devin-petersohn can we approve and merge this?

Makes sense! Ty for the explanations @vnlitvinov

modin/experimental/cloud/local_cluster.py
```diff
@@ -22,6 +22,10 @@
 from .tracing.tracing_connection import TracingWrappingConnection
 from modin.config import DoTraceRpyc
 
+_IS_PYTEST_DEBUG = (
+    any(arg.endswith("pytest") for arg in sys.argv) and "--pdb" in sys.argv
```
YarShev, line 26: Will `_IS_PYTEST_DEBUG` be True for these both cases?

```bash
$ python -m pytest --pdb ... # 1
$ pytest --pdb ... # 2
```

modin/conftest.py
```diff
@@ -129,6 +129,18 @@ def simulate_cloud(request):
     if mode == "off":
         yield
         return
+    if (
+        request.config.getoption("usepdb")
+        and request.config.getoption("capture") != "no"
```
YarShev, line 134: Should we check the presence of '-s' as well?

YarShev (commented):
@vnlitvinov, left a couple of comments. Also, it is probably worth adding some documentation on this. What do you think?

devin-petersohn (commented):
@vnlitvinov This looks pretty good to me, but there are a couple of questions from @YarShev. Will let @YarShev make the final approval.

modin/conftest.py
vnlitvinov: it's `capture` option 😉 checked at exactly the line you left the comment

vnlitvinov:
> Also, it is probably worth adding some documentation on this. What do you think?

I think all this feature is rather niche (and I don't expect a lot of contributions here), and I think anyone who's going to dive into developing it would find this place easy enough.

That said, I'm too lazy to write a proper "cloud debugging guide" now, and just stating "run `pytest --pdb -s` if your simulated cloud tests are failing" isn't that useful IMHO (especially given `pytest --pdb` _without_ `-s` would tell about that option anyway).

YarShev (approved):
@vnlitvinov, LGTM, thanks!

PR #4587, opened 2022-06-20
FIX-#4527: Log Modin Ray execution

### LINKED ISSUE #4527: Add Modin logging to engine layer
Add Modin logging to the Partition Managers and Partition (including Axis/virtual) layers for different execution engines (Ray, Dask to start). This will help us to debug parallelism related bugs, such as #4464 and others.

jeffreykennethli:
## What do these changes do?

Adds logging (via `ClassLogger`) to the Partition layer (partition manager, partition, axis partition layer), and execution (Ray remote functions), and disables logging for callback-functions like `read_callback` and `deploy_axis_func`, which are called from within a Ray worker and should be logged there.

Note:
Due to the complex interaction between python logging, Ray workers, and Modin, it is necessary to set the runtime environment variable for the Ray tasks. Ray tasks execute as separate processes, and while it is possible to share data between processes, such as a logger instance, file handler, and env vars like whether `LogMode == enabled`, it would require a rewrite of our interface with Ray. Currently, this implementation will set `LogMode` once for each task but never update it. This seems bad, but is also basically the behavior or Ray timeline (assumes envvars are static).

To run/test:
Run a script with `LogMode.enable()` that spawns Ray workers, such as `read_csv` on a semi-large file. I would also recommend limiting `NPartitions` to be 4 to reduce the number of spawned workers for testing. The resultant log file structure should be:

```
.modin/logs/job_<uuid>
    trace.log
    errors.log
    memory.log
    <worker_id_1>.log
    <worker_id_2>.log
```

- [ ] Resolves #4527 
- [ ] tests added and passing

### DIFF
--- a/docs/release_notes/release_notes-0.16.0.rst
+++ b/docs/release_notes/release_notes-0.16.0.rst
@@ -28,6 +28,7 @@ Key Features and Updates
   * PERF-#4288: Improve perf of `groupby.mean` for narrow data (#4591)
   * PERF-#4325: Improve perf of multi-column assignment in `__setitem__` when no new column names are assigning (#4455)
 * Benchmarking enhancements
+  *
 * Refactor Codebase
   * REFACTOR-#4530: Standardize access to physical data in partitions (#4563)
   * REFACTOR-#4534: Replace logging meta class with class decorator (#4535)
--- a/docs/release_notes/release_notes-0.16.0.rst
+++ b/docs/release_notes/release_notes-0.16.0.rst
@@ -28,7 +28,6 @@ Key Features and Updates
   * PERF-#4288: Improve perf of `groupby.mean` for narrow data (#4591)
   * PERF-#4325: Improve perf of multi-column assignment in `__setitem__` when no new column names are assigning (#4455)
 * Benchmarking enhancements
-  * FIX-#4527: Log Modin Ray execution (#4587)
 * Refactor Codebase
   * REFACTOR-#4530: Standardize access to physical data in partitions (#4563)
   * REFACTOR-#4534: Replace logging meta class with class decorator (#4535)
@@ -55,7 +54,7 @@ Key Features and Updates
 * New Features
   * FEAT-4463: Add experimental fuzzydata integration for testing against a randomized dataframe workflow (#4556)
   * FEAT-#4419: Extend virtual partitioning API to pandas on Dask (#4420)
-
+  * FIX-#4527: Log Modin Ray execution (#4587)
 
 Contributors
 ------------
--- a/docs/release_notes/release_notes-0.16.0.rst
+++ b/docs/release_notes/release_notes-0.16.0.rst
@@ -28,7 +28,7 @@ Key Features and Updates
   * PERF-#4288: Improve perf of `groupby.mean` for narrow data (#4591)
   * PERF-#4325: Improve perf of multi-column assignment in `__setitem__` when no new column names are assigning (#4455)
 * Benchmarking enhancements
-  *
+  * FIX-#4527: Log Modin execution (#4587)
 * Refactor Codebase
   * REFACTOR-#4530: Standardize access to physical data in partitions (#4563)
   * REFACTOR-#4534: Replace logging meta class with class decorator (#4535)
@@ -70,3 +70,4 @@ Contributors
 @anmyachev
 @d33bs
 @noloerino
+@jeffreykennethli
--- a/docs/release_notes/release_notes-0.16.0.rst
+++ b/docs/release_notes/release_notes-0.16.0.rst
@@ -28,7 +28,7 @@ Key Features and Updates
   * PERF-#4288: Improve perf of `groupby.mean` for narrow data (#4591)
   * PERF-#4325: Improve perf of multi-column assignment in `__setitem__` when no new column names are assigning (#4455)
 * Benchmarking enhancements
-  * FIX-#4527: Log Modin execution (#4587)
+  * FIX-#4527: Log Modin Ray execution (#4587)
 * Refactor Codebase
   * REFACTOR-#4530: Standardize access to physical data in partitions (#4563)
   * REFACTOR-#4534: Replace logging meta class with class decorator (#4535)
--- a/docs/usage_guide/advanced_usage/modin_logging.rst
+++ b/docs/usage_guide/advanced_usage/modin_logging.rst
@@ -33,7 +33,7 @@ In the example below, we enable logging for internal Modin API calls.
 In the next example, we add logging for not only internal Modin API calls, but also for partition metadata, remote execution, and memory profiling.
 We can set the granularity (in seconds) at which the system memory utilization is logged using ``LogMemoryInterval``.
 We can also set the maximum size of the logs (in MBs) using ``LogFileSize``.
-**Note:**: Logging should be enabled before any Modin calls, and remote execution calls cannot be disabled once enabled.
+**Note:**: Logging should be enabled before any Modin calls, and logging of remote execution calls cannot be disabled once enabled.
 
 .. code-block:: python
 
--- a/docs/usage_guide/advanced_usage/modin_logging.rst
+++ b/docs/usage_guide/advanced_usage/modin_logging.rst
@@ -30,10 +30,10 @@ In the example below, we enable logging for internal Modin API calls.
 
   # User code goes here
 
-In the next example, we add logging for not only internal Modin API calls, but also for partition metadata, remote execution, and memory profiling.
+In the next example, we add logging for not only internal Modin API calls, but also for partition metadata, remote execution (if using Ray), and memory profiling.
 We can set the granularity (in seconds) at which the system memory utilization is logged using ``LogMemoryInterval``.
 We can also set the maximum size of the logs (in MBs) using ``LogFileSize``.
-**Note:**: Logging should be enabled before any Modin calls, and logging of remote execution calls cannot be disabled once enabled.
+**Note:**: Logging should be enabled before any Modin calls, and logging of Ray remote execution calls cannot be disabled once enabled.
 
 .. code-block:: python
 
--- a/docs/usage_guide/advanced_usage/modin_logging.rst
+++ b/docs/usage_guide/advanced_usage/modin_logging.rst
@@ -30,9 +30,10 @@ In the example below, we enable logging for internal Modin API calls.
 
   # User code goes here
 
-In the next example, we add logging for not only internal Modin API calls, but also for partition metadata and memory profiling.
+In the next example, we add logging for not only internal Modin API calls, but also for partition metadata, remote execution, and memory profiling.
 We can set the granularity (in seconds) at which the system memory utilization is logged using ``LogMemoryInterval``.
 We can also set the maximum size of the logs (in MBs) using ``LogFileSize``.
+**Note:**: Logging should be enabled before any Modin calls, and remote execution calls cannot be disabled once enabled.
 
 .. code-block:: python
 
--- a/modin/core/dataframe/base/partitioning/axis_partition.py
+++ b/modin/core/dataframe/base/partitioning/axis_partition.py
@@ -14,9 +14,10 @@
 """Base class of an axis partition for a Modin Dataframe."""
 
 from abc import ABC
+from modin.logging import ClassLogger
 
 
-class BaseDataframeAxisPartition(ABC):  # pragma: no cover
+class BaseDataframeAxisPartition(ClassLogger, ABC):  # pragma: no cover
     """
     An abstract class that represents the parent class for any axis partition class.
 
--- a/modin/core/dataframe/base/partitioning/axis_partition.py
+++ b/modin/core/dataframe/base/partitioning/axis_partition.py
@@ -20,14 +20,13 @@ from modin.logging import ClassLogger
 class BaseDataframeAxisPartition(
     ABC,
     ClassLogger,
+    modin_layer="PARTITION"
 ):  # pragma: no cover
     """
     An abstract class that represents the parent class for any axis partition class.
 
     This class is intended to simplify the way that operations are performed.
     """
-
-    _modin_layer = "PARTITION"
 
     def apply(
         self,
--- a/modin/core/dataframe/base/partitioning/axis_partition.py
+++ b/modin/core/dataframe/base/partitioning/axis_partition.py
@@ -20,13 +20,13 @@ from modin.logging import ClassLogger
 class BaseDataframeAxisPartition(
     ABC,
     ClassLogger,
-    modin_layer="PARTITION",
 ):  # pragma: no cover
     """
     An abstract class that represents the parent class for any axis partition class.
 
     This class is intended to simplify the way that operations are performed.
     """
+    _modin_layer = "PARTITION"
 
     def apply(
         self,
--- a/modin/core/dataframe/base/partitioning/axis_partition.py
+++ b/modin/core/dataframe/base/partitioning/axis_partition.py
@@ -26,6 +26,7 @@ class BaseDataframeAxisPartition(
 
     This class is intended to simplify the way that operations are performed.
     """
+
     _modin_layer = "PARTITION"
 
     def apply(
--- a/modin/core/dataframe/base/partitioning/axis_partition.py
+++ b/modin/core/dataframe/base/partitioning/axis_partition.py
@@ -18,9 +18,7 @@ from modin.logging import ClassLogger
 
 
 class BaseDataframeAxisPartition(
-    ABC,
-    ClassLogger,
-    modin_layer="PARTITION"
+    ABC, ClassLogger, modin_layer="PARTITION"
 ):  # pragma: no cover
     """
     An abstract class that represents the parent class for any axis partition class.
--- a/modin/core/dataframe/base/partitioning/axis_partition.py
+++ b/modin/core/dataframe/base/partitioning/axis_partition.py
@@ -17,7 +17,11 @@ from abc import ABC
 from modin.logging import ClassLogger
 
 
-class BaseDataframeAxisPartition(ClassLogger, ABC):  # pragma: no cover
+class BaseDataframeAxisPartition(
+    ABC,
+    ClassLogger,
+    modin_layer="PARTITION",
+):  # pragma: no cover
     """
     An abstract class that represents the parent class for any axis partition class.
 
--- a/modin/core/dataframe/pandas/partitioning/axis_partition.py
+++ b/modin/core/dataframe/pandas/partitioning/axis_partition.py
@@ -125,6 +125,7 @@ class PandasDataframeAxisPartition(BaseDataframeAxisPartition):
         args.extend(self.list_of_blocks)
         return self._wrap_partitions(self.deploy_axis_func(*args, **kwargs))
 
+    # Disable logging for a function that is executed in a remote context, see ray/dask virtual_partition.py
     @classmethod
     @disable_logging
     def deploy_axis_func(
@@ -182,8 +183,9 @@ class PandasDataframeAxisPartition(BaseDataframeAxisPartition):
                     lengths = None
         return split_result_of_axis_func_pandas(axis, num_splits, result, lengths)
 
-    @disable_logging
+    # Disable logging for a function that is executed in a remote context, see ray/dask virtual_partition.py
     @classmethod
+    @disable_logging
     def deploy_func_between_two_axis_partitions(
         cls,
         axis,
--- a/modin/core/dataframe/pandas/partitioning/axis_partition.py
+++ b/modin/core/dataframe/pandas/partitioning/axis_partition.py
@@ -19,6 +19,7 @@ from modin.core.storage_formats.pandas.utils import split_result_of_axis_func_pa
 from modin.core.dataframe.base.partitioning.axis_partition import (
     BaseDataframeAxisPartition,
 )
+from modin.logging import disable_logging
 
 
 class PandasDataframeAxisPartition(BaseDataframeAxisPartition):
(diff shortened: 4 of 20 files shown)


modin/core/dataframe/pandas/partitioning/axis_partition.py
```diff
@@ -125,6 +126,7 @@ def shuffle(self, func, lengths, **kwargs):
         return self._wrap_partitions(self.deploy_axis_func(*args, **kwargs))
 
     @classmethod
+    @disable_logging
```
vnlitvinov, line 129: can you add a comment why it's disabled?

modin/core/execution/ray/common/task_wrapper.py
```diff
@@ -18,11 +18,15 @@
 """
 
 import asyncio
-
+from modin.config import LogMode
+from modin.logging import get_worker_logger, ClassLogger
+from modin.logging.config import JOB_ID
 import ray
 
+__WORKER_ENV__ = {"env_vars": {"MODIN_LOG_MODE": LogMode.get()}}
```
vnlitvinov, line 26: please add a note to the docs about "set before running, cannot change later" behaviour

modin/core/execution/ray/common/task_wrapper.py
```diff
@@ -39,10 +43,14 @@ def _deploy_ray_func(func, args):  # pragma: no cover
     ray.ObjectRef or list
         Ray identifier of the result being put to Plasma store.
     """
-    return func(**args)
+    logger = get_worker_logger(JOB_ID)
+    logger.info(f"START::PANDAS-API::deploy_ray_func.{func.__qualname__}")
+    res = func(**args)
```
vnlitvinov, line 48: I see no exception handling, it would lead to logging unpaired `START` if any error is raised in `func`

modin/core/execution/ray/common/task_wrapper.py
```diff
@@ -39,10 +43,14 @@ def _deploy_ray_func(func, args):  # pragma: no cover
     ray.ObjectRef or list
         Ray identifier of the result being put to Plasma store.
     """
-    return func(**args)
+    logger = get_worker_logger(JOB_ID)
+    logger.info(f"START::PANDAS-API::deploy_ray_func.{func.__qualname__}")
```
vnlitvinov, line 47:
```suggestion
    logger.info(f"START::RAY-REMOTE::deploy_ray_func:{func.__qualname__}")
```
I suggest changing both "modin layer" to signify it's Ray and to change dot to colon to highlight that `func.__qualname__` is not a child of `deploy_ray_func` (however the latter is questionable as it could potentially break log parsing if it exists already)

modin/core/dataframe/base/partitioning/axis_partition.py
```diff
@@ -14,9 +14,10 @@
 """Base class of an axis partition for a Modin Dataframe."""
 
 from abc import ABC
+from modin.logging import ClassLogger
 
 
-class BaseDataframeAxisPartition(ABC):  # pragma: no cover
+class BaseDataframeAxisPartition(ClassLogger, ABC):  # pragma: no cover
```
vnlitvinov, line 20: I wonder if we should set some specific `modin_layer="PARTITION"` logging layer here (as `ClassLogger` supports this), to make following logs easier

Here and in other classes, too. Stuffing everything under `PANDAS-API` umbrella feels wrong.

modin/core/execution/ray/implementations/pandas_on_ray/partitioning/partition.py
```diff
@@ -54,16 +58,6 @@ def __init__(self, data, length=None, width=None, ip=None, call_queue=None):
         self._ip_cache = ip
         self._identity = uuid.uuid4().hex
 
-        logger = get_logger()
```
vnlitvinov, line 57: Is there a reason to drop debug logging? `ClassLogger` only logs calls/exits at `.info` and it doesn't log particular details which might be interesting. I would just change the logging from eager formatting to lazy one (specifying the message template and format args as separate things, like `logger.debug("My message: {height}", height=self._length_cache)`, so `logger` would compose the message only when debug is enabled.

Here and below.

modin/core/execution/ray/implementations/pandas_on_ray/partitioning/partition.py
```diff
@@ -345,10 +323,18 @@ def _get_index_and_columns(df):
...
-    return len(df.index), len(df.columns)
+    logger = get_worker_logger(JOB_ID)
+    logger.info(
+        f"START::PANDAS-API::PandasOnRayDataframePartition._get_index_and_columns"
+    )
+    res = len(df.index), len(df.columns)
+    logger.info(
+        f"STOP::PANDAS-API::PandasOnRayDataframePartition._get_index_and_columns"
+    )
+    return res
```
vnlitvinov, line 334: I would rather we implement specific worker logger decorator which would do this instead of copy-pasting this pattern all over the place... plus why `PANDAS-API`?..

modin/core/execution/ray/implementations/pandas_on_ray/partitioning/partition.py
```diff
@@ -375,22 +361,28 @@ def _apply_func(partition, func, *args, **kwargs):  # pragma: no cover
...
         result = func(partition.copy(), *args, **kwargs)
-    return (
+    res = (
         result,
         len(result) if hasattr(result, "__len__") else 0,
         len(result.columns) if hasattr(result, "columns") else 0,
         get_node_ip_address(),
     )
+    logger.info(f"STOP::PANDAS-API::PandasOnRayDataframePartition._apply_func")
+    return res
```
vnlitvinov, line 382: you can keep `return` statement and add logging by using pattern like
```python
try:
    return foo
finally:
    logger.info("bar")
```
it's also better in terms of any exception raised during `return foo` line (in our case, for example, an exception in `get_node_ip_address()`) would _not_ prevent logging `STOP` message

modin/core/execution/ray/implementations/pandas_on_ray/partitioning/virtual_partition.py
```diff
@@ -504,7 +507,10 @@ class PandasOnRayDataframeRowPartition(PandasOnRayDataframeVirtualPartition):
     axis = 1
 
 
-@ray.remote
+__WORKER_ENV__ = {"env_vars": {"MODIN_LOG_MODE": LogMode.get()}}
```
vnlitvinov, line 510: I would suggest to create this in `modin/core/execution/ray/common/utils.py` and import from that everywhere it's needed... or even create a custom decorator in there, like
```python
import ray
import functools

__WORKER_ENV__ = {"env_vars": {"MODIN_LOG_MODE": LogMode.get()}}

logged_remote = functools.partial(ray.remote, runtime_env=__WORKER_ENV__)
```
and use it like
```python
from modin.core.execution.ray.common.utils import logged_remote

@logged_remote(num_returns=2)
def func(args):
    # do stuff
    return a, b
```

modin/core/execution/ray/implementations/pandas_on_ray/partitioning/virtual_partition.py
```diff
@@ -527,13 +533,21 @@ def deploy_ray_func(func, *args, **kwargs):  # pragma: no cover
...
     if isinstance(result, pandas.DataFrame):
-        return result, len(result), len(result.columns), ip
+        res = result, len(result), len(result.columns), ip
     elif all(isinstance(r, pandas.DataFrame) for r in result):
-        return [i for r in result for i in [r, len(r), len(r.columns), ip]]
+        res = [i for r in result for i in [r, len(r), len(r.columns), ip]]
     else:
-        return [i for r in result for i in [r, None, None, ip]]
+        res = [i for r in result for i in [r, None, None, ip]]
+    logger.info(
```
vnlitvinov, line 550: same about using
```python
try:
    if a: return foo
    if b: return bar
finally:
    logger.info("baz")
```
pattern

modin/core/storage_formats/pandas/parsers.py
```diff
@@ -140,6 +141,7 @@ def find_common_type_cat(types):
 class PandasParser(ClassLogger):
     """Base class for parser classes with pandas storage format."""
 
+    @disable_logging
```
vnlitvinov, line 144: why?

modin/pandas/accessor.py
```diff
@@ -178,6 +179,7 @@ def to_dense(self):
 
 @_inherit_docstrings(pandas.core.accessor.CachedAccessor)
 class CachedAccessor(ClassLogger):
+    @disable_logging
```
vnlitvinov, line 182: why?

vnlitvinov (commented):
Overall looks good, left a few comments, though.

And please fix the linting etc in the CI.

modin/core/dataframe/base/partitioning/axis_partition.py
devin-petersohn: +1 to @vnlitvinov I think we want to make sure that log column has some meaning 😄

modin/logging/__init__.py
```diff
@@ -12,12 +12,16 @@
 # governing permissions and limitations under the License.
 
 from .class_logger import ClassLogger  # noqa: F401
-from .config import get_logger  # noqa: F401
-from .logger_decorator import enable_logging, disable_logging  # noqa: F401
+from .config import get_logger, get_worker_logger  # noqa: F401
+from .logger_decorator import (
+    enable_logging,
+    disable_logging,
+)  # noqa: F401
```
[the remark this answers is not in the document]
jeffreykennethli: I think this was a black auto-format thing. I'll resolve it somehow!

modin/core/storage_formats/pandas/parsers.py
jeffreykennethli: I don't think this needs to be logged, will remove.

modin/pandas/accessor.py
jeffreykennethli: Removing this.

modin/core/execution/ray/implementations/pandas_on_ray/partitioning/partition.py
jeffreykennethli: Added a decorator `enable_remote_logging` which does this.

modin/core/dataframe/base/partitioning/axis_partition.py
jeffreykennethli: I tried doing `modin_layer="PARTITION"` in the class declaration as you mentioned, but the `modin_layer` variable was not being propagated to all subclasses. E.g. `PandasDataframePartition` would be `PARTITION` but `PandasOnRayDataframePartition` would be `PANDAS-API`.

I think the reason is that `__init_subclass__` is called from the perspective of `cls` (`PandasOnRayDataframePartition`) to `ClassLogger`, so it doesn't have the `modin_layer` of its immediate parent, `PandasDataframePartition`. Perhaps we could add this to the class declaration for each subclass. What I've implemented is another way, which is to "propagate" the var by making it a class variable, and making `ClassLogger` retrieve that class variable.

devin-petersohn (commented):
I think this looks good. Will let @vnlitvinov be the final approval since he has more context.

jeffreykennethli:
Per @vnlitvinov this PR can be refactored so that the Ray runtime env is set when we call `ray.init` instead of at the Task level init. Marking this PR as a draft until this is done.

jeffreykennethli:
Rebased on master to get #4583 and #4620. Now the inheritance is done as originally designed (not through a class variable), and type checking is added to `logging`. I also moved the Ray task `runtime_env` variable to the `ray.init` kwargs, to avoid having to use a partial in place of all `@ray.remote` decorators.

docs/release_notes/release_notes-0.16.0.rst
```diff
@@ -17,7 +17,7 @@ Key Features and Updates
   * PERF-#4182: Add cell-wise execution for binary ops, fix bin ops for empty dataframes (#4391)
   * PERF-#4288: Improve perf of `groupby.mean` for narrow data (#4591)
 * Benchmarking enhancements
-  *
+  * FIX-#4527: Log Modin Ray execution (#4587)
```
vnlitvinov, line 20: why it's under benchmarking?.. I think it's a feature

docs/usage_guide/advanced_usage/modin_logging.rst
```diff
@@ -30,9 +30,10 @@ In the example below, we enable logging for internal Modin API calls.
 
   # User code goes here
 
-In the next example, we add logging for not only internal Modin API calls, but also for partition metadata and memory profiling.
+In the next example, we add logging for not only internal Modin API calls, but also for partition metadata, remote execution, and memory profiling.
 We can set the granularity (in seconds) at which the system memory utilization is logged using ``LogMemoryInterval``.
 We can also set the maximum size of the logs (in MBs) using ``LogFileSize``.
+**Note:**: Logging should be enabled before any Modin calls, and remote execution calls cannot be disabled once enabled.
```
vnlitvinov, line 36:
```suggestion
**Note:**: Logging should be enabled before any Modin calls, and logging of remote execution calls cannot be disabled once enabled.
```

modin/core/execution/ray/common/utils.py
```diff
@@ -206,7 +208,14 @@ def initialize_ray(
                 "object_store_memory": object_store_memory,
                 "_redis_password": redis_password,
                 "_memory": object_store_memory,
+                "runtime_env": {
+                    "env_vars": {
+                        "MODIN_LOG_MODE": LogMode.get(),
+                        "MODIN_LOG_FILE_SIZE": str(LogFileSize.get()),
+                    }
+                },
```
vnlitvinov, line 216: I wonder if we should first merge https://github.com/modin-project/modin/pull/4603 and then augment the passed dictionary there (if logging is enabled) so that the configuration is checked for the case of externally-started clusters?

modin/core/execution/ray/implementations/pandas_on_ray/partitioning/partition.py
```diff
@@ -53,15 +53,13 @@ def __init__(self, data, length=None, width=None, ip=None, call_queue=None):
...
-                self._identity,
-                str(self._length_cache),
-                str(self._width_cache),
-                str(self._ip_cache),
-            )
+            "Partition ID: %s, Height: %s, Width: %s, Node IP: %s",
+            self._identity,
+            str(self._length_cache),
+            str(self._width_cache),
+            str(self._ip_cache),
```
vnlitvinov, line 62:
```suggestion
            self._length_cache,
            self._width_cache,
            self._ip_cache,
```
no need to explicitly `str()` things, as `%s` formatter calls that for us

modin/core/storage_formats/pandas/parsers.py
```diff
@@ -292,7 +293,9 @@ def single_worker_read(cls, fname, **kwargs):
 
 @doc(_doc_pandas_parser_class, data_type="CSV files")
 class PandasCSVParser(PandasParser):
+    # Disable logging for a function that is executed in a remote context, see TextFileDispatcher._launch_tasks
     @staticmethod
+    @disable_logging
```
vnlitvinov, line 298: this particular piece feels weird, could you please provide some more verbose reasoning behind it?

modin/logging/logger_decorator.py
```diff
@@ -143,3 +144,66 @@ def run_and_log(*args: Tuple, **kwargs: Dict) -> Any:
...
+        -------
+        Any
+        """
+        # Don't log remote calls if LogMode is disable or api_only.
+        if LogMode.get() != "enable":
+            return func(*args, **kwargs)
+
+        logger = get_worker_logger(JOB_ID)
+        start_line = f"START::{modin_layer.upper()}::{func.__name__}"
+        stop_line = f"STOP::{modin_layer.upper()}::{func.__name__}"
```
vnlitvinov, line 196: these two should be computed outside of `run_and_log` (see `enable_logging` implementation), they never change, we shouldn't be wasting time computing same things over and over again

modin/logging/logger_decorator.py
```diff
@@ -143,3 +144,66 @@ def run_and_log(*args: Tuple, **kwargs: Dict) -> Any:
...
+
+        funcs = [arg for arg in args if callable(arg)]
+        if funcs:
+            start_line += f":{funcs}"
+            stop_line += f":{funcs}"
```
vnlitvinov, line 201:
```suggestion
            start_line += ":".join(f.__name__ for f in funcs)
            stop_line += ":".join(f.__name__ for f in funcs)
```
so we get colon-separated functions?.. or any other join, 'cause right now what we would end up with is something like `"START::RAY-REMOTE::my_func:'[<function foo1 at 0x000002B3A2B5F040>, <function foo2 at 0x000002B3F2B5F040>]'"` which looks weird (IMHO)

modin/core/execution/ray/common/task_wrapper.py
```diff
@@ -18,11 +18,12 @@
 """
 
 import asyncio
-
+from modin.logging import ClassLogger, enable_remote_logging
 import ray
 
 
-@ray.remote
+@ray.remote(num_returns=2)
```
vnlitvinov, line 25: why change?.. also, as this is a general wrapper, I don't think we can know upfront how many items it would return

modin/core/storage_formats/pandas/parsers.py
jeffreykennethli: Added more context for this.

docs/release_notes/release_notes-0.16.0.rst
```diff
@@ -24,7 +24,6 @@ Key Features and Updates
   * PERF-#4288: Improve perf of `groupby.mean` for narrow data (#4591)
   * PERF-#4325: Improve perf of multi-column assignment in `__setitem__` when no new column names are assigning (#4455)
 * Benchmarking enhancements
-  *
```
vnlitvinov, line 27: why?

modin/core/execution/ray/common/task_wrapper.py
```diff
@@ -18,11 +18,12 @@
 """
 
 import asyncio
-
+from modin.logging import ClassLogger, enable_remote_logging
 import ray
```
vnlitvinov, line 22: we should import non-Modin first, then empty line, then Modin-defined modules

modin/core/execution/ray/common/utils.py
```diff
@@ -41,6 +43,60 @@
     ObjectIDType = (ray.ObjectRef, ClientObjectRef)
 
 
+def _move_stdlib_ahead_of_site_packages(*args):
```
vnlitvinov, line 46: this was removed in #4603, pls not re-add it

modin/core/execution/ray/common/utils.py
```diff
@@ -41,6 +43,60 @@
...
+        # - python3.x
+        #   - typing.py
+        #   - site-packages/
+        #     - pandas
+        # So extracting the dirname of the site_packages can point us
+        # to the directory containing standard libraries.
+        sys.path.insert(site_packages_path_index, os.path.dirname(site_packages_path))
+
+
+def _import_pandas(*args):
```
vnlitvinov, line 80: again, pls not re-add what was removed in #4603

modin/core/execution/ray/common/utils.py
```diff
@@ -64,7 +120,15 @@ def initialize_ray(
         What password to use when connecting to Redis.
         If not specified, ``modin.config.RayRedisPassword`` is used.
     """
-    extra_init_kw = {"runtime_env": {"env_vars": {"__MODIN_AUTOIMPORT_PANDAS__": "1"}}}
+    extra_init_kw = {
+        "runtime_env": {
+            "env_vars": {
+                "__MODIN_AUTOIMPORT_PANDAS__": "1",
+                "MODIN_LOG_MODE": LogMode.get(),
+                "MODIN_LOG_FILE_SIZE": str(LogFileSize.get()),
```
vnlitvinov, line 128: I would rather you only pass environment variables if they have non-default values, as this value is shown to the user in a warning

modin/core/storage_formats/pandas/parsers.py
vnlitvinov: I wonder what happens in non-remote context...

modin/logging/logger_decorator.py
vnlitvinov: this feels super weird, got a stacktrace?..

vnlitvinov (changes requested):

modin/core/storage_formats/pandas/parsers.py
jeffreykennethli: In a non-remote context (running a script that does call `PandasCSVParser.parse` without the remote logging enabled), the parse is not logged since it's invoked in a process (Ray process) that does not have access to the Modin logger.

modin/logging/logger_decorator.py
jeffreykennethli: https://gist.github.com/jeffreykennethli/4cd7606e68718000b9467e9f47b6c9a0. also see https://github.com/jeffreykennethli/modin/blob/log_engine_2/modin/logging/logger_decorator.py#L149-L215

modin/core/execution/ray/common/utils.py
```diff
@@ -65,6 +68,19 @@ def initialize_ray(
         If not specified, ``modin.config.RayRedisPassword`` is used.
     """
     extra_init_kw = {"runtime_env": {"env_vars": {"__MODIN_AUTOIMPORT_PANDAS__": "1"}}}
+
+    # Set Modin logging envvars if not default (if set).
+    if LogMode.get() != LogMode._get_default():
```
vnlitvinov, line 73:
```suggestion
    if LogMode.get_value_source() != ValueSource.DEFAULT:
```
without using protected methods (you'd have to import `ValueSource` from `modin.config` first, obviously)

for all configuration variables

YarShev:
Closed as outdated. Most of the changes have already been done. Remote logging is provided by such tools as ray timelime, for instance.

PR #4588, opened 2022-06-20
FIX-#4577: Set attribute of Modin dataframe to updated value

### LINKED ISSUE #4577: pd.Series.ffill() raise the error: AttributeError: 'numpy.ndarray' object has no attribute 'ffill'
### System information
- **OS Platform and Distribution - WIndows10**:
- **Modin version** (`0.15.0+7.g4ec7f634`):
- **Python version 3.9.12**:

```python
  import modin
  import modin.pandas as pd
  
  import numpy as np
  
  
  if __name__ == '__main__':
      if pd.__name__ == 'modin.pandas':
          client = Client(n_workers=3)
          print(modin.__version__)

    df = pd.DataFrame(
        dict(
            a=[1, 2, None, None, None, ],
            b=(1, None, 3, 4, 5,),
        )
    )

    df.a.ffill(inplace=True)
    print(df)

    df['tr_id'] = 0

    df.tr_id = np.where(
        (df.b <= 4),
        3,
        None,
    )

    df.tr_id.ffill(inplace=True)
```

### Describe the problem

I have the output:

```
0.15.0+7.g4ec7f634

    import ray
    ray.init()

     a    b
0  1.0  1.0
1  2.0  NaN
2  2.0  3.0
3  2.0  4.0
4  2.0  5.0
  File "d:\OD\OneDrive\Projects\Chud_Amaz\Soft_in_dev\moduled_way_OOP\modin_test_ffill.py", line 33, in <module>
    df.tr_id.ffill(inplace=True)
AttributeError: 'numpy.ndarray' object has no attribute 'ffill'
```

### **Problem**
When new column was filled with `np.where ` - method `ffill` does not work

pyrito:

## What do these changes do?

This PR addresses issue #4577 which ended up being quite a simple fix but involved a fair amount of soul-searching (thank you to @RehanSD for helping me with this on your day off). Previously, update to DataFrame columns via attribute manipulation would handle things correctly internally, but wouldn't set the attribute properly for the Modin DataFrame. As a result, if a column were to be set to a list, it would return a list when we tried to access the column by attribute (when it really should have been a pandas Series). The test included has a good example of this behavior. 

- [x] Resolves #4577 
- [x] tests added and passing

### DIFF
--- a/modin/pandas/dataframe.py
+++ b/modin/pandas/dataframe.py
@@ -2498,6 +2498,10 @@ class DataFrame(BasePandasDataset):
             pass
         elif key in self and key not in dir(self):
             self.__setitem__(key, value)
+            # Note: return immediately so we don't keep this `key` as dataframe state.
+            # `__getattr__` will return the columns not present in `dir(self)`, so we do not need
+            # to manually track this state in the `dir`.
+            return
         elif isinstance(value, pandas.Series):
             warnings.warn(
                 "Modin doesn't allow columns to be created via a new attribute name - see "
--- a/modin/pandas/test/dataframe/test_iter.py
+++ b/modin/pandas/test/dataframe/test_iter.py
@@ -248,7 +248,8 @@ def test_inplace_series_ops(data):
         df_equals(modin_df, pandas_df)
 
 
-def test___setattr__():
+# Note: Tests setting an attribute that is not an existing column label
+def test___setattr__not_column():
     pandas_df = pandas.DataFrame([1, 2, 3])
     modin_df = pd.DataFrame([1, 2, 3])
 
@@ -256,6 +257,28 @@ def test___setattr__():
     modin_df.new_col = [4, 5, 6]
 
     df_equals(modin_df, pandas_df)
+
+    # While `new_col` is not a column of the dataframe,
+    # it should be accessible with __getattr__.
+    assert modin_df.new_col == pandas_df.new_col
+
+
+def test___setattr__mutating_column():
+    # Use case from issue #4577
+    pandas_df = pandas.DataFrame([[1]], columns=["col0"])
+    modin_df = pd.DataFrame([[1]], columns=["col0"])
+
+    # Replacing a column with a list should mutate the column in place.
+    pandas_df.col0 = [3]
+    modin_df.col0 = [3]
+
+    df_equals(modin_df, pandas_df)
+    # Check that the col0 attribute reflects the value update.
+    df_equals(modin_df.col0, pandas_df.col0)
+    modin_df.col0 = pd.Series([5])
+    modin_df.loc[0, "col0"] = 4
+
+    assert modin_df.col0.equals(modin_df["col0"])
 
 
 @pytest.mark.parametrize("data", test_data_values, ids=test_data_keys)
--- a/modin/pandas/test/dataframe/test_iter.py
+++ b/modin/pandas/test/dataframe/test_iter.py
@@ -275,9 +275,18 @@ def test___setattr__mutating_column():
     df_equals(modin_df, pandas_df)
     # Check that the col0 attribute reflects the value update.
     df_equals(modin_df.col0, pandas_df.col0)
+
+    pandas_df.col0 = pd.Series([5])
     modin_df.col0 = pd.Series([5])
+
+    # Check that the col0 attribute reflects this update
+    df_equals(modin_df, pandas_df)
+
+    pandas_df.loc[0, "col0"] = 4
     modin_df.loc[0, "col0"] = 4
 
+    # Check that the col0 attribute reflects update via loc
+    df_equals(modin_df, pandas_df)
     assert modin_df.col0.equals(modin_df["col0"])
 
 
--- a/docs/release_notes/release_notes-0.16.0.rst
+++ b/docs/release_notes/release_notes-0.16.0.rst
@@ -10,6 +10,7 @@ Key Features and Updates
   * FIX-#4543: Fix `read_csv` in case skiprows=<0, []> (#4544)
   * FIX-#4059: Add cell-wise execution for binary ops, fix bin ops for empty dataframes (#4391)
   * FIX-#4589: Pin protobuf<4.0.0 to fix ray (#4590)
+  * FIX-#4577: Set attribute of Modin dataframe to updated value (#4588)
 * Performance enhancements
   * PERF-#4182: Add cell-wise execution for binary ops, fix bin ops for empty dataframes (#4391)
 * Benchmarking enhancements
@@ -40,3 +41,4 @@ Contributors
 @NickCrews
 @prutskov
 @vnlitvinov
+@pyrito


modin/pandas/dataframe.py
```diff
@@ -2498,6 +2498,9 @@ def __setattr__(self, key, value):
             pass
         elif key in self and key not in dir(self):
             self.__setitem__(key, value)
+            # Note: Get the most up to date value before setting the attribute since
+            # the type of value can change
+            value = self.__getitem__(key)
```
pyrito, line 2503: The alternative to this is to just return instead of setting the value. By doing so, `__getattr__` will actually get called since no attribute is registered and it will look up our "ghost" attribute for us. This is kind of hacky though, so I thought it would be at least more readable to approach things this was.

pyrito: @mvashishtha tagging you here so you're aware of why I went about this way

naren-ponder (commented):
@pyrito Can you add a release note to this PR?

vnlitvinov:
@pyrito I've fixed PR message to explicitly list it resolves the issue you mention in a free form.
GitHub is not smart enough to read your message entirely, but, when some keywords are used, it would understand that this PR is linked to that issue, and it would auto-close the issue when the PR is merged.

modin/pandas/dataframe.py
vnlitvinov: I like current approach more, it's explicit

modin/pandas/test/dataframe/test_iter.py
```diff
@@ -257,6 +257,17 @@ def test___setattr__():
...
+    pandas_df = pandas.DataFrame([[1]], columns=["col0"])
+    modin_df = pd.DataFrame([[1]], columns=["col0"])
+
+    pandas_df.col0 = [3]
+    modin_df.col0 = [3]
+
+    pandas_df.col0.ffill()
+    modin_df.col0.ffill()
+    df_equals(modin_df, pandas_df)
+
```
vnlitvinov, line 270: I would suggest turning it into a separate test

vnlitvinov (commented):
Overall looks great and simple, thanks @pyrito!

modin/pandas/test/dataframe/test_iter.py
devin-petersohn: +1, a new test would be good

modin/pandas/dataframe.py
devin-petersohn: I think in the interest of avoiding carrying extra state in the dataframe, it might be better to just `return`. The extra `__getitem__` isn't free, and if the column names change, I wonder if the behavior would still be correct.

```python
import modin.pandas as pd

df = pd.DataFrame([[1]], columns=['col0'])
df.col0 = [3]
df.columns = ["col1"]
print(df.col0)  # this will still work, but shouldn't
print(df.col1)  # this will also work, and it should
```

mvashishtha: +1 to @devin-petersohn . I think this fix also still suffers from the bug I pointed out [here](https://github.com/modin-project/modin/issues/4577#issuecomment-1158015868). If we mutate `col0` in place in the example above using `iloc`, I think we won't change `df.col0`.

modin/pandas/test/dataframe/test_iter.py
```diff
@@ -257,6 +257,17 @@ def test___setattr__():
...
     df_equals(modin_df, pandas_df)
 
+    # Use case from issue #4577
+    pandas_df = pandas.DataFrame([[1]], columns=["col0"])
+    modin_df = pd.DataFrame([[1]], columns=["col0"])
+
+    pandas_df.col0 = [3]
+    modin_df.col0 = [3]
+
+    pandas_df.col0.ffill()
```
mvashishtha, line 267: We should add a separate test case where we mutate the replaced column in place with something simple like `iloc`, and then check that `df.col0` has the updated value. I think something like my example [here](https://github.com/modin-project/modin/issues/4577#issuecomment-1158015868) will work.

modin/pandas/test/dataframe/test_iter.py
pyrito: Sounds good, I kept it as part of this because I felt like it is testing similar code paths, but I can do that as well.

modin/pandas/test/dataframe/test_iter.py
pyrito: Makes sense and that test seems to cover most things as well. I made changes to my PR to reflect this.

modin/pandas/test/dataframe/test_iter.py
```diff
@@ -258,6 +258,21 @@ def test___setattr__():
     df_equals(modin_df, pandas_df)
 
 
+def test___setattr___getattr__():
```
mvashishtha, line 261:
```suggestion
def test___setattr__mutating_column():
```

mvashishtha, line 261: Could you please rename the previous test case, `test___setattr__`, to `test___setattr__not_column`? Also please add a note to that test case saying that it's testing setting an attribute that is not an existing column label.

modin/pandas/test/dataframe/test_iter.py
```diff
@@ -258,6 +258,21 @@ def test___setattr__():
...
+    # Use case from issue #4577
+    pandas_df = pandas.DataFrame([[1]], columns=["col0"])
+    modin_df = pd.DataFrame([[1]], columns=["col0"])
+
+    pandas_df.col0 = [3]
```
mvashishtha, line 266:
```suggestion
    # Replacing a column with a list should mutate the column in place.
    pandas_df.col0 = [3]
```

modin/pandas/test/dataframe/test_iter.py
```diff
@@ -258,6 +258,21 @@ def test___setattr__():
...
+    modin_df.col0 = [3]
+
+    df_equals(modin_df, pandas_df)
```
mvashishtha, line 269:
```suggestion
    df_equals(modin_df, pandas_df)
    df_equals(modin_df.col0, pandas_df.col0)
```

modin/pandas/dataframe.py
```diff
@@ -2498,6 +2498,10 @@ def __setattr__(self, key, value):
             pass
         elif key in self and key not in dir(self):
             self.__setitem__(key, value)
+            # Note: Return so that we don't keep track of another attribute. Instead
+            # we will register a "ghost" attribute, that will be picked up when
```
devin-petersohn, line 2502: Sorry if this is pedantic, but what is the `"ghost"` attribute? We catch a missing attribute error in `__getattr__` and then return the column if it's a column name. Is this the `"ghost"` attribute?

modin/pandas/test/dataframe/test_iter.py
```diff
@@ -258,6 +258,21 @@ def test___setattr__():
...
+    modin_df.col0 = pd.Series([3])
+    modin_df.iloc[0, 0] = 4
```
mvashishtha, line 271:
```suggestion
    modin_df.loc['col0', 0] = 4
```

I think using the column label will make the test a little clearer.

modin/pandas/test/dataframe/test_iter.py
```diff
@@ -258,6 +258,21 @@ def test___setattr__():
...
+
+    assert modin_df.col0.equals(modin_df["col0"])
```
mvashishtha, line 273:
```suggestion
    df_equals(modin_df, pandas_df)
    # Check that the col0 attribute reflects the value update.
    df_equals(modin_df.col0, pandas_df.col0)
```

modin/pandas/dataframe.py
pyrito: I went ahead and rewrote this comment

modin/pandas/dataframe.py
```diff
@@ -2498,6 +2498,9 @@ def __setattr__(self, key, value):
             pass
         elif key in self and key not in dir(self):
             self.__setitem__(key, value)
+            # Note: Return so that we don't keep track of another attribute. Instead
+            # column names and such will be returned when __getattr__ is called
+            return
```
mvashishtha, line 2503:
```suggestion
            # Use __setitem__  to update an existing column with a new value.
            # In this case, we don't need to set the column label's attribute
            # explicitly, because __getattr__ will get the updated column
            # whenever it's needed. Return early to avoid setting the column
            # label's attribute later in this method.
            self.__setitem__(key, value)
            return
```
I think this is a clearer, if somewhat wordy, explanation.

modin/pandas/test/dataframe/test_iter.py
```diff
@@ -258,6 +258,21 @@ def test___setattr__():
...
+def test___setattr___getattr__():
+    # Use case from issue #4577
+    pandas_df = pandas.DataFrame([[1]], columns=["col0"])
+    modin_df = pd.DataFrame([[1]], columns=["col0"])
+
+    pandas_df.col0 = [3]
+    modin_df.col0 = [3]
+
+    df_equals(modin_df, pandas_df)
+    modin_df.col0 = pd.Series([3])
```
mvashishtha, line 270:
```suggestion
    modin_df.col0 = pd.Series([5])
```

Let's use a new value just in case something goes wrong when assigning a `Series`.

mvashishtha (commented):
This looks great! I left some style comments. This is a subtle method, so we should document the code and tests well for Modin developers :)

modin/pandas/dataframe.py
```diff
@@ -2498,6 +2498,9 @@ def __setattr__(self, key, value):
             pass
         elif key in self and key not in dir(self):
             self.__setitem__(key, value)
+            # Note: Return so that we don't keep track of another attribute. Instead
+            # column names and such will be returned when __getattr__ is called
```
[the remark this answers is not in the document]
pyrito: Yes, this sounds good

modin/pandas/dataframe.py
pyrito: @devin-petersohn had a suggestion from earlier that seemed to be fairly straightforward. Do you guys have a preference?

mvashishtha: I vote for his suggestion :)

modin/pandas/test/dataframe/test_iter.py
```diff
@@ -258,6 +259,24 @@ def test___setattr__():
     df_equals(modin_df, pandas_df)
```
mvashishtha, line 260:
```suggestion
        # While `new_col` is not a column of the dataframe,
        # it should be accessible with __getattr__.
        assert modin_df.new_col.equals(pandas_df.new_col)
```

This is not strictly part of the bug, but we should check that we're setting non-column attributes correctly after your fix.

pyrito: I don't think `equals` would work here since these are list types, so I just used the `==` operator which should work fine for these lists I think.

mvashishtha (commented):
I have one minor style suggestion.

mvashishtha (dismissed):
LGTM! Thanks for responding to all the comments.

modin/pandas/test/dataframe/test_iter.py
```diff
@@ -257,6 +258,37 @@ def test___setattr__():
...
+    # it should be accessible with __getattr__.
+    assert modin_df.new_col == pandas_df.new_col
+
+
+def test___setattr__mutating_column():
+    # Use case from issue #4577
+    pandas_df = pandas.DataFrame([[1]], columns=["col0"])
+    modin_df = pd.DataFrame([[1]], columns=["col0"])
+
+    # Replacing a column with a list should mutate the column in place.
```
RehanSD, line 271: I think we should add a test here ensuring that we can do something like `ffill` on `df.col0` after setting it to a list

RehanSD (changes requested):
LGTM but suggested one edit for the tests!

modin/pandas/test/dataframe/test_iter.py
mvashishtha: Let's just check that `pandas_df.col0.equals(modin_df.col0)`. That will check that the types are `series`. That's what we care about checking

pyrito: I had that before, but got rid of it due to @mvashishtha 's suggestion. You can see it here: https://github.com/modin-project/modin/pull/4588#discussion_r902620705

RehanSD (dismissed):
LGTM!

modin/pandas/test/dataframe/test_iter.py
RehanSD: Makes sense!

RehanSD (approved):
LGTM!

modin/pandas/dataframe.py
VasilijKolomiets: Thank for solving my [issue](https://github.com/modin-project/modin/issues/4577)

PR #4592, opened 2022-06-22
FIX-#4411: Fix binary_op between datetime64 Series and pandas timedelta

### LINKED ISSUE #4411: modin datetime Series fails when touches pandas timedeltas
### System information
- **OS Platform and Distribution**: Ubuntu 18.04.5 LTS
- **Modin version**: 0.14.0
- **Python version**: 3.8.5
- **Code to reproduce**:
    `modin.pandas.date_range('2000-12-12','2000-12-13').to_series().pipe(modin.pandas.Series) + modin.pandas.to_timedelta([1,2], unit='h')`

i ran into this problem when read excel file and parsed a column with `modin.pandas.to_datetime` (and `pandas.to_datetime` aswell!), the dataframe coulmn became a `modin.pandas.Series`. This series errors out everything i try to do with it.

Edit: i tried to create a DataFrame directly to check whether it is the read_excel that is the culprit. Spoiler: it isnt
```
df = modin.pandas.DataFrame(modin.pandas.date_range('2000-12-12','2000-12-13'))
df[0] + modin.pandas.to_timedelta([1,2], unit='h')
```
`TypeError: Cannot do operation non-numeric dtypes`

pyrito:
## What do these changes do?

This PR addresses #4411 which tried to add a pandas.timedeltas to a datetime64 Series. The bug ended up being tied up with how we validated binary operations, so I took the liberty to clean up code in `_validate_other` since I didn't see any of the dtype flags used anywhere else in the code. I also added in a test to validate the original posted issue. 

- [x] Resolves #4411 
- [x] tests added and passing

### DIFF
--- a/modin/pandas/base.py
+++ b/modin/pandas/base.py
@@ -222,7 +222,7 @@ class BasePandasDataset(ClassLogger):
         self,
         other,
         axis,
-        type_check=False,
+        dtype_check=False,
         compare_index=False,
     ):
         """
@@ -236,7 +236,7 @@ class BasePandasDataset(ClassLogger):
             Specifies axis along which to do validation. When `1` or `None`
             is specified, validation is done along `index`, if `0` is specified
             validation is done along `columns` of `other` frame.
-        type_check : bool, default: False
+        dtype_check : bool, default: False
             Validates that both frames have compatible dtypes.
         compare_index : bool, default: False
             Compare Index if True.
@@ -291,7 +291,7 @@ class BasePandasDataset(ClassLogger):
             if not self.index.equals(other.index):
                 raise TypeError("Cannot perform operation with non-equal index")
         # Do dtype checking.
-        if type_check:
+        if dtype_check:
             if not all(
                 (is_numeric_dtype(self_dtype) and is_numeric_dtype(other_dtype))
                 or (is_object_dtype(self_dtype) and is_object_dtype(other_dtype))
@@ -381,7 +381,7 @@ class BasePandasDataset(ClassLogger):
             return self._default_to_pandas(
                 getattr(self._pandas_class, op), other, **kwargs
             )
-        other = self._validate_other(other, axis, type_check=True)
+        other = self._validate_other(other, axis, dtype_check=True)
         exclude_list = [
             "__add__",
             "__radd__",
--- a/modin/pandas/base.py
+++ b/modin/pandas/base.py
@@ -235,14 +235,6 @@ class BasePandasDataset(ClassLogger):
             Specifies axis along which to do validation. When `1` or `None`
             is specified, validation is done along `index`, if `0` is specified
             validation is done along `columns` of `other` frame.
-        numeric_only : bool, default: False
-            Validates that both frames have only numeric dtypes.
-        numeric_or_time_only : bool, default: False
-            Validates that both frames have either numeric or time dtypes.
-        numeric_or_object_only : bool, default: False
-            Validates that both frames have either numeric or object dtypes.
-        comparison_dtypes_only : bool, default: False
-            Validates that both frames have either numeric or time or equal dtypes.
         compare_index : bool, default: False
             Compare Index if True.
 
--- a/modin/pandas/base.py
+++ b/modin/pandas/base.py
@@ -222,10 +222,6 @@ class BasePandasDataset(ClassLogger):
         self,
         other,
         axis,
-        numeric_only=False,
-        numeric_or_time_only=False,
-        numeric_or_object_only=False,
-        comparison_dtypes_only=False,
         compare_index=False,
     ):
         """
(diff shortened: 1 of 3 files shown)


modin/pandas/test/test_series.py
```diff
@@ -1141,6 +1141,20 @@ def test_between_time():
     )
 
 
+def test_time_ops():
+    # Make a pandas.core.indexes.timedeltas.TimedeltaIndex
+    deltas = pd.to_timedelta([1], unit="h")
+    modin_series = pd.Series(np.datetime64("2000-12-12")) + deltas
+    pandas_series = pandas.Series(np.datetime64("2000-12-12")) + deltas
+
+    df_equals(modin_series, pandas_series)
```
mvashishtha, line 1150:
```suggestion
    test_series = create_test_series(np.datetime64("2000-12-12"))
    eval_general(*test_series, lambda s: s + deltas)
```

modin/pandas/test/test_series.py
```diff
@@ -1141,6 +1141,20 @@ def test_between_time():
...
+
+    modin_series = pd.Series(np.datetime64("2000-12-12")) - deltas
+    pandas_series = pandas.Series(np.datetime64("2000-12-12")) - deltas
+
+    df_equals(modin_series, pandas_series)
```
mvashishtha, line 1155:
```suggestion
    eval_general(*test_series, lambda s: s - deltas)
```

modin/pandas/test/test_series.py
```diff
@@ -1141,6 +1141,20 @@ def test_between_time():
     )
 
 
+def test_time_ops():
```
mvashishtha, line 1144:
```suggestion
def test_add_series_to_timedeltaindex():
```

mvashishtha (commented):
Thanks for the fix @pyrito ! If `_validate_other` works like this, it's fine with me. I left some style comments.

modin/pandas/test/test_series.py
pyrito: Wow this is super clean, I'll try to make use of these idiom properly in the future.

mvashishtha (approved):
LGTM!

modin/pandas/base.py
```diff
@@ -222,10 +222,6 @@ def _validate_other(
         self,
         other,
         axis,
-        numeric_only=False,
-        numeric_or_time_only=False,
-        numeric_or_object_only=False,
-        comparison_dtypes_only=False,
```
vnlitvinov, line 228: I wonder why type checking was (usually) off before, given I don't see its values set as `True` in most places.
I also wonder how much of a performance hit this type checking introduces...

vnlitvinov (commented):
@pyrito please fill in the PR message template completely - it needs to mention the issue you're closing in this PR in the "Closes" checkbox, otherwise Github won't auto-close the issue when PR is merged.

modin/pandas/base.py
pyrito: @devin-petersohn thoughts?

devin-petersohn: The check here can (not always) introduce some performance issues, and requires the data types be materialized. That said, we probably need the data types to be materialized for the error checking here.

devin-petersohn (approved):
LGTM, thanks @pyrito. Will let @vnlitvinov give the final OK

modin/pandas/base.py
```diff
@@ -222,10 +222,7 @@ def _validate_other(
...
+        type_check=False,
```
vnlitvinov, line 225: I wonder if we should name this `dtype_check` (extra `d`) to highlight this is about data types?..

vnlitvinov (dismissed):
LGTM assuming CI passes!

pyrito:
CI looks green. Can someone from @modin-project/modin-core go ahead and merge this if everything looks good?

modin/pandas/base.py
```diff
@@ -300,44 +291,18 @@ def _validate_other(
...
-            ):
-                raise TypeError("Cannot do operation non-numeric dtypes")
-        elif comparison_dtypes_only:
-            if not all(
-                (is_numeric_dtype(self_dtype) and is_numeric_dtype(other_dtype))
                 or (
                     is_datetime_or_timedelta_dtype(self_dtype)
                     and is_datetime_or_timedelta_dtype(other_dtype)
                 )
                 or is_dtype_equal(self_dtype, other_dtype)
```
YarShev, line 302: I am also wondering if there are any performance implications from the changes introduced in this PR. I would suggest running a benchmark, which hits the last condition here, i.e. `is_dtype_equal(self_dtype, other_dtype)`, to see the performance of the worst case.

pyrito: I don't think there will be a large performance implication from this, it's just two extra calls, so it should be fine and the current usage should stay the same.

YarShev (approved):

PR #4603, opened 2022-06-24
FIX-#4564: Workaround import issues in Ray: auto-import pandas on python start if env var is set

### LINKED ISSUE #4564: BUG: ray 1.13 breaks CI, including test_binary and test_pickle, with circular import
### System information
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: macOS Monterey 12.2.1
- **Modin version** (`modin.__version__`): ba2b6ea47d81575e41af6c92ba4c2c3f8bcf54e2
- **Python version**: 3.9.12

With ray 1.13, run

`pytest modin/pandas/test/dataframe/test_pickle.py`

and

`pytest modin/pandas/test/dataframe/test_binary.py` 

and maybe more.

Sample CI failure: https://github.com/modin-project/modin/runs/6819134820?check_suite_focus=true

### Describe the problem
I can't tell what's going wrong. The ray tasks raise a `RaySystemError` with stack trace for ` modin/pandas/test/dataframe/test_pickle.py::test_column_pickle[True]`:

<details>

```
ray.exceptions.RaySystemError: System error: cannot import name 'DataFrame' from partially initialized module 'pandas.core.frame' (most likely due to a circular import) (/Users/maheshvashishtha/opt/anaconda3/envs/modin-latest/lib/python3.9/site-packages/pandas/core/frame.py)
traceback: Traceback (most recent call last):
  File "/Users/maheshvashishtha/opt/anaconda3/envs/modin-latest/lib/python3.9/site-packages/ray/serialization.py", line 340, in deserialize_objects
    obj = self._deserialize_object(data, metadata, object_ref)
  File "/Users/maheshvashishtha/opt/anaconda3/envs/modin-latest/lib/python3.9/site-packages/ray/serialization.py", line 237, in _deserialize_object
    return self._deserialize_msgpack_data(data, metadata_fields)
  File "/Users/maheshvashishtha/opt/anaconda3/envs/modin-latest/lib/python3.9/site-packages/ray/serialization.py", line 192, in _deserialize_msgpack_data
    python_objects = self._deserialize_pickle5_data(pickle5_data)
  File "/Users/maheshvashishtha/opt/anaconda3/envs/modin-latest/lib/python3.9/site-packages/ray/serialization.py", line 180, in _deserialize_pickle5_data
    obj = pickle.loads(in_band, buffers=buffers)
  File "/Users/maheshvashishtha/opt/anaconda3/envs/modin-latest/lib/python3.9/site-packages/pandas/core/frame.py", line 157, in <module>
    from pandas.core.generic import NDFrame
  File "/Users/maheshvashishtha/opt/anaconda3/envs/modin-latest/lib/python3.9/site-packages/pandas/core/generic.py", line 114, in <module>
    from pandas.core import (
  File "/Users/maheshvashishtha/opt/anaconda3/envs/modin-latest/lib/python3.9/site-packages/pandas/core/indexing.py", line 45, in <module>
    import pandas.core.common as com
  File "/Users/maheshvashishtha/opt/anaconda3/envs/modin-latest/lib/python3.9/site-packages/pandas/__init__.py", line 48, in <module>
    from pandas.core.api import (
  File "/Users/maheshvashishtha/opt/anaconda3/envs/modin-latest/lib/python3.9/site-packages/pandas/core/api.py", line 48, in <module>
    from pandas.core.groupby import (
  File "/Users/maheshvashishtha/opt/anaconda3/envs/modin-latest/lib/python3.9/site-packages/pandas/core/groupby/__init__.py", line 1, in <module>
    from pandas.core.groupby.generic import (
  File "/Users/maheshvashishtha/opt/anaconda3/envs/modin-latest/lib/python3.9/site-packages/pandas/core/groupby/generic.py", line 70, in <module>
    from pandas.core.frame import DataFrame
ImportError: cannot import name 'DataFrame' from partially initialized module 'pandas.core.frame' (most likely due to a circular import) (/Users/maheshvashishtha/opt/anaconda3/envs/modin-latest/lib/python3.9/site-packages/pandas/core/frame.py)
```
</details>

For `modin/pandas/test/dataframe/test_binary.py` for some reason I can't reproduce the failure on my laptop but here it is in CI: https://github.com/modin-project/modin/runs/6819134820?check_suite_focus=true

vnlitvinov:
## What do these changes do?
Based on what I found in https://github.com/modin-project/modin/issues/4564#issuecomment-1165419056, I think the real root cause of #4564 (and other similar issues with inability to import pandas) is the Python bug https://bugs.python.org/issue38884, which rises from the fact that they seem to have dropped the idea of the import lock in Python 3.3+.

This PR aims to workaround that by doing the following:
1. Create special `.pth` file to be shipped which, when package is installed, is placed under `site-packages` automatically and, when processed by Python interpreter, imports `pandas` if special environment variable is set
2. Change `setup.py` so this special file is part of both source and binary distributions
3. Initialize Ray so it sets this environment variable on each worker
4. Check Ray configuration if Ray was pre-initialized and warn the user to set the variables if they don't match

- [x] Resolves #4564 
- [x] tests added and passing

### DIFF
--- a/docs/getting_started/troubleshooting.rst
+++ b/docs/getting_started/troubleshooting.rst
@@ -298,6 +298,7 @@ If you're using some pre-configured Ray cluster to run Modin, it's possible you 
 be seeing spurious errors like
 
 .. code-block::
+
   ray.exceptions.RaySystemError: System error: partially initialized module 'pandas' has no attribute 'core' (most likely due to a circular import)
   traceback: Traceback (most recent call last):
     File "/usr/share/miniconda/envs/modin/lib/python3.8/site-packages/ray/serialization.py", line 340, in deserialize_objects
--- a/docs/getting_started/troubleshooting.rst
+++ b/docs/getting_started/troubleshooting.rst
@@ -293,3 +293,39 @@ or
 
 .. _issue: https://github.com/modin-project/modin/issues
 .. _Slack: https://modin.org/slack.html
+
+
+Spurious error "cannot import partially initialised pandas module" on custom Ray cluster
+""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""
+
+If you're using some pre-configured Ray cluster to run Modin, it's possible you would
+be seeing spurious errors like
+
+.. code-block::
+  ray.exceptions.RaySystemError: System error: partially initialized module 'pandas' has no attribute 'core' (most likely due to a circular import)
+  traceback: Traceback (most recent call last):
+    File "/usr/share/miniconda/envs/modin/lib/python3.8/site-packages/ray/serialization.py", line 340, in deserialize_objects
+      obj = self._deserialize_object(data, metadata, object_ref)
+    File "/usr/share/miniconda/envs/modin/lib/python3.8/site-packages/ray/serialization.py", line 237, in _deserialize_object
+      return self._deserialize_msgpack_data(data, metadata_fields)
+    File "/usr/share/miniconda/envs/modin/lib/python3.8/site-packages/ray/serialization.py", line 192, in _deserialize_msgpack_data
+      python_objects = self._deserialize_pickle5_data(pickle5_data)
+    File "/usr/share/miniconda/envs/modin/lib/python3.8/site-packages/ray/serialization.py", line 180, in _deserialize_pickle5_data
+      obj = pickle.loads(in_band, buffers=buffers)
+    File "/usr/share/miniconda/envs/modin/lib/python3.8/site-packages/pandas/__init__.py", line 135, in <module>
+      from pandas import api, arrays, errors, io, plotting, testing, tseries
+    File "/usr/share/miniconda/envs/modin/lib/python3.8/site-packages/pandas/testing.py", line 6, in <module>
+      from pandas._testing import (
+    File "/usr/share/miniconda/envs/modin/lib/python3.8/site-packages/pandas/_testing/__init__.py", line 979, in <module>
+      cython_table = pd.core.common._cython_table.items()
+  AttributeError: partially initialized module 'pandas' has no attribute 'core' (most likely due to a circular import)
+
+**Solution**
+
+Modin contains a workaround that should automatically do ``import pandas`` upon worker process starts.
+
+It is triggered by the presence of non-empty ``__MODIN_AUTOIMPORT_PANDAS__`` environment variable which
+Modin sets up automatically on the Ray clusters it spawns, but it might be missing on pre-configured clusters.
+
+So if you're seeing the issue like shown above, please make sure you set this environment variable on all
+worker nodes of your cluster before actually spawning the workers.
--- a/docs/getting_started/troubleshooting.rst
+++ b/docs/getting_started/troubleshooting.rst
@@ -291,10 +291,6 @@ or
     df = pd.DataFrame([0, 1, 2, 3])
     print(df)
 
-.. _issue: https://github.com/modin-project/modin/issues
-.. _Slack: https://modin.org/slack.html
-
-
 Spurious error "cannot import partially initialised pandas module" on custom Ray cluster
 """"""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""
 
@@ -329,3 +325,6 @@ Modin sets up automatically on the Ray clusters it spawns, but it might be missi
 
 So if you're seeing the issue like shown above, please make sure you set this environment variable on all
 worker nodes of your cluster before actually spawning the workers.
+
+.. _issue: https://github.com/modin-project/modin/issues
+.. _Slack: https://modin.org/slack.html
--- a/modin-autoimport-pandas.pth
+++ b/modin-autoimport-pandas.pth
@@ -1,0 +1,1 @@
+import os; os.environ.get("__MODIN_AUTOIMPORT_PANDAS_WORKAROUND__", None) and __import__("pandas")
--- a/modin-autoimport-pandas.pth
+++ b/modin-autoimport-pandas.pth
@@ -1,1 +1,1 @@
-import os; os.environ.get("__MODIN_AUTOIMPORT_PANDAS_WORKAROUND__", None) and __import__("pandas")
+import os; os.environ.get("__MODIN_AUTOIMPORT_PANDAS__", None) and __import__("pandas")
--- a/modin/core/execution/ray/common/utils.py
+++ b/modin/core/execution/ray/common/utils.py
@@ -172,7 +172,7 @@ def initialize_ray(
         for varname, varvalue in extra_init_kw["runtime_env"]["env_vars"].items():
             if str(env_vars.get(varname, "")) != str(varvalue):
                 ErrorMessage.single_warning(
-                    "If initialising Ray yourself, please ensure its runtime env "
+                    "When using a pre-initialized Ray cluster, please ensure that the runtime env "
                     + f"sets environment variable {varname} to {varvalue}"
                 )
 
--- a/modin/core/execution/ray/common/utils.py
+++ b/modin/core/execution/ray/common/utils.py
@@ -18,6 +18,7 @@ import sys
 import psutil
 from packaging import version
 import warnings
+import json
 
 import ray
 
@@ -117,6 +118,9 @@ def initialize_ray(
         What password to use when connecting to Redis.
         If not specified, ``modin.config.RayRedisPassword`` is used.
     """
+    extra_init_kw = {
+        "runtime_env": {"env_vars": {"__MODIN_AUTOIMPORT_PANDAS_WORKAROUND__": "1"}}
+    }
     if not ray.is_initialized() or override_is_cluster:
         cluster = override_is_cluster or IsRayCluster.get()
         redis_address = override_redis_address or RayRedisAddress.get()
@@ -138,6 +142,7 @@ def initialize_ray(
                 include_dashboard=False,
                 ignore_reinit_error=True,
                 _redis_password=redis_password,
+                **extra_init_kw,
             )
         else:
             from modin.error_message import ErrorMessage
@@ -146,9 +151,9 @@ def initialize_ray(
             # the warning message.
             ErrorMessage.not_initialized(
                 "Ray",
-                """
+                f"""
     import ray
-    ray.init()
+    ray.init({', '.join([f'{k}={v}' for k,v in extra_init_kw.items()])})
 """,
             )
             object_store_memory = Memory.get()
@@ -206,6 +211,7 @@ def initialize_ray(
                 "object_store_memory": object_store_memory,
                 "_redis_password": redis_password,
                 "_memory": object_store_memory,
+                **extra_init_kw,
             }
             ray.init(**ray_init_kwargs)
 
@@ -219,6 +225,20 @@ def initialize_ray(
             if not GPU_MANAGERS:
                 for i in range(GpuCount.get()):
                     GPU_MANAGERS.append(GPUManager.remote(i))
+    else:  # ray is already initialized, check runtime env config
+        main_cfg = ray.worker.global_worker.core_worker.get_job_config()
+        try:
+            runtime_env = json.loads(main_cfg.runtime_env_info.serialized_runtime_env)
+        except json.JSONDecodeError:
+            runtime_env = {}
+        env_vars = runtime_env.get("envVars", {})
+        for varname, varvalue in extra_init_kw["runtime_env"]["env_vars"].items():
+            if str(env_vars.get(varname, "")) != str(varvalue):
+                ErrorMessage.single_warning(
+                    "If initialising Ray yourself, please ensure its runtime env "
+                    + f"sets environment variable {varname} to {varvalue}"
+                )
+
     _move_stdlib_ahead_of_site_packages()
     ray.worker.global_worker.run_function_on_all_workers(
         _move_stdlib_ahead_of_site_packages
--- a/modin/core/execution/ray/common/utils.py
+++ b/modin/core/execution/ray/common/utils.py
@@ -75,26 +75,6 @@ def _move_stdlib_ahead_of_site_packages(*args):
         sys.path.insert(site_packages_path_index, os.path.dirname(site_packages_path))
 
 
-def _import_pandas(*args):
-    """
-    Import pandas to make sure all its machinery is ready.
-
-    This prevents a race condition between two threads deserializing functions
-    and trying to import pandas at the same time.
-
-    Parameters
-    ----------
-    *args : tuple
-        Ignored, added for compatibility with Ray.
-
-    Notes
-    -----
-    This function is expected to be run on all workers before any
-    serialization or deserialization starts.
-    """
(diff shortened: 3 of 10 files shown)


devin-petersohn (commented):
This looks pretty good, although it introduces minor friction for users who want to get started and initialize Ray themselves. Probably need extra documentation or something because we recommend users to start the engine themselves.

Can you add runtime env to an already running Ray cluster?

prutskov (commented):
How does it works in the case I didn't install modin? Suppose, I'm working in `modin`-root folder.

vnlitvinov:
> Can you add runtime env to an already running Ray cluster?

You can set runtime env even on per-task basis, but in this case it won't have any particular effect as the interpreter in each worker has already been initialized.

As for docs... I'm open for suggestions on where to put things.



> How does it work in the case I didn't install modin? Suppose, I'm working in `modin`-root folder.

Obviously, it won't have the workaround effect, but TBH this isn't the main use case, plus I personally never saw this issue locally. If someone is really bothered by this issue, one could `pip install .` to enable the workaround.

vnlitvinov:
For hinting the users on how to initialize Ray, this is how current (in this PR) warning look like:
```
>>> import modin.pandas as pd
UserWarning: The pandas version installed 1.4.2 does not match the supported pandas version in Modin 1.4.3. This may cause undesired side effects!
>>> df=pd.DataFrame({})

    import ray
    ray.init(runtime_env={'env_vars': {'__MODIN_AUTOIMPORT_PANDAS_WORKAROUND__': '1'}})

```

Garra1980:
Whoa, poor user, so many extra info...

vnlitvinov:
It was like this for quite a while, but before this PR it simply stated `ray.init()` without parameters:

```
UserWarning: The pandas version installed 1.4.2 does not match the supported pandas version in Modin 1.4.3. This may cause undesired side effects!

    import ray
    ray.init()

```

modin/core/execution/ray/common/utils.py
```diff
@@ -219,11 +204,25 @@ def initialize_ray(
             if not GPU_MANAGERS:
                 for i in range(GpuCount.get()):
                     GPU_MANAGERS.append(GPUManager.remote(i))
+    else:  # ray is already initialized, check runtime env config
+        main_cfg = ray.worker.global_worker.core_worker.get_job_config()
+        try:
+            runtime_env = json.loads(main_cfg.runtime_env_info.serialized_runtime_env)
+        except json.JSONDecodeError:
+            runtime_env = {}
+        env_vars = runtime_env.get("envVars", {})
```
iycheng, line 213: ray.worker is becoming private in the next release, please use the following:
```
ray.get_runtime_context().runtime_env
```

modin/core/execution/ray/common/utils.py
```diff
@@ -219,11 +204,25 @@ def initialize_ray(
...
+            if str(env_vars.get(varname, "")) != str(varvalue):
+                ErrorMessage.single_warning(
+                    "If initialising Ray yourself, please ensure its runtime env "
+                    + f"sets environment variable {varname} to {varvalue}"
+                )
+
     _move_stdlib_ahead_of_site_packages()
     ray.worker.global_worker.run_function_on_all_workers(
         _move_stdlib_ahead_of_site_packages
     )
```
YarShev, line 224: It seems to me we had kind of a similar problem with these lines. I wonder if we will have to add an additional config for ray init, when Ray removes `run_function_on_all_workers`?

devin-petersohn: This may not be a problem anymore (Python 3.6+). Previously when users had `typing` pypi package installed it would create conflicts with the stdlib `typing`.

vnlitvinov: Should I drop those lines then?

devin-petersohn: It's ok with me to drop.

YarShev:
What do you think if we leave the warning as is (just `import ray; ray.init()`) when initializing Ray by users themselves, and put a paragraph regarding race condition in the troubleshooting page?

modin/core/execution/ray/common/utils.py
```diff
@@ -138,17 +123,16 @@ def initialize_ray(
...
-
             # This string is intentionally formatted this way. We want it indented in
             # the warning message.
             ErrorMessage.not_initialized(
                 "Ray",
-                """
+                f"""
     import ray
-    ray.init()
+    ray.init({', '.join([f'{k}={v}' for k,v in extra_init_kw.items()])})
```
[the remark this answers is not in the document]
vnlitvinov: No, I'm imagining a case where there would be several parameters, so (to be understood by Python), they would have to be comma-separated.

modin/core/execution/ray/common/utils.py
```diff
@@ -117,6 +99,9 @@ def initialize_ray(
         What password to use when connecting to Redis.
         If not specified, ``modin.config.RayRedisPassword`` is used.
     """
+    extra_init_kw = {
+        "runtime_env": {"env_vars": {"__MODIN_AUTOIMPORT_PANDAS_WORKAROUND__": "1"}}
```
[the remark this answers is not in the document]
vnlitvinov: agreed, replaced

modin-autoimport-pandas.pth
```diff
@@ -0,0 +1 @@
+import os; os.environ.get("__MODIN_AUTOIMPORT_PANDAS_WORKAROUND__", None) and __import__("pandas")
```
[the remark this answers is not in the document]
vnlitvinov: agreed

vnlitvinov:
I don't think adding a little bit of info in our warning would make any harm, but I think it would harm user's experience when they would receive some spurious failures until they would realize that there is a troubleshooting guide on that...

modin/core/execution/ray/common/utils.py
```diff
@@ -138,17 +87,16 @@ def initialize_ray(
                 include_dashboard=False,
                 ignore_reinit_error=True,
                 _redis_password=redis_password,
+                **extra_init_kw,
```
prutskov, line 90: Did you checked this case? I can't find that extra parameters could be provided in case of existing cluster https://github.com/ray-project/ray/blob/master/python/ray/_private/worker.py#L1400-L1412

modin/core/execution/ray/common/utils.py
```diff
@@ -219,11 +168,15 @@ def initialize_ray(
...
-        _move_stdlib_ahead_of_site_packages
-    )
-    ray.worker.global_worker.run_function_on_all_workers(_import_pandas)
+    else:  # ray is already initialized, check runtime env config
+        env_vars = ray.get_runtime_context().runtime_env.get("env_vars", {})
+        for varname, varvalue in extra_init_kw["runtime_env"]["env_vars"].items():
+            if str(env_vars.get(varname, "")) != str(varvalue):
+                ErrorMessage.single_warning(
+                    "If initialising Ray yourself, please ensure its runtime env "
+                    + f"sets environment variable {varname} to {varvalue}"
```
prutskov, line 177:
```suggestion
                ErrorMessage.single_warning(
                    "If initializing Ray yourself, please ensure its runtime env "
                    + f"sets environment variable {varname} to {varvalue}."
```

vnlitvinov: I imitate British accent here 🙃 https://english.stackexchange.com/a/459561

modin/core/execution/ray/common/utils.py
iycheng: Thank you!

modin/core/execution/ray/common/utils.py
prutskov: Ok :)

setup.py
```diff
@@ -5,26 +5,42 @@
...
+# Distribute 'modin-autoimport-pandas.pth' along with binary and source distributions.
+# This file provides the "import pandas before Ray init" feature if specific
+# environment variable is set (see https://github.com/modin-project/modin/issues/4564).
+cmdclass = versioneer.get_cmdclass()
+
+
+class AddPthFileBuild(cmdclass["build_py"]):
+    def _get_data_files(self):
+        return (super()._get_data_files() or []) + [
+            (".", ".", self.build_lib, ["modin-autoimport-pandas.pth"])
```
YarShev, line 27: Should we declare a var for modin-autoimport-pandas.pth as we use it in both AddPthFileBuild and AddPthFileSDist?

YarShev:
> As for docs... I'm open for suggestions on where to put things.

What about modin.config - https://modin.readthedocs.io/en/stable/flow/modin/config.html?

modin/core/execution/ray/common/utils.py
vnlitvinov: > Can you elaborate why it would be useful?

This creates one central place to add more arguments to `ray.init()` instead of copy-pasting them in several different places.

> What the code do you mean?

https://github.com/modin-project/modin/blob/e426ce0331aa4a83f7e0507084c405a8a5256c7d/modin/core/execution/ray/common/utils.py#L170-L177

vnlitvinov:
> > As for docs... I'm open for suggestions on where to put things.
> 
> What about modin.config - https://modin.readthedocs.io/en/stable/flow/modin/config.html?

That place is about configuring Modin behaviour, not about workaround we employ.



> Maybe we should add a comment to the warning on what **MODIN_AUTOIMPORT_PANDAS** means or point to a doc page for reference?

I would actually like to see the request on which we _started_ giving this warning on "hey, you better init Ray yourself". For now I would prefer to stick to what I had already did - just give the (extended) warning.

modin/core/execution/ray/common/utils.py
YarShev: We don't hit this branch in case of cluster init.

YarShev:
> > > As for docs... I'm open for suggestions on where to put things.
> > 
> > 
> > What about modin.config - https://modin.readthedocs.io/en/stable/flow/modin/config.html?
> 
> That place is about configuring Modin behaviour, not about workaround we employ.

I think this workaround is kind of related to Modin configuration because Modin forces to import pandas on Ray workers. Another option to put some documentation on the workaround I see is https://modin.readthedocs.io/en/stable/getting_started/installation.html. Do you think which page is better? Maybe there is an other place you could propose?

>> Maybe we should add a comment to the warning on what MODIN_AUTOIMPORT_PANDAS means or point to a doc page for reference?
>
> I would actually like to see the request on which we started giving this warning on "hey, you better init Ray yourself". For now I would prefer to stick to what I had already did - just give the (extended) warning.

Not sure about that. I would like to look at other options someone has.

@modin-project/modin-core, @modin-project/modin-contributors

vnlitvinov:
The only place in the docs which makes sense to me is this: https://modin.readthedocs.io/en/stable/getting_started/troubleshooting.html

Thoughts?

devin-petersohn (commented):
@vnlitvinov looks great! There are a couple of comments here about cluster-related handling.

modin/core/execution/ray/common/utils.py
```diff
@@ -219,11 +167,15 @@ def initialize_ray(
...
-    ray.worker.global_worker.run_function_on_all_workers(
-        _move_stdlib_ahead_of_site_packages
-    )
-    ray.worker.global_worker.run_function_on_all_workers(_import_pandas)
+    else:  # ray is already initialized, check runtime env config
+        env_vars = ray.get_runtime_context().runtime_env.get("env_vars", {})
+        for varname, varvalue in extra_init_kw["runtime_env"]["env_vars"].items():
+            if str(env_vars.get(varname, "")) != str(varvalue):
+                ErrorMessage.single_warning(
+                    "If initialising Ray yourself, please ensure its runtime env "
```
RehanSD, line 175:
```suggestion
                    "When using a pre-initialized Ray cluster, please ensure that the runtime env "
```

modin/core/execution/ray/common/utils.py
```diff
@@ -219,11 +167,15 @@ def initialize_ray(
             if not GPU_MANAGERS:
                 for i in range(GpuCount.get()):
                     GPU_MANAGERS.append(GPUManager.remote(i))
-    _move_stdlib_ahead_of_site_packages()
-    ray.worker.global_worker.run_function_on_all_workers(
-        _move_stdlib_ahead_of_site_packages
-    )
-    ray.worker.global_worker.run_function_on_all_workers(_import_pandas)
+    else:  # ray is already initialized, check runtime env config
+        env_vars = ray.get_runtime_context().runtime_env.get("env_vars", {})
```
RehanSD, line 171: This only checks that the head nodes env vars are set correctly, right? Don't we need to set the environment variable on all of the workers/nodes, and if so, shouldn't we be checking that the env vars are set correctly on all of the workers? We could probably do that using the ray function that runs on all workers.

RehanSD (commented):
Left a quick change as well as a question

modin/core/execution/ray/common/utils.py
vnlitvinov: Huh?.. we either call `ray.init()` or hit this `else:` branch.

We hit `ray.init()` at line 84 (under `if cluster:`) or at line 158 (in `else` branch of that `if cluster`).

modin/core/execution/ray/common/utils.py
vnlitvinov: No, `.runtime_env` is the thing which _tells_ Ray which environment variables to set for its workers. It should set these variables for the workers automatically.

setup.py
vnlitvinov: fixed

modin/core/execution/ray/common/utils.py
YarShev: I was just wondering if we should check the runtime environment after ray.init() in case Modin itself initializes Ray? Because of this.

> Did you check this case? I can't find that extra parameters could be provided in case of existing cluster https://github.com/ray-project/ray/blob/master/python/ray/_private/worker.py#L1400-L1412
>> I hadn't, but apparently extra ones would just get ignored

docs/getting_started/troubleshooting.rst
```diff
@@ -293,3 +293,39 @@ or
 
 .. _issue: https://github.com/modin-project/modin/issues
 .. _Slack: https://modin.org/slack.html
+
+
+Spurious error "cannot import partially initialised pandas module" on custom Ray cluster
```
YarShev, line 298: This paragraph should be placed above the links (_issue, _Slack, etc.).

modin/core/execution/ray/common/utils.py
vnlitvinov: good point!

vnlitvinov:
Huh, when I look at the CI of the last commit (https://github.com/modin-project/modin/pull/4603/commits/887a5f467b8b24cbdd69efad77ad8516c443cee3) docs are built fine. Am I missing things?

YarShev:
Here is the job https://github.com/modin-project/modin/runs/7237231595?check_suite_focus=true.

YarShev (approved):
@vnlitvinov, LGTM, thanks!

Garra1980:
So green light for python 3.10 support? :)

Issue #4358, opened 2022-03-31
MultiIndex: DataFrame.loc with a tuple on columns defined as MultiIndex returns an error


neltacigreb:
### System information
- **OS Platform and Distribution**: MacOS Monterey 12.2.1
- **Modin version** 0.14
- **Python version**: 3.8.12
- **Code we can use to reproduce**:

### Using modin: 
```python
import ray
ray.init()
import modin.pandas as pd
import numpy as np

arrays = [
    ["bar", "bar", "baz", "baz", "foo", "foo", "qux", "qux"],
    ["one", "two", "one", "two", "one", "two", "one", "two"],
]
columns =  pd.MultiIndex.from_tuples(zip(*arrays), names=['a', 'b'])
data = [np.random.randint(10, 50, len(columns)) for _ in range(30)]
df = pd.DataFrame(data, columns=columns)

# bug happens here
df.loc[:, ("bar", "two")]
```

### Same operation in vanilla pandas 
```python
import pandas as pd
import numpy as np

arrays = [
    ["bar", "bar", "baz", "baz", "foo", "foo", "qux", "qux"],
    ["one", "two", "one", "two", "one", "two", "one", "two"],
]
columns =  pd.MultiIndex.from_tuples(zip(*arrays), names=['a', 'b'])
data = [np.random.randint(10, 50, len(columns)) for _ in range(30)]
df = pd.DataFrame(data, columns=columns)

df.loc[:, ("bar", "two")]
```

### Describe the problem
When using .loc with tuples on a multiIndex, a ValueError appears

### Source code / logs
**traceback:**
```shell
Traceback (most recent call last):
  File "/VIRTUALENV/lib/python3.8/site-packages/ipython-7.32.0-py3.8.egg/IPython/core/interactiveshell.py", line 3524, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-4-59ea95655168>", line 1, in <cell line: 1>
    df.loc[:, ("bar", "two")]
  File "/VIRTUALENV/lib/python3.8/site-packages/modin/pandas/indexing.py", line 674, in __getitem__
    result.columns = result.columns.droplevel(list(range(len(col_loc_as_list))))
  File "/VIRTUALENV/lib/python3.8/site-packages/pandas/core/indexes/base.py", line 2063, in droplevel
    return self._drop_level_numbers(levnums)
[...]

PR #4608, opened 2022-06-27
FIX-#4358: Return correct intermediate representation for MultiIndex DataFrames

pyrito:
## What do these changes do?
This is just a draft PR to demonstrate changes. 

- [ ] Resolves #? 
- [ ] tests added and passing


anmyachev:
@pyrito @mvashishtha you are trying to fix the consequences of the fact that in order to implement `loc`, we use `iloc` in our internals(`qc.view->qc.mask->partition.iloc`) and we have to do additional things like remove multi-index levels after that. What if we try to go the other way, change the implementation and try to use `partition.loc`?

mvashishtha:
> What if we try to go the other way, change the implementation and try to use `partition.loc`?

@anmyachev I think we should try that! @pyrito could you try it out?

mvashishtha:
@pyrito 

>  I agree that we are trying to fix the consequences of using iloc, but I think it's needed to maintain invariants with positioning though I could be wrong.

Which invariants are you thinking of? We already support label-based indexing in `PandasDataframe.Mask`. I think it should be possible to use that to select the correct subset of the dataframe.

pyrito:
@mvashishtha oh I see what you mean. I was digging more into the partition level `iloc` call in `PandasDataframePartition` and I thought that we needed the index-based lookups for when we partition the dataframe.

pyrito:
@mvashishtha digging into it more, it looks like when we try to call `get_indexer_for()` on `MultiIndex` with labels I only get `[-1 -1]` as my index. Reading the current mask implementation as it, it looks like we convert label lookup to index positions and rely on that via `iloc` to get the subset of data that we want. I think what's happening is that we are able to index the right rows and columns, but in the MultiIndex case this doesn't guarantee that we will drop the correct levels. 

Like you mentioned we could try writing a version that uses `loc`, but I'm not exactly sure how that would come together with all the partition logic, I'll have to dig deeper into understanding the lower levels.

modin/pandas/test/dataframe/test_indexing.py
```diff
@@ -562,6 +562,55 @@ def test_loc_multi_index():
     df_equals(modin_df.loc[modin_df.index[:7]], pandas_df.loc[pandas_df.index[:7]])
 
 
+# Better name suggestions?
+def test_loc_4358():
```
devin-petersohn, line 566:
```suggestion
def test_loc_mutli_index_with_tuples():
```

modin/pandas/indexing.py
```diff
@@ -330,14 +332,20 @@ def __getitem__(self, row_lookup, col_lookup, ndim):
...
                 None
-                if self.col_scalar and self.row_scalar
+                if (self.col_scalar and self.row_scalar)
+                or (self.row_multiindex and self.col_multiindex)
                 else 1
-                if self.col_scalar
+                if self.col_scalar or self.col_multiindex
                 else 0
             )
+
```
devin-petersohn, line 348:
```suggestion
```

devin-petersohn (commented):
Thanks @pyrito, the approach makes sense. The challenge here I think is that it adds more complexity inline, could we put this in a utility function and just call it from here?

I also added some nits. Thanks again!

pyrito:
@anmyachev @mvashishtha - I discussed this with @devin-petersohn and he suggested that we could merge in the bug fix for now and come around with a better design to make use of `loc` underneath the hood later since it will require more engineering effort.

modin/pandas/test/dataframe/test_indexing.py
```diff
@@ -562,6 +562,55 @@ def test_loc_multi_index():
...
+    )
+    data = [[100, 300, 900, 400], [200, 500, 300, 600]]
+    df = pd.DataFrame(data, columns=cols, index=multi_index)
+    pdf = pandas.DataFrame(data, columns=cols, index=multi_index)
+    df_equals(df.loc[("r0", "rA"), :], pdf.loc[("r0", "rA"), :])
+    df_equals(df.loc[("r0"), :], pdf.loc[("r0"), :])
+    df_equals(df.loc[:, ("Gasoline")], pdf.loc[:, ("Gasoline")])
+    df_equals(df.loc[:, ("Gasoline", "Toyota")], pdf.loc[:, ("Gasoline", "Toyota")])
+    # Test below fails
+    # df_equals(df.loc[("r0"), ("Gasoline", "Toyota")], pdf.loc[("r0"), ("Gasoline", "Toyota")])
```
anmyachev, line 611: What error appears here?

pyrito: There is a mismatch in the resulting DataFrames. Modin returns a single value and pandas returns a Series I believe. It seems to be an underlying issue with getting the value from `loc` with a DataFrame that has two MultiIndex columns. It didn't seem to be related to which axis we squeezed the result with.

modin/pandas/indexing.py
```diff
@@ -333,9 +335,10 @@ def __getitem__(self, row_lookup, col_lookup, ndim):
         else:
             axis = (
```
mvashishtha, line 336: Could you add a comment explaining how we decide `axis`?

modin/pandas/indexing.py
```diff
@@ -563,6 +600,23 @@ def __getitem__(self, key):
...
+        self.row_multiindex_full_lookup = self._multiindex_possibly_contains_key(
+            axis=0, key=row_loc
+        )
+        self.col_multiindex_full_lookup = self._multiindex_possibly_contains_key(
+            axis=1, key=col_loc
+        )
+        levels_already_dropped = (
+            True
+            if self.row_multiindex_full_lookup or self.col_multiindex_full_lookup
+            else False
```
mvashishtha, line 618:
```suggestion
            self.row_multiindex_full_lookup or self.col_multiindex_full_lookup
```

modin/pandas/test/dataframe/test_indexing.py
```diff
@@ -562,6 +562,46 @@ def test_loc_multi_index():
     df_equals(modin_df.loc[modin_df.index[:7]], pandas_df.loc[pandas_df.index[:7]])
 
 
+def test_loc_multi_index_with_tuples():
+    arrays = [
+        ["bar", "bar", "baz", "baz", "foo", "foo", "qux", "qux"],
+        ["one", "two", "one", "two", "one", "two", "one", "two"],
+    ]
+    columns = pd.MultiIndex.from_tuples(zip(*arrays), names=["a", "b"])
+    data = [np.random.randint(10, 50, len(columns)) for _ in range(30)]
```
mvashishtha, line 571: We shouldn't use random data in tests without seeding `np`. Anyway, can we use something like `np.arange().reshape()` instead?

modin/pandas/test/dataframe/test_indexing.py
```diff
@@ -562,6 +562,46 @@ def test_loc_multi_index():
     df_equals(modin_df.loc[modin_df.index[:7]], pandas_df.loc[pandas_df.index[:7]])
 
 
+def test_loc_multi_index_with_tuples():
+    arrays = [
+        ["bar", "bar", "baz", "baz", "foo", "foo", "qux", "qux"],
+        ["one", "two", "one", "two", "one", "two", "one", "two"],
+    ]
```
mvashishtha, line 569: Since we only need the `("bar", "two")` tuple, let's git rid of some of the extra tuples.

modin/pandas/test/dataframe/test_indexing.py
mvashishtha, line 571: We can use fewer rows-- maybe just 2?

mvashishtha (commented):
Thanks @pyrito ; I have some style comments.

modin/pandas/test/dataframe/test_indexing.py
pyrito: I want to keep a few of them for some future tests potentially. I'll remove some of them.

mvashishtha (approved):
LGTM!

modin/pandas/indexing.py
```diff
@@ -563,6 +604,22 @@ def __getitem__(self, key):
         self.row_scalar = is_scalar(row_loc)
         self.col_scalar = is_scalar(col_loc)
 
+        # The thought process here is that we should check to see that we have a full key lookup
+        # for a MultiIndex DataFrame. If that's the case, then we should not drop any levels
+        # since our resulting intermediate dataframe will have dropped these for us already.
+        # Thus, we need to make sure we don't try to drop these levels again. The logic here is
+        # kind of hacked together. Ideally, we should handle this properly in the lower-level
+        # implementations, but this will have to be engineering properly later.
```
helmeleegy, line 612:
```suggestion
        # implementations, but this will have to be engineered properly later.
```

modin/pandas/indexing.py
```diff
@@ -528,6 +536,39 @@ def _handle_boolean_masking(self, row_loc, col_loc):
...
+        Returns
+        -------
+        bool
+            If the MultiIndex possibly contains the given key.
+
+        Notes
+        -----
+        This function only returns False if we have a partial key lookup. It's
+        possibly that this function returns True for a key that does NOT exist
+        since we only check the length of the tuple to match the number of levels
```
vnlitvinov, line 563:
```suggestion
        possible that this function returns True for a key that does NOT exist
        since we only check the length of the `key` tuple to match the number of levels
```

modin/pandas/indexing.py
```diff
@@ -563,6 +604,22 @@ def __getitem__(self, key):
...
         self.col_scalar = is_scalar(col_loc)
 
+        # The thought process here is that we should check to see that we have a full key lookup
+        # for a MultiIndex DataFrame. If that's the case, then we should not drop any levels
+        # since our resulting intermediate dataframe will have dropped these for us already.
+        # Thus, we need to make sure we don't try to drop these levels again. The logic here is
+        # kind of hacked together. Ideally, we should handle this properly in the lower-level
+        # implementations, but this will have to be engineering properly later.
+        self.row_multiindex_full_lookup = self._multiindex_possibly_contains_key(
+            axis=0, key=row_loc
```
vnlitvinov, line 614: I wonder if we should use the `Axis` enum for clarity:
https://github.com/modin-project/modin/blob/cad123ed479a30898066fb4094fa413fbd06696c/modin/core/dataframe/base/dataframe/utils.py#L24-L36

modin/pandas/test/dataframe/test_indexing.py
```diff
@@ -562,6 +562,44 @@ def test_loc_multi_index():
...
+            ("Electric", "Tesla"),
+            ("Electric", "Nio"),
+        ]
+    )
+    data = [[100, 300, 900, 400], [200, 500, 300, 600]]
+    modin_df, pandas_df = create_test_dfs(data, columns=cols, index=multi_index)
+    eval_general(modin_df, pandas_df, lambda df: df.loc[("r0", "rA"), :])
+    eval_general(modin_df, pandas_df, lambda df: df.loc[:, ("Gasoline", "Toyota")])
+    # Test below bc of issue #4684
+    # df_equals(df.loc[("r0"), ("Gasoline", "Toyota")], pdf.loc[("r0"), ("Gasoline", "Toyota")])
```
vnlitvinov, line 600:
```suggestion
    # FIXME: Test below fails because of issue #4684
    # eval_general(modin_df, pandas_df, lambda df: df.loc[("r0"), ("Gasoline", "Toyota")]])
```
or, even better, parametrize the test by `.loc[]` parameter and xfail the test on one of those, like
```python
@pytest.mark.parametrize(
    "loc_label",
    [
        (("r0", "rA"), slice(None)),
        (slice(None), ("Gasoline", "Toyota")),
        pytest.param(
            (("r0"), ("Gasoline", "Toyota")),
            marks=pytest.mark.xfail("See Modin issue #4684"),
        ),
    ]
)
def test_loc_multi_index_both_axes(loc_label):
    # setup here the dataframes
    eval_general(modin_df, pandas_df, lambda df: df.loc[loc_label])
```
that way, when that issue is fixed if someone forgets to un-skip the label, CI should turn red on test which we expect to fail but it passed.

modin/pandas/indexing.py
pyrito: I think this is a good idea!

modin/pandas/test/dataframe/test_indexing.py
pyrito: Alternatively, I could just get rid of this test for now because I already mention it in the Github Issue.

vnlitvinov (approved):
LGTM!

anmyachev (approved):

YarShev (approved):
LGTM!

PR #4620, opened 2022-06-29
FEAT-#4619: Integrate mypy static type checking

### LINKED ISSUE #4619: Integrate mypy static type checking
We want to follow up on #74 to add type hints to the Modin source code. #3899 discusses the idea of using pyright as our tool for static type checking, but I propose that we make use of mypy instead. The main reasons are that mypy has a larger development community, more widespread industry adoption, and better tools to scale with larger codebases. 

I propose that we integrate mypy over a series of steps:

1. Add a simple, configurable `mypy.ini` file that only checks a subset of files
2. Make according type checks for the subset of files
3. Add mypy checks to the precommit hooks
4. Add mypy checks to CI

From here, we can start to add more type checks to the rest of the code base (perhaps at a module by module basis).

pyrito:
## What do these changes do?
This PR addresses #4619 and implements the first cut of mypy integration. The goal is to start off with small modules and work our way up to the majority of our codebase. I decided to start off with the logging module since the amount of code is fairly small. Please note that this PR is still in the works as I need to add CI integration and make other changes. 

- [x] Resolves #4619 
- [x] tests added and passing

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,13 +34,11 @@ jobs:
       - uses: actions/checkout@v2
         with:
           fetch-depth: 1
-      - uses: conda-incubator/setup-miniconda@v2
+      - uses: actions/setup-python@v2
         with:
-          activate-environment: modin
-          environment-file: environment-dev.yml
-          python-version: 3.8
-          channel-priority: strict
-          use-only-tar-bz2: true # IMPORTANT: This needs to be set for caching to work properly!
+          python-version: "3.8.x"
+          architecture: "x64"
+      - run: pip install -r requirements-dev.txt
       - run: mypy --config-file mypy.ini
 
   build-docs:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,6 +27,22 @@ jobs:
       - run: pip install black
       - run: black --check --diff modin/ asv_bench/benchmarks scripts/doc_checker.py
 
+  lint-mypy:
+    name: lint (mypy)
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 1
+      - uses: conda-incubator/setup-miniconda@v2
+        with:
+          activate-environment: modin
+          environment-file: environment-dev.yml
+          python-version: 3.8
+          channel-priority: strict
+          use-only-tar-bz2: true # IMPORTANT: This needs to be set for caching to work properly!
+      - run: mypy --config-file mypy.ini
+
   build-docs:
     name: build docs
     runs-on: ubuntu-latest
@@ -187,7 +203,7 @@ jobs:
         run: python -m pytest modin/test/test_headers.py
 
   test-clean-install-ubuntu:
-    needs: [lint-flake8, lint-black, test-api, test-headers]
+    needs: [lint-flake8, lint-black, lint-mypy, test-api, test-headers]
     runs-on: ubuntu-latest
     defaults:
       run:
@@ -208,7 +224,7 @@ jobs:
           MODIN_ENGINE=ray python -c "import modin.pandas as pd; print(pd.DataFrame([1,2,3]))"
 
   test-clean-install-windows:
-    needs: [lint-flake8, lint-black, test-api, test-headers]
+    needs: [lint-flake8, lint-black, lint-mypy, test-api, test-headers]
     runs-on: windows-latest
     defaults:
       run:
@@ -229,7 +245,7 @@ jobs:
           MODIN_ENGINE=ray python -c "import modin.pandas as pd; print(pd.DataFrame([1,2,3]))"
 
   test-internals:
-    needs: [lint-flake8, lint-black, test-api, test-headers]
+    needs: [lint-flake8, lint-black, lint-mypy, test-api, test-headers]
     runs-on: ubuntu-latest
     defaults:
       run:
@@ -265,7 +281,7 @@ jobs:
       - uses: codecov/codecov-action@v2
 
   test-defaults:
-    needs: [lint-flake8, lint-black, test-api, test-headers]
+    needs: [lint-flake8, lint-black, lint-mypy, test-api, test-headers]
     runs-on: ubuntu-latest
     defaults:
       run:
@@ -315,7 +331,7 @@ jobs:
       - uses: codecov/codecov-action@v2
 
   test-omnisci:
-    needs: [lint-flake8, lint-black, test-api, test-headers]
+    needs: [lint-flake8, lint-black, lint-mypy, test-api, test-headers]
     runs-on: ubuntu-latest
     defaults:
       run:
@@ -349,7 +365,7 @@ jobs:
       - uses: codecov/codecov-action@v2
 
   test-asv-benchmarks:
-    needs: [lint-flake8, lint-black, test-api, test-headers]
+    needs: [lint-flake8, lint-black, lint-mypy, test-api, test-headers]
     runs-on: ubuntu-latest
     defaults:
       run:
@@ -415,7 +431,7 @@ jobs:
         if: failure()
 
   test-all:
-    needs: [lint-flake8, lint-black, test-api, test-headers]
+    needs: [lint-flake8, lint-black, lint-mypy, test-api, test-headers]
     runs-on: ubuntu-latest
     defaults:
       run:
@@ -492,7 +508,7 @@ jobs:
       - uses: codecov/codecov-action@v2
 
   test-experimental:
-    needs: [lint-flake8, lint-black, test-api, test-headers]
+    needs: [lint-flake8, lint-black, lint-mypy, test-api, test-headers]
     runs-on: ubuntu-latest
     defaults:
       run:
@@ -523,7 +539,7 @@ jobs:
       - uses: codecov/codecov-action@v2
 
   test-cloud:
-    needs: [lint-flake8, lint-black, test-api, test-headers]
+    needs: [lint-flake8, lint-black, lint-mypy, test-api, test-headers]
     runs-on: ubuntu-latest
     defaults:
       run:
@@ -557,7 +573,7 @@ jobs:
       - uses: codecov/codecov-action@v2
 
   test-windows:
-    needs: [lint-flake8, lint-black, test-api, test-headers]
+    needs: [lint-flake8, lint-black, lint-mypy, test-api, test-headers]
     runs-on: windows-latest
     defaults:
       run:
@@ -612,7 +628,7 @@ jobs:
       - uses: codecov/codecov-action@v2
 
   test-pyarrow:
-    needs: [lint-flake8, lint-black, test-api, test-headers]
+    needs: [lint-flake8, lint-black, lint-mypy, test-api, test-headers]
     runs-on: ubuntu-latest
     defaults:
       run:
@@ -643,7 +659,7 @@ jobs:
       - run: python -m pytest modin/pandas/test/test_io.py::TestCsv --verbose
 
   test-spreadsheet:
-    needs: [lint-flake8, lint-black, test-api, test-headers]
+    needs: [lint-flake8, lint-black, lint-mypy, test-api, test-headers]
     runs-on: ubuntu-latest
     defaults:
       run:
--- a/contributing/pre-commit
+++ b/contributing/pre-commit
@@ -11,4 +11,7 @@ printf "running flake8. This script will preempt the commit if flake8 fails.\n"
 flake8 modin/ asv_bench/benchmarks scripts/doc_checker.py
 printf "flake8 passed!\n"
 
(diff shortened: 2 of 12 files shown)


modin/logging/config.py
```diff
@@ -83,14 +86,17 @@ def bytes_int_to_str(num_bytes, suffix="B"):
         Human-readable string format.
     """
     factor = 1000
+    bytes: float = num_bytes
```
mvashishtha, line 89: `bytes` is a built-in keyword, so we shouldn't use it.

modin/logging/logger_decorator.py
```diff
@@ -79,10 +79,10 @@ def enable_logging(
     log_level = log_level.lower()
     assert hasattr(Logger, log_level.lower()), f"Invalid log level: {log_level}"
 
-    def decorator(obj):
+    def decorator(obj: Any) -> Any:
         """Decorate function or class to add logs to Modin API function(s)."""
         if isinstance(obj, type):
-            seen = {}
+            seen: Dict[Any, Any] = {}
```
mvashishtha, line 85:
```suggestion
            seen: Dict[str, Any] = {}
```

Is this correct?

modin/logging/logger_decorator.py
```diff
@@ -103,11 +103,13 @@ def decorator(obj):
         elif isinstance(obj, staticmethod):
             return staticmethod(decorator(obj.__func__))
 
+        assert isinstance(modin_layer, str)
```
mvashishtha, line 106: If we assert that `modin_layer` is a string, why does the typing allow it not to be a string?

mypy.ini
```diff
@@ -0,0 +1,23 @@
+[mypy]
+python_version = 3.9
```
mvashishtha, line 2: Why 3.9? I know python 3.10 currently doesn't work with Modin. Will we have to update this?

mypy.ini
```diff
@@ -0,0 +1,23 @@
...
+follow_imports = silent
+
+# be strict
+disallow_untyped_calls=True
+disallow_untyped_defs=True
+strict_optional=True
+warn_no_return=True
+warn_redundant_casts=True
+warn_unused_ignores=True
+# Should we disallow generics?
```
mvashishtha, line 17: I think not.

contributing/pre-commit
```diff
@@ -11,4 +11,7 @@ printf "running flake8. This script will preempt the commit if flake8 fails.\n"
 flake8 modin/ asv_bench/benchmarks scripts/doc_checker.py
 printf "flake8 passed!\n"
 
+printf "running mypy. This script will preempt the commit if mypy fails.\n"
```
mvashishtha, line 14: I think I'm the only person who uses this hook :) We should add the MyPy check to [CI](https://github.com/modin-project/modin/blob/3982306a540bd09b43a21dcdca6b16b4741ed151/.github/workflows/ci.yml). Is there a way to have it check all changed lines in other files, in addition to all of `**/modin/logging/*.py`?

modin/logging/config.py
pyrito: This is a good point, I'll make the change

modin/logging/logger_decorator.py
pyrito: There's a bit of complexity here. Basically `enable_logging` returns a function since it's a decorator. If you see in the top of the function, we check to see whether `modin_layer` is a string or not. If it's not, then we call `enable_logging()(modin_layer)` which is basically doing `decorator(modin_layer)`. We can see in the implementation of `decorator` that we check to see what the type of `obj` is (and we claim earlier that it can be a string, type, or callable function). We check for these cases but I think `mypy` isn't convinced that we can get away with calling string methods on the union, so I added an assert for sanity. Does that make sense?

pyrito: mypy threw this error specifically:
```
modin/logging/logger_decorator.py:106:32: error: Item "function" of "Union[str, Callable[..., Any], Type[Any]]" has no attribute "upper"  [union-attr]
modin/logging/logger_decorator.py:107:30: error: Item "function" of "Union[str, Callable[..., Any], Type[Any]]" has no attribute "upper"  [union-attr]
```

contributing/pre-commit
mvashishtha: But we should also keep this change! I know I will want to add this to my pre-commit hook.

mypy.ini
pyrito: Reading the docs online, it looks like mypy will default to using whatever python version is used to run `mypy` itself. I set 3.9 because it seems it doesn't accept minor versions (e.g. 3.9.12) and I wasn't sure about 3.10. I could just get rid of this, but I figured that pinning a version is always better than not doing so.

contributing/pre-commit
jeffreykennethli: I use this hook too!

pyrito: Haha, I actually use this pre-commit hook. It's super useful! And yes, I intend on adding the MyPy check to CI. I'm not sure if we can have it check all changed lines in other files. I think it would be hard for mypy to do it only on recent diffs. I think we're going to have to slowly make type changes over time, but I'll try to see if I can find anything online about this.

modin/logging/logger_decorator.py
pyrito: I don't think so. If you see the usage of `seen` it seems to be looking things up with `attr_value` as the key, and the type for that seems to be `Any`.

modin/logging/config.py
```diff
@@ -37,7 +38,9 @@
 class ModinFormatter(logging.Formatter):  # noqa: PR01
     """Implement custom formatter to log at microsecond granularity."""
 
-    def formatTime(self, record, datefmt=None):
+    def formatTime(
+        self, record: logging.LogRecord, datefmt: Optional[str] = None
+    ) -> str:
```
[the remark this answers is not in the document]
pyrito: Yes, I was going to get to this. I was thinking of making a separate PR or I could just roll it into this one?

modin/logging/config.py
```diff
@@ -83,14 +86,17 @@ def bytes_int_to_str(num_bytes, suffix="B"):
...
-    return f"{num_bytes * 1000:.2f}P{suffix}"
+        if bytes < factor:
+            return f"{bytes:.2f}{unit}{suffix}"
+        bytes /= factor
+    return f"{bytes * 1000:.2f}P{suffix}"
 
 
-def _create_logger(namespace, job_id, log_name, log_level):
+def _create_logger(
+    namespace: str, job_id: str, log_name: str, log_level: int
```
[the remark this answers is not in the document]
pyrito: We would need to change some of the code to go from using the enum to a string. Is that a change we want to make?

modin/logging/logger_decorator.py
pyrito: @mvashishtha would it make sense to add a string to give more context on a potential failure on this assert here?

mvashishtha: Yeah, I think a short comment would be good.

modin/logging/logger_decorator.py
vnlitvinov: It should really be `Dict[Callable, Callable]`, as it's mapping unwrapped function to wrapped function, to handle cases like
```python
@enable_logging
class Foo:
    def bar(self):
        pass
    qux = bar

assert Foo.bar == Foo.qux
```

modin/logging/logger_decorator.py
vnlitvinov: I would advise against using bare `assert`-s, I think it's better to accompany each such statement with a proper assertion message which would double-serve as a comment

mypy.ini
vnlitvinov: I would rather we don't pin the version, as pinned version prohibits from running the checks with whatever version it is installed.

As for 3.10 - I think Modin should support that (we should be pretty agnostic on upper bound, requiring >=3.8), it's Ray which doesn't support 3.10 now.

requirements-dev.txt
```diff
@@ -46,3 +46,6 @@ flake8
 numpydoc==1.1.0
 # experimental version of fuzzydata requires at least 0.0.6 to successfully resolve all dependencies
 fuzzydata>=0.0.6
+# Pin versions of mypy and pandas stubs for stability
+mypy==0.961
+pandas-stubs==1.4.2.220626
```
vnlitvinov, line 51: we should now have a doc saying "how to update pandas version", as we have a bunch of different places where it is mentioned

vnlitvinov:
Unless `mypy` is run by CI, it won't protect anything really. Please add a CI check for that!

modin/logging/config.py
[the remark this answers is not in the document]
pyrito: I think we would also have to reflect the changes on all the callers of the function. Is that something. we want to do as well?

modin/logging/logger_decorator.py
pyrito: I had tried that before but mypy complained, so I kind of copped out to use the `Any`:
```
modin/logging/logger_decorator.py:91:40: error: Invalid index type "Union[FunctionType, MethodType, classmethod[Any], staticmethod[Any]]" for "Dict[Callable[..., Any], Callable[..., Any]]"; expected type "Callable[..., Any]"  [index]
modin/logging/logger_decorator.py:93:40: error: Invalid index type "Union[FunctionType, MethodType, classmethod[Any], staticmethod[Any]]" for "Dict[Callable[..., Any], Callable[..., Any]]"; expected type "Callable[..., Any]"  [index]
```

.github/workflows/ci.yml
```diff
@@ -27,6 +27,20 @@ jobs:
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 1
+      - uses: actions/setup-python@v2
+        with:
+          python-version: "3.8.x"
+          architecture: "x64"
+      - run: pip install -r requirements-dev.txt
```
mvashishtha, line 41: Can we just `pip install mypy` as we do with `black` above?

environment-dev.yml
```diff
@@ -36,6 +36,8 @@ dependencies:
   - scikit-learn
   - pymssql
   - psycopg2
+  - mypy==0.961
```
mvashishtha, line 39: We usually try not to pin. Is there a reason to pin this version? I see we don't pin `black` or `flake8`.

modin/logging/config.py
```diff
@@ -83,14 +86,17 @@ def bytes_int_to_str(num_bytes, suffix="B"):
...
+    n_bytes: float = num_bytes
     for unit in ["", "K", "M", "G", "T", "P"]:
-        if num_bytes < factor:
-            return f"{num_bytes:.2f}{unit}{suffix}"
-        num_bytes /= factor
-    return f"{num_bytes * 1000:.2f}P{suffix}"
+        if n_bytes < factor:
+            return f"{n_bytes:.2f}{unit}{suffix}"
+        n_bytes /= factor
+    return f"{n_bytes * 1000:.2f}P{suffix}"
```
naren-ponder, line 94: Why is this change necessary? Can we instead make `num_bytes` a float at the function signature and change the docstring? I'm just not a fan of `n_bytes: float = num_bytes`.

.github/workflows/ci.yml
pyrito: I decided to install all of it because I think `mypy` might require all packages to be installed so that it can do the type checking.

modin/logging/config.py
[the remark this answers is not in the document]
jeffreykennethli: Yes, but no need to do any of this in this PR, this was more of a general comment. If you think it's worth changing, feel free to open an issue for it.

environment-dev.yml
pyrito: I pinned the version to avoid any kind of surprises but I guess there's no reason to do so. I can get rid of the version pinning.

jeffreykennethli (approved):

modin/logging/config.py
pyrito: If I were to leave it as is, I'd get an error like this:
```
modin/logging/config.py:93:9: error: Incompatible types in assignment (expression has type "float", variable has type "int")  [assignment]
```

One alternative solution might be to do something like: 
```python
n_bytes = num_bytes # type: float
```
This is a little janky workaround and seems like a semantic difference. Maybe if we wanted to be more explicit we could do something like:
```python
n_bytes = float(num_bytes)
```

Does anyone else have any thoughts about this? @modin-project/modin-contributors

naren-ponder: Agreed with @mvashishtha.

modin/logging/config.py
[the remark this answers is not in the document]
pyrito: Yeah looking at it now, I think that we should probably unify the log_levels to either be `str` or `int` or maybe even both. I'll create an issue for this. It might be a good one for new folks to try working on.

PR #4621, opened 2022-06-30
FIX-#4593: Ensure Modin warns when setting columns via attributes

### LINKED ISSUE #4593: Modin should match pandas UserWarning for assigning columns via attributes
### System information
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: macOS 12.2.1
- **Modin version** (`modin.__version__`): 0.15.0+16.g4574746c
- **Python version**: 3.9.12

```python
import modin.pandas as pd
import pandas

pandas_df = pandas.DataFrame({"a": [1, 2, 3]})
modin_df = pd.DataFrame({"a": [1, 2, 3]})

# This throws a user warning
pandas_df.b = [1,2,3]

# This does not throw a user warning
modin_df.b = [1,2,3]
```

### Describe the problem
Modin doesn't throw a user warning for trying to assign new columns by setting an attribute. This isn't a super pressing issue, but we should at least mimic pandas behavior here.

helmeleegy:
Signed-off-by: Hazem Elmeleegy <hazem@ponder.io>

## What do these changes do?

- [x] Resolves #4593 
- [x] tests added and passing

### DIFF
--- a/modin/pandas/dataframe.py
+++ b/modin/pandas/dataframe.py
@@ -2496,18 +2496,18 @@ class DataFrame(BasePandasDataset):
         # - `_siblings`, which Modin initializes before it appears in __dict__
         if key in ["_query_compiler", "_siblings"] or key in self.__dict__:
             pass
-        elif key in self and key not in dir(self):
+        elif key not in dir(self):
+            if key not in self:
+                warnings.warn(
+                    "Modin doesn't allow columns to be created via a new attribute name - see "
+                    + "https://pandas.pydata.org/pandas-docs/stable/indexing.html#attribute-access",
+                    UserWarning,
+                )
             self.__setitem__(key, value)
             # Note: return immediately so we don't keep this `key` as dataframe state.
             # `__getattr__` will return the columns not present in `dir(self)`, so we do not need
             # to manually track this state in the `dir`.
             return
-        elif isinstance(value, pandas.Series):
-            warnings.warn(
-                "Modin doesn't allow columns to be created via a new attribute name - see "
-                + "https://pandas.pydata.org/pandas-docs/stable/indexing.html#attribute-access",
-                UserWarning,
-            )
         object.__setattr__(self, key, value)
 
     def __setitem__(self, key, value):
--- a/modin/pandas/dataframe.py
+++ b/modin/pandas/dataframe.py
@@ -2498,7 +2498,7 @@ class DataFrame(BasePandasDataset):
             pass
         elif key not in dir(self) and key not in self and not is_list_like(value):
             # Assigning a scalar to a non-existing attribute should not result in a
-            # broadcasted column, but rather just an an attribute
+            # broadcasted column, but rather just a new attribute.
             pass
         elif key not in dir(self):
             if key not in self:
--- a/modin/pandas/dataframe.py
+++ b/modin/pandas/dataframe.py
@@ -2496,6 +2496,10 @@ class DataFrame(BasePandasDataset):
         # - `_siblings`, which Modin initializes before it appears in __dict__
         if key in ["_query_compiler", "_siblings"] or key in self.__dict__:
             pass
+        elif key not in dir(self) and key not in self and not is_list_like(value):
+            # Assigning a scalar to a non-existing attribute should not result in a
+            # broadcasted column, but rather just an an attribute
+            pass
         elif key not in dir(self):
             if key not in self:
                 warnings.warn(
--- a/modin/pandas/dataframe.py
+++ b/modin/pandas/dataframe.py
@@ -2496,22 +2496,18 @@ class DataFrame(BasePandasDataset):
         # - `_siblings`, which Modin initializes before it appears in __dict__
         if key in ["_query_compiler", "_siblings"] or key in self.__dict__:
             pass
-        elif key not in dir(self) and key not in self and not is_list_like(value):
-            # Assigning a scalar to a non-existing attribute should not result in a
-            # broadcasted column, but rather just a new attribute.
-            pass
-        elif key not in dir(self):
-            if key not in self:
-                warnings.warn(
-                    "Modin doesn't allow columns to be created via a new attribute name - see "
-                    + "https://pandas.pydata.org/pandas-docs/stable/indexing.html#attribute-access",
-                    UserWarning,
-                )
+        elif key in self and key not in dir(self):
             self.__setitem__(key, value)
             # Note: return immediately so we don't keep this `key` as dataframe state.
             # `__getattr__` will return the columns not present in `dir(self)`, so we do not need
             # to manually track this state in the `dir`.
             return
+        elif is_list_like(value):
+            warnings.warn(
+                "Modin doesn't allow columns to be created via a new attribute name - see "
+                + "https://pandas.pydata.org/pandas-docs/stable/indexing.html#attribute-access",
+                UserWarning,
+            )
         object.__setattr__(self, key, value)
 
     def __setitem__(self, key, value):
--- a/modin/pandas/test/dataframe/test_iter.py
+++ b/modin/pandas/test/dataframe/test_iter.py
@@ -290,7 +290,8 @@ def test___setattr__mutating_column():
     df_equals(modin_df, pandas_df)
     assert modin_df.col0.equals(modin_df["col0"])
 
-    # Check that adding a new col via attributes raises warning
+    # Check that attempting to add a new col via attributes raises warning
+    # and adds the provided list as a new attribute and not a column.
     with pytest.warns(
         UserWarning,
         match="Modin doesn't allow columns to be created via a new attribute name - see",
@@ -298,9 +299,12 @@ def test___setattr__mutating_column():
         modin_df.col1 = [4]
 
     with warnings.catch_warnings():
-        warnings.simplefilter("error")
+        warnings.filterwarnings(
+            action="error", 
+            message="Modin doesn't allow columns to be created via a new attribute name - see")
         modin_df.col1 = [5]
         modin_df.new_attr = 6
+        modin_df.col0 = 7
 
     assert "new_attr" in dir(
         modin_df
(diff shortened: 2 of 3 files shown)


pyrito (commented):
Good start! Could you also check that this is working as mentioned in the issue by writing a small test? Also please don't forget to add your name to the release notes in `docs/release_notes`

modin/pandas/dataframe.py
```diff
@@ -2496,18 +2496,18 @@ def __setattr__(self, key, value):
...
         if key in ["_query_compiler", "_siblings"] or key in self.__dict__:
             pass
-        elif key in self and key not in dir(self):
+        elif key not in dir(self):
+            if key not in self:
+                warnings.warn(
+                    "Modin doesn't allow columns to be created via a new attribute name - see "
+                    + "https://pandas.pydata.org/pandas-docs/stable/indexing.html#attribute-access",
+                    UserWarning,
+                )
```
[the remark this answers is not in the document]
helmeleegy: Thanks for the comments! The logic is updated now, and it checks for `is_list_like`, which I believe is more general the pre-existing check for `pandas.Series`, and hence that old check is now removed.

Also, tests are added and `docs/release_notes` is updated. PTAL when you get a chance.

modin/pandas/dataframe.py
```diff
@@ -2496,18 +2496,22 @@ def __setattr__(self, key, value):
...
         if key in ["_query_compiler", "_siblings"] or key in self.__dict__:
             pass
-        elif key in self and key not in dir(self):
+        elif key not in dir(self) and key not in self and not is_list_like(value):
+            # Assigning a scalar to a non-existing attribute should not result in a
+            # broadcasted column, but rather just a new attribute.
+            pass
+        elif key not in dir(self):
+            if key not in self:
+                warnings.warn(
```
pyrito, line 2505: I believe the logic here is leading to test failures in CI, and this was actually a bug we encountered before where we were improperly setting the attributes. If the key is in self and not in dir, we need to call `__setitem__` and return so that we don't call the `object.__setattr__` at the bottom of the call. For example, in the failing test case we have this scenario where we are adding a new column as an attribute:
```python
def test___setattr__not_column():
    pandas_df = pandas.DataFrame([1, 2, 3])
    modin_df = pd.DataFrame([1, 2, 3])
    pandas_df.new_col = [4, 5, 6]
    modin_df.new_col = [4, 5, 6]
```

The current code as is will check if key is not in dir(self), which is true and will also see that key is not in self. We will print the warning, but we will also call `__setitem__` which is not what we want.

modin/pandas/dataframe.py
```diff
@@ -2496,18 +2496,22 @@ def __setattr__(self, key, value):
         # - `_siblings`, which Modin initializes before it appears in __dict__
         if key in ["_query_compiler", "_siblings"] or key in self.__dict__:
             pass
-        elif key in self and key not in dir(self):
+        elif key not in dir(self) and key not in self and not is_list_like(value):
```
pyrito, line 2499: This seems some what of confusing logic to me. I think the way it was expressed the comment above somewhat clearly: if we have an attribute that already exists (column that already exists), we can go ahead and assign the new value. However, if that's not the case, we should promptly let the user know that we can't set new columns as an attribute.

modin/pandas/test/dataframe/test_iter.py
```diff
@@ -289,6 +290,28 @@ def test___setattr__mutating_column():
     df_equals(modin_df, pandas_df)
     assert modin_df.col0.equals(modin_df["col0"])
 
+    # Check taht adding a new col via attributes raises warning
```
pyrito, line 293:
```suggestion
    # Check that adding a new col via attributes raises warning
```

modin/pandas/dataframe.py
```diff
@@ -2496,18 +2496,22 @@ def __setattr__(self, key, value):
...
+                    "Modin doesn't allow columns to be created via a new attribute name - see "
+                    + "https://pandas.pydata.org/pandas-docs/stable/indexing.html#attribute-access",
+                    UserWarning,
+                )
             self.__setitem__(key, value)
             # Note: return immediately so we don't keep this `key` as dataframe state.
             # `__getattr__` will return the columns not present in `dir(self)`, so we do not need
             # to manually track this state in the `dir`.
             return
-        elif isinstance(value, pandas.Series):
```
pyrito, line 2505: We could add an extra `elif` case that checks to see if the value the user is trying to set `is_list_like()` here and throw the warning in this case.

modin/pandas/test/dataframe/test_iter.py
```diff
@@ -289,6 +290,28 @@ def test___setattr__mutating_column():
...
+    with pytest.warns(
+        UserWarning,
+        match="Modin doesn't allow columns to be created via a new attribute name - see",
+    ):
+        modin_df.col1 = [4]
+
+    with warnings.catch_warnings():
+        warnings.simplefilter("error")
+        modin_df.col1 = [5]
+        modin_df.col0 = 6
```
pyrito, line 303: What does this bit do? Looks like this test is failing here in CI

helmeleegy: This was originally suggested by @RehanSD. I think the idea is to make sure that these statements are not raising warnings. The test does pass locally using pytest. I'm not sure why it doesn't in CI. I removed one of the checks in the last commit, but it's still failing. I'm not yet sure why honestly.

modin/pandas/dataframe.py
helmeleegy: I think replacement is sufficient rather than having both checks, right?

modin/pandas/test/dataframe/test_iter.py
```diff
@@ -289,6 +290,26 @@ def test___setattr__mutating_column():
...
+        modin_df.col1 = [5]
+        modin_df.new_attr = 6
+
+    assert "new_attr" in dir(
+        modin_df
+    ), "Modin attribute was not correctly added to the df."
+    assert (
+        "new_attr" not in modin_df
+    ), "New attribute was not correctly added to columns."
+    assert modin_df.new_attr == 7, "Modin attribute value was set incorrectly."
```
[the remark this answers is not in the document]
helmeleegy: Yes!

modin/pandas/dataframe.py
helmeleegy: I think this is outdated now. See below.

modin/pandas/test/dataframe/test_iter.py
pyrito: Oh I think I might know why. I think the base execution engine is Ray. The test you are failing uses a base execution engine of python so it's failing because it throws a warning that "we are defaulting to pandas". You can try running this locally yourself: `pytest -n 2 modin/pandas/test/dataframe/test_iter.py --execution=BaseOnPython`

helmeleegy: I think this is fixed now by matching on the warning message.

modin/pandas/dataframe.py
naren-ponder: Replacement is sufficient since `is_list_like()` will always be True for `pandas.Series`.

naren-ponder (approved):

modin/pandas/test/dataframe/test_iter.py
```diff
@@ -289,6 +290,34 @@ def test___setattr__mutating_column():
...
+    # Check that attempting to add a new col via attributes raises warning
+    # and adds the provided list as a new attribute and not a column.
+    with pytest.warns(
+        UserWarning,
+        match="Modin doesn't allow columns to be created via a new attribute name - see",
+    ):
+        modin_df.col1 = [4]
+
+    with warnings.catch_warnings():
+        warnings.filterwarnings(
```
pyrito, line 302: Why are we catching the warnings and filtering them out? Is this so ensure that only these warnings are thrown here?

pyrito (approved):
LGTM! Just had one question

modin/pandas/test/dataframe/test_iter.py
```diff
@@ -289,6 +290,34 @@ def test___setattr__mutating_column():
     df_equals(modin_df, pandas_df)
     assert modin_df.col0.equals(modin_df["col0"])
 
+    # Check that attempting to add a new col via attributes raises warning
+    # and adds the provided list as a new attribute and not a column.
+    with pytest.warns(
+        UserWarning,
+        match="Modin doesn't allow columns to be created via a new attribute name - see",
```
YarShev, line 297: This warning should be probably extended. "see ..." what?

modin/pandas/test/dataframe/test_iter.py
```diff
@@ -289,6 +290,34 @@ def test___setattr__mutating_column():
...
+    ):
+        modin_df.col1 = [4]
+
+    with warnings.catch_warnings():
+        warnings.filterwarnings(
+            action="error",
+            message="Modin doesn't allow columns to be created via a new attribute name - see",
```
YarShev, line 304: same

modin/pandas/test/dataframe/test_iter.py
helmeleegy: Yes, only the warnings we care about will cause the test to fail if they were raised (they should not!). Other types of warnings we have no control over. They may or may not be raised (possibly depending on the environment), so we want to ignore them (filter them out).

helmeleegy: Does it make sense?

pyrito: Gotcha, makes sense

modin/pandas/test/dataframe/test_iter.py
helmeleegy: Done.

modin/pandas/test/dataframe/test_iter.py
helmeleegy: Done.

pyrito (approved):
LGTM! @modin-project/modin-core  please take a look and give a stamp of approval if all looks good!

YarShev (approved):
@helmeleegy, LGTM, thanks!

PR #4647, opened 2022-07-06
FEAT-#4569: Add error message when `read_` function defaults to pandas

### LINKED ISSUE #4569: FEAT: Give users a better explanation for why read_ functions are defaulting to pandas.

[`check_parameters_support`](https://github.com/modin-project/modin/blob/45b65d99531d6c81fb787598caa4116c3bf11d9e/modin/core/io/text/fwf_dispatcher.py#L44) in `TextFileDispatcher`, common to `read_csv` and other file format readers, can default to pandas for many reasons. Right now users just see `UserWarning: Parameters provided defaulting to pandas implementation`. This makes it hard for the user to tell what to change, as in #4565. We should give a more informative reason. We can implement that by having `check_parameters_support` return both a bool and a reason, as omnisci's [`_read_csv_check_support`](https://github.com/modin-project/modin/blob/7c02bafa1badade0ba9d1916652ded74bc36a1b9/modin/experimental/core/execution/native/implementations/omnisci_on_native/io/io.py#L403) does.

noloerino:
## What do these changes do?
Adds an explanatory reason when `read_` functions default to the pandas implementation.

This PR adds a mandatory `reason` argument to [PandasParser.single_worker_read](https://github.com/modin-project/modin/blob/ce1d93a3ae397b4951ee81b9c971b2c9a72e30df/modin/core/storage_formats/pandas/parsers.py#L252). I'm not entirely sure if I have all the reasons described correctly, as I couldn't find docstrings for some of the arguments that were being checked. It may also be worthwhile for the error message for a missing path to distinguish whether the missing path wasn't found on S3, locally, or wherever else, but that may require some more difficult changes.

I have not added new tests; even the existing tests for the default-to-pandas cases seem a little brittle because of #3655. I did, however, add an extra check for the specific message being raised, which seems to work fine at least in the `read_csv` case.

- [x] Resolves #4569 
- [ ] tests added and passing

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -177,3 +177,7 @@ node_modules
 # Asv stuff
 asv_bench/.asv/
 asv_bench/modin/
+
+# Sublime stuff
+*.sublime-workspace
+*.sublime-project
--- a/docs/release_notes/release_notes-0.16.0.rst
+++ b/docs/release_notes/release_notes-0.16.0.rst
@@ -59,8 +59,8 @@ Key Features and Updates
 * New Features
   * FEAT-4463: Add experimental fuzzydata integration for testing against a randomized dataframe workflow (#4556)
   * FEAT-#4419: Extend virtual partitioning API to pandas on Dask (#4420)
-  * FEAT-#4569: Add error message when `read_` function defaults to pandas
   * FEAT-#4147: Add partial compatibility with Python 3.6 and pandas 1.1 (#4301)
+  * FEAT-#4569: Add error message when `read_` function defaults to pandas (#4647)
 
 Contributors
 ------------
--- a/docs/release_notes/release_notes-0.16.0.rst
+++ b/docs/release_notes/release_notes-0.16.0.rst
@@ -59,6 +59,7 @@ Key Features and Updates
 * New Features
   * FEAT-4463: Add experimental fuzzydata integration for testing against a randomized dataframe workflow (#4556)
   * FEAT-#4419: Extend virtual partitioning API to pandas on Dask (#4420)
+  * FEAT-#4569: Add error message when `read_` function defaults to pandas
   * FEAT-#4147: Add partial compatibility with Python 3.6 and pandas 1.1 (#4301)
 
 Contributors
--- a/modin/core/io/column_stores/hdf_dispatcher.py
+++ b/modin/core/io/column_stores/hdf_dispatcher.py
@@ -70,7 +70,7 @@ class HDFDispatcher(ColumnStoreDispatcher):  # pragma: no cover
         if cls._validate_hdf_format(path_or_buf=path_or_buf) is None:
             return cls.single_worker_read(
                 path_or_buf,
-                "File format seems to be `fixed`. For better distribution consider "
+                reason="File format seems to be `fixed`. For better distribution consider "
                 + "saving the file in `table` format. df.to_hdf(format=`table`).",
                 **kwargs
             )
--- a/modin/core/io/column_stores/hdf_dispatcher.py
+++ b/modin/core/io/column_stores/hdf_dispatcher.py
@@ -16,7 +16,6 @@
 import pandas
 
 from modin.core.io.column_stores.column_store_dispatcher import ColumnStoreDispatcher
-from modin.error_message import ErrorMessage
 
 
 class HDFDispatcher(ColumnStoreDispatcher):  # pragma: no cover
@@ -69,11 +68,12 @@ class HDFDispatcher(ColumnStoreDispatcher):  # pragma: no cover
             Query compiler with imported data for further processing.
         """
         if cls._validate_hdf_format(path_or_buf=path_or_buf) is None:
-            ErrorMessage.default_to_pandas(
+            return cls.single_worker_read(
+                path_or_buf,
                 "File format seems to be `fixed`. For better distribution consider "
-                + "saving the file in `table` format. df.to_hdf(format=`table`)."
+                + "saving the file in `table` format. df.to_hdf(format=`table`).",
+                **kwargs
             )
-            return cls.single_worker_read(path_or_buf, **kwargs)
 
         columns = kwargs.pop("columns", None)
         # Have to do this because of Dask's keyword arguments
--- a/modin/core/io/column_stores/parquet_dispatcher.py
+++ b/modin/core/io/column_stores/parquet_dispatcher.py
@@ -73,9 +73,9 @@ class ParquetDispatcher(ColumnStoreDispatcher):
             if len(partitioned_columns):
                 return cls.single_worker_read(
                     path,
-                    "Mixed partitioning columns in Parquet",
                     engine=engine,
                     columns=columns,
+                    reason="Mixed partitioning columns in Parquet",
                     **kwargs
                 )
 
--- a/modin/core/io/column_stores/parquet_dispatcher.py
+++ b/modin/core/io/column_stores/parquet_dispatcher.py
@@ -16,7 +16,6 @@
 import os
 
 from modin.core.io.column_stores.column_store_dispatcher import ColumnStoreDispatcher
-from modin.error_message import ErrorMessage
 from modin.utils import import_optional_dependency
 
 
@@ -72,9 +71,12 @@ class ParquetDispatcher(ColumnStoreDispatcher):
                     break
             partitioned_columns = list(partitioned_columns)
             if len(partitioned_columns):
-                ErrorMessage.default_to_pandas("Mixed Partitioning Columns in Parquet")
                 return cls.single_worker_read(
-                    path, engine=engine, columns=columns, **kwargs
+                    path,
+                    "Mixed partitioning columns in Parquet",
+                    engine=engine,
+                    columns=columns,
+                    **kwargs
                 )
 
         if not columns:
--- a/modin/core/io/file_dispatcher.py
+++ b/modin/core/io/file_dispatcher.py
@@ -343,5 +343,5 @@ class FileDispatcher(ClassLogger):
         )
 
     @classmethod
-    def _file_not_found_msg(filename):
+    def _file_not_found_msg(cls, filename):
         return f"No file with name '{filename}'"
--- a/modin/core/io/file_dispatcher.py
+++ b/modin/core/io/file_dispatcher.py
@@ -344,4 +344,17 @@ class FileDispatcher(ClassLogger):
 
     @classmethod
     def _file_not_found_msg(cls, filename):
+        """
+        Produce a message indicating that a file was not found.
+
+        Parameters
+        ----------
+        filename : str
+            The name of the file that was not found.
+
+        Returns
+        -------
+        str
+            A message indicating `filename` was not found.
+        """
         return f"No file with name '{filename}'"
--- a/modin/core/io/file_dispatcher.py
+++ b/modin/core/io/file_dispatcher.py
@@ -358,3 +358,7 @@ class FileDispatcher(ClassLogger):
             A message indicating `filename` was not found.
         """
         return f"No file with name '{filename}'"
+
+    @classmethod
+    def _buffer_unsupported_msg(cls):
+        return "Reading from buffers or other non-path-like objects is not supported"
--- a/modin/core/io/file_dispatcher.py
+++ b/modin/core/io/file_dispatcher.py
@@ -361,4 +361,5 @@ class FileDispatcher(ClassLogger):
 
     @classmethod
     def _buffer_unsupported_msg(cls):
+        """Produce a message indicating that buffers are not supported."""
         return "Reading from buffers or other non-path-like objects is not supported"
--- a/modin/core/io/file_dispatcher.py
+++ b/modin/core/io/file_dispatcher.py
@@ -341,3 +341,7 @@ class FileDispatcher(ClassLogger):
                 for i in range(len(partition_ids))
             ]
         )
+
+    @classmethod
+    def _file_not_found_msg(filename):
+        return f"No file with name '{filename}'"
--- a/modin/core/io/file_dispatcher.py
+++ b/modin/core/io/file_dispatcher.py
@@ -343,7 +343,7 @@ class FileDispatcher(ClassLogger):
         )
 
     @classmethod
-    def _file_not_found_msg(cls, filename):
+    def _file_not_found_msg(cls, filename: str):
         """
         Produce a message indicating that a file was not found.
 
--- a/modin/core/io/file_dispatcher.py
+++ b/modin/core/io/file_dispatcher.py
@@ -122,6 +122,10 @@ class FileDispatcher(ClassLogger):
     classes).
     """
 
+    BUFFER_UNSUPPORTED_MSG = (
+        "Reading from buffers or other non-path-like objects is not supported"
+    )
+
     frame_cls = None
     frame_partition_cls = None
     query_compiler_cls = None
@@ -343,23 +347,5 @@ class FileDispatcher(ClassLogger):
         )
 
     @classmethod
-    def _file_not_found_msg(cls, filename: str):
-        """
-        Produce a message indicating that a file was not found.
-
-        Parameters
-        ----------
-        filename : str
-            The name of the file that was not found.
-
-        Returns
-        -------
-        str
-            A message indicating `filename` was not found.
-        """
+    def _file_not_found_msg(cls, filename: str):  # noqa: GL08
         return f"No file with name '{filename}'"
-
-    @classmethod
-    def _buffer_unsupported_msg(cls):
-        """Produce a message indicating that buffers are not supported."""
-        return "Reading from buffers or other non-path-like objects is not supported"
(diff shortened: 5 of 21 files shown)


.gitignore
```diff
@@ -177,3 +177,7 @@ node_modules
 # Asv stuff
 asv_bench/.asv/
 asv_bench/modin/
+
```
noloerino, line 180: I included this by mistake when I ran `git add -u`, should I make a separate PR for it? (it doesn't really correspond to any issue)

docs/release_notes/release_notes-0.16.0.rst
```diff
@@ -44,6 +44,7 @@ Key Features and Updates
   * FEAT-#4619: Integrate mypy static type checking (#4620)
 * New Features
   * FEAT-4463: Add experimental fuzzydata integration for testing against a randomized dataframe workflow (#4556)
+  * FEAT-#4569: Add error message when `read_` function defaults to pandas
```
jeffreykennethli, line 47: Add in parenthesis the PR number `(#4647)`

modin/core/io/text/excel_dispatcher.py
```diff
@@ -48,15 +48,18 @@ def _read(cls, io, **kwargs):
...
                 + 'please specify `engine=None` or `engine="openpyxl"` to '
-                + "use Modin's parallel implementation."
+                + "use Modin's parallel implementation.",
+                **kwargs
             )
-            return cls.single_worker_read(io, **kwargs)
         if sys.version_info < (3, 7):
-            warnings.warn("Python 3.7 or higher required for parallel `read_excel`.")
-            return cls.single_worker_read(io, **kwargs)
+            warnings.warn()
```
pyrito, line 59: Stray warning message?

modin/core/io/pickle/pickle_dispatcher.py
```diff
@@ -47,9 +47,9 @@ def _read(cls, filepath_or_buffer, **kwargs):
         The number of partitions is equal to the number of input files.
         """
         if not (isinstance(filepath_or_buffer, str) and "*" in filepath_or_buffer):
-            warnings.warn("Defaulting to Modin core implementation")
             return cls.single_worker_read(
                 filepath_or_buffer,
+                "Defaulting to Modin core implementation",
```
jeffreykennethli, line 52: should we be saying Modin core? Does this mean anything to the user, or would it be better to just say "Modin".

modin/core/io/column_stores/parquet_dispatcher.py
```diff
@@ -72,9 +71,12 @@ def _read(cls, path, engine, columns, **kwargs):
                     break
             partitioned_columns = list(partitioned_columns)
             if len(partitioned_columns):
-                ErrorMessage.default_to_pandas("Mixed Partitioning Columns in Parquet")
                 return cls.single_worker_read(
-                    path, engine=engine, columns=columns, **kwargs
+                    path,
+                    "Mixed partitioning columns in Parquet",
```
pyrito, line 76: Do you know what this error message is referring to? It's not clear what "mixed partitioning columns" means to me at least.

modin/core/io/text/csv_glob_dispatcher.py
```diff
@@ -64,10 +64,18 @@ def _read(cls, filepath_or_buffer, **kwargs):
                     + f"files at once. Did you forget it? Passed filename: '{filepath_or_buffer}'"
                 )
             if not cls.file_exists(filepath_or_buffer, kwargs.get("storage_options")):
-                return cls.single_worker_read(filepath_or_buffer, **kwargs)
+                return cls.single_worker_read(
+                    filepath_or_buffer,
+                    cls._file_not_found_msg(filepath_or_buffer),
```
pyrito, line 69: `filepath_or_buffer` can be types other than strings so we should be careful of that in `_file_not_found_msg()`.

pyrito (commented):
I'm a fan of this approach overall! Just a few comments.

modin/core/io/file_dispatcher.py
```diff
@@ -338,3 +338,20 @@ def build_partition(cls, partition_ids, row_lengths, column_widths):
                 for i in range(len(partition_ids))
             ]
         )
+
+    @classmethod
+    def _file_not_found_msg(cls, filename):
```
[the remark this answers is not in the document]
pyrito: Yessss thank you @jeffreykennethli  :D

modin/core/io/column_stores/parquet_dispatcher.py
noloerino: I'm not sure (I just copied this from a warning in previous code).

modin/core/io/text/csv_glob_dispatcher.py
noloerino: Any suggestions on how best to handle the non-string case? Would printing the `repr` or `str` be acceptable?

pyrito: I think working with the `str` representation should be good enough since that is usually supported by most path objects. However, I would check with other folks and see what they have to say.

cc: @modin-project/modin-contributors

.gitignore
pyrito: I think this is ok

modin/core/io/text/csv_glob_dispatcher.py
noloerino: It looks like in this particular branch, the filename will always be a string. In other branches, I misunderstood the behavior -- if `not cls.pathlib_or_pypath(filepath_or_buffer)` holds, then that means the argument should be a buffer; I've now updated the error message in this branch accordingly.

In any case, the f-string should automatically interpolate the `str` representation of the argument.

modin/core/io/pickle/pickle_dispatcher.py
noloerino: Spoke with Rehan today; it seems this case occurs when a buffer or a file path with a wildcard (*) in it is passed in. I've updated the message accordingly.

modin/core/io/column_stores/hdf_dispatcher.py
```diff
@@ -69,11 +68,12 @@ def _read(cls, path_or_buf, **kwargs):
...
         """
         if cls._validate_hdf_format(path_or_buf=path_or_buf) is None:
-            ErrorMessage.default_to_pandas(
+            return cls.single_worker_read(
+                path_or_buf,
                 "File format seems to be `fixed`. For better distribution consider "
-                + "saving the file in `table` format. df.to_hdf(format=`table`)."
+                + "saving the file in `table` format. df.to_hdf(format=`table`).",
+                **kwargs
             )
```
vnlitvinov, line 76: I would rather you pass the message as named argument (here and everywhere else), i.e.
```pythonё
return cls.single_worker_read(path_or_buf, reason="File format ......", **kwargs)
```

I think this improved readability

modin/core/io/file_dispatcher.py
```diff
@@ -338,3 +338,25 @@ def build_partition(cls, partition_ids, row_lengths, column_widths):
...
+        ----------
+        filename : str
+            The name of the file that was not found.
+
+        Returns
+        -------
+        str
+            A message indicating `filename` was not found.
+        """
+        return f"No file with name '{filename}'"
```
vnlitvinov, line 357: IMHO it's easier for the reader if we omit this insanely large docstring for a one-line formatter, and just silence the doc_checker with correct `# noqa`-s

modin/core/io/file_dispatcher.py
```diff
@@ -338,3 +338,25 @@ def build_partition(cls, partition_ids, row_lengths, column_widths):
...
+
+    @classmethod
+    def _buffer_unsupported_msg(cls):
+        """Produce a message indicating that buffers are not supported."""
+        return "Reading from buffers or other non-path-like objects is not supported"
```
vnlitvinov, line 362: this should just be a constant attribute, no point of making it a method

vnlitvinov (commented):
Overall good direction, left a few comments

pyrito (dismissed):
LGTM! Small minor comments

modin/core/io/pickle/pickle_dispatcher.py
```diff
@@ -47,10 +47,10 @@ def _read(cls, filepath_or_buffer, **kwargs):
         The number of partitions is equal to the number of input files.
         """
         if not (isinstance(filepath_or_buffer, str) and "*" in filepath_or_buffer):
-            warnings.warn("Defaulting to Modin core implementation")
             return cls.single_worker_read(
                 filepath_or_buffer,
                 single_worker_read=True,
+                reason="Buffers and wildcard file names are not supported",
```
RehanSD, line 53: This is checking if a wildcard is in the buffer's path right? I think we should change the error message to something like
```suggestion
                reason="Buffers with wildcard file paths are not supported",
```

modin/core/io/sql/sql_dispatcher.py
```diff
@@ -56,18 +55,16 @@ def _read(cls, sql, con, index_col=None, **kwargs):
...
-            )
             return cls.single_worker_read(
                 sql,
                 con=con,
                 index_col=index_col,
                 read_sql_engine=ReadSqlEngine.get(),
+                reason="To use parallel implementation of `read_sql`, pass either "
+                + "the SQL connection string or a ModinDatabaseConnection "
+                + "with the arguments required to make a connection, instead "
+                + f"of {type(con)}. For documentation of ModinDatabaseConnection, see "
```
RehanSD, line 66:
```suggestion
                + f"of {type(con)}. For documentation on the ModinDatabaseConnection, see "
```

modin/core/io/sql/sql_dispatcher.py
```diff
@@ -56,18 +55,16 @@ def _read(cls, sql, con, index_col=None, **kwargs):
...
-                + "with the arguments required to make a connection, instead "
-                + f"of {type(con)}. For documentation of ModinDatabaseConnection, see "
-                + "https://modin.readthedocs.io/en/latest/supported_apis/io_supported.html#connecting-to-a-database-for-read-sql"
-            )
             return cls.single_worker_read(
                 sql,
                 con=con,
                 index_col=index_col,
                 read_sql_engine=ReadSqlEngine.get(),
+                reason="To use parallel implementation of `read_sql`, pass either "
```
RehanSD, line 63:
```suggestion
                reason="To use the parallel implementation of `read_sql`, pass either "
```

modin/core/io/file_dispatcher.py
```diff
@@ -338,3 +342,7 @@ def build_partition(cls, partition_ids, row_lengths, column_widths):
                 for i in range(len(partition_ids))
             ]
         )
+
+    @classmethod
+    def _file_not_found_msg(cls, filename: str):  # noqa: GL08
+        return f"No file with name '{filename}'"
```
RehanSD, line 348:
```suggestion
        return f"No such file: '{filename}'"
```
This way it looks more like the file not found error in python. What do people think?

modin/core/io/text/text_file_dispatcher.py
```diff
@@ -635,18 +634,19 @@ def check_parameters_support(
...
+            `None` if parameters are supported, otherwise an error
+            message describing why parameters are not supported.
         """
         skiprows = read_kwargs.get("skiprows")
         if isinstance(filepath_or_buffer, str):
-            if not cls.file_exists(
-                filepath_or_buffer, read_kwargs.get("storage_options")
-            ):
-                return False
+            if not cls.file_exists(filepath_or_buffer):
```
RehanSD, line 643: This if statement has been changed.
```suggestion
            if not cls.file_exists(filepath_or_buffer, read_kwargs.get("storage_options"):
```

modin/core/storage_formats/cudf/parser.py
```diff
@@ -70,8 +70,8 @@ def get_dtypes(cls, dtypes_ids):
         )
 
     @classmethod
-    def single_worker_read(cls, fname, **kwargs):
-        ErrorMessage.default_to_pandas("Parameters provided")
+    def single_worker_read(cls, fname, *, reason, **kwargs):
```
RehanSD, line 73:
```suggestion
    def single_worker_read(cls, fname, reason=None, **kwargs):
```
Why is there an additional star here?

modin/core/storage_formats/pandas/parsers.py
```diff
@@ -249,14 +249,16 @@ def get_dtypes(cls, dtypes_ids):
         return frame_dtypes
 
     @classmethod
-    def single_worker_read(cls, fname, **kwargs):
+    def single_worker_read(cls, fname, *, reason: str, **kwargs):
```
RehanSD, line 252: Same here.

modin/pandas/test/test_io.py
```diff
@@ -1534,7 +1548,7 @@ def test_read_json_categories(self):
         reason="The reason of tests fail in `cloud` mode is unknown for now - issue #3264",
     )
     def test_read_json_string_bytes(self, data):
-        with warns_that_defaulting_to_pandas():
+        with warns_that_defaulting_to_pandas(prefix="read_json"):
```
RehanSD, line 1551: This change is not related to this PR, right? We're just ensuring that we get the right warning here, even though this PR doesn't change the warning that should be raised?

RehanSD (changes requested):
LGTM - left some minor comments/changes.

modin/core/io/pickle/pickle_dispatcher.py
noloerino: Actually, it seems like I misunderstood the condition for this branch: `not (isinstance(filepath_or_buffer, str) and "*" in filepath_or_buffer))` will hold if (1) the argument is not a string (i.e. is a buffer), or (2) the argument is a string and does NOT contain a wildcard. If my interpretation is correct, then this error message should be to the effect of ""Buffers or filenames without wildcards are not supported."

modin/pandas/test/test_io.py
noloerino: This actually should now be different, since it should be affected by changes in `modin/core/io/text/json_dispatcher.py`. That said, the current error message for PandasOnRay is
```
UserWarning: Defaulting to pandas implementation.
Reason: Reading from buffers or other non-path-like objects is not supported
```
whereas the message for PandasOnPython is
```
UserWarning: `read_json` is not currently supported by PandasOnPython, defaulting to pandas implementation.
```
I'm not sure why this test case didn't fail in CI for PandasOnRay, Ill have to investigate more closely. Should I have the PandasOnRay (and other engine-specific) messages also explicitly mention `read_json` in the error message?

noloerino: I've removed this extra argument for now, let me know what you think of the changes in the above comment.

modin/core/io/text/csv_glob_dispatcher.py
```diff
@@ -92,15 +100,27 @@ def _read(cls, filepath_or_buffer, **kwargs):
                 # need python3.7 to .seek and .tell ZipExtFile
                 kwargs["compression"] = compression_type
             else:
-                return cls.single_worker_read(filepath_or_buffer, **kwargs)
+                return cls.single_worker_read(
+                    filepath_or_buffer,
+                    reason=f"Unsupported compression type {compression_type}",
```
RehanSD, line 105: Could we add something to this string that specifies what compression types are supported, since we do have some version dependent logic here?

RehanSD (commented):
Looks good to me - one quick change and I think we can merge!

RehanSD (approved):
LGTM!

modin/pandas/test/test_io.py
```diff
@@ -995,12 +995,21 @@ def test_read_csv_skiprows_names(self, names, skiprows):
             skiprows=skiprows,
         )
 
+    def _has_pandas_fallback_reason(self):
+        # The Python engine does not use custom IO dispatchers, so specialized error messages
+        # won't appear
+        return Engine.get() != "Python" and StorageFormat.get() != "Omnisci"
```
YarShev, line 1001: OmnisciOnNativeIO can read a CSV file with pyarrow and also can be defaulting to pandas. Doesn't it seem to you that the condition should be corrected?

noloerino: I believe adding Omnisci to this condition was necessary because `OmnisciOnNativeIO` manually defines [`read_csv`](https://github.com/modin-project/modin/blob/master/modin/experimental/core/execution/native/implementations/omnisci_on_native/io/io.py#L161) instead of using the implementation from `TextFileDispatcher` (which matters because of [this line](modin/pandas/test/dataframe/test_binary.py::test_math_functions_level[add]), which produces error messages). This is different form the Ray backend, for example, where [this line](modin/pandas/test/dataframe/test_binary.py::test_math_functions_level[add]) leads to the TextFileDispatcher method being used.

It does seem like Omnisci does its own [error validation](https://github.com/modin-project/modin/blob/master/modin/experimental/core/execution/native/implementations/omnisci_on_native/io/io.py#L403), so perhaps this test should be more refined to check that error message on the Omnisci format, instead of the standard TextFileDispatcher one.

PR #4651, opened 2022-07-07
FIX-#4641: Reindex pandas partitions in `df.describe()`

### LINKED ISSUE #4641: df.describe(include='all') generates error
### System information
- OS: macOS 12.4
- Modin version:  0.15.2
- 3.9.12 (Anaconda)

In Jupyter notebook:

```
import os
import modin.pandas as pd
import numpy as np
import ray
ray.init()
df = pd.DataFrame(np.zeros((2,60)))
df['new'] = "a"
df.describe(include='all').T
```

### Describe the problem
Error dump instead of same output as pandas

### Source code / logs

Here's the error dump:

```
Exception                                 Traceback (most recent call last)
Input In [4], in <cell line: 9>()
      7 df = pd.DataFrame(np.zeros((2,60)))
      8 df['new'] = "a"
----> 9 df.describe(include='all').T

File ~/anaconda3/lib/python3.9/site-packages/modin/pandas/base.py:3878, in BasePandasDataset.__getattribute__(self, item)
   3865 def __getattribute__(self, item):
   3866     """
   3867     Return item from the `BasePandasDataset`.
   3868 
   (...)
   3876     Any
   3877     """
-> 3878     attr = super().__getattribute__(item)
   3879     if item not in _DEFAULT_BEHAVIOUR and not self._query_compiler.lazy_execution:
   3880         # We default to pandas on empty DataFrames. This avoids a large amount of
   3881         # pain in underlying implementation and returns a result immediately rather
   3882         # than dealing with the edge cases that empty DataFrames have.
   3883         if callable(attr) and self.empty and hasattr(self._pandas_class, item):

File ~/anaconda3/lib/python3.9/site-packages/modin/pandas/dataframe.py:525, in DataFrame.transpose(self, copy, *args)
    520 """
    521 Transpose index and columns.
    522 """
    523 # FIXME: Judging by pandas docs `*args` serves only compatibility purpose
    524 # and does not affect the result, we shouldn't pass it to the query compiler.
--> 525 return DataFrame(query_compiler=self._query_compiler.transpose(*args))

File ~/anaconda3/lib/python3.9/site-packages/modin/logging/logger_metaclass.py:68, in logger_class_wrapper.<locals>.log_wrap(*args, **kwargs)
     66     logger.info(f"END::PANDAS-API::{class_name}.{method_name}")
     67     return result
---> 68 return method(*args, **kwargs)

File ~/anaconda3/lib/python3.9/site-packages/modin/core/storage_formats/pandas/query_compiler.py:672, in PandasQueryCompiler.transpose(self, *args, **kwargs)
    670 def transpose(self, *args, **kwargs):
    671     # Switch the index and columns and transpose the data within the blocks.
--> 672     return self.__constructor__(self._modin_frame.transpose())

File ~/anaconda3/lib/python3.9/site-packages/modin/logging/logger_metaclass.py:68, in logger_class_wrapper.<locals>.log_wrap(*args, **kwargs)
     66     logger.info(f"END::PANDAS-API::{class_name}.{method_name}")
     67     return result
---> 68 return method(*args, **kwargs)

File ~/anaconda3/lib/python3.9/site-packages/modin/core/dataframe/pandas/dataframe/dataframe.py:115, in lazy_metadata_decorator.<locals>.decorator.<locals>.run_f_on_minimally_updated_metadata(self, *args, **kwargs)
    113     elif apply_axis == "rows":
    114         obj._propagate_index_objs(axis=0)
--> 115 result = f(self, *args, **kwargs)
    116 if apply_axis is None and not transpose:
    117     result._deferred_index = self._deferred_index

File ~/anaconda3/lib/python3.9/site-packages/modin/core/dataframe/pandas/dataframe/dataframe.py:2887, in PandasDataframe.transpose(self)
   2885 else:
   2886     new_dtypes = None
-> 2887 return self.__constructor__(
   2888     new_partitions,
   2889     self.columns,
   2890     self.index,
   2891     self._column_widths,
   2892     self._row_lengths,
   2893     dtypes=new_dtypes,
   2894 )

File ~/anaconda3/lib/python3.9/site-packages/modin/logging/logger_metaclass.py:68, in logger_class_wrapper.<locals>.log_wrap(*args, **kwargs)
     66     logger.info(f"END::PANDAS-API::{class_name}.{method_name}")
     67     return result
---> 68 return method(*args, **kwargs)

File ~/anaconda3/lib/python3.9/site-packages/modin/core/dataframe/pandas/dataframe/dataframe.py:210, in PandasDataframe.__init__(self, partitions, index, columns, row_lengths, column_widths, dtypes)
    208     num_columns = sum(column_widths)
    209     if num_columns > 0:
--> 210         ErrorMessage.catch_bugs_and_request_email(
    211             num_columns != len(self._columns_cache),
    212             "Column widths: {} != {}".format(
    213                 num_columns, len(self._columns_cache)
    214             ),
    215         )
    216     ErrorMessage.catch_bugs_and_request_email(
    217         any(val < 0 for val in column_widths),
    218         "Column widths cannot be negative: {}".format(column_widths),
    219     )
    220 self._column_widths_cache = column_widths

File ~/anaconda3/lib/python3.9/site-packages/modin/error_message.py:70, in ErrorMessage.catch_bugs_and_request_email(cls, failure_condition, extra_log)
     68 if failure_condition:
     69     get_logger().info(f"Modin Error: Internal Error: {extra_log}")
---> 70     raise Exception(
     71         "Internal Error. "
     72         + "Please visit https://github.com/modin-project/modin/issues "
     73         + "to file an issue with the traceback and the command that "
     74         + "caused this error. If you can't file a GitHub issue, "
     75         + f"please email bug_reports@modin.org.\n{extra_log}"
     76     )

Exception: Internal Error. Please visit https://github.com/modin-project/modin/issues to file an issue with the traceback and the command that caused this error. If you can't file a GitHub issue, please email bug_reports@modin.org.
Column widths: 8 != 11
```

pyrito:
## What do these changes do?
`df.describe(include='all')` can sometimes encounter cases where partitions have different indexes as a result of having different data types (e.g. categorial data returns different summaries as opposed to numerical data). Since we already know what the index of the final DataFrame should look like, we can set the correct index for the partition via `reindex`. Special thanks to @mvashishtha and @RehanSD for helping me understand what was going on in the lower-levels of Modin. This was a fun one to track down. 

- [x] Resolves #4641 
- [x] tests added and passing

### DIFF
--- a/modin/core/storage_formats/pandas/query_compiler.py
+++ b/modin/core/storage_formats/pandas/query_compiler.py
@@ -1577,6 +1577,13 @@ class PandasQueryCompiler(BaseQueryCompiler):
 
         def describe_builder(df, internal_indices=[]):
             """Apply `describe` function to the subset of columns in a single partition."""
+            # The index of the resulting dataframe is the same amongst all partitions
+            # when dealing with the same data type. However, if we work with columns
+            # that contain strings, we can get extra values in our result index such as
+            # 'unique', 'top', and 'freq'. Since we call describe() on each partition,
+            # we can have cases where certain partitions do not contain any of the
+            # object string data leading to an index mismatch between partitions.
+            # Thus, we must reindex each partition with the global new_index.
             return df.iloc[:, internal_indices].describe(**kwargs).reindex(new_index)
 
         return self.__constructor__(
--- a/modin/core/storage_formats/pandas/query_compiler.py
+++ b/modin/core/storage_formats/pandas/query_compiler.py
@@ -1577,7 +1577,7 @@ class PandasQueryCompiler(BaseQueryCompiler):
 
         def describe_builder(df, internal_indices=[]):
             """Apply `describe` function to the subset of columns in a single partition."""
-            return df.iloc[:, internal_indices].describe(**kwargs)
+            return df.iloc[:, internal_indices].describe(**kwargs).reindex(new_index)
 
         return self.__constructor__(
             self._modin_frame.apply_full_axis_select_indices(
--- a/modin/pandas/test/dataframe/test_reduce.py
+++ b/modin/pandas/test/dataframe/test_reduce.py
@@ -182,12 +182,11 @@ def test_2195(datetime_is_numeric, has_numeric_column):
 
 
 def test_describe_different_index():
-    df = pd.DataFrame(np.zeros((2, 60)))
-    pdf = pandas.DataFrame(np.zeros((2, 60)))
-    df["new"] = "abc"
-    pdf["new"] = "abc"
-    df_equals(df.describe(include="all"), pdf.describe(include="all"))
-    df_equals(df.describe(include="all").T, pdf.describe(include="all").T)
+    modin_df = pd.DataFrame(np.zeros((2, 60)))
+    pandas_df = pandas.DataFrame(np.zeros((2, 60)))
+    modin_df["new"] = pandas_df["new"] = "abc"
+    df_equals(modin_df.describe(include="all"), pandas_df.describe(include="all"))
+    df_equals(modin_df.describe(include="all").T, pandas_df.describe(include="all").T)
 
 
 @pytest.mark.parametrize(
--- a/modin/pandas/test/dataframe/test_reduce.py
+++ b/modin/pandas/test/dataframe/test_reduce.py
@@ -181,12 +181,19 @@ def test_2195(datetime_is_numeric, has_numeric_column):
     )
 
 
-def test_describe_different_index():
-    modin_df = pd.DataFrame(np.zeros((2, 60)))
-    pandas_df = pandas.DataFrame(np.zeros((2, 60)))
-    modin_df["new"] = pandas_df["new"] = "abc"
-    df_equals(modin_df.describe(include="all"), pandas_df.describe(include="all"))
-    df_equals(modin_df.describe(include="all").T, pandas_df.describe(include="all").T)
+# Issue: https://github.com/modin-project/modin/issues/4641
+def test_describe_column_partition_has_different_index():
+    pandas_df = pandas.DataFrame(test_data["int_data"])
+    # The index of the resulting dataframe is the same amongst all partitions
+    # when dealing with only numerical data. However, if we work with columns
+    # that contain strings, we will get extra values in our result index such as
+    # 'unique', 'top', and 'freq'. Since we call describe() on each partition,
+    # we can have cases where certain partitions do not contain any of the
+    # object string data. Thus, we add an extra string column to make sure
+    # that we are setting the index correctly for all partitions.
+    pandas_df["string_column"] = "abc"
+    modin_df = pd.DataFrame(pandas_df)
+    eval_general(modin_df, pandas_df, lambda df: df.describe(include="all"))
 
 
 @pytest.mark.parametrize(
--- a/modin/pandas/test/dataframe/test_reduce.py
+++ b/modin/pandas/test/dataframe/test_reduce.py
@@ -181,6 +181,15 @@ def test_2195(datetime_is_numeric, has_numeric_column):
     )
 
 
+def test_describe_different_index():
+    df = pd.DataFrame(np.zeros((2, 60)))
+    pdf = pandas.DataFrame(np.zeros((2, 60)))
+    df["new"] = "abc"
+    pdf["new"] = "abc"
+    df_equals(df.describe(include="all"), pdf.describe(include="all"))
+    df_equals(df.describe(include="all").T, pdf.describe(include="all").T)
+
+
 @pytest.mark.parametrize(
     "exclude,include",
     [
--- a/modin/pandas/test/dataframe/test_reduce.py
+++ b/modin/pandas/test/dataframe/test_reduce.py
@@ -184,13 +184,8 @@ def test_2195(datetime_is_numeric, has_numeric_column):
 # Issue: https://github.com/modin-project/modin/issues/4641
 def test_describe_column_partition_has_different_index():
     pandas_df = pandas.DataFrame(test_data["int_data"])
-    # The index of the resulting dataframe is the same amongst all partitions
-    # when dealing with only numerical data. However, if we work with columns
-    # that contain strings, we will get extra values in our result index such as
-    # 'unique', 'top', and 'freq'. Since we call describe() on each partition,
-    # we can have cases where certain partitions do not contain any of the
-    # object string data. Thus, we add an extra string column to make sure
-    # that we are setting the index correctly for all partitions.
+    # We add a string column to test the case where partitions with mixed data
+    # types have different 'describe' rows, which causes an index mismatch.
     pandas_df["string_column"] = "abc"
     modin_df = pd.DataFrame(pandas_df)
     eval_general(modin_df, pandas_df, lambda df: df.describe(include="all"))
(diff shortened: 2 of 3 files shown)


modin/core/storage_formats/pandas/query_compiler.py
```diff
@@ -1577,7 +1577,7 @@ def describe(self, **kwargs):
 
         def describe_builder(df, internal_indices=[]):
             """Apply `describe` function to the subset of columns in a single partition."""
-            return df.iloc[:, internal_indices].describe(**kwargs)
+            return df.iloc[:, internal_indices].describe(**kwargs).reindex(new_index)
```
pyrito, line 1580: There may be some performance implications here but I presume that it should be ok given that we're working at partition granularity?

modin/pandas/test/dataframe/test_reduce.py
```diff
@@ -181,6 +181,15 @@ def test_2195(datetime_is_numeric, has_numeric_column):
     )
 
 
+def test_describe_different_index():
+    df = pd.DataFrame(np.zeros((2, 60)))
```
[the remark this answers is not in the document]
pyrito: Oops! Yes I can make that change.

modin/pandas/test/dataframe/test_reduce.py
```diff
@@ -181,6 +181,14 @@ def test_2195(datetime_is_numeric, has_numeric_column):
     )
 
 
+def test_describe_different_index():
+    modin_df = pd.DataFrame(np.zeros((2, 60)))
+    pandas_df = pandas.DataFrame(np.zeros((2, 60)))
+    modin_df["new"] = pandas_df["new"] = "abc"
```
mvashishtha, line 187:
```suggestion
    pandas_df = pandas.DataFrame(test_data["int_data"])
    pandas_df["string_column"] = "abc"
    modin_df = pd.DataFrame(pandas_df)
```

`test_data['int_data']` has int data spanning multiple column partitions. Still, could you also please double check that the new test case reproduces the bug when you revert your fix?

mvashishtha, line 187: Could you please add a comment explaining why we're adding the extra column?

modin/pandas/test/dataframe/test_reduce.py
```diff
@@ -181,6 +181,14 @@ def test_2195(datetime_is_numeric, has_numeric_column):
...
+    df_equals(modin_df.describe(include="all"), pandas_df.describe(include="all"))
```
mvashishtha, line 188:
```suggestion
    eval_general(modin_df, pandas_df, lambda df: df.describe(include='all'))
```

modin/pandas/test/dataframe/test_reduce.py
```diff
@@ -181,6 +181,14 @@ def test_2195(datetime_is_numeric, has_numeric_column):
...
+    df_equals(modin_df.describe(include="all").T, pandas_df.describe(include="all").T)
```
mvashishtha, line 189:
```suggestion
```

I know the original bug report had the transpose, but I don't think checking the transpose gives us any more coverage.

mvashishtha (commented):
@pyrito Thank you, the fix looks good! I left some style comments.

modin/core/storage_formats/pandas/query_compiler.py
pyrito: Thanks for running the quick benchmark as a sanity. Yeah, we need the extra rows either way, and I don't know of an easier way of doing that then with `reindex` either. I think the time of the reindex will also get larger pretty quickly with the number of additional rows we add. Hopefully it won't be terribly large in the case of `describe`.

modin/pandas/test/dataframe/test_reduce.py
pyrito: Yup, just checked that it failed without the fix. Printed stuff out and cross verified as well.

modin/pandas/test/dataframe/test_reduce.py
```diff
@@ -181,6 +181,21 @@ def test_2195(datetime_is_numeric, has_numeric_column):
...
 
+# Issue: https://github.com/modin-project/modin/issues/4641
+def test_describe_column_partition_has_different_index():
+    pandas_df = pandas.DataFrame(test_data["int_data"])
+    # The index of the resulting dataframe is the same amongst all partitions
+    # when dealing with only numerical data. However, if we work with columns
+    # that contain strings, we will get extra values in our result index such as
+    # 'unique', 'top', and 'freq'. Since we call describe() on each partition,
+    # we can have cases where certain partitions do not contain any of the
+    # categorical string data. Thus, we add an extra string column to make sure
```
mvashishtha, line 192: The type of the new string column is object, or `np.dtype('object_')`.

mvashishtha (approved):
LGTM!

modin/pandas/test/dataframe/test_reduce.py
```diff
@@ -181,6 +181,21 @@ def test_2195(datetime_is_numeric, has_numeric_column):
...
+# Issue: https://github.com/modin-project/modin/issues/4641
+def test_describe_column_partition_has_different_index():
+    pandas_df = pandas.DataFrame(test_data["int_data"])
+    # The index of the resulting dataframe is the same amongst all partitions
+    # when dealing with only numerical data. However, if we work with columns
+    # that contain strings, we will get extra values in our result index such as
+    # 'unique', 'top', and 'freq'. Since we call describe() on each partition,
+    # we can have cases where certain partitions do not contain any of the
+    # object string data. Thus, we add an extra string column to make sure
+    # that we are setting the index correctly for all partitions.
```
vnlitvinov, line 193: shouldn't this comment be in the implementation (not in a test)?..

mvashishtha: Maybe it's best for most of the comment to go in the implementation, and to keep a short note here saying that we're testing a case where different partitions have different `describe` rows.

pyrito: Yeah makes sense.

vnlitvinov (approved):
LGTM, thanks!

modin/core/storage_formats/pandas/query_compiler.py
RehanSD: We could potentially mitigate the performance hit by checking if the index of df.describe equals the desired index and only reinforcing if it doesn't?

RehanSD (commented):
Left a quick question, besides that, looks good to me!

modin/core/storage_formats/pandas/query_compiler.py
pyrito: @RehanSD is there a performance penalty for calling reindex if the index is already the same?

mvashishtha: @RehanSD I thought about that, but in practice it looks like the time to `describe` outweights the time to `reindex` with an equal index by a factor of over 1000:

```python
import pandas as pd
import numpy as np
np.random.seed(0)
df = pd.DataFrame(np.random.randint(0,100,(1,2 ** 12)))
%time df2 = df.describe()
 %time df2 = df2.reindex(df2.index)
```

I get 4.56 seconds for the describe and 282 microseconds for the reindex, so I don't think the extra optimization is worth the extra complexity.

pyrito: The docs say this: "A new object is produced unless the new index is equivalent to the current one and copy=False". In this case, copy is not False, so maybe the right thing to do here is to make copy=False?

pyrito: Yeah I just tried experimenting a bit with this case and came to the same conclusion.

PR #4656, opened 2022-07-07
FIX-#4491: Wait for all partitions in parallel in benchmark mode

### LINKED ISSUE #4491: BUG: benchmark mode should wait for all partitions in parallel

noloerino:
## What do these changes do?
When running in benchmark mode, results from partitions are waited for with a single call to `ray.get` rather than iterating over each partition and calling `ray.wait` under the hood. As the original issue (#4491) suggests, the performance improvement seems to be fairly negligible.

- [x] Resolves #4491 
- [ ] tests added and passing

### DIFF
--- a/modin/core/dataframe/pandas/partitioning/partition_manager.py
+++ b/modin/core/dataframe/pandas/partitioning/partition_manager.py
@@ -53,7 +53,7 @@ def wait_computations_if_benchmark_mode(func):
     if BenchmarkMode.get():
 
         @wraps(func)
-        def wait(*args, **kwargs):
+        def wait(cls, *args, **kwargs):
             """Wait for computation results."""
             result = func(*args, **kwargs)
             if isinstance(result, tuple):
@@ -69,8 +69,7 @@ def wait_computations_if_benchmark_mode(func):
             [part.drain_call_queue() for part in partitions.flatten()]
             # The partition manager invokes the relevant .wait() method under
             # the hood, which should wait in parallel for all computations to finish
-            partition_mgr_cls = args[0]
-            partition_mgr_cls.wait_partitions(partitions.flatten())
+            cls.wait_partitions(partitions.flatten())
             return result
 
         return wait
@@ -811,7 +810,7 @@ class PandasDataframePartitionManager(ABC):
     @classmethod
     def wait_partitions(cls, partitions):
         """
-        Wait on the objects wrapped by `partitions` in parallel, without materializing them.
+        Wait on the objects wrapped by `partitions`, without materializing them.
 
         This method will block until all computations in the list have completed.
 
--- a/modin/core/dataframe/pandas/partitioning/partition_manager.py
+++ b/modin/core/dataframe/pandas/partitioning/partition_manager.py
@@ -55,11 +55,6 @@ def wait_computations_if_benchmark_mode(func):
         @wraps(func)
         def wait(*args, **kwargs):
             """Wait for computation results."""
-            # Importing at the start of the file would cause a circular import
-            from modin.core.execution.ray.implementations.pandas_on_ray.partitioning import (
-                PandasOnRayDataframePartitionManager,
-            )
-
             result = func(*args, **kwargs)
             if isinstance(result, tuple):
                 partitions = result[0]
@@ -72,11 +67,9 @@ def wait_computations_if_benchmark_mode(func):
             # serially kick off all the deferred computations so that they can
             # all run asynchronously, then wait on all the results.
             [part.drain_call_queue() for part in partitions.flatten()]
-            # The PandasOnRay partition manager invokes ray.get under the hood, which waits
-            # in parallel for all computations to finish
-            PandasOnRayDataframePartitionManager.get_objects_from_partitions(
-                partitions.flatten()
-            )
+            # The partition manager invokes the relevant .wait() method under
+            # the hood, which should wait in parallel for all computations to finish
+            PandasDataframePartitionManager.wait_partitions(partitions.flatten())
             return result
 
         return wait
@@ -813,6 +806,26 @@ class PandasDataframePartitionManager(ABC):
         getting objects in parallel.
         """
         return [partition.get() for partition in partitions]
+
+    @classmethod
+    def wait_partitions(cls, partitions):
+        """
+        Wait on the objects wrapped by `partitions` in parallel, without materializing them.
+
+        This method will block until all computations in the list have completed.
+
+        Parameters
+        ----------
+        partitions : np.ndarray
+            NumPy array with ``PandasDataframePartition``-s.
+
+        Notes
+        -----
+        This method should be implemented in a more efficient way for engines that supports
+        waiting on objects in parallel.
+        """
+        for partition in partitions:
+            partition.wait()
 
     @classmethod
     def get_indices(cls, axis, partitions, index_func=None):
--- a/modin/core/dataframe/pandas/partitioning/partition_manager.py
+++ b/modin/core/dataframe/pandas/partitioning/partition_manager.py
@@ -53,7 +53,7 @@ def wait_computations_if_benchmark_mode(func):
     if BenchmarkMode.get():
 
         @wraps(func)
-        def wait(cls, *args, **kwargs):
+        def wait(*args, **kwargs):
             """Wait for computation results."""
             result = func(*args, **kwargs)
             if isinstance(result, tuple):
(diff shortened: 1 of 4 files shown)


modin/core/dataframe/pandas/partitioning/partition_manager.py
```diff
@@ -67,12 +72,11 @@ def wait(*args, **kwargs):
...
             [part.drain_call_queue() for part in partitions.flatten()]
-            # need to go through all the values of the map iterator
-            # since `wait` does not return anything, we need to explicitly add
-            # the return `True` value from the lambda
-            # TODO(https://github.com/modin-project/modin/issues/4491): Wait
-            # for all the partitions in parallel.
-            all(map(lambda partition: partition.wait() or True, partitions.flatten()))
+            # The PandasOnRay partition manager invokes ray.get under the hood, which waits
+            # in parallel for all computations to finish
+            PandasOnRayDataframePartitionManager.get_objects_from_partitions(
```
mvashishtha, line 77: `get_objects_from_partitions` will materialize all the partition data in the main process. We should avoid that as much as possible. It's not uncommon on a distributed cluster for the worker node to not have enough memory to hold all the data. Even when running ray on one machine we will have to pay for deserializing and copying all the data from the object store. For ray we want something like `ray.wait([partition._data for partition in partitions], num_returns=len(partitions))`.

mvashishtha, line 77: Note that we also need an implementation for the dask partition manager.

mvashishtha (commented):
@noloerino thank you for taking this! I left some comments.

modin/core/dataframe/pandas/partitioning/partition_manager.py
noloerino: Ah, I didn't realize that was the difference between `get` and `wait`. Is it correct to say that both methods will wait for all computations to finish, but `wait` doesn't copy the data from the object store?

Also, should we add a separate method to the partition manager to avoid having to access the `partition._data` field from here?

mvashishtha: > Is it correct to say that both methods will wait for all computations to finish, but wait doesn't copy the data from the object store?

That's the main difference.

> Also, should we add a separate method to the partition manager to avoid having to access the partition._data field from here?

I think so. That would also match `get_objects_from_partitions`, which will have a very similar purpose.

noloerino: Is it just Ray and Dask, or does this code need to be more general?

mvashishtha: Every class that subclasses `PandasDataframePartitionManager` needs an implementation. As far as I can tell, there's `PandasOnDaskDataframePartitionManager`,`PandasOnPythonDataframePartitionManager`, `GenericRayDataframePartitionManager` and 3 of its subclasses (but maybe you can implement this in `GenericRayDataframePartitionManager`) and `OmnisciOnNativeDataframePartitionManager`. For Python and Omnisci and Dask you can get an idea of the implementation by looking at how the `_partition_class` implements `wait`.

modin/core/dataframe/pandas/partitioning/partition_manager.py
```diff
@@ -67,12 +67,9 @@ def wait(*args, **kwargs):
...
             [part.drain_call_queue() for part in partitions.flatten()]
-            # need to go through all the values of the map iterator
-            # since `wait` does not return anything, we need to explicitly add
-            # the return `True` value from the lambda
-            # TODO(https://github.com/modin-project/modin/issues/4491): Wait
-            # for all the partitions in parallel.
-            all(map(lambda partition: partition.wait() or True, partitions.flatten()))
+            # The partition manager invokes the relevant .wait() method under
+            # the hood, which should wait in parallel for all computations to finish
+            PandasDataframePartitionManager.wait_partitions(partitions.flatten())
```
vnlitvinov, line 72: this calls `.wait_partitions()` on explicitly `PandasDataframePartitionManager`, not the manager for giving partitions (e.g. `PandasOnDaskDataframePartitionManager`), this seems to be wrong

vnlitvinov (changes requested):

modin/core/dataframe/pandas/partitioning/partition_manager.py
YarShev: We should probably extract the partition manager out of `*args`.

```python
partition_mgr_cls = args[0]
```

RehanSD (dismissed):
LGTM!

modin/core/dataframe/pandas/partitioning/partition_manager.py
```diff
@@ -67,12 +67,10 @@ def wait(*args, **kwargs):
...
             [part.drain_call_queue() for part in partitions.flatten()]
-            # need to go through all the values of the map iterator
-            # since `wait` does not return anything, we need to explicitly add
-            # the return `True` value from the lambda
-            # TODO(https://github.com/modin-project/modin/issues/4491): Wait
-            # for all the partitions in parallel.
-            all(map(lambda partition: partition.wait() or True, partitions.flatten()))
+            # The partition manager invokes the relevant .wait() method under
+            # the hood, which should wait in parallel for all computations to finish
+            partition_mgr_cls = args[0]
```
vnlitvinov, line 72: if this is always the case, we should probably change the signature of `wait()` to declare it a classmethod, i.e.
```python
@wraps(func)
def wait(cls, *args, **kwargs):
    result = func(cls, *args, **kwargs)
    # do the rest of the thing
    partitions = partitions.flatten()
    [part.drain_call_queue() for part in partitions]
    cls.wait_partitions(partitions)
```

I also wonder if it is the `.wait_partitions()` method which should actually be draining the call queues...

modin/core/dataframe/pandas/partitioning/partition_manager.py
```diff
@@ -806,6 +804,26 @@ def get_objects_from_partitions(cls, partitions):
         """
         return [partition.get() for partition in partitions]
 
+    @classmethod
+    def wait_partitions(cls, partitions):
+        """
+        Wait on the objects wrapped by `partitions` in parallel, without materializing them.
```
vnlitvinov, line 810:
```suggestion
        Wait on the objects wrapped by `partitions`, without materializing them.
```
as this specific implementation is _not_ waiting in parallel

modin/core/execution/ray/implementations/pandas_on_ray/partitioning/partition_manager.py
```diff
@@ -111,6 +111,22 @@ def get_objects_from_partitions(cls, partitions):
...
+
+        This method will block until all computations in the list have completed.
+
+        Parameters
+        ----------
+        partitions : np.ndarray
+            NumPy array with ``PandasDataframePartition``-s.
+        """
+        ray.wait(
+            [partition._data for partition in partitions], num_returns=len(partitions)
```
vnlitvinov, line 127: is there a reason to specify `num_returns` here?.. I think we can safely drop it

RehanSD: If you don't, I believe ray will not wait for all objectrefs to be ready.

prutskov: Rehan is right, `num_returns` means the number of elements which should finish the execution and put result to object storage (https://docs.ray.io/en/latest/ray-core/package-ref.html#ray-wait)

noloerino: and the default value for the argument is 1, so we do need to specify it

modin/core/dataframe/pandas/partitioning/partition_manager.py
noloerino: On further examination, it seems like this has to be `args[0]` rather than adding a `cls` argument to the decorator function (tests crash otherwise).

vnlitvinov: got a link to tests' failure? looks very suspicious

noloerino: [Here](https://github.com/modin-project/modin/runs/7307878853?check_suite_focus=true), I was able to reproduce it locally as well. I can't find the portion in the official Python docs describing why `*args` and `**kwargs` are the only allowable options, but all the examples I've seen through searching seem to only use those.

vnlitvinov: Can I see the link to the commit? I mean, maybe you forgot to pass this `cls` as the argument? (like I did in my suggestion)

noloerino: https://github.com/modin-project/modin/pull/4656/commits/4f91ce8bf9557efd2384fec30059883537c7533f

It seems you're right, I missed `func(cls, *args, **kwargs)`. I've put the explicit `cls` argument back.

vnlitvinov (approved):
LGTM

mvashishtha (approved):
LGTM! Thank you @noloerino

PR #4671, opened 2022-07-12
FEAT-#4670: Implement convert_dtypes by mapping across partitions

### LINKED ISSUE #4670: FEAT: Implement `convert_dtypes` by mapping across partitions instead of defaulting to pandas
As in title, the [`convert_dtypes`](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.convert_dtypes.html) currently defaults to pandas, but can be implemented more efficiently by mapping across column partitions.

noloerino:
## What do these changes do?

Implements `convert_dtypes` by mapping across partitions instead of defaulting to pandas.

- [x] Resolves #4670 
- [x] tests added and passing

### DIFF
--- a/docs/release_notes/release_notes-0.16.0.rst
+++ b/docs/release_notes/release_notes-0.16.0.rst
@@ -30,7 +30,8 @@ Key Features and Updates
   * REFACTOR-#4530: Standardize access to physical data in partitions (#4563)
   * REFACTOR-#4534: Replace logging meta class with class decorator (#4535)
 * Pandas API implementations and improvements
-  *
+  * FEAT-#4670: Implement convert_dtypes by mapping across partitions (#4671)
+
 * OmniSci enhancements
   *
 * XGBoost enhancements
@@ -52,7 +53,6 @@ Key Features and Updates
 * New Features
   * FEAT-4463: Add experimental fuzzydata integration for testing against a randomized dataframe workflow (#4556)
   * FEAT-#4419: Extend virtual partitioning API to pandas on Dask (#4420)
-  * FEAT-#4670:  Implement convert_dtypes by mapping across partitions (#4671)
 
 
 Contributors
--- a/docs/release_notes/release_notes-0.16.0.rst
+++ b/docs/release_notes/release_notes-0.16.0.rst
@@ -52,6 +52,7 @@ Key Features and Updates
 * New Features
   * FEAT-4463: Add experimental fuzzydata integration for testing against a randomized dataframe workflow (#4556)
   * FEAT-#4419: Extend virtual partitioning API to pandas on Dask (#4420)
+  * FEAT-#4670:  Implement convert_dtypes by mapping across partitions (#4671)
 
 
 Contributors
@@ -66,3 +67,4 @@ Contributors
 @helmeleegy
 @anmyachev
 @d33bs
+@noloerino
--- a/docs/release_notes/release_notes-0.16.0.rst
+++ b/docs/release_notes/release_notes-0.16.0.rst
@@ -31,7 +31,6 @@ Key Features and Updates
   * REFACTOR-#4534: Replace logging meta class with class decorator (#4535)
 * Pandas API implementations and improvements
   * FEAT-#4670: Implement convert_dtypes by mapping across partitions (#4671)
-
 * OmniSci enhancements
   *
 * XGBoost enhancements
--- a/modin/core/storage_formats/base/query_compiler.py
+++ b/modin/core/storage_formats/base/query_compiler.py
@@ -1397,15 +1397,38 @@ class BaseQueryCompiler(ClassLogger, abc.ABC):
     def convert_dtypes(
         self,
         infer_objects: bool = True,
-        convert_objects: bool = True,
+        convert_string: bool = True,
         convert_integer: bool = True,
         convert_boolean: bool = True,
         convert_floating: bool = True,
     ):
+        """
+        Convert columns to best possible dtypes using dtypes supporting ``pd.NA``.
+
+        Parameters
+        ----------
+        infer_objects : bool, default: True
+            Whether object dtypes should be converted to the best possible types.
+        convert_string : bool, default: True
+            Whether object dtypes should be converted to ``pd.StringDtype()``.
+        convert_integer : bool, default: True
+            Whether, if possbile, conversion should be done to integer extension types.
+        convert_boolean : bool, default: True
+            Whether  object dtypes should be converted to ``pd.BooleanDtype()``.
+        convert_floating : bool, default: True
+            Whether, if possible, conversion can be done to floating extension types.
+            If `convert_integer` is also True, preference will be give to integer dtypes
+            if the floats can be faithfully casted to integers.
+
+        Returns
+        -------
+        BaseQueryCompiler
+            New QueryCompiler with updated dtypes.
+        """
         return DataFrameDefault.register(pandas.DataFrame.convert_dtypes)(
             self,
             infer_objects=infer_objects,
-            convert_objects=convert_objects,
+            convert_string=convert_string,
             convert_integer=convert_integer,
             convert_boolean=convert_boolean,
             convert_floating=convert_floating,
--- a/modin/core/storage_formats/base/query_compiler.py
+++ b/modin/core/storage_formats/base/query_compiler.py
@@ -1394,6 +1394,23 @@ class BaseQueryCompiler(ClassLogger, abc.ABC):
             self, dtype=col_dtypes, **kwargs
         )
 
+    def convert_dtypes(
+        self,
+        infer_objects: bool = True,
+        convert_objects: bool = True,
+        convert_integer: bool = True,
+        convert_boolean: bool = True,
+        convert_floating: bool = True,
+    ):
+        return DataFrameDefault.register(pandas.DataFrame.convert_dtypes)(
+            self,
+            infer_objects=infer_objects,
+            convert_objects=convert_objects,
+            convert_integer=convert_integer,
+            convert_boolean=convert_boolean,
+            convert_floating=convert_floating,
+        )
+
     @property
     def dtypes(self):
         """
--- a/modin/core/storage_formats/base/query_compiler.py
+++ b/modin/core/storage_formats/base/query_compiler.py
@@ -1414,7 +1414,7 @@ class BaseQueryCompiler(ClassLogger, abc.ABC):
         convert_integer : bool, default: True
             Whether, if possbile, conversion should be done to integer extension types.
         convert_boolean : bool, default: True
-            Whether  object dtypes should be converted to ``pd.BooleanDtype()``.
+            Whether object dtypes should be converted to ``pd.BooleanDtype()``.
         convert_floating : bool, default: True
             Whether, if possible, conversion can be done to floating extension types.
             If `convert_integer` is also True, preference will be give to integer dtypes
--- a/modin/core/storage_formats/pandas/query_compiler.py
+++ b/modin/core/storage_formats/pandas/query_compiler.py
@@ -1347,6 +1347,7 @@ class PandasQueryCompiler(BaseQueryCompiler):
     abs = Map.register(pandas.DataFrame.abs, dtypes="copy")
     applymap = Map.register(pandas.DataFrame.applymap)
     conj = Map.register(lambda df, *args, **kwargs: pandas.DataFrame(np.conj(df)))
+    convert_dtypes = Map.register(pandas.DataFrame.convert_dtypes)
     invert = Map.register(pandas.DataFrame.__invert__)
     isin = Map.register(pandas.DataFrame.isin, dtypes=np.bool_)
     isna = Map.register(pandas.DataFrame.isna, dtypes=np.bool_)
--- a/modin/pandas/test/dataframe/test_map_metadata.py
+++ b/modin/pandas/test/dataframe/test_map_metadata.py
@@ -605,6 +605,7 @@ def test_convert_dtypes_multiple_row_partitions():
     # Column 0 should have an int dtype
     modin_part2 = pd.DataFrame([1]).convert_dtypes()
     modin_df = pd.concat([modin_part1, modin_part2])
+    assert modin_df._query_compiler._modin_frame._partitions.shape == (2, 1)
     pandas_df = pandas.DataFrame(["a", 1], index=[0, 0])
     # The initial dataframes should be the same
     df_equals(modin_df, pandas_df)
@@ -614,7 +615,6 @@ def test_convert_dtypes_multiple_row_partitions():
     modin_result = modin_df.convert_dtypes()
     pandas_result = pandas_df.convert_dtypes()
     df_equals(modin_result, pandas_result)
-    assert modin_df._query_compiler._modin_frame._partitions.shape == (2, 1)
     assert modin_result.dtypes.equals(pandas_result.dtypes)
 
 
--- a/modin/pandas/test/dataframe/test_map_metadata.py
+++ b/modin/pandas/test/dataframe/test_map_metadata.py
@@ -557,11 +557,11 @@ def test_astype_category_large():
     assert modin_result.dtypes.equals(pandas_result.dtypes)
 
 
-@pytest.mark.parametrize("infer_objects", (True, False))
-@pytest.mark.parametrize("convert_string", (True, False))
-@pytest.mark.parametrize("convert_integer", (True, False))
-@pytest.mark.parametrize("convert_boolean", (True, False))
-@pytest.mark.parametrize("convert_floating", (True, False))
+@pytest.mark.parametrize("infer_objects", (True, False), ids=("infer_objects", ""))
+@pytest.mark.parametrize("convert_string", (True, False), ids=("convert_string", ""))
+@pytest.mark.parametrize("convert_integer", (True, False), ids=("convert_integer", ""))
+@pytest.mark.parametrize("convert_boolean", (True, False), ids=("convert_boolean", ""))
+@pytest.mark.parametrize("convert_floating", (True, False), ids=("convert_floating", ""))
 def test_convert_dtypes(
     infer_objects, convert_string, convert_integer, convert_boolean, convert_floating
 ):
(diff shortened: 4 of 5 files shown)


mvashishtha:
@noloerino I think implementing this function by mapping all the Modin block partitions won't work, because we need access to all the data in a column at once to choose the right dtype for that column. For example suppose a dataframe has two row partitions, one of which has only strings in column 0, and the other of which has only ints:

```python
import pandas as pd

part1 = pd.DataFrame(['a'])
part2 = pd.DataFrame([1])
print(part1.convert_dtypes().dtypes)
print(part2.convert_dtypes().dtypes)
print(pd.concat([part1, part2]).convert_dtypes().dtypes)
```

We get `string` dtype for `part1` and `Int64` for `part2`. But the combined frame would have dtype 'object'.

I suggest adding a test case like this, where different row partitions have different types in a single column. Once that fails, you can reimplement the function as column-parallel. Once your row numbers go past `modin.config.MinPartitionSize`, you should get a new row partition.

noloerino:
Thanks for the suggestion. Are you saying that the code right now (where the operator is implemented by a naive `Map.register`) might be applying the function across block/row partitions rather than column partitions, so it could lead to the wrong result?

I'm trying this abridged version of your example to get multiple row partitions:
```
part_col_size = MinPartitionSize.get()
modin_part1 = pd.DataFrame(["a"] * part_col_size).convert_dtypes()  # Column 0 should have string dtype
modin_part2 = pd.DataFrame([1] * part_col_size).convert_dtypes()  # Column 0 should have an int dtype
modin_result = pd.concat([modin_part1, modin_part2]).convert_dtypes()
```
...but `modin_result.dtype` is still 'object', which we expect. Is there a way to check that the new `modin_result` dataframe is actually stored across multiple row partitions (like some property of `modin_result._query_compiler`)?

noloerino:
I looked into this some more, and by checking `modin_result._query_compiler._modin_frame._partitions`, I found that the frame did indeed have multiple row partitions. However, I don't think this is an issue, since the process for constructing a new dataframe from `convert_dtypes` looks something like this:
- Because the method was registered with `Map.register`, the `pandas.convert_dtypes` function is applied to individual partitions by [`PandasDataframe.map`](https://github.com/modin-project/modin/blob/5af9832d7fad3d17f05d63908bc377e61542d953/modin/core/dataframe/pandas/dataframe/dataframe.py#L1527).
- After the function is applied to each partition, the last line of `PandasDataFrame.map` calls the dataframe constructor with the new partitions.
- Because the `dtypes` argument passed to the constructor should always be `None`, the `_dtype` attribute of the dataframe is also set to `None` [here](https://github.com/modin-project/modin/blob/5af9832d7fad3d17f05d63908bc377e61542d953/modin/core/dataframe/pandas/dataframe/dataframe.py#L221).
- When needed, the dtype of the new frame is lazily computed by [`_compute_dtypes`](https://github.com/modin-project/modin/blob/5af9832d7fad3d17f05d63908bc377e61542d953/modin/core/dataframe/pandas/dataframe/dataframe.py#L286), which finds the common ancestor of all the types present in a particular column.

Let me know if my reasoning is correct. If necessary, I can also add the test case you described just to make sure the type computation behavior I described here gets changed in the future.

mvashishtha:
@noloerino 

> I don't think this is an issue, since the process for constructing a new dataframe from convert_dtypes looks something like this:

That's what I was thinking, too, but then I realized that we don't always use the overall frame's inferred `dtypes`. Operations like `iloc` can get results from partitions without using the overall frame's dtypes. Here's an example that fails on your branch but not on current modin:

```python
import modin.pandas as pd
df1 = pd.DataFrame([1])
df2 = pd.DataFrame(['a'])
df = pd.concat([df1, df2])
print(type(df.convert_dtypes().iloc[0,0]))
```

I get `<class 'numpy.int64'>` from your branch, but `<class 'int'>` from pandas.

That being said, the bugs with internal and external dtype synchronization go beyond just your branch, as I've explained in #4675. The snippet above happens to work in current Modin because Modin converts the entire frame to pandas, does `convert_dtypes`, then converts the frame back to Modin. In fact, current Modin (5af9832d7fad3d17f05d63908bc377e61542d953) is already messing up the dtypes if we don't include `convert_dtypes`:

```python
import modin.pandas as pd
df1 = pd.DataFrame([1])
df2 = pd.DataFrame(['a'])
df = pd.concat([df1, df2])
print(type(df.iloc[0,0]))
```

It gives `numpy.int64` instead of the `int` we get with pandas, because the internal partition doesn't know its data is of `object` type.

My conclusion is that your `Map` implementation is fine, and we can tackle the dtype synchronization problem in issues like #4675.

>  If necessary, I can also add the test case you described just to make sure the type computation behavior I described here gets changed in the future.

Yes, let's add this test case anyway. 

P.S. your script does give multiple partitions, but FYI you can't count on `concat` to preserve partitions, because we sometimes [rebalance partitions](https://github.com/modin-project/modin/blob/8f35ab57996d18f2c1f001d624682952853454cd/modin/core/execution/ray/implementations/pandas_on_ray/partitioning/partition_manager.py#L148) after concat.

modin/core/storage_formats/pandas/query_compiler.py
```diff
@@ -1347,6 +1347,7 @@ def stack(self, level, dropna):
     abs = Map.register(pandas.DataFrame.abs, dtypes="copy")
     applymap = Map.register(pandas.DataFrame.applymap)
     conj = Map.register(lambda df, *args, **kwargs: pandas.DataFrame(np.conj(df)))
+    convert_dtypes = Map.register(pandas.DataFrame.convert_dtypes)
```
RehanSD, line 1350: Convert_dtypes requires full column access in order to determine the most appropriate dtype right? Doesn't that mean we have to do a apply_axis_partitions call on our Modin frame rather than relying on the algebra map which is cell wise?

RehanSD (commented):
Left a quick q - looks good to me besides that

modin/core/storage_formats/pandas/query_compiler.py
RehanSD: Sorry - just saw your discussion with Mahesh - this is just a rehash of that, so feel free to ignore/resolve

modin/pandas/test/dataframe/test_map_metadata.py
```diff
@@ -557,6 +558,57 @@ def test_astype_category_large():
...
 
 
+@pytest.mark.parametrize("infer_objects", (True, False), ids=("infer_objects", ""))
+@pytest.mark.parametrize("convert_string", (True, False), ids=("convert_string", ""))
+@pytest.mark.parametrize("convert_integer", (True, False), ids=("convert_integer", ""))
+@pytest.mark.parametrize("convert_boolean", (True, False), ids=("convert_boolean", ""))
+@pytest.mark.parametrize(
+    "convert_floating", (True, False), ids=("convert_floating", "")
+)
+def test_convert_dtypes(
```
mvashishtha, line 568:
```suggestion
def test_convert_dtypes_single_partition(
```

modin/pandas/test/dataframe/test_map_metadata.py
```diff
@@ -557,6 +558,57 @@ def test_astype_category_large():
...
+    modin_df = pd.DataFrame(data)
+    pandas_df = pandas.DataFrame(data)
+    modin_result = modin_df.convert_dtypes(**kwargs)
+    pandas_result = pandas_df.convert_dtypes(**kwargs)
+    assert modin_result.dtypes.equals(pandas_result.dtypes)
+
+
+def test_convert_dtypes_row_partition():
+    # Tests `convert_dtypes` when values in the same column may come from different
+    # row partitions.
```
mvashishtha, line 597:
```suggestion
```

modin/pandas/test/dataframe/test_map_metadata.py
```diff
@@ -557,6 +558,57 @@ def test_astype_category_large():
...
+    part_col_size = MinPartitionSize.get()
```
mvashishtha, line 598: `part_col_size` isn't useful now that we're forcing multiple partitions with `concat`. We can initialize each dataframe with a single element.

modin/pandas/test/dataframe/test_map_metadata.py
```diff
@@ -557,6 +558,57 @@ def test_astype_category_large():
...
+    # Column 0 should have string dtype
+    modin_part1 = pd.DataFrame(["a"] * part_col_size).convert_dtypes()
+    # Column 0 should have an int dtype
+    modin_part2 = pd.DataFrame([1] * part_col_size).convert_dtypes()
+    modin_result = pd.concat([modin_part1, modin_part2]).convert_dtypes()
+    pandas_part1 = pandas.DataFrame(["a"] * part_col_size).convert_dtypes()
+    pandas_part2 = pandas.DataFrame([1] * part_col_size).convert_dtypes()
+    pandas_result = pandas.concat([pandas_part1, pandas_part2]).convert_dtypes()
```
mvashishtha, line 606:
```suggestion
    modin_df = pd.concat([modin_part1, modin_part2])
    pandas_df = pandas.DataFrame(["a", 1])
    # The initial dataframes should be the same
    df_equals(modin_df, pandas_df)
    # TODO(https://github.com/modin-project/modin/pull/3805): delete
    # this assert once df_equals checks dtypes    
    assert modin_df.dtypes.equals(pandas_df.dtypes)
    modin_result = modin_df.convert_dtypes()
    pandas_result = pandas_df.convert_dtypes()
    df_equals(modin_result, pandas_result)
    assert modin_result.dtypes.equals(pandas_result.dtypes)    
```

modin/pandas/test/dataframe/test_map_metadata.py
```diff
@@ -557,6 +558,57 @@ def test_astype_category_large():
...
+    # After exceeding MinPartitionSize, the dataframe should have 2 row partitions
+    # (though this is not guaranteed)
```
mvashishtha, line 608: Instead of this comment, let's assert after creating `modin_df` that `modin_df._query_compiler._modin_frame._partitions.shape == (2, 1)`

mvashishtha (commented):
@noloerino I left some style comments on the test.

modin/pandas/test/dataframe/test_map_metadata.py
noloerino: Since as you said above, having multiple partitions isn't guaranteed with `concat`, would it make more sense to directly construct a single dataframe with something like `pd.DataFrame((["a"] * part_col_size) + ([1] * part_col_size))`?

mvashishtha: We should never treat multiple partitions as guaranteed, though generally we assume in our unit tests that constructing a frame with more than `MinPartitionSize` elements on an axis forces multiple partitions on that axis. I think the way you have it is fine for now. Let's just add the assert I mentioned in my other comment.

modin/pandas/test/dataframe/test_map_metadata.py
noloerino: It seems that the pandas_df also has to be constructed with `pandas.concat`, or else the index types don't match: the one constructed by concat uses `Int64Index([0, 0], dtype='int64')`, whereas the one constructed directly from a 2-element list has `RangeIndex(start=0, stop=2, step=1)`.

mvashishtha: Let's just do `pandas_df = pandas.DataFrame(["a", 1], index=[0, 0])` then.

modin/pandas/test/dataframe/test_map_metadata.py
```diff
@@ -592,20 +599,22 @@ def test_convert_dtypes(
...
+    pandas_df = pandas.DataFrame(["a", 1], index=[0, 0])
+    # The initial dataframes should be the same
+    df_equals(modin_df, pandas_df)
+    # TODO(https://github.com/modin-project/modin/pull/3805): delete
+    # this assert once df_equals checks dtypes
+    assert modin_df.dtypes.equals(pandas_df.dtypes)
+    modin_result = modin_df.convert_dtypes()
+    pandas_result = pandas_df.convert_dtypes()
+    df_equals(modin_result, pandas_result)
+    assert modin_df._query_compiler._modin_frame._partitions.shape == (2, 1)
```
mvashishtha, line 617: Could you please put this right after we make `modin_df`? The check works here, but I think it would make more sense there.

mvashishtha (commented):
@noloerino thanks for addressing my comments. I left one minor comment.

mvashishtha (approved):
LGTM!

YarShev:
`ci / Test BaseOnPython execution` job failed, please take a look at it - https://github.com/modin-project/modin/runs/7412162027?check_suite_focus=true.

noloerino:
It seems like on BaseOnPython, this is failing the multiple partition assertion @mvashishtha suggested inserting. I see a couple other tests (like [this one](https://github.com/modin-project/modin/blob/8289d54b54a41af14582eebe24f85052697b22b3/modin/pandas/test/dataframe/test_map_metadata.py#L291)) just ignore for BaseOnPython, is there a reason for that + is it acceptable to do so here as well?

mvashishtha (dismissed):

noloerino:
Thanks, I've modified the test accordingly.

docs/release_notes/release_notes-0.16.0.rst
```diff
@@ -52,6 +52,7 @@ Key Features and Updates
 * New Features
   * FEAT-4463: Add experimental fuzzydata integration for testing against a randomized dataframe workflow (#4556)
   * FEAT-#4419: Extend virtual partitioning API to pandas on Dask (#4420)
+  * FEAT-#4670:  Implement convert_dtypes by mapping across partitions (#4671)
```
YarShev, line 55:
```suggestion
  * FEAT-#4670: Implement convert_dtypes by mapping across partitions (#4671)
```

YarShev: This release note should probably go under `Pandas API implementations and improvements`.

modin/core/storage_formats/base/query_compiler.py
```diff
@@ -1394,6 +1394,46 @@ def astype(self, col_dtypes, **kwargs):  # noqa: PR02
...
+        Parameters
+        ----------
+        infer_objects : bool, default: True
+            Whether object dtypes should be converted to the best possible types.
+        convert_string : bool, default: True
+            Whether object dtypes should be converted to ``pd.StringDtype()``.
+        convert_integer : bool, default: True
+            Whether, if possbile, conversion should be done to integer extension types.
+        convert_boolean : bool, default: True
+            Whether  object dtypes should be converted to ``pd.BooleanDtype()``.
```
YarShev, line 1417:
```suggestion
            Whether object dtypes should be converted to ``pd.BooleanDtype()``.
```

modin/pandas/test/dataframe/test_map_metadata.py
```diff
@@ -16,9 +16,9 @@
 import pandas
 from pandas.testing import assert_index_equal, assert_series_equal
 import matplotlib
+
 import modin.pandas as pd
 from modin.utils import get_current_execution
-
```
YarShev, line 21: There is no need for this change. Please revert it.

modin/pandas/test/dataframe/test_map_metadata.py
```diff
@@ -557,6 +557,71 @@ def test_astype_category_large():
...
+    get_current_execution() == "BaseOnPython",
+    reason="BaseOnPython cannot not have multiple partitions.",
+)
+def test_convert_dtypes_multiple_row_partitions():
+    # Column 0 should have string dtype
+    modin_part1 = pd.DataFrame(["a"]).convert_dtypes()
+    # Column 0 should have an int dtype
+    modin_part2 = pd.DataFrame([1]).convert_dtypes()
+    modin_df = pd.concat([modin_part1, modin_part2])
+    assert modin_df._query_compiler._modin_frame._partitions.shape == (2, 1)
```
YarShev, line 612: This can be removed or put under a condition when we resolve the issue mentioned above.

modin/pandas/test/dataframe/test_map_metadata.py
```diff
@@ -557,6 +557,71 @@ def test_astype_category_large():
...
+    modin_df = pd.DataFrame(data)
+    pandas_df = pandas.DataFrame(data)
+    modin_result = modin_df.convert_dtypes(**kwargs)
+    pandas_result = pandas_df.convert_dtypes(**kwargs)
+    assert modin_result.dtypes.equals(pandas_result.dtypes)
+
+
+@pytest.mark.skipif(
+    get_current_execution() == "BaseOnPython",
+    reason="BaseOnPython cannot not have multiple partitions.",
```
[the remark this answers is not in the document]
mvashishtha: Is this really an issue with `BaseOnPython`? Do we ever want multiple partitions in `BaseOnPython`? I would think not.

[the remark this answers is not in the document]
YarShev: We do not want multiple partitions in `BaseOnPython` because it is like defaulting to pandas. My concern is that the test should work not only with multiple partitions, but even with one partition.

docs/release_notes/release_notes-0.16.0.rst
```diff
@@ -30,7 +30,8 @@ Key Features and Updates
   * REFACTOR-#4530: Standardize access to physical data in partitions (#4563)
   * REFACTOR-#4534: Replace logging meta class with class decorator (#4535)
 * Pandas API implementations and improvements
-  *
+  * FEAT-#4670: Implement convert_dtypes by mapping across partitions (#4671)
+
```
YarShev, line 34:
```suggestion
```

modin/pandas/test/dataframe/test_map_metadata.py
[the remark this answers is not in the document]
mvashishtha: The test should pass when there's just one partition, but on the other hand I think it's confusing to run `test_convert_dtypes_multiple_row_partitions` when there can only be one partition, and I think it's good to check that we really are testing for the condition of multiple row partitions.

But I won't insist. I'm okay with getting rid of the check and the `skipif`.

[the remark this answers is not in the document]
YarShev: Having looked at the line where the test failed for BaseOnPython - https://github.com/modin-project/modin/runs/7412162027?check_suite_focus=true, I think it is okay to keep the test as is. I just thought about other reason of the test failure.

YarShev (approved):
LGTM!

mvashishtha (approved):

PR #4674, opened 2022-07-13
FIX-#4660: Fix `fillna` when Modin series object is an argument

### LINKED ISSUE #4660: Fillna doesn't work when series object is an argument
### System information
- **Python version**:

```python
import modin.pandas as pd

df = pd.DataFrame({'a': ['a'], 'b': ['b'],}, index=['row1'])
df['c'] = pd.NA
df['c'] = df['c'].fillna(df['b'])   # df['c'].fillna('c') works fine
print(df)
```

### Describe the problem

when fillna is called with series object as argument seems nothing happens while pandas does actual na replacement

anmyachev:

## What do these changes do?

- [x] Resolves #4660 
- [x] tests added and passing


modin/core/storage_formats/pandas/query_compiler.py
```diff
@@ -1865,15 +1865,17 @@ def fillna_builder(series):
                     )
                 else:
 
-                    def fillna_builder(series, value_arg):
-                        # Both arguments for this function are 1-column `DataFrames` which denote `Series` type.
-                        # Because they are both of the same type, it is not necessary to convert either of them into
-                        # `Series` by squeezing since `fillna` works perfectly in the same way on 1-column `DataFrame`
-                        # objects (when `limit` parameter is absent) as it works on two `Series`.
```
anmyachev, line 1872: This's no longer true.

modin/core/storage_formats/pandas/query_compiler.py
```diff
@@ -1865,15 +1865,17 @@ def fillna_builder(series):
...
-                        return series.fillna(value=value_arg, **kwargs)
+                    def fillna_builder(df, value_arg):
+                        if isinstance(value_arg, pandas.DataFrame):
+                            value_arg = value_arg.squeeze(axis=1)
+                        res = df.squeeze(axis=1).fillna(value=value_arg, **kwargs)
```
anmyachev, line 1871: One of the two major issues (another one is column reindexing) that need to be fixed, details here: https://github.com/modin-project/modin/issues/4660#issuecomment-1182008362

modin/pandas/test/dataframe/test_window.py
```diff
@@ -206,6 +206,13 @@ def test_fillna_downcast():
     df_equals(modin_df, result)
 
 
+def test_fillna_4660():
```
jbrockmendel, line 209: out of curiosity, what is the convention for deciding where this test goes?  i wouldn't have guessed this file

anmyachev: Good question, I just found similar tests in this file and added my test here. Do you have any suggestions where to move these tests? I would gladly do it.

jbrockmendel: in pandas this would live in pandas/tests/frame/methods/test_fillna.py; i expect implementing that file-per-method setup in modin would make a lot of noise

anmyachev: So, in the file there are no windowing operations according to https://pandas.pydata.org/docs/user_guide/window.html#overview.

I think if we give it a different name, it will already be better. For example, `test_window.py` -> `test_methods.py`. With an increase in the number of tests, it will be possible to divide it into files in the future. What do you think about?

modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -2362,7 +2362,9 @@ def broadcast_apply_full_axis(
             result.synchronize_labels(axis=1)
         return result
 
-    def _copartition(self, axis, other, how, sort, force_repartition=False):
+    def _copartition(
+        self, axis, other, how, sort, force_repartition=False, make_reindex=True
```
vnlitvinov, line 2366:
```suggestion
        self, axis, other, how, sort, force_repartition=False, perform_reindex=True
```
you don't "make" an operation so it is applied, you "perform", "apply", "do" or "execute" it 🙃

modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -2428,7 +2433,9 @@ def _copartition(self, axis, other, how, sort, force_repartition=False):
         base_index = base_frame.axes[axis]
 
         # define conditions for reindexing and repartitioning `self` frame
-        do_reindex_base = not base_index.equals(joined_index)
+        do_reindex_base = False
+        if make_reindex:
+            do_reindex_base = not base_index.equals(joined_index)
```
vnlitvinov, line 2438:
```suggestion
        do_reindex_base = make_reindex and not base_index.equals(joined_index)
```

modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -2483,10 +2496,15 @@ def _copartition(self, axis, other, how, sort, force_repartition=False):
...
+        return (
+            reindexed_frames[0],
+            reindexed_frames[1:],
+            joined_index if make_reindex else None,
+            base_lengths,
+        )
 
     @lazy_metadata_decorator(apply_axis="both")
-    def binary_op(self, op, right_frame, join_type="outer"):
+    def binary_op(self, op, right_frame, join_type="outer", make_column_reindex=True):
```
vnlitvinov, line 2507: same about "doing" a reindex or "performing" or whatever

modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -2532,11 +2554,12 @@ def binary_op(self, op, right_frame, join_type="outer"):
                 left_parts, op, right_parts[0]
             )
         )
-
+        if not make_column_reindex:
+            joined_columns = self.columns.join(right_frame.columns, how=join_type)
```
vnlitvinov, line 2558: could you please explain why?

modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -2532,11 +2554,12 @@ def binary_op(self, op, right_frame, join_type="outer"):
...
         return self.__constructor__(
             new_frame,
             joined_index,
-            joined_columns,
+            joined_columns if make_column_reindex else joined_columns,
```
vnlitvinov, line 2562: huh? 🤣

modin/pandas/test/dataframe/test_window.py
vnlitvinov: `test_methods` feels to generic, can we determine a subset of methods we test here? or maybe split `fillna` in a separate test file...

modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -2383,6 +2385,8 @@ def _copartition(self, axis, other, how, sort, force_repartition=False):
             this method will skip repartitioning if it is possible. This is because
             reindexing is extremely inefficient. Because this method is used to
             `join` or `append`, it is vital that the internal indices match.
+        make_reindex : bool, default: True
```
anmyachev, line 2388:
```suggestion
        perform_reindex : bool, default: True
```

modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -2400,9 +2404,10 @@ def _copartition(self, axis, other, how, sort, force_repartition=False):
 
         self_index = self.axes[axis]
         others_index = [o.axes[axis] for o in other]
-        joined_index, make_reindexer = self._join_index_objects(
-            axis, [self_index] + others_index, how, sort
-        )
+        if make_reindex:
```
anmyachev, line 2407:
```suggestion
        if perform_reindex:
```

modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -2428,7 +2433,9 @@ def _copartition(self, axis, other, how, sort, force_repartition=False):
         base_index = base_frame.axes[axis]
 
         # define conditions for reindexing and repartitioning `self` frame
-        do_reindex_base = not base_index.equals(joined_index)
+        do_reindex_base = False
+        if make_reindex:
```
anmyachev, line 2437:
```suggestion
        if perform_reindex:
```

modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -2438,7 +2445,9 @@ def _copartition(self, axis, other, how, sort, force_repartition=False):
             reindexed_base = base_frame._partition_mgr_cls.map_axis_partitions(
                 axis,
                 base_frame._partitions,
-                make_reindexer(do_reindex_base, base_frame_idx),
+                make_reindexer(do_reindex_base, base_frame_idx)
+                if make_reindex
```
anmyachev, line 2449:
```suggestion
                if perform_reindex
```

modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -2473,7 +2484,9 @@ def _copartition(self, axis, other, how, sort, force_repartition=False):
                 ]._partition_mgr_cls.map_axis_partitions(
                     axis,
                     other_frames[i]._partitions,
-                    make_reindexer(do_repartition_others[i], base_frame_idx + 1 + i),
+                    make_reindexer(do_repartition_others[i], base_frame_idx + 1 + i)
+                    if make_reindex
```
anmyachev, line 2488:
```suggestion
                    if perform_reindex
```

modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -2483,10 +2496,15 @@ def _copartition(self, axis, other, how, sort, force_repartition=False):
             + [reindexed_base]
             + reindexed_other_list
         )
-        return (reindexed_frames[0], reindexed_frames[1:], joined_index, base_lengths)
+        return (
+            reindexed_frames[0],
+            reindexed_frames[1:],
+            joined_index if make_reindex else None,
```
anmyachev, line 2502:
```suggestion
            joined_index if perform_reindex else None,
```

modin/core/dataframe/pandas/dataframe/dataframe.py
anmyachev, line 2507:
```suggestion
    def binary_op(self, op, right_frame, join_type="outer", perform_column_reindex=True):
```

modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -2498,6 +2516,8 @@ def binary_op(self, op, right_frame, join_type="outer"):
             Modin DataFrame to join with.
         join_type : str, default: "outer"
             Type of join to apply.
+        make_column_reindex : bool, default: True
```
anmyachev, line 2519:
```suggestion
        perform_column_reindex : bool, default: True
```

modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -2523,7 +2543,9 @@ def binary_op(self, op, right_frame, join_type="outer"):
             right_parts,
             joined_columns,
             column_widths,
-        ) = new_left_frame._copartition(1, new_right_frame, join_type, sort=True)
+        ) = new_left_frame._copartition(
+            1, new_right_frame, join_type, sort=True, make_reindex=make_column_reindex
```
anmyachev, line 2547:
```suggestion
            1, new_right_frame, join_type, sort=True, perform_reindex=perform_column_reindex
```

modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -2532,11 +2554,12 @@ def binary_op(self, op, right_frame, join_type="outer"):
                 left_parts, op, right_parts[0]
             )
         )
-
+        if not make_column_reindex:
```
anmyachev, line 2557:
```suggestion
        if not perform_column_reindex:
```

modin/core/storage_formats/pandas/query_compiler.py
```diff
@@ -1865,15 +1865,17 @@ def fillna_builder(series):
...
+                        return pandas.DataFrame(res)
 
                     new_modin_frame = self._modin_frame.binary_op(
-                        fillna_builder, value._modin_frame, join_type="left"
+                        fillna_builder,
+                        value._modin_frame,
+                        join_type="left",
+                        make_column_reindex=False,
```
anmyachev, line 1878:
```suggestion
                        perform_column_reindex=False,
```

modin/core/dataframe/pandas/dataframe/dataframe.py
anmyachev: I rewrote this piece of code. `joined_columns` is now obtained from `_copartition` as before.

modin/pandas/test/dataframe/test_window.py
anmyachev: Let's discuss it here https://github.com/modin-project/modin/issues/4688

modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -2483,10 +2493,17 @@ def _copartition(self, axis, other, how, sort, force_repartition=False):
             + [reindexed_base]
             + reindexed_other_list
         )
-        return (reindexed_frames[0], reindexed_frames[1:], joined_index, base_lengths)
+        return (
+            reindexed_frames[0],
+            reindexed_frames[1:],
+            joined_index,
+            base_lengths,
+        )
```
vnlitvinov, line 2501: what is the reason for this change?

vnlitvinov (approved):
LGTM with a small question

modin/core/dataframe/pandas/dataframe/dataframe.py
anmyachev: No reason, returned changes

modin/pandas/test/dataframe/test_window.py
prutskov, line 209: Why is this added in `test_window.py`?

prutskov: Oh, I found the thread with conversation related this https://github.com/modin-project/modin/pull/4674#discussion_r925605315. So, I resolve this conversation

vnlitvinov (commented):
I now wonder what is the reason for _not_ reindexing in the case of `df1["c"].fillna(df2)`? What if partition boundaries do not match for `df1` and `df2`? What happens if `df1.index` does not match `df2.index`, how would `.fillna()` work in that case?

modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -2498,6 +2510,8 @@ def binary_op(self, op, right_frame, join_type="outer"):
             Modin DataFrame to join with.
         join_type : str, default: "outer"
             Type of join to apply.
+        perform_column_reindex : bool, default: True
```
YarShev, line 2513: I wonder if we can avoid introducing a new parameter to `binary_op` and compute a condition inside of the operator when `perform_reindex` is equal to `True` or `False` in order to pass it in to `_copartition`?

prutskov:
> I now wonder what is the reason for _not_ reindexing in the case of `df1["c"].fillna(df2)`? What if partition boundaries do not match for `df1` and `df2`? What happens if `df1.index` does not match `df2.index`, how would `.fillna()` work in that case?

The case `df1["c"].fillna(df2)` is not possible, because `value` for `Series.fillna` must be `a scalar, dict or Series`, but not `DataFrame`:

```python
import pandas as pd
# import modin.pandas as pd

df = pd.DataFrame({'a': ['a'], 'b': ['b'],}, index=['row1'])
df['c'] = pd.NA

df2_0 = pd.DataFrame({"a": [0], "b": [5]}, index=['row1'])
df2_1 = pd.DataFrame({"c": ["c"]}, index=['row1'])
df2 = pd.concat([df2_0, df2_1], axis=1)
df = df["c"].fillna(df2)
print(df)
```

```shell
  File "test_fillna.py", line 12, in <module>
    df = df["c"].fillna(df2)
  File "C:\Users\79049\.conda\envs\modin\lib\site-packages\pandas\util\_decorators.py", line 311, in wrapper
    return func(*args, **kwargs)
  File "C:\Users\79049\.conda\envs\modin\lib\site-packages\pandas\core\series.py", line 4908, in fillna
    return super().fillna(
  File "C:\Users\79049\.conda\envs\modin\lib\site-packages\pandas\core\generic.py", line 6461, in fillna
    raise TypeError(
TypeError: "value" parameter must be a scalar, dict or Series, but you passed a "DataFrame"
```

modin/core/dataframe/pandas/dataframe/dataframe.py
prutskov: I believe, we can avoid introducing the new parameter. Instead of this we just need to check amount of columns of `self`/`right_frame` of equality to `1` (this is only one case when `fillna` uses `binary_op`, @anmyachev correct me if I'm wrong) and make reindexing only by rows in this case.

anmyachev: It'll be complex condition because there's another case where `binary_op` is used. For example, let's take `add` binary operation. Behavior of the function is different where type of left and right operands is `DataFrame` and amount of `self`/`right_frame` is `1`. In that case we need to do column `reindex`.

Code:
```python
import modin.pandas as pd
df1 = pd.DataFrame([1,2,3,4], columns=["a"])
df2 = pd.DataFrame([7,6,4,9], columns=["b"])
print(df1 + df2)
```

However we can't differentiate `fillna` and `add` cases in `binary_op` function because we can't check type of operands in `binary_op`. So I'm even not sure that is possible to define `perform_reindex` condition in `binary_op` layer.

anmyachev: @prutskov @YarShev hope I answered you.

YarShev: I am still thinking about to avoid introducing the parameter. What if we do column reindex?

anmyachev: If we do column reindex, then this simply will not allow us to correctly apply the fillna function. In the case when the series have different names, this will cause the right operand to consist entirely of nans (after reindex), which in the case of this function will simply do nothing with the left operand.

anmyachev: I can try to enter a new value for parameter `join_type`. Something like: `left_only_rows`.

anmyachev: @YarShev I managed to get rid of the new parameter in the `_copartition` function.

modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -2658,6 +2660,8 @@ def n_ary_op(self, op, right_frames: list, join_type="outer"):
             Modin DataFrames to join with.
         join_type : str, default: "outer"
             Type of join to apply.
+        perform_column_reindex : bool, default: True
+            For some ops, for example `fillna`, column reindexing of right operand isn't needed.
```
YarShev, line 2664:
```suggestion
            Whether to perform copartitioning along columns or not.
            For some ops this isn't needed (e.g., `fillna`).
```

modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -2658,6 +2660,8 @@ def n_ary_op(self, op, right_frames: list, join_type="outer"):
             Modin DataFrames to join with.
         join_type : str, default: "outer"
             Type of join to apply.
+        perform_column_reindex : bool, default: True
```
YarShev, line 2663:
```suggestion
        copartition_along_columns : bool, default: True
```

modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -2646,7 +2646,9 @@ def _copartition(self, axis, other, how, sort, force_repartition=False):
         return (reindexed_frames[0], reindexed_frames[1:], joined_index, base_lengths)
 
     @lazy_metadata_decorator(apply_axis="both")
-    def n_ary_op(self, op, right_frames: list, join_type="outer"):
+    def n_ary_op(
+        self, op, right_frames: list, join_type="outer", perform_column_reindex=True
```
YarShev, line 2650:
```suggestion
        self, op, right_frames: list, join_type="outer", copartition_along_columns=True
```

modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -2667,26 +2671,35 @@ def n_ary_op(self, op, right_frames: list, join_type="outer"):
...
-            left_parts, joined_index, self.columns, row_lengths, self.column_widths
-        )
-        new_right_frames = [
-            self.__constructor__(
-                right_parts,
-                joined_index,
-                right_frame.columns,
-                row_lengths,
-                right_frame.column_widths,
+        if perform_column_reindex:
```
YarShev, line 2674:
```suggestion
        if copartition_along_columns:
```

docs/release_notes/release_notes-0.16.0.rst
```diff
@@ -15,6 +15,7 @@ Key Features and Updates
   * FIX-#4604: Fix `groupby` + `agg` in case when multicolumn can arise (#4642)
   * FIX-#4582: Inherit custom log layer (#4583)
   * FIX-#4639: Fix `storage_options` usage for `read_csv` and `read_csv_glob` (#4644)
+  * FIX-#4660: Fix `fillna` when Modin series object is an argument (#4674)
```
YarShev, line 18:
```suggestion
```

YarShev (approved):

PR #4495, opened 2022-05-25
PERF-#4493: Use partition size caches more in Modin dataframe.

mvashishtha:
Signed-off-by: mvashishtha <mahesh@ponder.io>

## What do these changes do?

- [x] commit message follows format outlined [here](https://modin.readthedocs.io/en/latest/development/contributing.html#commit-message-formatting)
- [x] passes `flake8 modin/ asv_bench/benchmarks scripts/doc_checker.py`
- [x] passes `black --check modin/ asv_bench/benchmarks scripts/doc_checker.py`
- [x] signed commit with `git commit -s` 
- [x] Resolves #4493 
- [x] tests added and passing
- [x] module layout described at `docs/development/architecture.rst` is up-to-date 
- [x] added (Issue Number: PR title (PR Number)) and github username to release notes for next major release

Issue #4493, opened 2022-05-25
PERF: Try width and length caches before materializing all partition lengths/widths in Modin frame


mvashishtha:
### System information
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: macOS Monterey 12.2.1
- **Modin version** (`modin.__version__`): 0f70e828ed127229f7160075d2ad06a6b16b1002
- **Python Version**: 3.9.12

### Describe the problem
Currently I see two points in the Modin frame where we compute either all partition lengths for the first column of partitions, or all column widths for the first row of partitions:

1) https://github.com/modin-project/modin/blob/c736defeae77e963e5421b1c60d6d020d53a8220/modin/core/dataframe/pandas/dataframe/dataframe.py#L2380

2) https://github.com/modin-project/modin/blob/c736defeae77e963e5421b1c60d6d020d53a8220/modin/core/dataframe/pandas/dataframe/dataframe.py#L2152

The first line, in `_copartition`, recently caused single-threaded execution for a frame with partitions of `Decimal` objects. Each frame had a transpose on the queue. Executing a multiply then caused the widths to be computed serially, so each partition's call queue was drained in sequence. The result was that serially, each partition slowly put the transpose result in the object store. (The objects took a while to put in the object store because `Decimal` data is slow to serialize.) However, in this case the lengths and widths were cache, so there was no need to compute lengths and widths at all.

Attached is a [ray timeline](https://github.com/modin-project/modin/files/8770352/out.json.zip) and here is an image of the single threaded execution for the transpose in the middle (from a similar script). ![Screen Shot 2022-05-25 at 5 32 32 AM](https://user-images.githubusercontent.com/10712109/170242622-64642576-6b76-415f-9463-0c07ab3e7c2a.png)

**Reproduction script**

```python
import numpy as np
import modin.pandas as pd
from decimal import Decimal

height = 50_000
width = 751
im = pd.DataFrame(np.random.randint(0, 2, size=(height, width)))
im = im.applymap(lambda x: Decimal(str(x)))
weight_vec = pd.Series(np.random.rand(height)).apply(lambda x: Decimal(str(x)))
print(im.T.multiply(weight_vec))
```

mvashishtha:
It turns out that in the case of 1) `reindexed_base` has unknown axis lengths because we might have to add elements along the axis to align with the other frame e.g. for 

```python
import pandas as pd

A =  pd.DataFrame([[1]])
[...]

Issue #2814, opened 2021-03-04
Improve scaling of to_pandas


vnlitvinov:

jbrockmendel:
> I wasn't able to find out a good sequence of incantations for Pandas to not copy blocks around during concatenation

Can you give an example of what you're trying to do?  This may be fixed in newer pandas:

```
arr = np.random.randn(4, 2)
arr2 = np.random.randn(4, 3)
df = pd.DataFrame(arr)
df2 = pd.DataFrame(arr2)

res = pd.concat([df, df2], axis=1)
>>> res._mgr.nblocks
2
```

PR #4683, opened 2022-07-18
PERF-#4494: Get partition widths/lengths in parallel instead of serially

### LINKED ISSUE #4494: PERF: get all partition widths/lengths in parallel instead of serially.
The reproducing script is the same as in #4493, but the solution here is different: instead of getting all lengths/widths serially, we should do so in parallel. We will need to add a new method at the physical layer for that. 

This solution will not save us the cost of serializing the call queue drain result after computing length, but it will let us get all the lengths/widths at once instead of serially.

noloerino:
## What do these changes do?

- [ ] Resolves #4494 
- [ ] tests added and passing


pyrito:
Haven't taken a closer look at the implementation details, but do you have any benchmarks or performance measurements to compare with master?

noloerino:
Sadly no, and I’d appreciate some suggestions on what code to run. Rehan
suggested manually invalidating the ._row_lengths_cache and .length_cache
fields on a dataframe and its partitions, then ensuring they’re recomputed
properly. It succeeds for simple examples, but I had trouble producing a
Ray timeline, and I’m not sure how else to benchmark it (most API-level
dataframe manipulations would probably hit the cached length/width).

modin/core/dataframe/pandas/partitioning/partition.py
```diff
@@ -273,13 +274,42 @@ def length(self):
...
+            just computed.
+        """
         if self._length_cache is None:
             cls = type(self)
             func = cls._length_extraction_fn()
             preprocessed_func = cls.preprocess_func(func)
             self._length_cache = self.apply(preprocessed_func)
         return self._length_cache
 
+    def try_set_length_cache(self, length: int):
```
jeffreykennethli, line 298: nit: this function doesn't really "try" anything. Could it be named just `set_length_cache`? The same applies for `try_set_width_cache`.

modin/core/dataframe/pandas/partitioning/partition.py
```diff
@@ -273,13 +274,42 @@ def length(self):
         int
             The length of the object.
         """
+        self.try_build_length_cache()
```
jeffreykennethli, line 277: this can be `return self.try_build_length_cache()`.

modin/core/dataframe/pandas/partitioning/partition_manager.py
```diff
@@ -870,6 +870,25 @@ def get_indices(cls, axis, partitions, index_func=None):
...
+        Build and set the length and width caches of each _physical_ partition in the given list of partitions.
+
+        Parameters
+        ----------
+        partitions : np.ndarray
+            The partitions for which to update length caches.
+
+        Notes
+        -----
+        For backends that support parallel computations, these caches should be computed asynchronously.
```
jeffreykennethli, line 885: nit: "these caches *are* computed...". IMO this is clearer because sometimes when we say "should", we mean "should theoretically" or "should eventually (when I get around to implementing it)".

modin/core/execution/ray/implementations/pandas_on_ray/partitioning/partition_manager.py
```diff
@@ -223,6 +225,42 @@ def map_axis_partitions(
...
+            w_cache = part.try_build_width_cache()
+            if isinstance(w_cache, ObjectIDType):
+                promises.append(w_cache)
+                part_idxs.append(i)
+                is_lens.append(False)
+        new_dims = ray.get(promises)
+        for i, new_cache in enumerate(new_dims):
+            part = partitions[part_idxs[i]]
+            if is_lens[i]:
+                part.try_set_length_cache(new_dims[i])
```
jeffreykennethli, line 260: should `new_dims[i]` be `new_cache`?

mvashishtha:
@noloerino @pyrito 

> Sadly no, and I’d appreciate some suggestions on what code to run.

I spent a while today trying to get a script that showcases the performance here without breaking anything in Modin, but I failed. Getting a reproducer is hard for a few reasons.

For one thing, this optimization is only useful for unusual cases like in #4493 where the partitions' call queues include costly operations. When there is no call queue, the partitions will execute all dataframe functions eagerly, simultaneously calculating shapes. The call queues are generally meant to carry cheap operations like transpose and reindexing, but the [reproducer](https://github.com/modin-project/modin/issues/4493#issue-1247911681) in that issue has a frame that is very expensive to serialize, so that even the transpose was expensive. There the slow code was in `_copartition`, which unnecessarily calculated the widths of the base frame. #4495 fixed that unnecessary recalculation, so that script no longer works. 

Looking at all the serial shape computations I listed [here](https://github.com/modin-project/modin/issues/4494#issuecomment-1182614520), most are in internal length computations. One is `_copartition`, and I spent a while trying to get around the cache fix in #4495 with a pair of frames that really needed copartitioning, but in that case the [`map_axis_partitions`](https://github.com/modin-project/modin/blob/81648e711f7525d87dbc58f53e67a2fd22650ae7/modin/core/dataframe/pandas/dataframe/dataframe.py#L2438) in `_copartition` triggers parallel computation. The last type of length computation is in `apply_func_to_indices_both_axis`, which as far as I can tell is only used in `melt`. We could try engineering an example that bypasses the cache for `melt`, but I don't think it's worth the time...

I think it's good practice to get multiple ray objects in parallel (see also [this note](https://github.com/modin-project/modin/issues/2814#issue-822049379) about a similar improvement in `_to_pandas`). Also, if our caches fail for any reason later on, we can have faster length computation as a backup.

vnlitvinov:
This adds a certain bit of complexity (judging by the number of lines change, haven't looked at the diff yet), and I haven't yet seen any performance proof for that. I would like to see some measurements before increasing our (already huge) codebase...

modin/core/dataframe/pandas/partitioning/partition.py
```diff
@@ -273,13 +274,42 @@ def length(self):
...
+        return self._length_cache
```
RehanSD, line 278: We need to unwrap `_length_cache` here, since its type will be `PandasDataframePartition`

modin/core/execution/dask/implementations/pandas_on_dask/partitioning/virtual_partition.py
```diff
@@ -372,6 +372,20 @@ def length(self):
         """
         if self._length_cache is None:
             if self.axis == 0:
+                caches = [
+                    obj.try_build_length_cache()
+                    for obj in self.list_of_partitions_to_combine
+                ]
+                new_lengths = DaskWrapper.materialize(
+                    [cache for cache in caches if isinstance(promise, Future)]
```
RehanSD, line 380:
```suggestion
                    [cache for cache in caches if isinstance(cache, Future)]
```

modin/core/execution/dask/implementations/pandas_on_dask/partitioning/virtual_partition.py
```diff
@@ -372,6 +372,20 @@ def length(self):
...
+                )
+                dask_idx = 0
+                for i, cache in enumerate(caches):
+                    if isinstance(cache, Future):
+                        self.list_of_partitions_to_combine[i].try_set_length_cache(
+                            new_lengths[dask_idx]
```
RehanSD, line 386: Shouldn't this just be `i` as well?

modin/core/execution/dask/implementations/pandas_on_dask/partitioning/virtual_partition.py
```diff
@@ -392,6 +406,20 @@ def width(self):
         """
         if self._width_cache is None:
             if self.axis == 1:
+                caches = [
+                    obj.try_build_width_cache()
+                    for obj in self.list_of_partitions_to_combine
+                ]
+                new_lengths = DaskWrapper.materialize(
```
RehanSD, line 413:
```suggestion
                new_widths = DaskWrapper.materialize(
```

modin/core/execution/ray/implementations/pandas_on_ray/partitioning/partition_manager.py
```diff
@@ -223,6 +225,42 @@ def map_axis_partitions(
...
+        """
+        Build and set the length and width caches of each _physical_ partition in the given list of partitions.
+
+        Parameters
+        ----------
+        partitions : np.ndarray
+            The partitions for which to update length caches.
+        """
+        promises = []
+        # If `ray_idxs[i] = j`, that means `promises[i]` represents a computation corresponding
```
RehanSD, line 239:
```suggestion
        # If `part_idxs[i] = j`, that means `promises[i]` represents a computation corresponding
```

?

modin/core/execution/dask/implementations/pandas_on_dask/partitioning/partition.py
```diff
@@ -271,12 +286,25 @@ def width(self):
...
             The width of the object.
         """
-        if self._width_cache is None:
-            self._width_cache = self.apply(lambda df: len(df.columns))._data
+        self.try_build_width_cache()
         if isinstance(self._width_cache, Future):
             self._width_cache = DaskWrapper.materialize(self._width_cache)
         return self._width_cache
 
+    def try_build_width_cache(self) -> Union[Future, int]:
```
RehanSD, line 294: Shouldn't this and the length one just be `build_..._cache`?

modin/core/execution/ray/implementations/pandas_on_ray/partitioning/partition.py
```diff
@@ -278,15 +280,30 @@ def length(self):
...
+        Returns
+        -------
+        ray.ObjectRef | int
+            Either a Ray object ID representing the length of the object wrapped by this partition,
+            or the concrete value of the length if it was already cached.
+        """
         if self._length_cache is None:
             if len(self.call_queue):
                 self.drain_call_queue()
             else:
```
RehanSD, line 303: This should not be in an else.

modin/core/execution/ray/implementations/pandas_on_ray/partitioning/partition_manager.py
```diff
@@ -223,6 +225,42 @@ def map_axis_partitions(
...
+    def _update_partition_dimension_caches(cls, partitions: np.ndarray):
+        """
+        Build and set the length and width caches of each _physical_ partition in the given list of partitions.
+
+        Parameters
+        ----------
+        partitions : np.ndarray
+            The partitions for which to update length caches.
+        """
+        promises = []
```
RehanSD, line 238: We may want to use a different term than promises here, since we don't really use that term anywhere else in the codebase.

modin/core/execution/ray/implementations/pandas_on_ray/partitioning/partition_manager.py
RehanSD: +1

modin/core/execution/ray/implementations/pandas_on_ray/partitioning/virtual_partition.py
```diff
@@ -377,6 +377,20 @@ def length(self):
         """
         if self._length_cache is None:
             if self.axis == 0:
+                caches = [
+                    obj.try_build_length_cache()
+                    for obj in self.list_of_partitions_to_combine
+                ]
+                new_lengths = ray.get(
+                    [cache for cache in caches if isinstance(promise, ObjectIDType)]
```
RehanSD, line 385:
```suggestion
                    [cache for cache in caches if isinstance(cache, ObjectIDType)]
```

RehanSD (changes requested):
Left some comments, but great work!

modin/core/dataframe/pandas/partitioning/partition.py
noloerino: What do you mean by `unwrap`? Also, as far as I can tell, the logic for this method should be the same as it originally was (the code was just moved into the `try_build_length_cache`, so does this mean the original code returned `PandasDataframePartition` as well?

modin/core/dataframe/pandas/partitioning/partition_manager.py
```diff
@@ -1214,6 +1233,19 @@ def apply_func_to_indices_both_axis(
         if col_widths is None:
             col_widths = [None] * len(col_partitions_list)
 
+        if row_lengths is None and col_widths is None:
```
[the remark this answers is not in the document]
noloerino: The `length` and `width` values of each partition are accessed in the local `compute_part_size`, defined immediately below. The double `for` loop structure where `compute_part_size` is called makes it hard to parallelize the computation of these dimensions, so I thought it would be simplest to precompute the relevant dimensions before the loop.

modin/core/execution/dask/implementations/pandas_on_dask/partitioning/virtual_partition.py
noloerino: No, since `new_lengths` may have fewer elements than `caches` in the case where some length values were already computed (and are filtered out by the `isinstance(cache, Future)` check). The value computed at `new_lengths[dask_idx]` should correspond to the promise at `caches[i]`.

modin/core/dataframe/pandas/partitioning/partition.py
noloerino: changed

modin/core/dataframe/pandas/partitioning/partition_manager.py
noloerino: changed, in my head "should" implies that it is a contract for implementors to follow, but I'll adhere to the standard of the codebase

modin/core/execution/dask/implementations/pandas_on_dask/partitioning/partition.py
noloerino: changed

modin/core/execution/ray/implementations/pandas_on_ray/partitioning/partition_manager.py
noloerino: changed to `futures`

modin/core/execution/ray/implementations/pandas_on_ray/partitioning/partition_manager.py
noloerino: good catch, I forgot that's why I used the `enumerate`

noloerino:
@vnlitvinov that makes sense, I'll look into coming up with concrete benchmarks.

vnlitvinov:
@pyrito please have a look at https://github.com/vnlitvinov/modin/tree/speedup-masking, it might be doing somewhat the same in terms of getting the sizes in parallel

YarShev:
Related discussion on handling metadata (index and columns) in https://github.com/modin-project/modin/issues/3673.

PR #4685, opened 2022-07-19
REFACTOR-#4629: Add type annotations to `modin/config`

### LINKED ISSUE #4629: Add mypy checks for modin.config
We need to expand our mypy checks to cover Python files in `modin/config/`. There may be a lot of files to modify at once so folks should focus on type annotations one file at a time. Please make sure to add files to `mypy.ini`!

pyrito:
## What do these changes do?

This PR adds type annotations and various fixes found by `mypy` to `modin/config`. 

- [x] Resolves #4629 
- [x] tests added and passing

### DIFF
--- a/modin/config/__main__.py
+++ b/modin/config/__main__.py
@@ -57,9 +57,9 @@ def export_config_help(filename: str) -> None:
                 if obj.__name__ != "RayRedisPassword"
                 else "random string",
                 # `Notes` `-` underlining can't be correctly parsed inside csv table by sphinx
-                "Description": dedent(obj.__doc__).replace("Notes\n-----", "Notes:\n")
-                if obj.__doc__
-                else "",
+                "Description": dedent(obj.__doc__ or "").replace(
+                    "Notes\n-----", "Notes:\n"
+                ),
                 "Options": obj.choices,
             }
             configs_data.append(data)
--- a/modin/config/__main__.py
+++ b/modin/config/__main__.py
@@ -27,7 +27,7 @@ import pandas
 import argparse
 
 
-def print_config_help():
+def print_config_help() -> None:
     """Print configs help messages."""
     for objname in sorted(globals()):
         obj = globals()[objname]
@@ -35,7 +35,7 @@ def print_config_help():
             print(f"{obj.get_help()}\n\tCurrent value: {obj.get()}")  # noqa: T201
 
 
-def export_config_help(filename: str):
+def export_config_help(filename: str) -> None:
     """
     Export all configs help messages to the CSV file.
 
@@ -57,7 +57,9 @@ def export_config_help(filename: str):
                 if obj.__name__ != "RayRedisPassword"
                 else "random string",
                 # `Notes` `-` underlining can't be correctly parsed inside csv table by sphinx
-                "Description": dedent(obj.__doc__).replace("Notes\n-----", "Notes:\n"),
+                "Description": dedent(obj.__doc__).replace("Notes\n-----", "Notes:\n")
+                if obj.__doc__
+                else "",
                 "Options": obj.choices,
             }
             configs_data.append(data)
--- a/modin/config/pubsub.py
+++ b/modin/config/pubsub.py
@@ -122,15 +122,15 @@ class Parameter(object):
 
     Attributes
     ----------
-    choices : sequence of str
+    choices : Optional[Sequence[str]]
         Array with possible options of ``Parameter`` values.
     type : str
         String that denotes ``Parameter`` type.
-    default : Any
+    default : Optional[Any]
         ``Parameter`` default value.
     is_abstract : bool, default: True
         Whether or not ``Parameter`` is abstract.
-    _value_source : int
+    _value_source : Optional[int]
         Source of the ``Parameter`` value, should be set by
         ``ValueSource``.
     """
--- a/modin/config/pubsub.py
+++ b/modin/config/pubsub.py
@@ -13,7 +13,7 @@
 
 """Module houses ``Parameter`` class - base class for all configs."""
 
-import collections
+from collections import defaultdict
 from typing import Any, Callable, NamedTuple, Optional, Sequence, Union
 
 
@@ -140,6 +140,9 @@ class Parameter(object):
     default: Optional[Any] = None
     is_abstract = True
     _value_source: Optional[int] = None
+    _value: Any = _UNSET
+    _subs: list = []
+    _once: defaultdict[Any, list] = defaultdict(list)
 
     @classmethod
     def _get_raw_from_config(cls) -> str:
@@ -196,7 +199,7 @@ class Parameter(object):
         cls.is_abstract = abstract
         cls._value = _UNSET
         cls._subs = []
-        cls._once = collections.defaultdict(list)
+        cls._once = defaultdict(list)
         super().__init_subclass__(**kw)
 
     @classmethod
--- a/modin/config/pubsub.py
+++ b/modin/config/pubsub.py
@@ -106,7 +106,7 @@ _TYPE_PARAMS = {
 _UNSET = object()
 
 
-class ValueSource(IntEnum):
+class ValueSource(IntEnum):  # noqa: PR01
     """Class that describes the method of getting the value for a parameter."""
 
     # got from default, i.e. neither user nor configuration source had the value
--- a/modin/config/pubsub.py
+++ b/modin/config/pubsub.py
@@ -14,7 +14,6 @@
 """Module houses ``Parameter`` class - base class for all configs."""
 
 import collections
-from subprocess import call
 from typing import Any, Callable, NamedTuple, Optional, Sequence, Union
 
 
@@ -71,7 +70,7 @@ _TYPE_PARAMS = {
     ),
     int: TypeDescriptor(
         decode=lambda value: int(value.strip()),
-        normalize=int, # type: ignore
+        normalize=int,  # type: ignore
         verify=lambda value: isinstance(value, int)
         or (isinstance(value, str) and value.strip().isdigit()),
         help="an integer value",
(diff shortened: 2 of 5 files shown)


modin/config/pubsub.py
```diff
@@ -48,7 +48,7 @@ class ExactStr(str):
 _TYPE_PARAMS = {
     str: TypeDescriptor(
         decode=lambda value: value.strip().title(),
-        normalize=lambda value: value.strip().title(),
+        normalize=lambda value: str(value).strip().title(),
```
vnlitvinov, line 51: `value` is expected to already be `str` here tbh...

modin/config/pubsub.py
```diff
@@ -135,11 +135,14 @@ class Parameter(object):
...
     """
 
-    choices: typing.Sequence[str] = None
+    choices: Optional[Sequence[str]] = None
     type = str
-    default = None
+    default: Optional[Any] = None
     is_abstract = True
-    _value_source = None
+    _value_source: Optional[int] = None
```
vnlitvinov, line 142: ideally it should be an enum I think

modin/config/pubsub.py
```diff
@@ -70,7 +70,7 @@ class ExactStr(str):
     ),
     int: TypeDescriptor(
         decode=lambda value: int(value.strip()),
-        normalize=int,
+        normalize=int,  # type: ignore
```
jeffreykennethli, line 73: could you explain the reasoning for the ignore here?

modin/config/pubsub.py
```diff
@@ -135,11 +135,14 @@ class Parameter(object):
         ``ValueSource``.
     """
 
-    choices: typing.Sequence[str] = None
+    choices: Optional[Sequence[str]] = None
     type = str
-    default = None
+    default: Optional[Any] = None
     is_abstract = True
```
jeffreykennethli, line 141: does `is_abstract` need a type hint `bool`?

jeffreykennethli (approved):

modin/config/__main__.py
```diff
@@ -57,7 +57,9 @@ def export_config_help(filename: str):
                 if obj.__name__ != "RayRedisPassword"
                 else "random string",
                 # `Notes` `-` underlining can't be correctly parsed inside csv table by sphinx
-                "Description": dedent(obj.__doc__).replace("Notes\n-----", "Notes:\n"),
+                "Description": dedent(obj.__doc__).replace("Notes\n-----", "Notes:\n")
+                if obj.__doc__
+                else "",
```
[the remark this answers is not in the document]
pyrito: Ha! Learned something new about python today

modin/config/pubsub.py
pyrito: It is expected, but it looks like `object` is not enough for `mypy` to be satisfied. I wasn't sure how to get around this.

modin/config/pubsub.py
```diff
@@ -224,7 +227,7 @@ def _get_default(cls):
         return cls.default
 
     @classmethod
-    def get_value_source(cls):
+    def get_value_source(cls) -> Union[int, None]:
```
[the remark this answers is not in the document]
pyrito: Yes, that was a brainfart moment

modin/config/pubsub.py
```diff
@@ -35,9 +35,9 @@ class TypeDescriptor(typing.NamedTuple):
         Class description string.
     """
 
-    decode: typing.Callable[[str], object]
-    normalize: typing.Callable[[object], object]
-    verify: typing.Callable[[object], bool]
+    decode: Callable[[str], object]
+    normalize: Callable[[object], object]
```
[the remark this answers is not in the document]
pyrito: `Any` would disregard the return type of the function. `object` is somewhat more specific, so wouldn't that be better?

modin/config/pubsub.py
pyrito: Technically yes. However I think `mypy` still treats it as an int, so it complains when we try to assign enum values to `_value_source` when it is defined as `Optional[ValueSource]`

modin/config/pubsub.py
pyrito: It didn't need the type hint, I think it was able to figure it out on its own.

modin/config/pubsub.py
pyrito: Yes, this was the error that I was getting:
```
modin/config/pubsub.py:73:19: error: Argument "normalize" to "TypeDescriptor" has incompatible type "Type[int]"; expected "Callable[[object], object]"  [arg-type]
```

I think `mypy` is assuming `int` to be a type instead of using the hint to make use of it as a function to cast something to an int. I tried to make it more explicit by writing something like this:
```python
normalize=lambda value: int(value)
```
but then I got hit with these warnings:
```
modin/config/pubsub.py:73:33: error: No overload variant of "int" matches argument type "object"  [call-overload]
modin/config/pubsub.py:73:33: note: Possible overload variants:
modin/config/pubsub.py:73:33: note:     def int(cls, Union[str, bytes, array[Any], mmap, _CData, PickleBuffer, SupportsInt, SupportsIndex, SupportsTrunc] = ...) -> int
modin/config/pubsub.py:73:33: note:     def int(cls, Union[str, bytes], base: SupportsIndex) -> int
```

modin/config/pubsub.py
```diff
@@ -135,11 +135,14 @@ class Parameter(object):
         ``ValueSource``.
     """
 
-    choices: typing.Sequence[str] = None
+    choices: Optional[Sequence[str]] = None
```
noloerino, line 138: Should fields like this that are now optional also be reflected in the docstring?

modin/config/pubsub.py
noloerino: Or an [`IntEnum`](https://docs.python.org/3/library/enum.html#enum.IntEnum) if you still want it to be an int value.

modin/config/pubsub.py
pyrito: @noloerino yes thank you for pointing this out!

modin/config/pubsub.py
mvashishtha: I see the input type of `normalize` in `normalize: Callable[[object], object]`. Is that why?

pyrito: I guess a cop-out would be to do something like `Callable[[Any], object]`

modin/config/pubsub.py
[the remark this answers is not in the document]
vnlitvinov: I wonder if there's any real difference, as in some cases `decode` can e.g. return a `bool` - is that considered an `object`?..

modin/config/__main__.py
```diff
@@ -57,9 +57,9 @@ def export_config_help(filename: str) -> None:
                 if obj.__name__ != "RayRedisPassword"
                 else "random string",
                 # `Notes` `-` underlining can't be correctly parsed inside csv table by sphinx
-                "Description": dedent(obj.__doc__).replace("Notes\n-----", "Notes:\n")
-                if obj.__doc__
-                else "",
+                "Description": dedent(obj.__doc__ or "").replace(
+                    "Notes\n-----", "Notes:\n"
+                ),
```
vnlitvinov, line 62: huh, this didn't make the code any shorter because of `black`... I wonder which version looks easier to read, maybe the original one is easier

modin/config/pubsub.py
vnlitvinov: can we get this done as part of this PR?

modin/config/pubsub.py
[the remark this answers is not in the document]
pyrito: There is a difference because `object` is a real supertype where as using `Any` will allow any kind of type mess. Returning a `bool` is considered an `object`.

modin/config/pubsub.py
pyrito: @vnlitvinov yes I can!

modin/config/pubsub.py
```diff
@@ -224,21 +228,21 @@ def _get_default(cls):
         return cls.default
 
     @classmethod
-    def get_value_source(cls):
+    def get_value_source(cls) -> Optional[ValueSource]:
```
vnlitvinov, line 231: I wonder why it's optional here, it shouldn't be returning `None`

pyrito: It's possible that if we try to call `get` and `cls._value` is not `_UNSET`, we can still have `cls._value_source` stay as `None` right? That's why `mypy` complained about it earlier.

mvashishtha (approved):

modin/config/pubsub.py
vnlitvinov: This feels like a bug in logic somewhere... if `cls._value` is set to something, its value source should be defined.

I think `mypy` complains here because `._value` is not a _private_ attribute, so in theory someone can set it from outside.
Maybe we should `assert cls._value_source is not None` before returning, so that `get_value_source()` returns non-optional?

pyrito: Yeah I think that makes sense. While we can have such a case, it's better that we catch and report the error via an assert instead of allowing that kind of behavior.

modin/config/pubsub.py
```diff
@@ -228,17 +228,18 @@ def _get_default(cls) -> Any:
...
 
         Returns
         -------
-        Optional[ValueSource]
+        ValueSource
         """
         if cls._value_source is None:
             # dummy call to .get() to initialize the value
             cls.get()
+        assert cls._value_source is not None, "_value_source should not be set to None"
```
mvashishtha, line 242: The comment is saying the same thing as the assertion. If you understand the context enough to give a reason why `_value_source` is `None`, I think you should add that.

jeffreykennethli (approved):

modin/config/pubsub.py
vnlitvinov:
```suggestion
        assert cls._value_source is not None, "_value_source must be initialized by now"
```

vnlitvinov (dismissed):
LGTM

mvashishtha (dismissed):

mvashishtha (approved):

mvashishtha:
@pyrito thank you very much for this improvement to the codebase!

PR #4697, opened 2022-07-21
FIX-#4686: Propagate metadata and drain call queue in unwrap_partitions.

### LINKED ISSUE #4686: Column rename operation fails when converting from Modin DataFrame to Ray Dataset
### System information
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: MacOS 12.3.1
- **Modin version** (`modin.__version__`): 0.15.2
- **Python version**: 3.8.13
- **Code we can use to reproduce**: https://gist.github.com/geoffreyangus/5e9fdfd711d9776e76053840fb332274

### Describe the problem

I am attempting to do the following:

1. Convert Modin Series into a Modin DataFrame, while changing the column name using the `to_frame(name="value")`
2. Convert the Modin DataFrame to a Ray Dataset
3. run `map_batches` using the new column name (in this example, "value")
4. Convert the Ray Dataset back to a Modin DataFrame

I am getting an error in step (3) because "value" cannot be found in the Pandas DataFrame objects operated upon in the `map_batches` call.

### Expected Behavior

I expect column name assignment when calling `to_frame(name=<NAME>)`.

Interestingly enough, if the DataFrame is materialized beforehand (using `DataFrame._to_pandas()`), the column name is correctly renamed as a side effect.

### Source code / logs

A minimal repro can be found in the Gist above.

mvashishtha:

## What do these changes do?

FIX-#4686: Propagate metadata and drain call queue in unwrap_partitions.

- [x] Resolves #4686 
- [x] tests added and passing

### DIFF
--- a/modin/distributed/dataframe/pandas/partitions.py
+++ b/modin/distributed/dataframe/pandas/partitions.py
@@ -53,6 +53,11 @@ def unwrap_partitions(api_layer_object, axis=None, get_ip=False):
     if axis is None:
 
         def _unwrap_partitions():
+            api_layer_object._query_compiler._modin_frame._propagate_index_objs(None)
+            [
+                p.drain_call_queue()
+                for p in api_layer_object._query_compiler._modin_frame._partitions.flatten()
+            ]
             if get_ip:
                 return [
                     [(partition._ip_cache, partition._data) for partition in row]
@@ -76,6 +81,7 @@ def unwrap_partitions(api_layer_object, axis=None, get_ip=False):
             f"Do not know how to unwrap '{actual_engine}' underlying partitions"
         )
     else:
+        api_layer_object._query_compiler._modin_frame._propagate_index_objs(None)
         partitions = api_layer_object._query_compiler._modin_frame._partition_mgr_cls.axis_partition(
             api_layer_object._query_compiler._modin_frame._partitions, axis ^ 1
         )
--- a/modin/distributed/dataframe/pandas/partitions.py
+++ b/modin/distributed/dataframe/pandas/partitions.py
@@ -50,28 +50,24 @@ def unwrap_partitions(api_layer_object, axis=None, get_ip=False):
             f"Only API Layer objects may be passed in here, got {type(api_layer_object)} instead."
         )
 
+    modin_frame = api_layer_object._query_compiler._modin_frame
     if axis is None:
 
         def _unwrap_partitions():
-            api_layer_object._query_compiler._modin_frame._propagate_index_objs(None)
-            [
-                p.drain_call_queue()
-                for p in api_layer_object._query_compiler._modin_frame._partitions.flatten()
-            ]
+            modin_frame._propagate_index_objs(None)
+            [p.drain_call_queue() for p in modin_frame._partitions.flatten()]
             if get_ip:
                 return [
                     [(partition._ip_cache, partition._data) for partition in row]
-                    for row in api_layer_object._query_compiler._modin_frame._partitions
+                    for row in modin_frame._partitions
                 ]
             else:
                 return [
                     [partition._data for partition in row]
-                    for row in api_layer_object._query_compiler._modin_frame._partitions
+                    for row in modin_frame._partitions
                 ]
 
-        actual_engine = type(
-            api_layer_object._query_compiler._modin_frame._partitions[0][0]
-        ).__name__
+        actual_engine = type(modin_frame._partitions[0][0]).__name__
         if actual_engine in (
             "PandasOnRayDataframePartition",
             "PandasOnDaskDataframePartition",
@@ -81,9 +77,9 @@ def unwrap_partitions(api_layer_object, axis=None, get_ip=False):
             f"Do not know how to unwrap '{actual_engine}' underlying partitions"
         )
     else:
-        api_layer_object._query_compiler._modin_frame._propagate_index_objs(None)
-        partitions = api_layer_object._query_compiler._modin_frame._partition_mgr_cls.axis_partition(
-            api_layer_object._query_compiler._modin_frame._partitions, axis ^ 1
+        modin_frame._propagate_index_objs(None)
+        partitions = modin_frame._partition_mgr_cls.axis_partition(
+            modin_frame._partitions, axis ^ 1
         )
         return [
             part.force_materialization(get_ip=get_ip).unwrap(
--- a/modin/distributed/dataframe/pandas/partitions.py
+++ b/modin/distributed/dataframe/pandas/partitions.py
@@ -51,10 +51,10 @@ def unwrap_partitions(api_layer_object, axis=None, get_ip=False):
         )
 
     modin_frame = api_layer_object._query_compiler._modin_frame
+    modin_frame._propagate_index_objs(None)
     if axis is None:
 
         def _unwrap_partitions():
-            modin_frame._propagate_index_objs(None)
             [p.drain_call_queue() for p in modin_frame._partitions.flatten()]
             if get_ip:
                 return [
@@ -79,7 +79,6 @@ def unwrap_partitions(api_layer_object, axis=None, get_ip=False):
             f"Do not know how to unwrap '{actual_engine}' underlying partitions"
         )
     else:
-        modin_frame._propagate_index_objs(None)
         partitions = modin_frame._partition_mgr_cls.axis_partition(
             modin_frame._partitions, axis ^ 1
         )
--- a/modin/distributed/dataframe/pandas/partitions.py
+++ b/modin/distributed/dataframe/pandas/partitions.py
@@ -15,6 +15,12 @@
 
 import numpy as np
 
+from modin.core.execution.ray.implementations.pandas_on_ray.partitioning import (
+    PandasOnRayDataframePartition,
+)
+from modin.core.execution.dask.implementations.pandas_on_dask.partitioning.partition import (
+    PandasOnDaskDataframePartition,
+)
 from modin.core.storage_formats.pandas.query_compiler import PandasQueryCompiler
 from modin.pandas.dataframe import DataFrame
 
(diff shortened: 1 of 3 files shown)


modin/test/test_partition_api.py
```diff
@@ -54,27 +54,50 @@
 
 
 @pytest.mark.parametrize("axis", [None, 0, 1])
-def test_unwrap_partitions(axis):
+@pytest.mark.parametrize("reset_index", [True, False])
+@pytest.mark.parametrize("reset_columns", [True, False])
+def test_unwrap_partitions(axis, reset_index, reset_columns):
```
vnlitvinov, line 59: maybe we should name those `reverse_index` and `reverse_columns` to better reflect what they're doing?

modin/test/test_partition_api.py
```diff
@@ -54,27 +54,50 @@
...
+    df = get_df(data)
+    # Propagate any deferred updates to the partitions and drain call
+    # queues before getting the data.
+    expected_df = get_df(data)
+    expected_df._query_compiler._modin_frame._propagate_index_objs(None)
+    [
+        p.drain_call_queue()
+        for p in expected_df._query_compiler._modin_frame._partitions.flatten()
+    ]
+    expected_partitions = expected_df._query_compiler._modin_frame._partitions
```
vnlitvinov, line 84: Maybe we should better create this "reversed" df by building then reversing it in pandas and then converting to Modin? For now this test seems to be repeating what is done in the code, what is the point of checking that?

vnlitvinov (commented):
Please rebase taking https://github.com/modin-project/modin/pull/4699 to fix CI failures

modin/test/test_partition_api.py
mvashishtha: Done. I agree that that's better. I was uncomfortable doing the same thing in the unit test and the code, too.

mvashishtha (commented):
@vnlitvinov 

> Please rebase taking https://github.com/modin-project/modin/pull/4699 to fix CI failures

Done.

modin/distributed/dataframe/pandas/partitions.py
```diff
@@ -50,23 +50,24 @@ def unwrap_partitions(api_layer_object, axis=None, get_ip=False):
             f"Only API Layer objects may be passed in here, got {type(api_layer_object)} instead."
         )
 
+    modin_frame = api_layer_object._query_compiler._modin_frame
     if axis is None:
 
         def _unwrap_partitions():
+            modin_frame._propagate_index_objs(None)
+            [p.drain_call_queue() for p in modin_frame._partitions.flatten()]
```
RehanSD, line 58: Shouldn't we drain call queue before propagating index objs? (trying to future proof as we begin to explore more lazy eval)

modin/test/test_partition_api.py
```diff
@@ -54,27 +54,42 @@
...
+    def get_df(lib, data):
+        df = lib.DataFrame(data)
+        if reverse_index:
+            df.index = df.index[::-1]
+        if reverse_columns:
+            df.columns = df.columns[::-1]
+        return df
+
+    df = get_df(pd, data)
+    # `df` should not have propagatated the index and column updates to its
```
RehanSD, line 71:
```suggestion
    # `df` should not have propagated the index and column updates to its
```

RehanSD (commented):
Left 2 quick questions, but lgtm!

modin/distributed/dataframe/pandas/partitions.py
```diff
@@ -50,23 +50,24 @@ def unwrap_partitions(api_layer_object, axis=None, get_ip=False):
...
                 return [
                     [partition._data for partition in row]
-                    for row in api_layer_object._query_compiler._modin_frame._partitions
+                    for row in modin_frame._partitions
                 ]
 
-        actual_engine = type(
-            api_layer_object._query_compiler._modin_frame._partitions[0][0]
-        ).__name__
+        actual_engine = type(modin_frame._partitions[0][0]).__name__
```
jeffreykennethli, line 70: Would this be better if it compared just `type(modin_frame._partitions[0][0])` to the actual types, as opposed to the strings?

jeffreykennethli (dismissed):

modin/distributed/dataframe/pandas/partitions.py
mvashishtha: `_propagate_index_objs` [adds](https://github.com/modin-project/modin/blob/88c3c33cf9b8aab5f5b7077555bea9f2d8466534/modin/core/dataframe/pandas/dataframe/dataframe.py#L500-L544) the index updates to the call queue, so we should call `_propagate_index_objs` first.

modin/distributed/dataframe/pandas/partitions.py
mvashishtha: Done.

modin/distributed/dataframe/pandas/partitions.py
```diff
@@ -15,6 +15,12 @@
 
 import numpy as np
 
+from modin.core.execution.ray.implementations.pandas_on_ray.partitioning import (
+    PandasOnRayDataframePartition,
+)
+from modin.core.execution.dask.implementations.pandas_on_dask.partitioning.partition import (
+    PandasOnDaskDataframePartition,
+)
```
vnlitvinov, line 23: this feels wrong, it would make _both_ `ray` and `dask` hard requirements installed before running things regardless of which engine we plan to use

modin/distributed/dataframe/pandas/partitions.py
```diff
@@ -50,34 +56,36 @@ def unwrap_partitions(api_layer_object, axis=None, get_ip=False):
...
-        actual_engine = type(
-            api_layer_object._query_compiler._modin_frame._partitions[0][0]
-        ).__name__
-        if actual_engine in (
-            "PandasOnRayDataframePartition",
-            "PandasOnDaskDataframePartition",
+        partition_type = modin_frame._partitions[0][0]
+        if isinstance(
+            modin_frame._partitions[0][0],
+            (PandasOnRayDataframePartition, PandasOnDaskDataframePartition),
```
vnlitvinov, line 79: I suggest keeping the original logic of checking class name

vnlitvinov (changes requested):

modin/distributed/dataframe/pandas/partitions.py
mvashishtha: Reverted.

modin/test/test_partition_api.py
```diff
@@ -54,27 +54,42 @@
...
+    # `df` should not have propagated the index and column updates to its
+    # partitions yet. The partitions of `expected_df` should have the updated
+    # metadata because we construct `expected_df` directly from the updated
+    # pandas dataframe.
+    expected_df = pd.DataFrame(get_df(pandas, data))
+    expected_partitions = expected_df._query_compiler._modin_frame._partitions
     if axis is None:
-        expected_partitions = df._query_compiler._modin_frame._partitions
         actual_partitions = np.array(unwrap_partitions(df, axis=axis))
         assert (
```
RehanSD, line 79: Couldn't this assert just check that shape == shape rather than checking each axis of shape individually?

modin/test/test_partition_api.py
```diff
@@ -54,27 +54,42 @@
...
             expected_partitions.shape[0] == actual_partitions.shape[0]
             and expected_partitions.shape[1] == expected_partitions.shape[1]
         )
         for row_idx in range(expected_partitions.shape[0]):
             for col_idx in range(expected_partitions.shape[1]):
-                assert (
-                    expected_partitions[row_idx][col_idx]._data
-                    == actual_partitions[row_idx][col_idx]
+                df_equals(
+                    get_func(expected_partitions[row_idx][col_idx]._data),
```
RehanSD, line 86: Is this an info leak? We should probably have a group discussion about layers of abstraction wrt `_data` since this has come up in other PRs too

RehanSD (commented):
Left some questions - LGTM otherwise!

modin/test/test_partition_api.py
mvashishtha: Yes, this is an information leak. We can track it in #4530.

modin/test/test_partition_api.py
mvashishtha: Sure, I can fix that while I'm here. Done.

modin/distributed/dataframe/pandas/partitions.py
```diff
@@ -50,18 +50,21 @@ def unwrap_partitions(api_layer_object, axis=None, get_ip=False):
             f"Only API Layer objects may be passed in here, got {type(api_layer_object)} instead."
         )
 
+    modin_frame = api_layer_object._query_compiler._modin_frame
     if axis is None:
 
         def _unwrap_partitions():
+            modin_frame._propagate_index_objs(None)
```
YarShev, line 57:
```suggestion
    modin_frame = api_layer_object._query_compiler._modin_frame
    modin_frame._propagate_index_objs(None)
    if axis is None:

        def _unwrap_partitions():
```

modin/distributed/dataframe/pandas/partitions.py
```diff
@@ -76,8 +79,9 @@ def _unwrap_partitions():
             f"Do not know how to unwrap '{actual_engine}' underlying partitions"
         )
     else:
-        partitions = api_layer_object._query_compiler._modin_frame._partition_mgr_cls.axis_partition(
-            api_layer_object._query_compiler._modin_frame._partitions, axis ^ 1
+        modin_frame._propagate_index_objs(None)
```
YarShev, line 82:
```suggestion
```

modin/test/test_partition_api.py
```diff
@@ -54,27 +54,39 @@
...
+    expected_df = pd.DataFrame(get_df(pandas, data))
+    expected_partitions = expected_df._query_compiler._modin_frame._partitions
     if axis is None:
-        expected_partitions = df._query_compiler._modin_frame._partitions
         actual_partitions = np.array(unwrap_partitions(df, axis=axis))
-        assert (
-            expected_partitions.shape[0] == actual_partitions.shape[0]
-            and expected_partitions.shape[1] == expected_partitions.shape[1]
-        )
+        expected_partitions.shape == expected_partitions.shape
```
YarShev, line 79:
```suggestion
        expected_partitions.shape == actual_partitions.shape
```

modin/test/test_partition_api.py
```diff
@@ -54,27 +54,39 @@
...
+    expected_df = pd.DataFrame(get_df(pandas, data))
+    expected_partitions = expected_df._query_compiler._modin_frame._partitions
     if axis is None:
-        expected_partitions = df._query_compiler._modin_frame._partitions
         actual_partitions = np.array(unwrap_partitions(df, axis=axis))
-        assert (
-            expected_partitions.shape[0] == actual_partitions.shape[0]
-            and expected_partitions.shape[1] == expected_partitions.shape[1]
-        )
+        expected_partitions.shape == actual_partitions.shape
```
RehanSD, line 79: This is missing an assert!

RehanSD (commented):
Left a quick comment but LGTM otherwise!

modin/test/test_partition_api.py
mvashishtha: Fixed.

YarShev (approved):
LGTM!

Issue #1366, opened 2020-04-15
KeyError(f"None of [{key}] are in the [{axis_name}]")


jadhosn:
### System information
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**:
- **Modin version** (`modin.__version__`): 0.7.2
- **Python version**: 3.7
- **Code we can use to reproduce**:
```
import pandas as pd
import modin.pandas as mpd

df = pd.DataFrame({'idx': [1,4,5], 'A': [4,5,6], 'B':['a','b','c']})
df = df.set_index("idx")
# write out parquet with index using pandas
df.to_parquet("./data/tmp.parquet")
# read the same parquet using modin.pandas
mdf = mpd.read_parquet("./data/tmp.parquet")
```
returns 
>"None of [Index(['idx'], dtype='object')] are in the [columns]"

### Describe the problem

Parquet files that are written using Pandas, throws this error when read using modin's `read_parquet()`

### Source code / logs

> KeyError: "None of [Index(['idx'], dtype='object')] are in the [columns]"

devin-petersohn:
Hi @jadhosn thanks for the report!

This looks like a bug specific to setting the index, I will dig into this, thanks again for the report!

jadhosn:
@devin-petersohn thank you! 
In the meantime, which of the following workarounds would you recommend to use to still gain an efficiency boost using modin:
* Write out the original parquet file with `idx` as a column, read the parquet file using modin, then `set_index` using modin
```
import pandas as pd
import modin.pandas as mpd

df = pd.DataFrame({'idx': [1,4,5], 'A': [4,5,6], 'B':['a','b','c']})
# write out pandas df with idx as column
df.to_parquet("./data/tmp.parquet")

# read the parquet file using modin
mdf = mpd.read_parquet("./data/tmp.parquet")

# set index
mdf = mdf.set_index("idx")
```
* Read the indexed parquet file using pandas, transform the pandas dataframe into a modin dataframe, then proceed with workflow
```
import pandas as pd
import modin.pandas as mpd

df = pd.DataFrame({'idx': [1,4,5], 'A': [4,5,6], 'B':['a','b','c']})
df = df.set_index("idx")

# write out pandas df with idx as column
df.to_parquet("./data/tmp.parquet")

# read the parquet file using pandas
df = pd.read_parquet("./data/tmp.parquet") 

# transform into modin dataframe
mdf = mpd.DataFrame(df)
```

[...]

PR #4700, opened 2022-07-21
PERF-#4305: Parallelize `read_parquet` over row groups

### LINKED ISSUE #4305: Add parallel implementation for `read_parquet` over rows
### System information
- **Modin installed from (source or binary)**:  source
- **Modin version**: e67db1dd3ee448eec23a05ffdb3ca38a2b2b709e
- **Python version**: 3.8.11

<details><summary>Benchmark code</summary>

```python
import time

import numpy as np
import pandas as pd

import modin.pandas as mpd
import modin.config as cfg
import ray

cfg.BenchmarkMode.put(True)
ray.init()

def generate_data(path, nrows=5000, n_random_cols=100):
    data = {f"col{i}":  np.random.rand(nrows) for i in range(n_random_cols)}

    df = mpd.DataFrame(data)
    df.to_parquet(path)


if __name__ == "__main__":

    path = "/localdisk/aprutsko/modin/data.parquet"
    generate_data(path, nrows=5_000_000_000, n_random_cols=1)

    t = time.time()
    pdf = pd.read_parquet(path)
    print(f"pandas read_parquet time: {time.time() - t} s")

    t = time.time()
    mdf = mpd.read_parquet(path)
    print(f"modin read_parquet time: {time.time() - t} s")

    print(f"Original shape: {pdf.shape}")

```

</details>

### Describe the problem

Modin `read_parquet` does't have a parallel implementation over row groups.

In the table below performance comparison Modin vs pandas vs bodo for `read_parquet` is shown.
| Shape(rows, cols) | (100k, 1), s | (1m, 1), s | (10m, 1), s | (100m, 1), s | (1b, 1), s | (5b, 1), s |
| ----------- | ----------- | ----------- | ----------- | ----------- | ----------- |----------- |
| pandas | 0.0247 |0.0459|0.19805|1.4098|15.0055|82.6288|
| modin |0.0988  |0.1807|0.4486|2.0881|21.6647|138.677|
| bodo (including compile-time)|3.329|3.3411|3.3052|3.3854| 4.5883|14.0036|

As a start point of row-parallel implementation we can try to use `row groups` from parquet metadata or some other metainfo.

pyrito:
## What do these changes do?

This PR introduces optimizations to `read_parquet` by parallelizing reads over row groups. The following table outlines the **speedups over pandas** when we test with datasets that have many row groups.

|          **Shape (rows, cols)**         | **(50 million, 1)** | **(500 million, 1)** | **(750 million, 1)** | **(900 million, 1)** |
|:---------------------------------------:|:-------------------:|:--------------------:|:--------------------:|:--------------------:|
|              modin (0.15.2)             |        1.383        |         0.784        |         0.799        |         0.713        |
| modin (with read-group parallelization) |        3.760        |         2.113        |         2.268        |         2.064        |

You can see the scripts used for benchmarking here: https://github.com/pyrito/parquet-benchmarks (see `run.sh` for what was run and `benchmark.py` for specifics). Everything was run on a t2.2xlarge AWS instance (8 vCPUs and 32 GB memory). 

The majority of the implementation was spearheaded by the one and only @mvashishtha. My main responsibilities were to benchmark our implementation and compare it with pandas as well as shepherd this PR across the finish line (hopefully). 

- [x] Resolves #4305 
- [x] tests added and passing


modin/core/io/column_stores/parquet_dispatcher.py
```diff
@@ -23,6 +37,317 @@
...
+        new_query_compiler : BaseQueryCompiler
+            Query compiler with imported data for further processing.
+        """
+        col_partitions, column_widths = cls.build_columns(columns)
+        partition_ids = cls.call_deploy(path, col_partitions, **kwargs)
+        index, needs_index_sync = cls.build_index(path, kwargs["storage_options"])
+        remote_parts = cls.build_partition(partition_ids, column_widths)
+        if len(partition_ids) > 0 and len(partition_ids[0]) > 0:
+            first_row = partition_ids[0]
+            dtypes = cls.build_dtypes(
```
vnlitvinov, line 331: what if dtypes are different in other partitions?..

modin/core/io/column_stores/parquet_dispatcher.py
```diff
@@ -23,6 +37,317 @@
...
+            index,
+            columns,
+            row_lengths=row_lengths,
+            column_widths=column_widths,
+            dtypes=dtypes,
+        )
+        new_query_compiler = cls.query_compiler_cls(frame)
+        if needs_index_sync:
+            frame.synchronize_labels(axis=0)
+        return new_query_compiler
```
vnlitvinov, line 349: I would rather swap the statements (for mostly aesthetical purposes of finishing `frame` construction before building a query compiler from it), and do smth like
```suggestion
        if needs_index_sync:
            frame.synchronize_labels(axis=0)
        return cls.query_compiler_cls(frame)
```

modin/core/io/column_stores/parquet_dispatcher.py
```diff
@@ -15,6 +15,20 @@
 
 import os
 
+import fsspec
+import fsspec.core
+from fsspec.core import split_protocol
+import io
+from fsspec.registry import get_filesystem_class
```
vnlitvinov, line 22: let's not mix imports from modules (import stuff from `fsspec` without injecting `import io` between):
```suggestion
import io

import fsspec
import fsspec.core
from fsspec.core import split_protocol
from fsspec.registry import get_filesystem_class
```

also I would like to make imports in 3 different groups separated by an empty line:
1. builtin and stdlib (`os` and `io` in this case)
2. installed as dependency (`fsspec`, `pandas`, etc.)
3. Modin modules (`modin.config`, etc.)

modin/core/storage_formats/pandas/parsers.py
```diff
@@ -635,13 +638,31 @@ def parse(fname, **kwargs):
...
+    row_group_end: int
+
+
 @doc(_doc_pandas_parser_class, data_type="PARQUET data")
 class PandasParquetParser(PandasParser):
     @staticmethod
     @doc(_doc_parse_func, parameters=_doc_parse_parameters_common)
     def parse(fname, **kwargs):
         num_splits = kwargs.pop("num_splits", None)
-        columns = kwargs.get("columns", None)
```
vnlitvinov, line 644: why remove, isn't it used later on?

modin/core/storage_formats/pandas/parsers.py
```diff
@@ -653,9 +674,49 @@ def parse(fname, **kwargs):
...
+        for file_for_parser in files_for_parser:
+            if isinstance(file_for_parser.path, IOBase):
+                context = nullcontext(file_for_parser.path)
+            else:
+                context = fsspec.open(file_for_parser.path, **storage_options)
+            with context as f:
+                chunk = (
+                    ParquetFile(f)
+                    .read_row_groups(
+                        list(
```
vnlitvinov, line 702: why do you want to materialize `range()` here? in most cases it just eats memory and slows down code

modin/core/storage_formats/pandas/parsers.py
```diff
@@ -653,9 +674,49 @@ def parse(fname, **kwargs):
...
+                            range(
+                                file_for_parser.row_group_start,
+                                file_for_parser.row_group_end,
+                            )
+                        ),
+                        columns=columns,
+                        use_pandas_metadata=True,
+                    )
+                    .to_pandas()
```
vnlitvinov, line 711: why?

modin/core/storage_formats/pandas/parsers.py
```diff
@@ -653,9 +674,49 @@ def parse(fname, **kwargs):
...
+                )
+            columns = [
+                c for c in columns if c not in chunk.index.names and c in chunk.columns
+            ]
+            chunks.append(chunk)
+        df = pandas.concat(chunks)
```
vnlitvinov, line 717: why?

modin/core/storage_formats/pandas/parsers.py
```diff
@@ -653,9 +674,49 @@ def parse(fname, **kwargs):
...
+                                file_for_parser.row_group_end,
+                            )
+                        ),
+                        columns=columns,
+                        use_pandas_metadata=True,
+                    )
+                    .to_pandas()
+                )
+            columns = [
+                c for c in columns if c not in chunk.index.names and c in chunk.columns
```
vnlitvinov, line 714: this reduced `columns` (potentially) for each chunk, and each subsequent chunk receives (potentially) smaller amount of columns, is this desired?

modin/pandas/test/test_io.py
```diff
@@ -199,6 +200,26 @@ def eval_to_file(modin_obj, pandas_obj, fn, extension, **fn_kwargs):
...
+def make_parquet_dir():
+    list_of_path: list[str] = []
+
+    def _make_parquet_dir(
+        path: str, dfs_by_filename: Dict[str, pandas.DataFrame], row_group_size: int
+    ):
+        if os.path.exists(path):
+            shutil.rmtree(path)
+        else:
+            os.makedirs(path)
```
vnlitvinov, line 213:
```suggestion
        os.makedirs(path)
```
otherwise if `path` exists you'd wipe it but won't make an empty directory to save to

modin/pandas/test/test_io.py
```diff
@@ -199,6 +200,26 @@ def eval_to_file(modin_obj, pandas_obj, fn, extension, **fn_kwargs):
...
+        for filename, df in dfs_by_filename.items():
+            df.to_parquet(os.path.join(path, filename), row_group_size=row_group_size)
+        list_of_path.append(path)
+
+    yield _make_parquet_dir
+
+    shutil.rmtree(list_of_path[0])
```
vnlitvinov, line 220: This appending (to the end) and removal of the first (but not removing this first item from the list) all looks really weird.
What was wrong with just `shutil.rmtree(path)` _without_ keeping a `list_of_path` variable around?..

modin/core/io/column_stores/parquet_dispatcher.py
```diff
@@ -23,6 +37,317 @@
...
+            Paramters for specific storage engine.
+        **kwargs : dict
+            Parameters of deploying read_* function.
+
+        Returns
+        -------
+        np.ndarray
+            Array with references to the task deploy result for each partition.
+        """
+        from modin.core.storage_formats.pandas.parsers import ParquetFileToRead
```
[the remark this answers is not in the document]
pyrito: No particular reason. We just thought it would be cleaner to encapsulate it in the function that was needed, but can move it back if needed.

pyrito:
Thanks for the comments @vnlitvinov , working through them now :)

modin/core/io/column_stores/parquet_dispatcher.py
```diff
@@ -23,6 +37,317 @@
...
+            Array with references to the partitions data.
+        columns : list
+            List of columns that should be read from file.
+
+        Returns
+        -------
+        dtypes : pandas.Series
+            Series with dtypes for columns.
+        """
+        dtypes = pandas.concat(cls.materialize(list(partition_ids)), axis=0)
```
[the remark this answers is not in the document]
pyrito: The "magic" of getting the `dtypes` comes from the `parse_fsspec_files`, which returns the dataframe partition, the respective dtypes and length of the partition. These are all returned when we materialize the object reference.

modin/core/io/column_stores/parquet_dispatcher.py
pyrito: The way we are dividing up the partitions right now is by row-group, so we are splitting things horizontally. Thus, just getting the first partition should be enough. This is poorly named, and I will push some changes that reflect a better naming convention.

modin/core/io/column_stores/parquet_dispatcher.py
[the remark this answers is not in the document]
pyrito: What we pass into this function is a list of these dtypes, which we concatenate and get for the entire resulting DataFrame.

modin/core/io/column_stores/parquet_dispatcher.py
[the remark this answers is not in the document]
pyrito: There was a circular dependency with the imports otherwise.

modin/core/storage_formats/pandas/parsers.py
pyrito: I think I should actually get rid of this entire function altogether since it won't be used anymore.

modin/core/storage_formats/pandas/parsers.py
pyrito: We are trying to parallelize by reading over row groups so this function is only called when the partition is being materialized. We need to do this to get the resulting dataframe.

modin/core/storage_formats/pandas/parsers.py
pyrito: See above - does it make sense as to what we are doing here? Did you have a different approach in mind?

modin/core/io/column_stores/parquet_dispatcher.py
pyrito: This is the assumption, but we may have to revisit this.

modin/core/io/column_stores/parquet_dispatcher.py
```diff
@@ -23,6 +37,317 @@
...
+            index_columns = pandas_metadata.get("index_columns", index_columns)
+        column_names_to_read = []
+        for column in index_columns:
+            if isinstance(column, str):
+                column_names_to_read.append(column)
+            elif column["name"] is not None:
+                column_names_to_read.append(column["name"])
+        complete_index = (
+            read_table(path, columns=column_names_to_read).to_pandas().index
+        )
```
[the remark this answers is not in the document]
pyrito: No, this will not read the entire dataset. This will only read in columns that are index columns. There are cases where parquet will have a separate column that is the index column I believe.

modin/core/io/column_stores/parquet_dispatcher.py
pyrito: @vnlitvinov I think you may be right on this - I think it may not be worth even trying to compute dtypes ahead of time and instead lazily compute them when needed.

modin/pandas/test/test_io.py
mvashishtha: I copied this pattern from the existing [`make_parquet_file`](https://github.com/modin-project/modin/blob/71815691339ea5a61be4c2f3e448fd51ea97b77e/modin/conftest.py#L437). The inner function that we `yield` has to communicate the `path` up to the `make_parquet_dir` for cleanup. The simplest way I could think of at the time was to mutate a list that's in the `make_parquet_dir` scope.

pyrito: I'm not entirely sure.  But looking at this now I think we can just use something like `tempdir` for all of this logic.

modin/core/storage_formats/pandas/parsers.py
mvashishtha: @vnlitvinov I don't see any other way to [read](https://arrow.apache.org/docs/python/generated/pyarrow.parquet.ParquetFile.html#pyarrow.parquet.ParquetFile.read_row_groups) a range of row groups from a parquet file. It seems very unlikely that the number of groups in a single partition will be over 1000, so I think this will be fast enough.

modin/core/io/column_stores/parquet_dispatcher.py
```diff
@@ -23,6 +37,317 @@
...
+        return complete_index, len(index_columns) == 0 or any(
+            not isinstance(c, str) for c in index_columns
+        )
+
+    @classmethod
+    def build_columns(cls, columns):
```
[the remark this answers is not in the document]
pyrito: Do you remember which one?

modin/core/io/column_stores/parquet_dispatcher.py
```diff
@@ -23,6 +37,292 @@
...
+            Parameters of deploying read_* function.
+
+        Returns
+        -------
+        np.ndarray
+            Array with references to the task deploy result for each partition.
+        """
+        from modin.core.storage_formats.pandas.parsers import ParquetFileToRead
+
+        storage_options = storage_options or {}
```
RehanSD, line 94: Isn't this line unnecessary since storage_options isn't optional? Do we expect folks to pass `None` or other falsey values for storage_options instead of an empty dict?

modin/core/io/column_stores/parquet_dispatcher.py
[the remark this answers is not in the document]
RehanSD: @vnlitvinov the `read_table` here is from pyarrow - so presumably pyarrow can do an efficient read of just the columns we want, and then the `to_pandas` converts it from a pyarrow table to a pandas df, and we grab the index from there.

modin/core/io/column_stores/parquet_dispatcher.py
```diff
@@ -23,6 +37,292 @@
...
+            read_table(path, columns=column_names_to_read).to_pandas().index
+        )
+        return complete_index, len(index_columns) == 0 or any(
+            not isinstance(c, str) for c in index_columns
+        )
+
+    @classmethod
+    def build_columns(cls, columns):
+        """
+        Split columns into chunks that should be read be workers.
```
RehanSD, line 256:
```suggestion
        Split columns into chunks that should be read by workers.
```

modin/core/io/column_stores/parquet_dispatcher.py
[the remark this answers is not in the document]
RehanSD: This is defined by the parent class, and I'm not sure that it makes sense to override, since the code hasn't seemed to change?

modin/core/io/column_stores/parquet_dispatcher.py
```diff
@@ -62,7 +362,7 @@ def _read(cls, path, engine, columns, **kwargs):
             # parquet directories have a unique column at each directory level.
             # Thus, we can use os.walk(), which does a dfs search, to walk
             # through the different columns that the data is partitioned on
-            for (root, dir_names, files) in os.walk(path):
+            for (_, dir_names, files) in os.walk(path):
```
RehanSD, line 365: Was this change made because `root` is no longer used?

modin/core/storage_formats/pandas/parsers.py
```diff
@@ -653,9 +674,49 @@ def parse(fname, **kwargs):
         columns = [c for c in columns if c not in df.index.names and c in df.columns]
         if columns is not None:
             df = df[columns]
-        # Append the length of the index here to build it externally
```
RehanSD, line 656: Why was this removed?

modin/core/storage_formats/pandas/parsers.py
RehanSD: To be clear - the `to_pandas` here is converting from PyArrow to pandas (which we want the partitions to be)

modin/core/storage_formats/pandas/parsers.py
```diff
@@ -653,9 +674,49 @@ def parse(fname, **kwargs):
...
+                        use_pandas_metadata=True,
+                    )
+                    .to_pandas()
+                )
+            columns = [
+                c for c in columns if c not in chunk.index.names and c in chunk.columns
+            ]
+            chunks.append(chunk)
+        df = pandas.concat(chunks)
+        return df, df.dtypes, len(df)
```
RehanSD, line 718: Is there a reason for this return being different than the return of `parse`? Also, if we're not going to be using the code in `parse`, we should just name this function `parse` to keep parity with the other parsers.

modin/core/storage_formats/pandas/parsers.py
RehanSD: Not sure what this does - could someone please clarify? Seems to be removing the columns in the index of the current chunk for subsequent chunks - does that mean that the columns in the index are the ones that are actually materialized in the current chunk, while the ones in the columns list are the ones that are in the other chunks?

modin/pandas/test/test_io.py
```diff
@@ -200,6 +201,22 @@ def eval_to_file(modin_obj, pandas_obj, fn, extension, **fn_kwargs):
...
+
+    def _make_parquet_dir(
+        dfs_by_filename: Dict[str, pandas.DataFrame], row_group_size: int
+    ):
+        for filename, df in dfs_by_filename.items():
+            df.to_parquet(os.path.join(path, filename), row_group_size=row_group_size)
+        return path
+
+    with tempfile.TemporaryDirectory() as tmp_dir:
+        path = tmp_dir
```
RehanSD, line 216: Won't this override the `path` arg specified when calling `make_parquet_dir`?

RehanSD (commented):
Left some questions and comments. Great work @mvashishtha and @pyrito!

modin/core/io/column_stores/parquet_dispatcher.py
pyrito: Yes

modin/conftest.py
```diff
@@ -465,17 +465,14 @@ def _make_parquet_file(
             df = pandas.DataFrame(
                 {f"col{x + 1}": np.arange(nrows) for x in range(ncols)}
             )
-            if directory:
```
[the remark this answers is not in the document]
pyrito: `tempfile.TemporaryDirectory` should handle this for me now.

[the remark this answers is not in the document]
YarShev: `directory` parameter of the function should be removed.

modin/core/io/column_stores/parquet_dispatcher.py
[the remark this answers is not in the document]
YarShev: Yes, this look like a duplicate.
https://github.com/modin-project/modin/blob/f5f4c16d545b671d58929cf44c8174e4dad9af4e/modin/core/io/column_stores/column_store_dispatcher.py#L144

modin/core/io/column_stores/parquet_dispatcher.py
```diff
@@ -23,6 +37,292 @@
...
+                        storage_options=storage_options,
+                        **kwargs,
+                    )
+                    for cols in col_partitions
+                ]
+            )
+        return all_partitions
+
+    @classmethod
+    def build_partition(cls, partition_ids, column_widths):
```
YarShev, line 176: Can we move this method to the parent class (ColumnStoreDispatcher)? Please look at other newly added methods here, can they be moved to the parent?

modin/core/storage_formats/pandas/parsers.py
pyrito: I don't remember, but I think this entire function needs to get removed either way.

modin/conftest.py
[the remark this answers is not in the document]
pyrito: This code chunk was removed because we implemented a separate version for directories in `test_io.py`.

modin/pandas/test/test_io.py
pyrito: Yes, that's what I want to happen. It's kind of hacky but I'm doing this on purpose to set the path to the tmp_dir path, and allow the context manager to delete the directory after we return from the function yield.

modin/core/io/column_stores/parquet_dispatcher.py
pyrito: It actually is optional, so folks can pass `None` instead of an empty dict.

modin/core/io/column_stores/parquet_dispatcher.py
pyrito: I don't think this specific method can be moved back to the parent because of the way the list of partitions is handled here.

modin/core/storage_formats/pandas/parsers.py
pyrito: Yeah I think this was copied over from the previous method. I don't think this is actually safe to keep, but I don't understand why it was here even in the old code.

mvashishtha:
@pyrito there's a black error: https://github.com/modin-project/modin/runs/7549527343?check_suite_focus=true

and an LGTM.com error: https://lgtm.com/projects/g/modin-project/modin/rev/pr-c024f67ef6e42b0dd7e18448cc55349bc3ede9cb 

Please resolve those.

pyrito:
@mvashishtha thanks, I should add `black` checks as part of my precommit hook.

modin/core/storage_formats/pandas/parsers.py
vnlitvinov: you don't have to call `list(range(x))` to iterate over `range(x)`... my question is why you want to call `list()`, why `range()` is not enough?

modin/core/io/column_stores/parquet_dispatcher.py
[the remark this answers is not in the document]
vnlitvinov: it's not a "partition id" then, so we should fix the parameters and the docstring, right?

YarShev:
@pyrito, there are some failures in CI, please take a look.

modin/core/io/column_stores/parquet_dispatcher.py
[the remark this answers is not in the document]
pyrito: Yes but I ended up removing this function either way.

modin/core/storage_formats/pandas/parsers.py
pyrito: `read_row_groups` takes a list, not an iterable it seems

modin/core/storage_formats/pandas/parsers.py
pyrito: I now do understand why this was kept in the code. Please see: https://github.com/modin-project/modin/issues/1366

This case should now be addressed when we process columns in `_read` instead.

modin/core/storage_formats/pandas/parsers.py
pyrito: We need to concatenate the row groups assigned here so the chunk is valid.

modin/core/storage_formats/pandas/parsers.py
```diff
@@ -686,15 +672,24 @@ def parse(fname, **kwargs):
...
+        if isinstance(files_for_parser, str):
+            return pandas.read_parquet(files_for_parser, **kwargs)
+
         for file_for_parser in files_for_parser:
             if isinstance(file_for_parser.path, IOBase):
-                context = nullcontext(file_for_parser.path)
+                if sys.version_info < (3, 7):
+                    context = NullContextManager(file_for_parser.path)
+                else:
+                    from contextlib import nullcontext
```
vnlitvinov, line 690: if we import `NullContextManager` regardless of Python version above, why can't we use it regardless of the version?

modin/core/storage_formats/pandas/utils.py
```diff
@@ -18,6 +18,24 @@
...
+    """
+
+    def __init__(self, resource=None):
+        self.resource = resource
+
+    def __enter__(self):  # noqa: GL08
+        return self.resource
+
+    def __exit__(self, *args):  # noqa: GL08
+        pass
```
vnlitvinov, line 36: I think you can move this one and generalize, and it would take less code to do so:
https://github.com/modin-project/modin/blob/23c40a666dd8af7f781f1c10e0efea15f562d72a/modin/pandas/test/test_general.py#L32-L35

modin/pandas/test/test_io.py
vnlitvinov: I would rather you declare `path` and `_make_parquet_dir()` under the same context manager, as right now this closure looks strange (it might work given how closures compute their cell variables, but it still reads really weird):
```python
with tempfile.TemporaryDirectory() as path:
    def _make_parquet_dir(....):
        for filename, ....:
            df.to_parquet(os.path.join(path, filename), ....)
    yield _make_parquet_dir
```

modin/pandas/test/test_io.py
```diff
@@ -1448,18 +1442,36 @@ def test_read_parquet_pandas_index(self):
...
             }
         )
-        try:
-            pandas_df.set_index("idx").to_parquet(unique_filename)
-            # read the same parquet using modin.pandas
-            df_equals(
-                pd.read_parquet(unique_filename), pandas.read_parquet(unique_filename)
+        # Older version of pyarrow do not support Arrow to Parquet
+        # schema conversion for duration[ns]
+        if sys.version_info >= (3, 7):
```
vnlitvinov, line 1460: The comment talks about pyarrow but you check Python version... feels weird. Can we explicitly check something like `version.parse(pyarrow.__version__) >= version.parse("some.known.version")` instead?

modin/pandas/test/test_io.py
```diff
@@ -1448,18 +1442,36 @@ def test_read_parquet_pandas_index(self):
...
+                start="1 day", periods=2000
             )
 
+        try:
+            for col in pandas_df.columns:
+                if col.startswith("idx"):
+                    pandas_df.set_index(col).to_parquet(unique_filename)
+                    # read the same parquet using modin.pandas
+                    df_equals(
+                        pd.read_parquet(unique_filename),
```
vnlitvinov, line 1471: can we also check that read was done in parallel, not defaulting to single-thread pandas call?

modin/core/io/column_stores/parquet_dispatcher.py
```diff
@@ -323,29 +282,23 @@ def build_query_compiler(cls, path, columns, **kwargs):
...
-            dtypes = None
             row_lengths = None
         frame = cls.frame_cls(
             remote_parts,
             index,
             columns,
             row_lengths=row_lengths,
             column_widths=column_widths,
-            dtypes=dtypes,
+            dtypes=None,
```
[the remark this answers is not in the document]
pyrito: @mvashishtha and I discussed this earlier. There can be some cases where some partitions infer the wrong dtypes for more ambiguous data types and based on some profiling the dtype computation didn't take up that much time.

modin/core/storage_formats/pandas/parsers.py
pyrito: This is probably the way to go. I didn't know we could do something as elegant as you suggested before. My solution was too verbose and I wasn't sure about its correctness.

modin/core/storage_formats/pandas/utils.py
pyrito: Yes this is much nicer to use.

modin/pandas/test/test_io.py
pyrito: Yes, that's fair. I did that for `fsspec` just now so I'll do the same for `pyarrow`. Thanks for pointing it out.

modin/pandas/test/test_io.py
pyrito: That is a good idea!

pyrito: I think we can make such a check once we have this merged: https://github.com/modin-project/modin/pull/4724/files

modin/core/io/column_stores/parquet_dispatcher.py
```diff
@@ -15,13 +15,309 @@
...
+        # was only added relatively recently:
+        # https://github.com/fsspec/filesystem_spec/pull/828
+        def _unstrip_protocol(protocol, path):
+            protos = (protocol,) if isinstance(protocol, str) else protocol
+            for protocol in protos:
+                if path.startswith(f"{protocol}://"):
+                    return path
+            return f"{protos[0]}://{path}"
+
+        if isinstance(path, io.IOBase):
```
mvashishtha, line 66: Not all `IOBase` instances have an `fs`. Should we be checking for the fsspec IOBase subclass instead?

modin/core/io/column_stores/parquet_dispatcher.py
```diff
@@ -15,13 +15,309 @@
...
+
+    @classmethod
+    def build_partition(cls, partition_ids, column_widths):
+        """
+        Build array with partitions of `cls.frame_partition_cls` class.
+
+        Parameters
+        ----------
+        partition_ids : list
+            Array with references to the partitions data.
```
mvashishtha, line 202: I think a brief specification of what's in `partition_ids[i][j][k]` would make this function less opaque.

modin/core/io/column_stores/parquet_dispatcher.py
```diff
@@ -15,13 +15,309 @@
...
+            )
+        # Empty DataFrame case
+        elif len(partition_ids) == 0:
+            return [], False
+        else:
+            index_ids = [part_id[0][1] for part_id in partition_ids if len(part_id) > 0]
+            index_objs = cls.materialize(index_ids)
+            complete_index = index_objs[0].append(index_objs[1:])
+        return complete_index, len(index_columns) == 0 or any(
+            not isinstance(c, str) for c in index_columns
```
mvashishtha, line 277:
```suggestion
        return complete_index, range_index or (len(index_columns) == 0)
```

mvashishtha (commented):
@pyrito I left some minor style comments.

modin/core/storage_formats/pandas/parsers.py
```diff
@@ -669,13 +668,13 @@ class PandasParquetParser(PandasParser):
...
-storage_options : dict
-    Parameters for specific storage engine.""",
+""",
     )
-    def parse(files_for_parser, storage_options, **kwargs):
+    def parse(files_for_parser, **kwargs):
         from pyarrow.parquet import ParquetFile
 
         columns = kwargs.get("columns", None)
+        storage_options = kwargs.pop("storage_options", {}) or {}
```
vnlitvinov, line 677:
```suggestion
        storage_options = kwargs.pop("storage_options", None) or {}
```
this would be sliiiiiiiiiiiiiiiiightly faster (you won't have to construct a dict at each call to `.pop()`, only when it returns you something empty)
or
```suggestion
        storage_options = kwargs.pop("storage_options", {})
```
looks shorter. Pick any you like :)

modin/core/storage_formats/pandas/utils.py
```diff
@@ -14,26 +14,19 @@
...
-class NullContextManager(object):  # noqa: PR01
+@contextlib.contextmanager
+def _nullcontext(dummy_value=None):  # noqa: PR01
     """
-    Dummy context manager that tries to mimic `nullcontext` for older Python versions.
+    Act as a replacement for contextlib.nullcontext missing in older Python.
 
-    Python 3.6 currently does not support `nullcontext` which is used in parsers.py.
-    This class tries provide some backwards compatibility.
+    contextlib.nullcontext is only available from Python 3.7.
```
vnlitvinov, line 27: this should be a `Note` section IMHO

vnlitvinov (commented):
Changes look good, please also address what @mvashishtha found

modin/core/io/column_stores/parquet_dispatcher.py
```diff
@@ -15,13 +15,309 @@
...
+            Parameters of deploying read_* function.
+
+        Returns
+        -------
+        new_query_compiler : BaseQueryCompiler
+            Query compiler with imported data for further processing.
+        """
+        col_partitions, column_widths = cls.build_columns(columns)
+        partition_ids = cls.call_deploy(path, col_partitions, **kwargs)
+        index, needs_index_sync = cls.build_index(path, partition_ids, index_columns)
```
YarShev, line 303:
```suggestion
        index, sync_index = cls.build_index(path, partition_ids, index_columns)
```
Maybe just so?

Also, can we right away get `row_lengths` from `build_index` as it is done in the parent in order for us not to run additional `ray.get()` below in `part.length()`?

modin/core/storage_formats/pandas/parsers.py
pyrito: I think I actually need to implement it this way. I encountered the case where `storage_options` wasn't in `kwargs` which then returns the empty dict. If we have `storage_options` that maps to nothing, then we get none, which we don't handle, so we need the second `or {}`.

modin/core/io/column_stores/parquet_dispatcher.py
pyrito: Yeah we should be checking for `AbstractBufferedFile`

modin/core/io/column_stores/parquet_dispatcher.py
pyrito: Yes we can get the `row_lengths` for some cases.

modin/core/io/column_stores/parquet_dispatcher.py
pyrito: @mvashishtha added some comments in the Notes section. Let me know what you think.

modin/core/io/column_stores/parquet_dispatcher.py
```diff
@@ -208,6 +208,18 @@ def build_partition(cls, partition_ids, column_widths):
         np.ndarray
             array with shape equals to the shape of `partition_ids` and
             filed with partition objects.
+
+        Notes
+        -----
+        partition_ids contains three levels of nesting. The first level contains lists
+        of files to read with corresponding columns as such:
+        partition_ids[i] -> [read(files, cols[0..n]), read(files, cols[n..m]), ...].
```
mvashishtha, line 216: Why include a description of the top level? I think the part about "The second level contains a list of object references for each read call:" is sufficient. It shouldn't be part of the API that `partition_ids[i]` comes from `read`.

modin/core/storage_formats/pandas/utils.py
```diff
@@ -24,6 +24,8 @@ def _nullcontext(dummy_value=None):  # noqa: PR01
     """
     Act as a replacement for contextlib.nullcontext missing in older Python.
 
+    Notes
+    -----
```
mvashishtha, line 28: Do you need a notes section?

mvashishtha (commented):
I have two minor comments. Otherwise LGTM.

modin/core/storage_formats/pandas/utils.py
pyrito: @mvashishtha told me to add this section

mvashishtha: You mean @vnlitvinov ? Ok.

pyrito: Yes, sorry!

modin/core/io/column_stores/parquet_dispatcher.py
pyrito: Fair enough

modin/core/io/column_stores/parquet_dispatcher.py
YarShev: I think it does matter because we need to call ray.get/client.gather (these are blocking operations) again. To avoid this we could return `row_lengths` from `build_index`. This will also conform the parent behavior.

mvashishtha: The blocking call shouldn't matter for performance because we pass the length to the partition class constructor in `build_partition`, so the partition should have already cached that length.

YarShev: IIUC we pass the length as a `ray.ObjectRef` to the partition class constructor in `build_partition`. Next we call `ray.get` (let's consider Ray for instance) in `build_index` to get index objs of the partitions. After that we again call `ray.get` for every partition to get its length. I think that (calling `ray.get` twice in the main thread) incurs some perf and memory implications. If you could check this, it would be great. I am not against doing this as part of a separate issue.

YarShev (approved):
LGTM!

modin/core/io/column_stores/parquet_dispatcher.py
mvashishtha: It's a good point that we're passing an object ref instead of the integer width to the partition constructor in `build_partition`. Still, if we're blocking on the index in `build_index`, the same remote function call should also compute length, so the length should be ready and the `ray.get` should be cheap. But like you, I don't know what is the cost of `ray.get` for a small object that's already ready in the object store. I filed #4743 for further work.

mvashishtha (approved):

PR #4710, opened 2022-07-25
FIX-#4657: Use `fsspec` for handling s3/http-like paths instead of `s3fs`

### LINKED ISSUE #4657: `read_csv` accepts only `s3://...`-like address pattern for reading from remote storages
### System information
- **Modin version** (`modin.__version__`): 06ec64bedc2ace639af09759053182472fdd6544
- **Python version**: 3.8.13

```python
import modin.pandas as pd

s3_path = "https://modin-test.s3.us-west-1.amazonaws.com/yellow_tripdata_2015-01.csv"
df = pd.read_csv(s3_path, parse_dates=["tpep_pickup_datetime", "tpep_dropoff_datetime"], quoting=3) # defaults to pandas
```

### Describe the problem

`read_csv` function accepts only single pattern of remote `s3` storages:
https://github.com/modin-project/modin/blob/06ec64bedc2ace639af09759053182472fdd6544/modin/core/io/file_dispatcher.py#L257

https://github.com/modin-project/modin/blob/06ec64bedc2ace639af09759053182472fdd6544/modin/core/io/file_dispatcher.py#L29

Pandas does the next: https://github.com/pandas-dev/pandas/blob/fce0953d079e297d90289ff6371b32dd799d70a5/pandas/io/common.py#L350-L367

prutskov:

## What do these changes do?

This PR changes handling of `s3`-like paths from direct `s3fs` to `fsspec` for `read_csv`/`read_csv_glob`. Also some tests for `read_csv_glob` was stabilizied.

In the result:
1. `s3fs` package is used in test system only now
2. New approach allows to switch on parallel implementation of `read_csv` in case `http/https`-like paths instead of defaulting to pandas.

Exceptions: `read_csv_glob` still doesn't support `http`-like paths because `fsspec.HTTPFileSystem.glob` isn't working as expected (always gives empty list instead of all files in remote directory).

- [x] Resolves #4657  
- [x] tests added and passing

### DIFF
--- a/examples/tutorial/jupyter/execution/pandas_on_dask/requirements.txt
+++ b/examples/tutorial/jupyter/execution/pandas_on_dask/requirements.txt
@@ -1,5 +1,4 @@
 fsspec
-s3fs
 jupyterlab
 ipywidgets
 modin[dask]
--- a/modin/core/io/file_dispatcher.py
+++ b/modin/core/io/file_dispatcher.py
@@ -284,11 +284,12 @@ class FileDispatcher(ClassLogger):
                     EndpointConnectionError,
                     ConnectTimeoutError,
                 ):
-                    pass
-                fs, _ = fsspec.core.url_to_fs(
-                    file_path, anon=True, **new_storage_options
-                )
-                return exists or fs.exists(file_path)
+                    fs, _ = fsspec.core.url_to_fs(
+                        file_path, anon=True, **new_storage_options
+                    )
+                    exists = fs.exists(file_path)
+
+                return exists
         return os.path.exists(file_path)
 
     @classmethod
--- a/modin/core/io/file_dispatcher.py
+++ b/modin/core/io/file_dispatcher.py
@@ -279,9 +279,9 @@ class FileDispatcher(ClassLogger):
                     new_storage_options = {}
 
                 fs, _ = fsspec.core.url_to_fs(file_path, **new_storage_options)
-                is_exist = False
+                exists = False
                 try:
-                    is_exist = fs.exists(file_path)
+                    exists = fs.exists(file_path)
                 except (
                     NoCredentialsError,
                     PermissionError,
@@ -292,7 +292,7 @@ class FileDispatcher(ClassLogger):
                 fs, _ = fsspec.core.url_to_fs(
                     file_path, anon=True, **new_storage_options
                 )
-                return is_exist or fs.exists(file_path)
+                return exists or fs.exists(file_path)
         return os.path.exists(file_path)
 
     @classmethod
--- a/modin/core/io/file_dispatcher.py
+++ b/modin/core/io/file_dispatcher.py
@@ -257,7 +257,12 @@ class FileDispatcher(ClassLogger):
             Whether file exists or not.
         """
         if isinstance(file_path, str):
-            if fsspec.core.split_protocol(file_path)[0] in ("s3", "S3", "https", "http"):
+            if fsspec.core.split_protocol(file_path)[0] in (
+                "s3",
+                "S3",
+                "https",
+                "http",
+            ):
                 if file_path.startswith("S"):
                     file_path = f"s{file_path[1:]}"
 
--- a/modin/core/io/file_dispatcher.py
+++ b/modin/core/io/file_dispatcher.py
@@ -258,42 +258,40 @@ class FileDispatcher(ClassLogger):
             Whether file exists or not.
         """
         if (
-            isinstance(file_path, str)
-            and fsspec.core.split_protocol(file_path)[0] in _SUPPORTED_PROTOCOLS
+            not isinstance(file_path, str)
+            or fsspec.core.split_protocol(file_path)[0] not in _SUPPORTED_PROTOCOLS
         ):
-            # `file_path` may start with a capital letter, which isn't supported by `fsspec.core.url_to_fs` used below.
-            file_path = file_path[0].lower() + file_path[1:]
-
-            from botocore.exceptions import (
-                NoCredentialsError,
-                EndpointConnectionError,
-                ConnectTimeoutError,
-            )
-
-            if storage_options is not None:
-                new_storage_options = dict(storage_options)
-                new_storage_options.pop("anon", None)
-            else:
-                new_storage_options = {}
-
-            fs, _ = fsspec.core.url_to_fs(file_path, **new_storage_options)
-            exists = False
-            try:
-                exists = fs.exists(file_path)
-            except (
-                NoCredentialsError,
-                PermissionError,
-                EndpointConnectionError,
-                ConnectTimeoutError,
-            ):
-                fs, _ = fsspec.core.url_to_fs(
-                    file_path, anon=True, **new_storage_options
-                )
-                exists = fs.exists(file_path)
-
-            return exists
-        else:
             return os.path.exists(file_path)
+
+        # `file_path` may start with a capital letter, which isn't supported by `fsspec.core.url_to_fs` used below.
+        file_path = file_path[0].lower() + file_path[1:]
+
+        from botocore.exceptions import (
+            NoCredentialsError,
+            EndpointConnectionError,
+            ConnectTimeoutError,
+        )
+
+        if storage_options is not None:
+            new_storage_options = dict(storage_options)
+            new_storage_options.pop("anon", None)
+        else:
+            new_storage_options = {}
+
+        fs, _ = fsspec.core.url_to_fs(file_path, **new_storage_options)
+        exists = False
+        try:
+            exists = fs.exists(file_path)
+        except (
+            NoCredentialsError,
+            PermissionError,
+            EndpointConnectionError,
+            ConnectTimeoutError,
+        ):
+            fs, _ = fsspec.core.url_to_fs(file_path, anon=True, **new_storage_options)
+            exists = fs.exists(file_path)
+
+        return exists
 
     @classmethod
     def deploy(cls, func, *args, num_returns=1, **kwargs):  # noqa: PR01
--- a/modin/core/io/file_dispatcher.py
+++ b/modin/core/io/file_dispatcher.py
@@ -257,40 +257,43 @@ class FileDispatcher(ClassLogger):
         bool
             Whether file exists or not.
         """
-        if isinstance(file_path, str):
-            if fsspec.core.split_protocol(file_path)[0] in _SUPPORTED_PROTOCOLS:
-                # `file_path` may start with a capital letter, which isn't supported by `fsspec.core.url_to_fs` used below.
-                file_path = file_path[0].lower() + file_path[1:]
+        if (
+            isinstance(file_path, str)
+            and fsspec.core.split_protocol(file_path)[0] in _SUPPORTED_PROTOCOLS
+        ):
+            # `file_path` may start with a capital letter, which isn't supported by `fsspec.core.url_to_fs` used below.
+            file_path = file_path[0].lower() + file_path[1:]
 
-                from botocore.exceptions import (
-                    NoCredentialsError,
-                    EndpointConnectionError,
-                    ConnectTimeoutError,
+            from botocore.exceptions import (
+                NoCredentialsError,
+                EndpointConnectionError,
+                ConnectTimeoutError,
+            )
+
+            if storage_options is not None:
+                new_storage_options = dict(storage_options)
+                new_storage_options.pop("anon", None)
+            else:
+                new_storage_options = {}
+
+            fs, _ = fsspec.core.url_to_fs(file_path, **new_storage_options)
+            exists = False
+            try:
+                exists = fs.exists(file_path)
+            except (
+                NoCredentialsError,
+                PermissionError,
+                EndpointConnectionError,
+                ConnectTimeoutError,
+            ):
+                fs, _ = fsspec.core.url_to_fs(
+                    file_path, anon=True, **new_storage_options
                 )
+                exists = fs.exists(file_path)
 
-                if storage_options is not None:
-                    new_storage_options = dict(storage_options)
-                    new_storage_options.pop("anon", None)
-                else:
-                    new_storage_options = {}
-
-                fs, _ = fsspec.core.url_to_fs(file_path, **new_storage_options)
-                exists = False
-                try:
-                    exists = fs.exists(file_path)
-                except (
-                    NoCredentialsError,
-                    PermissionError,
-                    EndpointConnectionError,
-                    ConnectTimeoutError,
-                ):
-                    fs, _ = fsspec.core.url_to_fs(
-                        file_path, anon=True, **new_storage_options
-                    )
-                    exists = fs.exists(file_path)
-
-                return exists
-        return os.path.exists(file_path)
+            return exists
+        else:
+            return os.path.exists(file_path)
 
     @classmethod
     def deploy(cls, func, *args, num_returns=1, **kwargs):  # noqa: PR01
--- a/modin/core/io/file_dispatcher.py
+++ b/modin/core/io/file_dispatcher.py
@@ -282,7 +282,6 @@ class FileDispatcher(ClassLogger):
                     )
                     exists = fs.exists(path)
                 return exists
-
         return os.path.exists(file_path)
 
     @classmethod
--- a/modin/core/io/file_dispatcher.py
+++ b/modin/core/io/file_dispatcher.py
@@ -275,9 +275,9 @@ class FileDispatcher(ClassLogger):
                     new_storage_options = {}
 
                 fs, _ = fsspec.core.url_to_fs(file_path, **new_storage_options)
-                exists = False
+                is_exist = False
                 try:
-                    exists = fs.exists(file_path)
+                    is_exist = fs.exists(file_path)
                 except (
                     NoCredentialsError,
                     PermissionError,
@@ -288,7 +288,7 @@ class FileDispatcher(ClassLogger):
                 fs, _ = fsspec.core.url_to_fs(
                     file_path, anon=True, **new_storage_options
                 )
-                return exists or fs.exists(file_path)
+                return is_exist or fs.exists(file_path)
         return os.path.exists(file_path)
 
     @classmethod
--- a/modin/core/io/file_dispatcher.py
+++ b/modin/core/io/file_dispatcher.py
@@ -22,9 +22,9 @@ import fsspec
 import os
 import re
 from modin.config import StorageFormat
-from modin.utils import import_optional_dependency
 from modin.logging import ClassLogger
 import numpy as np
+from pandas.io.common import is_fsspec_url, is_url
 
 S3_ADDRESS_REGEX = re.compile("[sS]3://(.*?)/(.*)")
 NOT_IMPLEMENTED_MESSAGE = "Implement in children classes!"
@@ -213,7 +213,9 @@ class FileDispatcher(ClassLogger):
         if `file_path` is an S3 bucket, parameter will be returned as is, otherwise
         absolute path will be returned.
         """
-        if isinstance(file_path, str) and S3_ADDRESS_REGEX.search(file_path):
+        if isinstance(file_path, str) and (
+            is_fsspec_url(file_path) or is_url(file_path)
+        ):
             return file_path
         else:
             return os.path.abspath(file_path)
@@ -258,13 +260,7 @@ class FileDispatcher(ClassLogger):
             Whether file exists or not.
         """
         if isinstance(file_path, str):
-            match = S3_ADDRESS_REGEX.search(file_path)
-            if match is not None:
-                if file_path[0] == "S":
-                    file_path = "{}{}".format("s", file_path[1:])
-                S3FS = import_optional_dependency(
-                    "s3fs", "Module s3fs is required to read S3FS files."
-                )
+            if is_fsspec_url(file_path) or is_url(file_path):
                 from botocore.exceptions import (
                     NoCredentialsError,
                     EndpointConnectionError,
@@ -276,14 +272,17 @@ class FileDispatcher(ClassLogger):
                 else:
                     new_storage_options = {}
 
-                s3fs = S3FS.S3FileSystem(anon=False, **new_storage_options)
+                fs, path = fsspec.core.url_to_fs(file_path, **new_storage_options)
                 exists = False
                 try:
-                    exists = s3fs.exists(file_path) or exists
+                    exists = fs.exists(path)
                 except (NoCredentialsError, PermissionError, EndpointConnectionError):
-                    pass
-                s3fs = S3FS.S3FileSystem(anon=True, **new_storage_options)
-                return exists or s3fs.exists(file_path)
+                    fs, path = fsspec.core.url_to_fs(
+                        file_path, anon=True, **new_storage_options
+                    )
+                    exists = fs.exists(path)
+                return exists
+
         return os.path.exists(file_path)
 
     @classmethod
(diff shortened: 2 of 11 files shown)


modin/core/io/file_dispatcher.py
```diff
@@ -258,16 +258,14 @@ def file_exists(cls, file_path, storage_options=None):
...
         if isinstance(file_path, str):
-            match = S3_ADDRESS_REGEX.search(file_path)
-            if match is not None:
-                if file_path[0] == "S":
-                    file_path = "{}{}".format("s", file_path[1:])
-                S3FS = import_optional_dependency(
-                    "s3fs", "Module s3fs is required to read S3FS files."
-                )
+            if is_fsspec_url(file_path) or is_url(file_path):
+                if file_path.startswith("S"):
```
pyrito, line 262: You can use `split_protocol` and check for an exact match for the protocol.

modin/core/io/file_dispatcher.py
```diff
@@ -276,14 +274,21 @@ def file_exists(cls, file_path, storage_options=None):
                 else:
                     new_storage_options = {}
 
-                s3fs = S3FS.S3FileSystem(anon=False, **new_storage_options)
-                exists = False
+                fs, _ = fsspec.core.url_to_fs(file_path, **new_storage_options)
+                is_exist = False
```
pyrito, line 278: why `is_exist` vs `exists`?

modin/core/io/text/csv_glob_dispatcher.py
```diff
@@ -330,16 +330,14 @@ def file_exists(cls, file_path: str, storage_options=None) -> bool:
...
-            match = S3_ADDRESS_REGEX.search(file_path)
-            if match is not None:
-                if file_path[0] == "S":
-                    file_path = "{}{}".format("s", file_path[1:])
-                S3FS = import_optional_dependency(
-                    "s3fs", "Module s3fs is required to read S3FS files."
-                )
+            if is_fsspec_url(file_path):
+                if file_path.startswith("S"):
+                    file_path = f"s{file_path[1:]}"
```
pyrito, line 335: ditto from before

modin/core/io/text/csv_glob_dispatcher.py
```diff
@@ -348,14 +346,27 @@ def file_exists(cls, file_path: str, storage_options=None) -> bool:
                 else:
                     new_storage_options = {}
 
-                s3fs = S3FS.S3FileSystem(anon=False, **new_storage_options)
-                exists = False
+                fs, _ = fsspec.core.url_to_fs(file_path, **new_storage_options)
+                is_exist = False
```
pyrito, line 350: ditto from before

modin/core/io/text/csv_glob_dispatcher.py
```diff
@@ -373,28 +384,30 @@ def get_path(cls, file_path: str) -> list:
...
         """
-        if S3_ADDRESS_REGEX.search(file_path):
-            # S3FS does not allow captial S in s3 addresses.
-            if file_path[0] == "S":
-                file_path = "{}{}".format("s", file_path[1:])
-
-            S3FS = import_optional_dependency(
-                "s3fs", "Module s3fs is required to read S3FS files."
+        if is_fsspec_url(file_path) or is_url(file_path):
+            if file_path.startswith("S"):
```
pyrito, line 388: ditto

modin/core/io/file_dispatcher.py
prutskov: `split_protocol` is used now for more accurate checking on supported paths.

Line `if file_path.startswith("S"):` is needed to making `S`->`s` because `fsspec` doesn't work with path starting from capital letter:

```python
In [1]: import fsspec

In [2]: path = "S3://modin-datasets/testing/multiple_csv/test_data*.csv"

In [3]: fsspec.core.url_to_fs(path)
Input In [3], in <cell line: 1>()
----> 1 fsspec.core.url_to_fs(path)

File ~/miniconda3/envs/modin/lib/python3.8/site-packages/fsspec/core.py:408, in url_to_fs(url, **kwargs)
    406 else:
    407     protocol = split_protocol(url)[0]
--> 408     cls = get_filesystem_class(protocol)
    410     options = cls._get_kwargs_from_urls(url)
    411     update_storage_options(options, kwargs)

File ~/miniconda3/envs/modin/lib/python3.8/site-packages/fsspec/registry.py:225, in get_filesystem_class(protocol)
    223 if protocol not in registry:
    224     if protocol not in known_implementations:
--> 225         raise ValueError("Protocol not known: %s" % protocol)
    226     bit = known_implementations[protocol]
    227     try:

ValueError: Protocol not known: S3
```

modin/core/io/text/csv_glob_dispatcher.py
prutskov: See https://github.com/modin-project/modin/pull/4710#discussion_r932988588

modin/core/io/text/csv_glob_dispatcher.py
prutskov: See https://github.com/modin-project/modin/pull/4710#discussion_r932988588

modin/core/io/file_dispatcher.py
prutskov: I prefer to start naming of boolean values by `is`/`can`/`has`/etc because it additionally highlights type of variable and looks more clearer in subsequent code.

modin/core/io/text/csv_glob_dispatcher.py
prutskov: See https://github.com/modin-project/modin/pull/4710#discussion_r932994389

modin/core/io/file_dispatcher.py
pyrito: Could we try `does_exist`? I think that would flow better here :D

modin/core/io/file_dispatcher.py
pyrito: Nit: what if we did something like this? Just trying to see if we can default to lowercasing the first letter so that we don't have to check for the upper case 'S'.
```python
if fsspec.core.split_protocol(file_path)[0] in (
    "s3",
    "S3",
    "https",
    "http",
):
    file_path = file_path[0].lower() + file_path[1:]
```
We can default to lowercasing the path

modin/core/io/text/csv_glob_dispatcher.py
pyrito: See: https://github.com/modin-project/modin/pull/4710#discussion_r933379025

modin/core/io/text/csv_glob_dispatcher.py
pyrito: See: https://github.com/modin-project/modin/pull/4710#discussion_r933379025

modin/core/io/text/csv_glob_dispatcher.py
pyrito: See https://github.com/modin-project/modin/pull/4710#discussion_r933379025

modin/core/io/file_dispatcher.py
prutskov: It looks more clearer, I have used your suggestion + added a comment why we this do

modin/core/io/text/csv_glob_dispatcher.py
prutskov: see https://github.com/modin-project/modin/pull/4710#discussion_r933442381

modin/core/io/text/csv_glob_dispatcher.py
prutskov: see https://github.com/modin-project/modin/pull/4710#discussion_r933442381

modin/core/io/file_dispatcher.py
prutskov: Why would it be better? We already use approach when naming starts from `is_` for boolean values/methods/properties (functions from pandas like [these](https://github.com/modin-project/modin/blob/05bf659b67d2b7153d8aaa004dd95252377eed51/modin/core/storage_formats/pandas/query_compiler.py#L28-L31), for example)

pyrito: Grammar wise I think `does_exist` or even just `exists` is technically more correct than `is_exist`. We could do like `is_valid` or something like that if you want to stick with the `is_*`

modin/experimental/pandas/test/test_io_exp.py
```diff
@@ -164,24 +168,25 @@ def test_read_single_csv_with_parse_dates(self, parse_dates):
...
-    ]
-    pandas_df = pd.concat(pandas_dfs)
-
-    # Indexes get messed up when concatting so we reset both.
-    pandas_df = pandas_df.reset_index(drop=True)
-    modin_df = modin_df.reset_index(drop=True)
+    path = "S3://modin-datasets/testing/multiple_csv/test_data*.csv"
 
-    df_equals(modin_df, pandas_df)
+    def _pandas_read_csv_glob(path, storage_options):
```
pyrito, line 173: Could we make this a helper function?

modin/core/io/file_dispatcher.py
prutskov: I'm not a native speaker so I reverted previous variant of naming for this variable (`exists`), if this is more grammatically correct

modin/experimental/pandas/test/test_io_exp.py
prutskov: Probably if would be better to create helper function like this `pandas_read_csv_glob` for using in all our `read_csv_glob` tests, because the most part of tests in `test_io_exp.py` emulate pandas `read_csv_glob` behavior in several places. It should be resolved as a separate issue because it requires a lot of code changes in `test_io_exp.py`

modin/core/io/file_dispatcher.py
pyrito: No worries! Not a problem at all :D

modin/experimental/pandas/test/test_io_exp.py
pyrito: Sounds reasonable to me. Could you create the issue for it?

pyrito (dismissed):
LGTM - just curious but do we not support `gcsfs` currently?

prutskov:
> LGTM - just curious but do we not support `gcsfs` currently?

Parallel implementation before this PR supported only `s3`/`S3` file system (regular expressions were used to determine fs). For now, support of different remote storages might be extended by changing the set of supported protocols (`s3`, `S3`, `https`, `http`).

We can try to add `gcsfs` in this set with corresponed tests, but as a part of a separate issue.

pyrito:
Sounds good. We should try to support it in the future, but yes, it should be a separate issue.

modin/core/io/file_dispatcher.py
```diff
@@ -213,7 +210,9 @@ def get_path(cls, file_path):
         if `file_path` is an S3 bucket, parameter will be returned as is, otherwise
         absolute path will be returned.
         """
-        if isinstance(file_path, str) and S3_ADDRESS_REGEX.search(file_path):
+        if isinstance(file_path, str) and (
+            fsspec.core.split_protocol(file_path)[0] in ("s3", "S3", "https", "http")
```
vnlitvinov, line 214: let's make a module-level constant like `_SUPPORTED_PROTOS = {"s3", "S3", "http", "https"}` and use it everywhere

modin/core/io/file_dispatcher.py
```diff
@@ -258,16 +257,19 @@ def file_exists(cls, file_path, storage_options=None):
...
-                    "s3fs", "Module s3fs is required to read S3FS files."
-                )
+            if fsspec.core.split_protocol(file_path)[0] in (
+                "s3",
+                "S3",
+                "https",
+                "http",
+            ):
+                # `file_path` may start with a capital letter, which isn't supported by `fsspec.core.url_to_fs` used below.
+                file_path = file_path[0].lower() + file_path[1:]
```
vnlitvinov, line 267: this looks like an `fsspec` bug

modin/core/io/file_dispatcher.py
```diff
@@ -276,14 +278,21 @@ def file_exists(cls, file_path, storage_options=None):
...
+                    PermissionError,
+                    EndpointConnectionError,
+                    ConnectTimeoutError,
+                ):
                     pass
-                s3fs = S3FS.S3FileSystem(anon=True, **new_storage_options)
-                return exists or s3fs.exists(file_path)
+                fs, _ = fsspec.core.url_to_fs(
+                    file_path, anon=True, **new_storage_options
+                )
```
vnlitvinov, line 294: I would rather we check for `anon=True` only in `except:` block - there's no need to try an anonymous access if normal one worked fine

modin/core/io/text/csv_glob_dispatcher.py
```diff
@@ -348,14 +345,27 @@ def file_exists(cls, file_path: str, storage_options=None) -> bool:
...
+                    PermissionError,
+                    EndpointConnectionError,
+                    ConnectTimeoutError,
+                ):
                     pass
-                s3fs = S3FS.S3FileSystem(anon=True, **new_storage_options)
-                return exists or len(s3fs.glob(file_path)) > 0
+                fs, _ = fsspec.core.url_to_fs(
+                    file_path, anon=True, **new_storage_options
+                )
```
vnlitvinov, line 361: let's do anonymous check in `except` block only; also it might make sense to see if `anon` wasn't already in `new_storage_options`

modin/core/io/text/csv_glob_dispatcher.py
```diff
@@ -348,14 +345,27 @@ def file_exists(cls, file_path: str, storage_options=None) -> bool:
...
+                return exists or len(fs.glob(file_path)) > 0
+
+            if fsspec.core.split_protocol(file_path)[0] in ("http", "https"):
```
vnlitvinov, line 364: why it's not a condition like `if proto not in ("s3", "S3")`, what if I specify a URL like `ftp://foo/bar.csv`?..

examples/tutorial/jupyter/execution/pandas_on_dask/requirements.txt
```diff
@@ -1,5 +1,4 @@
 fsspec
-s3fs
```
vnlitvinov, line 2: did you search for all files listing `s3fs` dependency? what about `env_omnisci.yml` or other requirement files?

prutskov: We use `s3fs` for testing here: https://github.com/modin-project/modin/blob/adb16a17f721048005520388080627975c6852d8/modin/pandas/test/test_io.py#L1527

By this reason `s3fs` wasn't removed from dev-env files, but removed from other (like example deps), because modin package no longer requires `s3fs`

modin/core/io/text/csv_glob_dispatcher.py
pyrito: See: https://github.com/modin-project/modin/issues/4742

prutskov: Condition was changed to `if is_url(file_path):` to throw exceptions on each URL type

modin/core/io/file_dispatcher.py
prutskov: The constants were added for `file_dispatcher` and `csv_glob_dispatcher`

modin/core/io/file_dispatcher.py
prutskov: Moved to `except` block

modin/core/io/text/csv_glob_dispatcher.py
prutskov: Moved to `except` block. `new_storage_options` doesn't have `anon` because we have specific check above to handling this

modin/core/io/file_dispatcher.py
prutskov: `fsspec` answered in this issue https://github.com/fsspec/filesystem_spec/issues/1009

modin/core/io/file_dispatcher.py
```diff
@@ -276,14 +274,22 @@ def file_exists(cls, file_path, storage_options=None):
...
+                    EndpointConnectionError,
+                    ConnectTimeoutError,
+                ):
+                    fs, _ = fsspec.core.url_to_fs(
+                        file_path, anon=True, **new_storage_options
+                    )
+                    exists = fs.exists(file_path)
+
+                return exists
         return os.path.exists(file_path)
```
mvashishtha, line 293: I'm confused about when we hit this line. I think it's when `isinstance(file_path, str)` is False? Will `os.path.exists` ever return True in that case? If we do want this `os.path.exists` line still, can we move it up and return early if it's True? that will make the code less indented. https://blog.codinghorror.com/flattening-arrow-code/

modin/core/io/text/csv_glob_dispatcher.py
```diff
@@ -348,14 +348,27 @@ def file_exists(cls, file_path: str, storage_options=None) -> bool:
...
-                s3fs = S3FS.S3FileSystem(anon=False, **new_storage_options)
+                fs, _ = fsspec.core.url_to_fs(file_path, **new_storage_options)
                 exists = False
                 try:
-                    exists = len(s3fs.glob(file_path)) > 0 or exists
-                except (NoCredentialsError, PermissionError, EndpointConnectionError):
-                    pass
-                s3fs = S3FS.S3FileSystem(anon=True, **new_storage_options)
-                return exists or len(s3fs.glob(file_path)) > 0
+                    exists = fs.exists(file_path)
```
mvashishtha, line 354: the `try` block is checking `fs.exists(file_path)`, while the except block is checking `len(fs.glob(file_path)) > 0`. Both blocks will check `len(fs.glob(file_path)) > 0` if their respective check fails. Could you make the `try` and `except` branches check the same things (in the original code they are both just checking `len(s3fs.glob(file_path)) > 0`).

modin/core/io/text/csv_glob_dispatcher.py
```diff
@@ -373,28 +386,29 @@ def get_path(cls, file_path: str) -> list:
...
+                ConnectTimeoutError,
             )
-            from botocore.exceptions import NoCredentialsError, EndpointConnectionError
 
             def get_file_path(fs_handle) -> List[str]:
                 file_paths = fs_handle.glob(file_path)
-                s3_addresses = ["{}{}".format("s3://", path) for path in file_paths]
+                if len(file_paths) == 0 and not fs_handle.exists(file_path):
+                    raise FileNotFoundError(f"Path <{file_path}> isn't available.")
+                s3_addresses = [f"s3://{path}" for path in file_paths]
```
mvashishtha, line 403: Should we change this to not assume the addresses are all for s3?

mvashishtha (commented):
@prutskov thanks, I have some minor comments.

modin/core/io/text/csv_glob_dispatcher.py
prutskov: For now we can leave as it is, because this block of code is used only in case s3 storage (`_SUPPORTED_PROTOCOLS` for read_csv_glob is {s3, S3} only).

modin/core/io/text/csv_glob_dispatcher.py
prutskov: I've aligned try-except block (now both check using `fs.exists`). `exists` is faster than `fs.glob` because it just check availability of `file_path`. On `return` stage I think it's better to use `fs.glob` in case `exists` is False (leave return phase as it is).

modin/core/io/file_dispatcher.py
prutskov: I've changed if-else branching to see when each flow happens. Please, take a look again

mvashishtha: Can we do `if not isinstance(file_path, str): return os.path.exists(file_path)` and then dedent the code that's currently under the if clause? I think the code will be easier to read if it's less indented.

modin/core/io/text/csv_glob_dispatcher.py
mvashishtha: Okay, but still we check `len(fs.glob(file_path)) > 0` twice if it's false and we hit the `except`? Is there a good way to avoid that? It's confusing to read. 

I know this problem was there before your PR...

modin/core/io/file_dispatcher.py
prutskov: > Can we do `if not isinstance(file_path, str): return os.path.exists(file_path)` and then dedent the code that's currently under the if clause? I think the code will be easier to read if it's less indented.

But we do `return os.path.exists(file_path)` not only in case `if not isinstance(file_path, str)`. With your suggestion this block will look like:
```python
if not isinstance(file_path, str):
    return os.path.exists(file_path)`
elif isinstance(file_path, str) and fsspec.core.split_protocol(file_path)[0] in _SUPPORTED_PROTOCOLS:
   # protocols processing stuff
else:
    return os.path.exists(file_path)` # processing the last case when `file_path` just a local filesystem path
```

mvashishtha: Can we do this:

```python
if not isinstance(file_path, str) or fsspec.core.split_protocol(file_path)[0] not in _SUPPORTED_PROTOCOLS:
  return os.path.exists(file_path)
# protocols processing stuff not inside "if" or "else"
```

modin/core/io/text/csv_glob_dispatcher.py
prutskov: I think that we should leave this logic with trying to check with `exists()` (and in case it's `False` use `fs.glob`). The reason of check with `fs.glob` is addresses like `s3://url_path/*`. In this case `exists` returns `False`, but `fs.glob` returns all files placed in  s3://url_path/

modin/core/io/file_dispatcher.py
prutskov: I understand what you mean! We can do this. I've applied this for `FileDispatcher.file_exists` and `CsvGlobDispatcher.file_exists/get_path`

mvashishtha (approved):
@prutskov LGTM, thank you for addressing the comments.

mvashishtha (approved):

modin/core/io/text/csv_glob_dispatcher.py
```diff
@@ -383,35 +384,39 @@ def get_path(cls, file_path: str) -> list:
...
+            NoCredentialsError,
+            EndpointConnectionError,
+            ConnectTimeoutError,
+        )
+
+        def get_file_path(fs_handle) -> List[str]:
+            file_paths = fs_handle.glob(file_path)
+            if len(file_paths) == 0 and not fs_handle.exists(file_path):
+                raise FileNotFoundError(f"Path <{file_path}> isn't available.")
+            s3_addresses = [f"s3://{path}" for path in file_paths]
```
vnlitvinov, line 405: This feels wrong, what if original protocol wasn't `S3`-like? Previous version was explicitly checking for S3 only...

prutskov: This version also explicitly checking on supported protocols in the beginning of `get_path`. In case protocol is n't {"s3", "S3"} this line won't be run.

pyrito: Could we add a TODO here to change this to be more generalizable in the future? Or it might be good to at least mention this in the issue I created when we try to add more support to things other than s3.

vnlitvinov: oh, I've missed that this file has different `_SUPPORTED_PROTOCOLS`, sorry! 🙃

vnlitvinov (approved):
LGTM!

modin/core/io/text/csv_glob_dispatcher.py
prutskov: @pyrito I mentioned this thread in #4742 instead of TODO because the PR was already merged.

pyrito: @prutskov all good. Thanks!

PR #4715, opened 2022-07-25
FIX-#4597: Refactor Partition handling of func, args, kwargs

### LINKED ISSUE #4597: Refactor Partition handling of func, args, kwargs
### Describe the problem

[Issue #4464](https://github.com/modin-project/modin/issues/4464) captured an issue in our partition layer introduced by our poor handling of func, args, and kwargs in that layer. https://github.com/modin-project/modin/pull/4490 was a quick fix for this issue, but also introduced some unclean code, such as https://github.com/modin-project/modin/pull/4490/files#r88838177. 

And we have areas still where the kwargs of the func to apply are mixed with the kwargs of the parent func, for example https://github.com/modin-project/modin/blob/b22b93df20ad25ae7a11f0c89d32fb2f234d4641/modin/core/dataframe/pandas/partitioning/axis_partition.py#L155-L157 and https://github.com/modin-project/modin/blob/fde216da6ffcc72e2815e776e796f9aafac42ef5/modin/core/dataframe/pandas/partitioning/axis_partition.py#L121-L122

The main issue is that we implement a different `apply` for virtual partitions than for physical partitions, see: https://github.com/modin-project/modin/blob/fde216da6ffcc72e2815e776e796f9aafac42ef5/modin/core/dataframe/base/partitioning/axis_partition.py#L26-L32

vs

https://github.com/modin-project/modin/blob/4ec7f6347903f9133c65ebc5b6e0e15553b98577/modin/core/dataframe/pandas/partitioning/partition.py#L54

so in places where we call `apply` on a partition, where the partition can be virtual or physical (this may happen a lot, since we want to treat them identically), such as https://github.com/modin-project/modin/blob/fde216da6ffcc72e2815e776e796f9aafac42ef5/modin/core/dataframe/pandas/partitioning/partition_manager.py#L462-L472 and https://github.com/modin-project/modin/blob/fde216da6ffcc72e2815e776e796f9aafac42ef5/modin/core/dataframe/pandas/partitioning/partition_manager.py#L376-L383

we can (and have previously) passed these arbitrary positional args (`*args`) to `num_splits`.

This PR https://github.com/modin-project/modin/pull/4473 was supposed to handle this, but I'll open a new one that tries to address a broader mixing of the `*args` and `**kwargs` of a func to be applied, and the parent func.

noloerino:
## What do these changes do?
(picking up from #4600)

This introduces the `Invocable` object, which wraps tuples of `func, *args, **kwargs` that were previously being passed around the codebase, and led to possible mixing of modin-internal arguments with API-level ones.

- [x] Resolves #4597 
- [ ] tests added and passing


modin/core/execution/dask/common/engine_wrapper.py
```diff
@@ -13,36 +13,46 @@
...
-        *args : list
-            Additional positional arguments to be passed in `func`.
+        invocable : Invocable
+            The function with its args and kwargs to be deployed in a worker process.
+            In order to avoid spawning a worker process from within another worker process,
+            this invocable must have a callable func rather than a future of a function.
         num_returns : int, default: 1
             The number of returned objects.
-        pure : bool, optional
+        pure : bool, default: True
```
noloerino, line 43: `pure=True` is the default behavior of Dask ([see here](https://docs.dask.org/en/stable/futures.html#distributed.Client.submit)); making this the case in our codebase as well may help avoid potential confusion in the future.

modin/core/execution/dask/implementations/pandas_on_dask/partitioning/partition.py
```diff
@@ -84,27 +85,27 @@ def apply(self, func, *args, **kwargs):
         -----
         The keyword arguments are sent as a dictionary.
         """
-        call_queue = self.call_queue + [[func, args, kwargs]]
+        call_queue = self.call_queue + [Invocable(func=func, args=args, kwargs=kwargs)]
+        call_queue = [DaskWrapper.materialize_invocable(inv) for inv in call_queue]
```
noloerino, line 89: Adding this line was necessary because if we don't materialize here, `DaskWrapper.deploy` ends up getting Dask futures instead of `Invocable` objects, and those `Invocable` objects also end up having futures in its arguments.

modin/core/execution/dask/implementations/pandas_on_dask/partitioning/partition.py
```diff
@@ -131,34 +132,32 @@ def add_to_apply_calls(self, func, *args, **kwargs):
...
+            call_queue=self.call_queue
+            + [Invocable(func=func, args=args, kwargs=kwargs)],
         )
 
     def drain_call_queue(self):
         """Execute all operations stored in the call queue on the object wrapped by this partition."""
         if len(self.call_queue) == 0:
             return
-        call_queue = self.call_queue
+        call_queue = [DaskWrapper.materialize_invocable(inv) for inv in self.call_queue]
```
noloerino, line 144: see above comment on why we materialize early here

modin/core/execution/dask/implementations/pandas_on_dask/partitioning/partition.py
jeffreykennethli: Thanks for figuring this out. I wonder if you could run some test workflows to benchmark whether this change has any meaningful performance regression - although I agree with your intuition that it likely won't matter.

Another really messy (and possibly confusing) way is to not compose `call_queue` with `Invocables`, and instead with the original `(func, args, kwargs)`.

That means we go from `(func, args, kwargs)` in the API layer, convert to `Invocable` when passing them around the partition layer, and unpack them when dealing with the execution layer, which seems bad. Again, all of this is to make the partition layer cleaner, at the cost of... making the partition layer complex again. What do you think?

vnlitvinov: this should be a comment in-code somewhere for sure

modin/core/execution/ray/implementations/pandas_on_ray/partitioning/partition.py
```diff
@@ -13,6 +13,7 @@
 
 """Module houses class that wraps data (block partition) and its metadata."""
 
+from modin.utils import Invocable
```
vnlitvinov, line 16: please re-order the imports - we usually first import builtin and stdlib, then empty line, then from non-standard dependencies (like `pandas` or `ray`), then empty line, then Modin-defined packages

modin/core/execution/ray/implementations/pandas_on_ray/partitioning/partition.py
```diff
@@ -107,15 +108,15 @@ def apply(self, func, *args, **kwargs):
...
         if len(call_queue) > 1:
             logger.debug(f"SUBMIT::_apply_list_of_funcs::{self._identity}")
             result, length, width, ip = _apply_list_of_funcs.remote(call_queue, data)
         else:
             # We handle `len(call_queue) == 1` in a different way because
             # this dramatically improves performance.
-            func, args, kwargs = call_queue[0]
-            result, length, width, ip = _apply_func.remote(data, func, *args, **kwargs)
+            invocable = call_queue[0]
+            result, length, width, ip = _apply_func.remote(data, invocable)
```
vnlitvinov, line 119:
```suggestion
            result, length, width, ip = _apply_func.remote(data, call_queue[0])
```

modin/core/execution/ray/implementations/pandas_on_ray/partitioning/partition.py
```diff
@@ -166,14 +167,14 @@ def drain_call_queue(self):
         else:
             # We handle `len(call_queue) == 1` in a different way because
             # this dramatically improves performance.
-            func, args, kwargs = call_queue[0]
+            invocable = call_queue[0]
```
vnlitvinov, line 170:
```suggestion
```

modin/core/execution/ray/implementations/pandas_on_ray/partitioning/partition.py
```diff
@@ -166,14 +167,14 @@ def drain_call_queue(self):
...
             logger.debug(f"SUBMIT::_apply_func::{self._identity}")
             (
                 self._data,
                 self._length_cache,
                 self._width_cache,
                 self._ip_cache,
-            ) = _apply_func.remote(data, func, *args, **kwargs)
+            ) = _apply_func.remote(data, invocable)
```
vnlitvinov, line 177:
```suggestion
            ) = _apply_func.remote(data, call_queue[0])
```

modin/core/execution/ray/implementations/pandas_on_ray/partitioning/virtual_partition.py
```diff
@@ -154,48 +157,56 @@ def deploy_axis_func(
...
         num_splits : int
             The number of splits to return (see ``split_result_of_axis_func_pandas``).
         maintain_partitioning : bool
             If True, keep the old partitioning if possible.
             If False, create a new partition layout.
         *partitions : iterable
             All partitions that make up the full axis (row or column).
-        **kwargs : dict
-            Additional keywords arguments to be passed in `func`.
+        lengths : list, default: None
```
vnlitvinov, line 169:
```suggestion
        lengths : list, optional
```

modin/core/execution/ray/implementations/pandas_on_ray/partitioning/virtual_partition.py
```diff
@@ -154,48 +157,56 @@ def deploy_axis_func(
...
+            The list of lengths to shuffle the object.
+        manual_partition : bool, default: False
+            If True, partition the result with lengths.
```
vnlitvinov, line 172:
```suggestion
            If True, partition the result with `lengths`.
```

modin/core/execution/ray/implementations/pandas_on_ray/partitioning/virtual_partition.py
```diff
@@ -283,6 +297,10 @@ def apply(
             In this case, we have to return the partitioning to its previous
             orientation (the lengths will remain the same). This is ignored between
             two axis partitions.
+        lengths : list, default: None
+            The list of lengths to shuffle the object.
+        manual_partition : bool, default: False
+            If True, partition the result with lengths.
```
vnlitvinov, line 303:
```suggestion
        lengths : list, optional
            The list of lengths to shuffle the object.
        manual_partition : bool, default: False
            If True, partition the result with `lengths`.
```

modin/utils.py
```diff
@@ -40,6 +40,34 @@
...
+    """
+
+    __slots__ = ()
+
+    def __new__(cls, func, args=None, kwargs=None):  # noqa: GL08
+        if args is None:
+            args = ()
+        if kwargs is None:
+            kwargs = {}
+        return super().__new__(cls, func, args, kwargs)
```
vnlitvinov, line 68:
```suggestion
        return super().__new__(cls, func, args or (), kwargs or {})
```

vnlitvinov (commented):
Overall looks nice, left a bunch of small suggestions

modin/core/execution/dask/implementations/pandas_on_dask/partitioning/partition.py
noloerino: @vnlitvinov makes sense, I'll add it in.

@jeffreykennethli I'll see if I can find any performance changes. I'm not sure if it's possible to decouple `Invocable` from the call queue for Dask without making any changes in the `PandasDataframeAxisPartition` superclass, since that class constructs `Invocable`s and passes it to overridden functions; making the Dask virtual partition methods not use `Invocables` in the call queue would make reasoning about nested `Invocable`s (which take other `Invocable`s as function arguments) difficult.

modin/utils.py
noloerino: Wouldn't this technically change the behavior if the caller of `Invocable` modifies args/kwargs after the fact? For example,
```
func = ... # something random
kwargs = {}
obj = Invocable(func=func, kwargs=kwargs)
kwargs["key1"] = 0
```
In your suggested change, since `{}` is falsey, the `Invocable` would get a new dict as its field, whereas in the current code it would share a reference to the originally passed-in dictionary.
This case is pretty pathological and I don't think we'd have situations where args/kwargs get modified after this constructor is called, but I'd just like to check whether it's reasonable to program defensively here.

modin/core/execution/dask/implementations/pandas_on_dask/partitioning/virtual_partition.py
```diff
@@ -157,41 +159,61 @@ def deploy_axis_func(
...
+            If True, partition the result with lengths.
 
         Returns
         -------
         list
             A list of distributed.Future.
         """
-        lengths = kwargs.get("_lengths", None)
         result_num_splits = len(lengths) if lengths else num_splits
         return DaskWrapper.deploy(
```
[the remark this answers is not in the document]
noloerino: I think this would be a little counter-intuitive, since `deploy` is meant to return a Future. This behavior is consistent with the [`RayTask`](https://github.com/modin-project/modin/pull/4715/files#diff-c4ce653a77903b2db19f7ed67000840fcdd4ce2ec60610766df77a7cb48a7090) class, which currently also requires explicit calls to `ray.gather` on arguments being passed in.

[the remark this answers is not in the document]
noloerino: Actually, on further examination it seems like this problem is avoided altogether in Ray, and we just need to call `deserialize` at the call site ([here](https://github.com/modin-project/modin/blob/d04ab97e0227e2cc4c9bd46c6577c09851e94131/modin/core/execution/ray/implementations/pandas_on_ray/partitioning/partition.py#L390)).

modin/core/execution/dask/implementations/pandas_on_dask/partitioning/virtual_partition.py
```diff
@@ -157,41 +159,61 @@ def deploy_axis_func(
...
-            *partitions,
+            Invocable(
+                func=deploy_dask_func,
+                args=(
+                    DaskWrapper.materialize_invocable(
+                        Invocable(
+                            func=PandasDataframeAxisPartition.deploy_axis_func,
+                            args=(
+                                axis,
+                                DaskWrapper.materialize_invocable(invocable),
```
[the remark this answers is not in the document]
noloerino: You're right, I'll remove this. It's left over from earlier commits where I didn't have `materialize_invocable` work recursively.

modin/core/execution/ray/implementations/pandas_on_ray/partitioning/partition.py
noloerino: done

modin/core/execution/dask/implementations/pandas_on_dask/partitioning/partition.py
noloerino: Here's a quick and dirty benchmark for ten runs of this code, where the call queue reaches depth ~1000:
```
df = pd.concat([pd.Series(np.ones(10)), pd.Series(np.ones(10))], axis=1, ignore_index=True)
for i in range(1000):
    df = pd.concat([df, pd.Series(np.ones(10))], axis=1, ignore_index=True)
repr(df)
```
This branch (`a7ce90b7`): 19.60 s
Current master (`daa50b12`): 20.16 s

This branch is 3% faster over these samples, but I don't think the setup is rigorous enough to really consider it a speedup -- at the very least, this test seems to indicate no appreciable drop in performance.

modin/utils.py
```diff
@@ -40,6 +40,34 @@
...
+
```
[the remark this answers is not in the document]
noloerino: Good idea, I've now made this change.

jeffreykennethli (approved):

modin/core/execution/dask/common/engine_wrapper.py
```diff
@@ -13,36 +13,46 @@
...
 class DaskWrapper:
     """The class responsible for execution of remote operations."""
 
     @classmethod
-    def deploy(cls, func, *args, num_returns=1, pure=None, **kwargs):
+    def deploy(
+        cls,
+        invocable,
+        num_returns=1,
+        pure=True,
```
YarShev, line 30: Why is this changed to pure=True?

modin/core/execution/dask/implementations/pandas_on_dask/partitioning/partition.py
YarShev: I am not sure we want to materialize these here instead of allowing Dask to do that in a worker process. This causes inter-process communication to get materialized data in the main thread instead of a worker process where we are not blocked. Imagine a case when we have a cluster related case. The data may be transferred between nodes to get the data in the main thread.

This might be a related issue - https://github.com/dask/distributed/issues/4842. Probably, we can ping Dask guys to replay activity there.

modin/core/execution/dask/implementations/pandas_on_dask/partitioning/virtual_partition.py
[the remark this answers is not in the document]
YarShev: `client.submit` auto-materializes all futures contained in most of python containers. `ray_func.remote` can't do that so we have to call `ray.get` inside of a worker process. However, we shouldn't do data materialization in the main thread.

YarShev (changes requested):
I am not sure we should merge the changes until we check performance. As I see there might be performance penalties both for Ray and Dask engines.

Also, I was wondering if we could do preprocessing (`ray.put`. `client.scatter`) for the arguments to be passed in a ray/dask remote function? That may drastically increase performance.

modin/core/execution/dask/common/engine_wrapper.py
noloerino: I added a comment about this earlier, but it seems to have disappeared after a rebase. `pure=True` is the default behavior for Dask ([see here](https://docs.dask.org/en/stable/futures.html#distributed.Client.submit)); making this the case in our codebase as well may help avoid potential confusion in the future.

modin/core/execution/dask/implementations/pandas_on_dask/partitioning/partition.py
noloerino: In that case, it seems like this is something that has to be fixed on Dask's end, since the tree-recursive structure induced by `Invocable(func, args=[Invocable(...), ...])` isn't something we can easily schedule ourselves without significant changes.

Do you have any suggestions for a workaround @YarShev? We'd like to get this PR in to make the `apply` interface uniform between block and axis partitions, and I'm not sure what a clean way to do that without materializing early would be.

noloerino:
> Also, I was wondering if we could do preprocessing (ray.put. client.scatter) for the arguments to be passed in a ray/dask remote function? That may drastically increase performance.

From what I ran into while testing, there are some points in the codebase where the arguments are already Ray object IDs/Dask futures when they're passed into a function. I can try investigating further to see if this can be made uniform, perhaps such that  `Invocable` or some other wrapper function is always called with concrete Python objects as arguments, and then calls `ray.put`/`client.scatter` within its constructor.

YarShev:
As to the performance penalties I said earlier, you can see that in the following simple example.

```python
if __name__ == "__main__":
    import modin.pandas as pd
    import numpy as np
    from timeit import default_timer as dt

    data = np.random.randint(0, 100, size=(2**16, 2**12))
    df = pd.DataFrame(data)
    start = dt()
    result = df.abs()
    end = dt()
    print("time =", end - start)
```

**Ray, master**
time = 0.024973500025225803

**Ray, PR4715**
time = 0.047648999985540286

**Dask, master**
time = 0.4006148000189569

**Dask, PR4715**
time = 1.6876778000150807

I ran it in my laptop with 8 cores available.

modin/core/execution/dask/implementations/pandas_on_dask/partitioning/partition.py
YarShev: I do not think we should introduce the Invocable concept now because that hits perfromance. I suggest keeping just `func, *args, **kwargs`, which can be handled both Ray and Dask automatically, for partition.apply.

YarShev:
> > Also, I was wondering if we could do preprocessing (ray.put. client.scatter) for the arguments to be passed in a ray/dask remote function? That may drastically increase performance.
> 
> From what I ran into while testing, there are some points in the codebase where the arguments are already Ray object IDs/Dask futures when they're passed into a function. I can try investigating further to see if this can be made uniform, perhaps such that `Invocable` or some other wrapper function is always called with concrete Python objects as arguments, and then calls `ray.put`/`client.scatter` within its constructor.

Please look at my comment first [here](https://github.com/modin-project/modin/pull/4715#discussion_r942355255).

Yes, some of the arguments can already be futures. We should probably go through `*args` and `**kwargs`, find non-future objects and make those futures.

noloerino:
Thanks for the pointers. I've removed the `Invocable` class and instead replaced it with a type alias for a 3-tuple of `(func, args, kwargs)`. Here's the speeds:

**Ray, master:** 0.009s
**Dask, master:** 0.129s
**Ray, old PR version:** 0.012s
**Dask, old PR version:** 0.537s
**Ray, latest PR version:** 0.011s
**Dask, latest PR version:** 0.132s

I'll look into how easy it is to call `ray.put`/`client.scatter` over `args` and `kwargs`, and see if it's simple enough to include in this PR.

noloerino:
It seems that CI is still failing because when Dask encounters a nested tuple, where functions are embedded deeper inside the tuple, it seems to try to evaluate the functions in a depth-first fashion. Here's a simple example of what I mean:
```python

c = Client()

def f1(f, arg):
    print("in f1")
    return f(arg)

def f2(arg):
    print("in f2")
    return arg + 1

future = c.submit(f1, f2, 100) # passes
print(future)
print(c.gather(future))
print("passed")

future = c.submit(f1, (f2, 100)) # crashes
print(future)

print(c.gather(future))
```

In the first configuration, where `f1, f2, 100` are the arguments passed to `c.submit`, everything passes fine; `f1` is properly called with `f2, 100` as its arguments. In the second, where the arguments are instead `f1, (f2, 100)`, Dask attempts to evaluate `f2()` before `f1`.

<details>
<summary>stdout/stderr for evidence:</summary>
<pre>
<Future: pending, key: f1-fb4e274f293ee6ef2d392253a6fb1096>
in f1
in f2
101
passed
<Future: pending, key: f1-db5fb5becdfd2e9b2edf6ef903a81100>
in f2
2022-08-10 18:44:09,542 - distributed.worker - WARNING - Compute Failed
Key:       f1-db5fb5becdfd2e9b2edf6ef903a81100
Function:  execute_task
args:      ((<function f1 at 0x7fd1a8af67a0>, (<function f2 at 0x7fd1a8af6710>, 100)))
kwargs:    {}
Exception: 'TypeError("f1() missing 1 required positional argument: \'arg\'")'

  File "<string>", line 21, in <module>
  File "/Users/jhshi/opt/miniconda3/envs/modin-dev/lib/python3.10/site-packages/distributed/client.py", line 2210, in gather
    return self.sync(
  File "/Users/jhshi/opt/miniconda3/envs/modin-dev/lib/python3.10/site-packages/distributed/utils.py", line 338, in sync
    return sync(
  File "/Users/jhshi/opt/miniconda3/envs/modin-dev/lib/python3.10/site-packages/distributed/utils.py", line 405, in sync
    raise exc.with_traceback(tb)
  File "/Users/jhshi/opt/miniconda3/envs/modin-dev/lib/python3.10/site-packages/distributed/utils.py", line 378, in f
    result = yield future
  File "/Users/jhshi/opt/miniconda3/envs/modin-dev/lib/python3.10/site-packages/tornado/gen.py", line 762, in run
    value = future.result()
  File "/Users/jhshi/opt/miniconda3/envs/modin-dev/lib/python3.10/site-packages/distributed/client.py", line 2073, in _gather
    raise exception.with_traceback(traceback)
TypeError: f1() missing 1 required positional argument: 'arg'
</pre>
</details>

I'm not familiar enough with Dask to say whether or not this is a bug on Dask's end, since I think it's reasonable for the scheduler to try to evaluate depth-first, although this behavior is somewhat counterintuitive. It looks like we may have to go back to @jeffreykennethli 's original suggestion of just leaving the Dask call queue of flattened tuples.
> Another really messy (and possibly confusing) way is to not compose call_queue with Invocables, and instead with the original (func, args, kwargs).
> That means we go from (func, args, kwargs) in the API layer, convert to Invocable when passing them around the partition layer, and unpack them when dealing with the execution layer, which seems bad. Again, all of this is to make the partition layer cleaner, at the cost of... making the partition layer complex again.

YarShev:
@noloerino, please note that you are getting the error because f1 didn't get enough arguments. The following message says about that - `Exception: 'TypeError("f1() missing 1 required positional argument: \'arg\'")'`. You should unwrap the tuple with `*(f2, 100)` to get this worked.

YarShev:
Also, note that Dask may not work with tuples passed in to a remote function. That's why we have used a list for the call queue so far.

noloerino:
@YarShev, you're right, I set up my minimized example somewhat incorrectly, and I would have needed to `*` the tuple. That said, I think it still proves my point about Dask's evaluation order being different, as the `in f2` print statement is hit before `f1` is even called, which seems to imply that Dask is depth-first evaluating functions within tuples. As you say, this doesn't seem to be the case with lists, so I've now returned `call_queue` to using lists rather than tuples.

YarShev:
> the `in f2` print statement is hit before `f1` is even called, which seems to imply that Dask is depth-first evaluating functions within tuples

I don't see this behavior. My output is `in f1` first and `in f2` next.

noloerino:
What about this example?
```

c = Client()

def f1(arg):
    print("in f1, received", arg)

def f2(arg):
    print("in f2, received", arg)
    return arg + 1

future = c.submit(f1, (f2, 100))
```

My output:
```
in f2, received 100
in f1, received 101
```

Environment info, in case it's platform-specific:
- MacOS Monterey (12.4) on Apple Silicon
- Python 3.10.4
- Dask 2022.8.0

modin/utils.py
```diff
@@ -47,6 +47,58 @@
 """
 
 
+Invocable = list
```
vnlitvinov, line 50: this feels super-weird, but I guess it's because Dask is weird...

vnlitvinov (approved):
LGTM, but please also post the latest perf figures here as well.

noloerino:
Updated benchmarks on the above `df.abs()` code (avg of 10 runs for each):
- **Ray, master (fd225c52):** 0.014s
- **Dask, master:** 0.329s
- **Ray, latest PR version:** 0.017s
- **Dask, latest PR version:** 0.267s

It seems like some recent change to master has caused a slowdown with the Dask backend (compared to [my numbers from last week](https://github.com/modin-project/modin/pull/4715#issuecomment-1211312373)). There also was a sizable variance on my runs on master, ranging from 0.208 to 0.544 seconds per run, but regardless it seems like this PR no longer causes a large performance hit.

modin/utils.py
noloerino: Yes, unfortunately making this `Invocable = tuple` didn't work (I think because of the depth-first evaluation behavior I mentioned above).

YarShev:
> future = c.submit(f1, (f2, 100))

@noloerino, yes, now I see the same as you but I am not sure that is correct enough arguments pass to a remote function. Anyway, we understand that dask may not work with tuples.

YarShev:
> Updated benchmarks on the above `df.abs()` code (avg of 10 runs for each):
> 
> * **Ray, master (fd225c52):** 0.014s
> * **Dask, master:** 0.329s
> * **Ray, latest PR version:** 0.017s
> * **Dask, latest PR version:** 0.267s
> 
> It seems like some recent change to master has caused a slowdown with the Dask backend (compared to [my numbers from last week](https://github.com/modin-project/modin/pull/4715#issuecomment-1211312373)). There also was a sizable variance on my runs on master, ranging from 0.208 to 0.544 seconds per run, but regardless it seems like this PR no longer causes a large performance hit.

Can you also run a full axis operation?

modin/utils.py
```diff
@@ -47,6 +47,58 @@
...
+
+def call_invocable(invocable, *args, **kwargs):  # noqa: PR01, RT01
+    """
+    Call an ``Invocable`` with the provided arguments.
+
+    The ``*args`` and ``**kwargs`` passed to this function will be passed to ``func`` before
+    the ``Invocable``'s ``args`` and `kwargs` fields. Specifically, ``call_invocable(i, *args, **kwargs)``
+    is equivalent to ``i[0](*args, *i[1], **kwargs, **i[2])``.
+    """
+    return invocable[0](*args, *invocable[1], **kwargs, **invocable[2])
```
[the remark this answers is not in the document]
YarShev: Why only [0], [1], [2]?

noloerino:
@YarShev here's benchmarks for the average time of `df.describe()` over 5 runs:
**Ray, master (fd225c52):** 2.52s
**Dask, master:** 2.86s
**Ray, latest PR version:** 2.51s
**Dask, latest PR version:** 2.89s

modin/utils.py
noloerino: Sure, it might make sense to move it. Did you have a better place in mind for this?

modin/utils.py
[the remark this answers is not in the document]
noloerino: As described in `make_invocable`, and `Invocable` is a three-element list of `[func, args, kwargs]`, where `args` is the list of positional arguments, and `kwargs` is the list of keyword arguments. As such, `invocable[0]` gets the function to be called, `*invocable[1]` will destructure the list of positional args, and `**invocable[2]` will destructure the ilst of keyword args.

modin/core/execution/ray/implementations/pandas_on_ray/partitioning/partition.py
```diff
@@ -382,6 +380,7 @@ def _apply_func(partition, func, *args, **kwargs):  # pragma: no cover
     str
         The node IP address of the worker process.
     """
+    func, args, kwargs = deserialize(invocable)
```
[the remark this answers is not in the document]
noloerino: I ran `df.describe()` on a 20_000 x 10_000 dataframe of random ints (about as large as I could get before spilling to disk). On average, this PR finished the code in 6.029 seconds, while current master (af8600) finished in 6.058.

I do agree that there may be potential for lost parallelism here, and in this PR's [change to `deserialize` itself](https://github.com/modin-project/modin/pull/4715/files/456b885bc7d5756d8b89d12108cfb92b6c9c8663#diff-7ddcb1cabfe012c33c62919f2da1e7e4cb3d783907e45e5a1492c73d7bb377cdR221). However, since `Invocable` is now a simple `list` again, I should be able to revert this with little trouble.

YarShev:
I suggest getting rid of the Invocable stuff for now because I am still seeing the significant performance penalties on the workload I am currently profiling. You can find it in https://github.com/YarShev/daal4py/tree/nyc-taxi/examples/notebooks/NYCTaxi-E2E-Demo. The script is large_data.py, which is based on NYCTaxi-E2E-Large-Data-Optimized.ipynb.

112 CPUs
**master (ray)**
whole time = 383.454846611013
**PR 4715 (ray)**
701.9466575289844

noloerino:
I did some more exploring, and I'm having issues reproducing your perf issues locally -- it seems like the NYC TLC has changed [their data](https://www1.nyc.gov/site/tlc/about/tlc-trip-record-data.page) to all be Parquet files, and have removed lat/long information from their datasets. Running a modified (and smaller) version of your large_data.py on my laptop doesn't exhibit any slowdowns, though since you said your benchmark was on 112 cores, I would guess this is something I'd have to run on AWS to reproduce.

Regardless, I agree that removing the invocable abstraction might be the way to go for now, as I would guess the extra `deserialize` calls are the source of the slowdowns, and unlike Dask, Ray seems to not recursively resolve promises (`ray.get([ray.put([ray.put(0)])])` returns `[[ObjectRef(...)]]` for me). I'll see if I can find a good way to preserve the separation of user args/kwargs without passing around an explicit Invocable object.

noloerino:
@YarShev I've removed the extraneous `deserialize` calls, which should bring performance back to expected levels. Unfortunately I still had trouble running your benchmark code since I couldn't find the CSV dataset with lat/long information, so could you please try running things again on your end? If the slowdown is still around, I'll see if I can find a comparably large dataset to test against.

YarShev:
@noloerino, please make CI green, after that I will run the benchmark in order not to encounter any trouble.

noloerino:
Should be fixed now @YarShev, (the failing FuzzyData on Dask check seems to be unrelated, and is failing on current master for me locally).

YarShev:
**master**
whole time:  236.58521463605575

**PR4715**
whole time:  281.4897154511418

YarShev:
Let me look into the changes.

modin/core/execution/ray/implementations/pandas_on_ray/partitioning/virtual_partition.py
```diff
@@ -279,26 +298,39 @@ def apply(
...
             num_splits = 1
         if len(self.call_queue) > 0:
             self.drain_call_queue()
-        kwargs["args"] = args
         result = super(PandasOnRayDataframeVirtualPartition, self).apply(
             func,
-            num_splits,
-            other_axis_partition,
-            maintain_partitioning,
+            *deserialize(args),
```
YarShev, line 328: This might be one of the places where we get slow as we are blocked in the main thread by calling ray.get. We likely want to call ray.get in worker processes in order for us not to be blocked. Maybe we should be thinking of a more graceful handling of the parameters.

noloerino: Yes, that makes sense. Since this issue seems to be in a virtual partition `apply` call, I've managed to find a mini-benchmark on AWS which exhibits a slowdown, so I'll continue looking into this.

noloerino:
I've refactored `deploy_ray_func`, which used to rely on some implicit behavior of `deserialize`, to be a little less confusing, and I've moved the extraneous `deserialize` call previously pointed out to be more precise and occur on a worker node.

I tried looking at a few micro-benchmarks on an AWS t2.2xlarge (8 cores, 32GB RAM) instance, but I couldn't find any substantial perf differences between this branch, the previous version of this PR, and latest master on `.groupby` or repeated `.apply` calls, though I was able to observe that adding a few extra unnecessary `deserialize` calls did cause a large slowdown.

@YarShev assuming CI passes, please run your benchmark again to see if anything has changed. If performance continues to be an issue, it would be helpful if we could set up a short meeting some time next week so I can have a reliable benchmark that I can check performance issues with myself.

YarShev:
Now the perf is on par with master on Ray engine, but there is a little slowdown on Dask engine.

modin/core/execution/dask/implementations/pandas_on_dask/partitioning/partition.py
```diff
@@ -308,20 +298,16 @@ def ip(self):
         return self._ip_cache
 
 
-def apply_func(partition, func, *args, **kwargs):
+def apply_func(partition, call_queue_entry):
```
YarShev, line 301: Let's revert this back. It might be a cause of the slowdown.
```suggestion
def apply_func(partition, func, *args, **kwargs):
```

noloerino: Fixed, and added a note about this in the docstring.

modin/core/execution/dask/implementations/pandas_on_dask/partitioning/partition.py
```diff
@@ -308,7 +300,7 @@ def ip(self):
         return self._ip_cache
 
 
-def apply_func(partition, func, *args, **kwargs):
+def apply_func(partition, func, f_args, f_kwargs):
```
YarShev, line 303:
```suggestion
def apply_func(partition, func, *args, **kwargs):
```
Why doesn't this work?

noloerino: Doing this made it easier to pass arguments to `DaskWrapper.deploy`, as [here](https://github.com/modin-project/modin/pull/4715/files/0939080168bad7f612e704adc45178401b9b2093#diff-9982d8740602ab4ffe8104fb6590b98d10b065e8731fa2cc4b286bc1e84c20d5R158), and this change fixed the performance difference between this PR and master on my micro-benchmark (calling `df.describe()` on a 2^16 x 2^12 dataframe). I'll refactor it to your version for consistency with Ray.

pyrito (commented):
@noloerino can you rebase this on master so that you can redo the OmniSci runs?

noloerino:
I've already rebased on latest master; Mahesh initiated a rerun of CI for me and it looks like Omnisci tests are passing.

pyrito:
I think Mahesh re-ran some other stuff. This seems to be failing: https://github.com/modin-project/modin/actions/runs/3048755188/jobs/4914151049

modin/core/execution/ray/implementations/pandas_on_ray/partitioning/virtual_partition.py
```diff
@@ -482,31 +514,46 @@ class PandasOnRayDataframeRowPartition(PandasOnRayDataframeVirtualPartition):
 
 
 @ray.remote
-def deploy_ray_func(func, *args, **kwargs):  # pragma: no cover
+def deploy_ray_func(
```
YarShev, line 517: Please make similar changes for `deploy_dask_func`.

modin/core/execution/ray/implementations/pandas_on_ray/partitioning/virtual_partition.py
```diff
@@ -293,26 +312,39 @@ def apply(
...
         -------
         list
             A list of `PandasOnRayDataframeVirtualPartition` objects.
+
+        Notes
+        -----
+        The ``args`` iterable is deserialized before being passed to the superclass ``apply``
+        method. In older versions of Modin, ``args`` was instead passed internally as
+        ``kwargs["args"]``, and this deserialization was handled in a special case in
+        ``deploy_ray_func``, making control flow difficult to follow.
```
YarShev, line 332: Is this actual?

modin/core/execution/ray/common/task_wrapper.py
```diff
@@ -23,51 +23,55 @@
...
-        **kwargs : dict
-            Additional keyword arguments to be passed in `func`.
 
         Returns
         -------
         ray.ObjectRef or list
             Ray identifier of the result being put to Plasma store.
         """
+        args = [] if f_args is None else f_args
+        kwargs = [] if f_kwargs is None else f_kwargs
```
YarShev, line 72:
```suggestion
        kwargs = {} if f_kwargs is None else f_kwargs
```

modin/core/execution/dask/implementations/pandas_on_dask/partitioning/virtual_partition.py
```diff
@@ -473,25 +510,25 @@ class PandasOnDaskDataframeRowPartition(PandasOnDaskDataframeVirtualPartition):
     axis = 1
 
 
-def deploy_dask_func(func, *args, **kwargs):
+def deploy_dask_func(func, f_args, f_kwargs):
```
YarShev, line 513: Please make the changes for this function as for `deploy_ray_func`.

modin/core/execution/dask/implementations/pandas_on_dask/partitioning/partition.py
```diff
@@ -317,31 +311,36 @@ def apply_func(partition, func, *args, **kwargs):
...
     -------
     pandas.DataFrame
         The resulting pandas DataFrame.
     str
         The node IP address of the worker process.
+
+    Notes
+    -----
+    Directly passing a call queue entry (i.e. a list of [func, args, kwargs]) instead of
+    destructuring it causes a performance penalty.
```
YarShev, line 330: This is applicable for Ray too so let's add the changes for Ray engine as well.

noloerino: Added this comment to ray's _apply_func.

modin/core/execution/dask/implementations/pandas_on_dask/partitioning/virtual_partition.py
noloerino: I'm assuming you mean to have the signature match (e.g. `deployer, axis, func_to_deploy, ...` etc.). I've now made this change as well.

modin/core/execution/ray/common/task_wrapper.py
noloerino: Fixed.

modin/core/execution/ray/implementations/pandas_on_ray/partitioning/virtual_partition.py
noloerino: `args` isn't deserialized here (this was left over from when I was messing around with deserializing in different spots), and instead deserialization takes place in `deploy_ray_func` as before. However, it is true that we were previously passing positional arguments as `kwargs["args"]` (you can see it in the diff, and [here](https://github.com/modin-project/modin/blob/7c260ad55d40bd31de27a21c66f47202e430b66e/modin/core/execution/ray/implementations/pandas_on_ray/partitioning/virtual_partition.py#L310)).

noloerino: I've changed the comments to reflect this.

modin/core/execution/dask/implementations/pandas_on_dask/partitioning/virtual_partition.py
```diff
@@ -164,39 +167,59 @@ def deploy_axis_func(
...
 
         Returns
         -------
         list
             A list of distributed.Future.
         """
-        lengths = kwargs.get("_lengths", None)
         result_num_splits = len(lengths) if lengths else num_splits
         return DaskWrapper.deploy(
             deploy_dask_func,
```
YarShev, line 193:
```suggestion
            func=deploy_dask_func,
```
for consistency

YarShev (approved):
Left a minor comment, other than that, LGTM!

modin/core/execution/ray/implementations/pandas_on_ray/partitioning/virtual_partition.py
```diff
@@ -482,31 +514,46 @@ class PandasOnRayDataframeRowPartition(PandasOnRayDataframeVirtualPartition):
...
+    axis : {0, 1}
+        The axis to perform the function along.
+    f_to_deploy : callable or RayObjectID
+        The function to deploy.
+    f_args : list or tuple
+        Positional arguments to pass to ``f_to_deploy``.
+    f_kwargs : dict
+        Keyword arguments to pass to ``f_to_deploy``.
+    *args : list
+        Positional arguments to pass to ``func``.
```
YarShev, line 542: What is `func` here?

modin/core/execution/ray/implementations/pandas_on_ray/partitioning/virtual_partition.py
```diff
@@ -482,31 +514,46 @@ class PandasOnRayDataframeRowPartition(PandasOnRayDataframeVirtualPartition):
...
     **kwargs : dict
-        Additional keyword arguments to be passed in `func`.
+        Keyword arguments to pass to ``func``.
```
YarShev, line 544: Same, please check all the docstrings.

modin/core/execution/ray/implementations/pandas_on_ray/partitioning/virtual_partition.py
noloerino: This should be `deployer`.

noloerino:
Thanks, I've gone through and cleaned up the docstrings.

YarShev (approved):
LGTM, assuming tests pass. Thanks.

PR #4718, opened 2022-07-26
REFACTOR-#4717: Improve PartitionMgr.get_indices() usage

### LINKED ISSUE #4717: Improve PartitionManager.get_indices() usage
### Describe the problem
`.get_indices()` requires a non-empty `index_func` argument to extract the index, though in most places can be defaulted to `lambda df: df.axes[axis]`.

Also it might be useful to return the indices of partitions as well as merged index along the axis.

vnlitvinov:

## What do these changes do?

Make some sane default for `index_func=None` argument of `PartitionManager.get_indices()`, which improves its usability.
Also return internal partition indices as well as the total frame index - this allows to compute the `row_lengths` / `column_widths` information later on without some separate remote calls.

- [x] Resolves #4717
- [ ] tests added and passing

### DIFF
--- a/modin/core/dataframe/pandas/partitioning/partition_manager.py
+++ b/modin/core/dataframe/pandas/partitioning/partition_manager.py
@@ -848,6 +848,8 @@ class PandasDataframePartitionManager(ClassLogger, ABC):
         -------
         pandas.Index
             A pandas Index object.
+        list of pandas.Index
+            The list of internal indices for each partition.
 
         Notes
         -----
@@ -855,21 +857,16 @@ class PandasDataframePartitionManager(ClassLogger, ABC):
         when you have deleted rows/columns internally, but do not know
         which ones were deleted.
         """
+        if index_func is None:
+            index_func = lambda df: df.axes[axis]  # noqa: E731
         ErrorMessage.catch_bugs_and_request_email(not callable(index_func))
         func = cls.preprocess_func(index_func)
-        if axis == 0:
-            new_idx = (
-                [idx.apply(func) for idx in partitions.T[0]]
-                if len(partitions.T)
-                else []
-            )
-        else:
-            new_idx = (
-                [idx.apply(func) for idx in partitions[0]] if len(partitions) else []
-            )
+        target = partitions.T if axis == 0 else partitions
+        new_idx = [idx.apply(func) for idx in target[0]] if len(target) else []
         new_idx = cls.get_objects_from_partitions(new_idx)
         # TODO FIX INFORMATION LEAK!!!!1!!1!!
-        return new_idx[0].append(new_idx[1:]) if len(new_idx) else new_idx
+        total_idx = new_idx[0].append(new_idx[1:]) if new_idx else new_idx
+        return total_idx, new_idx
 
     @classmethod
     def _apply_func_to_list_of_partitions_broadcast(
--- a/modin/experimental/batch/pipeline.py
+++ b/modin/experimental/batch/pipeline.py
@@ -356,40 +356,30 @@ class PandasQueryPipeline(object):
                 outs.append(output_partitions)
             else:
                 outs[node.output_id] = output_partitions
-        if not self.is_output_id_specified:
-            final_results = []
-            for df in outs:
-                partitions = []
-                for row_partition in df:
-                    partitions.append(row_partition.list_of_partitions_to_combine)
-                partitions = np.array(partitions)
-                partition_mgr_class = PandasOnRayDataframe._partition_mgr_cls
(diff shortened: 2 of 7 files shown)


modin/core/dataframe/pandas/partitioning/partition_manager.py
```diff
@@ -848,28 +848,25 @@ def get_indices(cls, axis, partitions, index_func=None):
...
-                else []
-            )
-        else:
-            new_idx = (
-                [idx.apply(func) for idx in partitions[0]] if len(partitions) else []
-            )
+        target = partitions.T if axis == 0 else partitions
+        new_idx = [idx.apply(func) for idx in target[0]] if len(target) else []
         new_idx = cls.get_objects_from_partitions(new_idx)
         # TODO FIX INFORMATION LEAK!!!!1!!1!!
```
mvashishtha, line 867: I don't think there's any information leak here anymore :) I think the PR that added `get_objects_from_partitions` deleted it the access to `partitions[i]._data`. Could you delete this warning?

mvashishtha (commented):
@vnlitvinov thank you, this is excellent! I left one minor cleanup comment.

modin/experimental/batch/pipeline.py
```diff
@@ -356,40 +356,30 @@ def compute_batch(
...
+            result_modin_frame = PandasOnRayDataframe(
+                partitions,
+                index,
+                columns,
+                row_lengths=tuple(map(len, internal_rows)),
+                column_widths=tuple(map(len, internal_cols)),
+            )
+            query_compiler = PandasQueryCompiler(result_modin_frame)
+            result_df = pd.DataFrame(query_compiler=query_compiler)
+            final_results[id] = result_df
```
RehanSD, line 383: Thank you - this refactor is way way neater lol!

modin/core/dataframe/pandas/partitioning/partition_manager.py
RehanSD: I just double checked, and the info leak is still there in the `PandasOnRayDataframePartitionManager`:
https://github.com/modin-project/modin/blob/49c03981c2cc55495fa0f95e17e6fd2b6e421d9b/modin/core/execution/ray/implementations/pandas_on_ray/partitioning/partition_manager.py#L110

RehanSD: I think it's doing this because `partition.get` will materialize partitions in serial, so it makes sense for it to do it, but the problem is if the partition has a call queue that will be ignored. Perhaps this is ok though?

RehanSD (approved):
LGTM - lets merge after @mvashishtha's comments are resolved!

modin/core/dataframe/pandas/partitioning/partition_manager.py
mvashishtha: I dug into the history of this comment, and I still don't know what it's talking about. The diff from the original PR is [here](https://github.com/modin-project/modin/pull/721/files#diff-0c6caf26a29eab528096f123ec3ac58888509dcd47817c0a4d45a781e60ed06d). I don't see any information leak there, where the code uses `get()` to get the data.

Regarding the `_data` access in the ray partition manager, I think you're correct that it's incorrect because it doesn't drain the call queue. This use in `get_indices` happens to work because `apply` drain the call queue. I think we need to take a stance as to whether we should allow accessing the `_data` of  block/non-full-axis-virtual partitions. We should probably use `get()`, which drains the call queue, instead.

So back to this comment, I think the right thing would be to file an issue to drain the call queue before getting `_data` (or maybe just to get rid of all accesses to `_data`), and also to move this comment to `get_objects_from_partitions` until that issue is fixed. But I don't think that's a responsibility for this PR.

YarShev: Related issue #4530

YarShev (approved):
LGTM!

mvashishtha (approved):

modin/core/dataframe/pandas/partitioning/partition_manager.py
RehanSD: Sgtm - I'll go ahead and approve and merge then.

RehanSD (approved):

RehanSD (approved):
LGTM!

modin/core/dataframe/pandas/partitioning/partition_manager.py
vnlitvinov: Yeah, I think we should get rid of `._data` access anywhere but in the partition class itself (ideally).

PR #4721, opened 2022-07-26
FIX-#4672: Fix incorrect warning when setting frame.columns or frame.index

### LINKED ISSUE #4672: BUG: setting dataframe index or columns raises UserWarning: Modin doesn't allow columns to be created via a new attribute name
### System information
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: MacOS Monterey 12.4
- **Modin version** (`modin.__version__`): 05933a5f27fb96f5a7ff6025ae2573d033a31b11
- **Python version**: 3.9.12

```python
import modin.pandas as pd
pd.DataFrame().index = []
```

### Describe the problem
Modin raises `UserWarning: Distributing <class 'NoneType'> object. This may take some time.
UserWarning: Modin doesn't allow columns to be created via a new attribute name - see https://pandas.pydata.org/pandas-docs/stable/indexing.html#attribute-access` when we set `column` in the code above, but pandas does not.

### Source code / logs
We should fix the check [here](https://github.com/modin-project/modin/blob/eddfda4b521366c628596dcb5c21775c7f50eec1/modin/pandas/dataframe.py#L2506) so that it only warns when setting an attribute like `index` or columns` that is already in `dir(self)`.

jbrockmendel:
Signed-off-by: Brock <jbrockmendel@gmail.com>

## What do these changes do?

Stop emitting incorrect warning.

- [x] Resolves #4672 
- [x] tests added and passing

### DIFF
--- a/modin/pandas/test/dataframe/test_default.py
+++ b/modin/pandas/test/dataframe/test_default.py
@@ -16,9 +16,13 @@ import numpy as np
 import pandas
 import matplotlib
 import modin.pandas as pd
-from modin.utils import to_pandas
+from modin.utils import (
+    to_pandas,
+    get_current_execution,
+)
 from numpy.testing import assert_array_equal
 import io
+import warnings
 
 from modin.pandas.test.utils import (
     df_equals,
@@ -1171,3 +1175,19 @@ def test_hasattr_sparse(is_sparse_data):
         else create_test_dfs(test_data["float_nan_data"])
     )
     eval_general(modin_df, pandas_df, lambda df: hasattr(df, "sparse"))
+
+
+def test_setattr_axes():
+    # Test that setting .index or .columns does not warn
+    df = pd.DataFrame([[1, 2], [3, 4]])
+    with warnings.catch_warnings():
+        if get_current_execution() != "BaseOnPython":
+            # In BaseOnPython, setting columns raises a warning because get_axis
+            #  defaults to pandas.
+            warnings.simplefilter("error")
+        df.index = ["foo", "bar"]
+        df.columns = [9, 10]
+
+    # Check that ensure_index was called
+    pandas.testing.assert_index_equal(df.index, pandas.Index(["foo", "bar"]))
+    pandas.testing.assert_index_equal(df.columns, pandas.Index([9, 10]))
--- a/docs/release_notes/release_notes-0.16.0.rst
+++ b/docs/release_notes/release_notes-0.16.0.rst
@@ -24,6 +24,7 @@ Key Features and Updates
   * FIX-#4491: Wait for all partitions in parallel in benchmark mode (#4656)
   * FIX-#4358: MultiIndex `loc` shouldn't drop levels for full-key lookups (#4608)
   * FIX-#4658: Expand exception handling for `read_*` functions from s3 storages (#4659)
+  * FIX-#4672: Fix incorrect warning when setting `frame.index` or `frame.columns` (#4721)
 * Performance enhancements
   * PERF-#4182: Add cell-wise execution for binary ops, fix bin ops for empty dataframes (#4391)
   * PERF-#4288: Improve perf of `groupby.mean` for narrow data (#4591)
--- a/modin/pandas/dataframe.py
+++ b/modin/pandas/dataframe.py
@@ -2367,7 +2367,7 @@ class DataFrame(DataFrameCompat, BasePandasDataset):
             # `__getattr__` will return the columns not present in `dir(self)`, so we do not need
             # to manually track this state in the `dir`.
             return
-        elif is_list_like(value):
+        elif is_list_like(value) and key not in ["index", "columns"]:
             warnings.warn(
                 "Modin doesn't allow columns to be created via a new attribute name - see "
                 + "https://pandas.pydata.org/pandas-docs/stable/indexing.html#attribute-access",


modin/pandas/test/dataframe/test_default.py
```diff
@@ -1171,3 +1171,15 @@ def test_hasattr_sparse(is_sparse_data):
         else create_test_dfs(test_data["float_nan_data"])
     )
     eval_general(modin_df, pandas_df, lambda df: hasattr(df, "sparse"))
+
+
+def test_setattr_axes():
+    # Test that setting .index or .columns does not warn
+    df = pd.DataFrame(np.random.randn(2, 2))
```
mvashishtha, line 1178:
```suggestion
    df = pd.DataFrame([[1, 2], [3, 4]])
```

We should keep unit test data deterministic. (Even though the data really shouldn't affect whether this test case passes.)

mvashishtha (commented):
@jbrockmendel thank you so much for fixing this. I have a minor comment.

mvashishtha (dismissed):
LGTM!

modin/pandas/test/dataframe/test_default.py
```diff
@@ -1171,3 +1171,15 @@ def test_hasattr_sparse(is_sparse_data):
         else create_test_dfs(test_data["float_nan_data"])
     )
     eval_general(modin_df, pandas_df, lambda df: hasattr(df, "sparse"))
+
+
+def test_setattr_axes():
+    # Test that setting .index or .columns does not warn
+    df = pd.DataFrame([[1, 2], [3, 4]])
+    with pytest.warns(None):
```
mvashishtha, line 1179: I learned while working on something similar today that `pytest.warn(None)` [doesn't check](https://stackoverflow.com/q/45671803/17554722) that we are not issuing warnings. The [recommended way](
https://docs.pytest.org/en/7.0.x/how-to/capture-warnings.html#additional-use-cases-of-warnings-in-tests) to check for no warnings is with `warnings.simplefilter`, but we do the equivalent in the entire project's setup.cfg [here](https://github.com/modin-project/modin/blob/38450e10ad13851b61ccc0690c3a98f63b4dc640/setup.cfg#L210), but we skip that check for this entire test [here](https://github.com/modin-project/modin/blob/38450e10ad13851b61ccc0690c3a98f63b4dc640/modin/pandas/test/dataframe/test_default.py#L54) due to #3655.

The end result is that this test doesn't reproduce the bug #4672. Sadly, the [only way](https://stackoverflow.com/a/53363394/17554722) to check that we are NOT warning here but leave the carvout for the rest of the test cases in the file is to do `pytest.mark.parametrize` for every test in the file. We could put every single other test in this file under a new test class, and wrap that class with `@pytest.mark.filterwarnings(default_to_pandas_ignore_string)`.

Let me know if you see a better solution.

modin/pandas/test/dataframe/test_default.py
```diff
@@ -1171,3 +1171,15 @@ def test_hasattr_sparse(is_sparse_data):
...
+        df.index = ["foo", "bar"]
```
mvashishtha, line 1180: `pytest.warns(None)` [doesn't check](https://stackoverflow.com/q/45671803/17554722) that we're not warning, so this test case doesn't reproduce the bug. The [recommended solution](https://docs.pytest.org/en/7.0.x/how-to/capture-warnings.html#additional-use-cases-of-warnings-in-tests) is to use `warnings.catch_warnings()` and `warnings.simplefilter("error")`. I just checked that that reproduces the bug.

```suggestion
    with warnings.catch_warnings():
        warnings.simplefilter("error")
        df.index = ["foo", "bar"]
```

You'll also have to `import warnings` at the top of the file.

jbrockmendel: could use `pandas.testing.assert_produces_warning(None):`

mvashishtha (dismissed):

modin/pandas/test/dataframe/test_default.py
mvashishtha: looks like `assert_produces_warning` doesn't live in `pandas.testing` anymore, and it was [moved](https://stackoverflow.com/a/63532540/17554722) to [`pandas/_testing`](https://github.com/pandas-dev/pandas/blob/c9c6685c51ead26bbbb9a0dd565e82967cd839e8/pandas/_testing/_warnings.py#L19). I guess we should use `catch_warnings` after all.

modin/pandas/test/dataframe/test_default.py
```diff
@@ -1171,3 +1172,16 @@ def test_hasattr_sparse(is_sparse_data):
         else create_test_dfs(test_data["float_nan_data"])
     )
     eval_general(modin_df, pandas_df, lambda df: hasattr(df, "sparse"))
+
+
+def test_setattr_axes():
+    # Test that setting .index or .columns does not warn
+    df = pd.DataFrame([[1, 2], [3, 4]])
+    with warnings.catch_warnings():
+        warnings.simplefilter("error")
```
mvashishtha, line 1181:
```suggestion
        # In BaseOnPython, setting columns raises a warning because get_axis
        # defaults to pandas.
        if get_current_execution() != "BaseOnPython"
                warnings.simplefilter("error")
```

This should fix the CI failure in BaseOnPython: https://github.com/modin-project/modin/runs/7545798761?check_suite_focus=true

modin/pandas/test/dataframe/test_default.py
```diff
@@ -1171,3 +1172,19 @@ def test_hasattr_sparse(is_sparse_data):
...
+
+
+def test_setattr_axes():
+    # Test that setting .index or .columns does not warn
+    df = pd.DataFrame([[1, 2], [3, 4]])
+    with warnings.catch_warnings():
+        # In BaseOnPython, setting columns raises a warning because get_axis
+        # defaults to pandas.
+        if get_current_execution() != "BaseOnPython"
+                warnings.simplefilter("error")
```
mvashishtha, line 1184:
```suggestion
            warnings.simplefilter("error")
```

modin/pandas/test/dataframe/test_default.py
```diff
@@ -1171,3 +1172,19 @@ def test_hasattr_sparse(is_sparse_data):
...
     eval_general(modin_df, pandas_df, lambda df: hasattr(df, "sparse"))
+
+
+def test_setattr_axes():
+    # Test that setting .index or .columns does not warn
+    df = pd.DataFrame([[1, 2], [3, 4]])
+    with warnings.catch_warnings():
+        # In BaseOnPython, setting columns raises a warning because get_axis
+        # defaults to pandas.
+        if get_current_execution() != "BaseOnPython"
```
mvashishtha, line 1183:
```suggestion
        if get_current_execution() != "BaseOnPython":
```

modin/pandas/test/dataframe/test_default.py
```diff
@@ -1171,3 +1172,19 @@ def test_hasattr_sparse(is_sparse_data):
...
     eval_general(modin_df, pandas_df, lambda df: hasattr(df, "sparse"))
+
+
+def test_setattr_axes():
+    # Test that setting .index or .columns does not warn
+    df = pd.DataFrame([[1, 2], [3, 4]])
+    with warnings.catch_warnings():
+        # In BaseOnPython, setting columns raises a warning because get_axis
+        # defaults to pandas.
+        if get_current_execution() != "BaseOnPython":
```
mvashishtha, line 1183: @jbrockmendel could you please change `from modin.utils import to_pandas` to `from modin.utils import to_pandas, get_current_execution` to the top of this file? I think that should fix the tests at last.

You can run the BaseOnPython test `pytest modin/pandas/test/dataframe/test_default.py -k setattr --execution=BaseOnPython`

mvashishtha (approved):

PR #4724, opened 2022-07-27
FIX-#3180: Stop defaulting to pandas for paths relative to home.

### LINKED ISSUE #3180: `read_csv` fallbacks to pandas implementation if file path given relative to the home directory
Initially was found in the https://github.com/modin-project/modin/issues/3170#issuecomment-866513968.

mvashishtha:

## What do these changes do?

FIX-#3180: Stop defaulting to pandas for paths relative to home.

- [x] Resolves #3180
- [x] tests added and passing

### DIFF
--- a/modin/core/io/file_dispatcher.py
+++ b/modin/core/io/file_dispatcher.py
@@ -216,7 +216,7 @@ class FileDispatcher(ClassLogger):
         if isinstance(file_path, str) and S3_ADDRESS_REGEX.search(file_path):
             return file_path
         else:
-            return os.path.abspath(file_path)
+            return os.path.abspath(os.path.expanduser(file_path))
 
     @classmethod
     def file_size(cls, f):
@@ -284,7 +284,7 @@ class FileDispatcher(ClassLogger):
                     pass
                 s3fs = S3FS.S3FileSystem(anon=True, **new_storage_options)
                 return exists or s3fs.exists(file_path)
-        return os.path.exists(file_path)
+        return os.path.exists(os.path.expanduser(file_path))
 
     @classmethod
     def deploy(cls, func, *args, num_returns=1, **kwargs):  # noqa: PR01
--- a/modin/pandas/test/test_io.py
+++ b/modin/pandas/test/test_io.py
@@ -197,7 +197,6 @@ def patch_expanduser(input_path_to_fake, fake_expansion, monkeypatch):
     original_expanduser = os.path.expanduser
 
     def patched_expanduser(path):
-        print(f"calling patched expanduser with path {path}")
         if path == input_path_to_fake:
             return fake_expansion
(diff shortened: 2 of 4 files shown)


modin/pandas/test/test_io.py
```diff
@@ -1291,6 +1306,21 @@ def test_to_csv_with_index(self):
...
 
+# Leave this test apart from the test classes, which skip the default to pandas
+# warning check. We want to make sure we are NOT defaulting to pandas for a
+# path relative to user home.
+# TODO(https://github.com/modin-project/modin/issues/3655): Get rid of this
+# commment once we turn all default to pandas messages into errors.
+def test_read_csv_relative_to_user_home(monkeypatch, make_csv_file):
+    unique_filename = get_unique_filename()
+
+    patch_expanduser("~/testfile.csv", unique_filename, monkeypatch)
```
vnlitvinov, line 1317: If we would ever change our implementation to stop using `os.path.expanduser()` but expand that `~` in some other way this test would break, but the implementation would be correct...

Why don't we change `HOME` environment variable to some temp directory for the duration of this test, and let that ~/testfile.csv` point to a file in that directory instead?

mvashishtha: Ah, I see that's a great idea because both windows and Posix should count `$HOME` as home. Done.

modin/core/io/file_dispatcher.py
```diff
@@ -216,7 +216,7 @@ def get_path(cls, file_path):
         if isinstance(file_path, str) and S3_ADDRESS_REGEX.search(file_path):
             return file_path
         else:
-            return os.path.abspath(file_path)
+            return os.path.abspath(os.path.expanduser(file_path))
```
pyrito, line 219: What's the general rule on trying to support something that `pandas` currently doesn't support? For example, if you pass in a path like `$HOME/blah/blahhhh/` that isn't automatically expanded in pandas, but is something that we could try to handle.

pyrito (approved):
LGTM!

modin/core/io/file_dispatcher.py
vnlitvinov: We generally use `modin.experimental` namespace for that. That said, on Windows `$HOME` is not expanded to anything, and I'm not sure where to draw the line at variables expansion on *nix, either - should we then try to support bash-ims like `$(date)/foo/a.csv`?

vnlitvinov (approved):
LGTM, thanks!

modin/core/io/file_dispatcher.py
pyrito: Fair enough!

vnlitvinov:
@mvashishtha you have some CI failures and a merge conflict, please fix

PR #4726, opened 2022-07-27
FEAT-#4725: Make index and columns lazy in Modin DataFrame

### LINKED ISSUE #4725: Support lazy initialization of index and columns in Modin Dataframe
Current logic requires passing a fully computed index (and columns) when creating `modin.core.dataframe.pandas.dataframe.dataframe.PandasDataframe`, however there might be cases when we do not know the index of resulting operation (e.g. when making `df[mask]` call) until that operation is complete. This requires to wait for the operation to complete.

The proposal is to allow making index be lazily computed if it wasn't provided upon `PandasDataframe` creation.

vnlitvinov:

## What do these changes do?
Allow _not_ specifying `index` and `columns` when constructing `PandasDataframe`, they would be computed on-demand when accessing `.index` and `.columns`.

- [x] Resolves #4725
- [x] tests added and passing

### DIFF
--- a/modin/core/dataframe/pandas/dataframe/dataframe.py
+++ b/modin/core/dataframe/pandas/dataframe/dataframe.py
@@ -351,9 +351,9 @@ class PandasDataframe(ClassLogger):
             An index object containing the row labels.
         """
         if self._index_cache is None:
-            self._index_cache, internal_sizes = self._compute_axis_labels(0)
+            self._index_cache, row_lengths = self._compute_axis_labels(0)
             if self._row_lengths_cache is None:
-                self._row_lengths_cache = internal_sizes
+                self._row_lengths_cache = row_lengths
         return self._index_cache
 
     def _get_columns(self):
@@ -366,9 +366,9 @@ class PandasDataframe(ClassLogger):
             An index object containing the column labels.
         """
         if self._columns_cache is None:
-            self._columns_cache, internal_sizes = self._compute_axis_labels(1)
+            self._columns_cache, column_widths = self._compute_axis_labels(1)
             if self._column_widths_cache is None:
-                self._column_widths_cache = internal_sizes
+                self._column_widths_cache = column_widths
         return self._columns_cache
 
     def _set_index(self, new_index):
--- a/modin/core/dataframe/pandas/dataframe/dataframe.py
+++ b/modin/core/dataframe/pandas/dataframe/dataframe.py
@@ -222,7 +222,7 @@ class PandasDataframe(ClassLogger):
                 any(val < 0 for val in column_widths),
                 "Column widths cannot be negative: {}".format(column_widths),
             )
-        self._filter_empties(compute=False)
+        self._filter_empties(compute_lengths=False)
 
     @property
     def _row_lengths(self):
@@ -436,7 +436,7 @@ class PandasDataframe(ClassLogger):
         -------
         pandas.Index
             Labels for the specified `axis`.
-        tuple of int
+        List of int
             Size of partitions alongsize specified `axis`.
         """
         if partitions is None:
@@ -444,16 +444,16 @@ class PandasDataframe(ClassLogger):
         new_index, internal_idx = self._partition_mgr_cls.get_indices(axis, partitions)
         return new_index, list(map(len, internal_idx))
 
-    def _filter_empties(self, compute=True):
+    def _filter_empties(self, compute_lengths=True):
         """
         Remove empty partitions from `self._partitions` to avoid triggering excess computation.
 
         Parameters
         ----------
-        compute : bool, default: True
+        compute_lengths : bool, default: True
             Trigger the computations for partition sizes if they're not done already.
         """
-        if not compute and (
+        if not compute_lengths and (
             self._index_cache is None
             or self._columns_cache is None
             or self._row_lengths_cache is None
--- a/modin/core/dataframe/pandas/dataframe/dataframe.py
+++ b/modin/core/dataframe/pandas/dataframe/dataframe.py
@@ -192,10 +192,10 @@ class PandasDataframe(ClassLogger):
         self._column_widths_cache = column_widths
         self._dtypes = dtypes
 
-        self._validate_axes_splits()
+        self._validate_axes_lengths()
         self._filter_empties(compute_metadata=False)
 
-    def _validate_axes_splits(self):
+    def _validate_axes_lengths(self):
         """Validate that labels are split correctly if split is known."""
         if self._row_lengths_cache is not None and len(self.index) > 0:
             # An empty frame can have 0 rows but a nonempty index. If the frame
@@ -238,9 +238,10 @@ class PandasDataframe(ClassLogger):
         """
         if self._row_lengths_cache is None:
             if len(self._partitions) > 0:
-                index, self._row_lengths_cache = self._compute_axis_labels_and_lengths(
-                    0
-                )
+                (
+                    index,
+                    self._row_lengths_cache,
+                ) = self._compute_axis_labels_and_lengths(0)
                 if self._index_cache is None:
                     self._index_cache = index
             else:
@@ -446,7 +447,7 @@ class PandasDataframe(ClassLogger):
         pandas.Index
             Labels for the specified `axis`.
         List of int
-            Size of partitions alongsize specified `axis`.
+            Size of partitions alongside specified `axis`.
         """
         if partitions is None:
             partitions = self._partitions
@@ -2198,7 +2199,7 @@ class PandasDataframe(ClassLogger):
             passed_len += len(internal)
         return result_dict
 
-    def __make_constructor_args(self, partitions, index, columns) -> dict:
+    def __make_init_labels_args(self, partitions, index, columns) -> dict:
         kw = {}
         kw["index"], kw["row_lengths"] = (
             self._compute_axis_labels_and_lengths(0, partitions)
@@ -2291,7 +2292,7 @@ class PandasDataframe(ClassLogger):
             keep_remaining,
         )
 
-        kw = self.__make_constructor_args(new_partitions, new_index, new_columns)
+        kw = self.__make_init_labels_args(new_partitions, new_index, new_columns)
         return self.__constructor__(new_partitions, **kw)
 
     @lazy_metadata_decorator(apply_axis="both")
@@ -2359,7 +2360,7 @@ class PandasDataframe(ClassLogger):
             keep_partitioning=True,
         )
         # Index objects for new object creation. This is shorter than if..else
-        kw = self.__make_constructor_args(new_partitions, new_index, new_columns)
+        kw = self.__make_init_labels_args(new_partitions, new_index, new_columns)
         if dtypes == "copy":
             kw["dtypes"] = self._dtypes
         elif dtypes is not None:
@@ -2751,7 +2752,7 @@ class PandasDataframe(ClassLogger):
         new_partitions = self._partition_mgr_cls.groupby_reduce(
             axis, self._partitions, by_parts, map_func, reduce_func, apply_indices
         )
-        kw = self.__make_constructor_args(new_partitions, new_index, new_columns)
+        kw = self.__make_init_labels_args(new_partitions, new_index, new_columns)
         return self.__constructor__(new_partitions, **kw)
 
     @classmethod
--- a/modin/core/dataframe/pandas/dataframe/dataframe.py
+++ b/modin/core/dataframe/pandas/dataframe/dataframe.py
@@ -143,10 +143,12 @@ class PandasDataframe(ClassLogger):
     ----------
     partitions : np.ndarray
         A 2D NumPy array of partitions.
-    index : sequence
+    index : sequence, optional
         The index for the dataframe. Converted to a ``pandas.Index``.
-    columns : sequence
+        Is computed from partitions on demand if not specified.
+    columns : sequence, optional
         The columns object for the dataframe. Converted to a ``pandas.Index``.
+        Is computed from partitions on demand if not specified.
     row_lengths : list, optional
         The length of each partition in the rows. The "height" of
         each of the block partitions. Is computed if not provided.
@@ -177,15 +179,19 @@ class PandasDataframe(ClassLogger):
     def __init__(
         self,
         partitions,
-        index,
-        columns,
+        index=None,
+        columns=None,
         row_lengths=None,
         column_widths=None,
         dtypes=None,
     ):
         self._partitions = partitions
-        self._index_cache = ensure_index(index)
-        self._columns_cache = ensure_index(columns)
+        self._index_cache = ensure_index(index) if index is not None else None
+        self._columns_cache = ensure_index(columns) if columns is not None else None
+        self._row_lengths_cache = row_lengths
+        self._column_widths_cache = column_widths
+        self._dtypes = dtypes
+
         if row_lengths is not None and len(self.index) > 0:
             # An empty frame can have 0 rows but a nonempty index. If the frame
             # does have rows, the number of rows must equal the size of the
@@ -200,7 +206,6 @@ class PandasDataframe(ClassLogger):
                 any(val < 0 for val in row_lengths),
                 "Row lengths cannot be negative: {}".format(row_lengths),
             )
-        self._row_lengths_cache = row_lengths
         if column_widths is not None and len(self.columns) > 0:
             # An empty frame can have 0 column but a nonempty column index. If
             # the frame does have columns, the number of columns must equal the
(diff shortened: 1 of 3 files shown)


YarShev:
Related discussion on handling metadata (index and columns) in #3673.

modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -423,14 +436,32 @@ def _compute_axis_labels(self, axis: int, partitions=None):
         -------
         pandas.Index
             Labels for the specified `axis`.
+        tuple of int
```
mvashishtha, line 439:
```suggestion
        List of int
```

modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -423,14 +436,32 @@ def _compute_axis_labels(self, axis: int, partitions=None):
...
+            Size of partitions alongsize specified `axis`.
         """
         if partitions is None:
             partitions = self._partitions
-        new_index, _ = self._partition_mgr_cls.get_indices(axis, partitions)
-        return new_index
+        new_index, internal_idx = self._partition_mgr_cls.get_indices(axis, partitions)
+        return new_index, list(map(len, internal_idx))
+
+    def _filter_empties(self, compute=True):
```
mvashishtha, line 447:
```suggestion
    def _filter_empties(self, compute_lengths=True):
```

I think `compute_lengths` is clearer (note you'll have to replace it in several places if you accept this suggestion).

modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -2154,6 +2185,18 @@ def _prepare_frame_to_broadcast(self, axis, indices, broadcast_all):
             passed_len += len(internal)
         return result_dict
 
+    def __make_constructor_args(self, partitions, index, columns) -> dict:
```
mvashishtha, line 2188: This is a good idea.

mvashishtha (commented):
I have some minor style comments. Thanks @vnlitvinov !

vnlitvinov:
> Related discussion on handling metadata (index and columns) in #3673.

That issue is talking about improving pivot speed if we can omit computing index and labels... well, after this PR we will be able to! 😄

modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -177,15 +179,19 @@ def __constructor__(self):
...
         column_widths=None,
         dtypes=None,
     ):
         self._partitions = partitions
-        self._index_cache = ensure_index(index)
-        self._columns_cache = ensure_index(columns)
+        self._index_cache = ensure_index(index) if index is not None else None
+        self._columns_cache = ensure_index(columns) if columns is not None else None
+        self._row_lengths_cache = row_lengths
+        self._column_widths_cache = column_widths
```
YarShev, line 192: I assume we should check `self._row_lengths_cache` (`self._column_widths_cache`) instead of `row_lengths` (column_widths) in the conditions below as we do with `self._index_cache` (`self._columns_cache`).

modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -345,6 +350,10 @@ def _get_index(self):
         pandas.Index
             An index object containing the row labels.
         """
+        if self._index_cache is None:
+            self._index_cache, internal_sizes = self._compute_axis_labels(0)
```
YarShev, line 354:
```suggestion
            self._index_cache, row_lengths = self._compute_axis_labels(0)
```

modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -356,6 +365,10 @@ def _get_columns(self):
         pandas.Index
             An index object containing the column labels.
         """
+        if self._columns_cache is None:
+            self._columns_cache, internal_sizes = self._compute_axis_labels(1)
```
YarShev, line 369:
```suggestion
            self._columns_cache, column_widths = self._compute_axis_labels(1)
```

modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -356,6 +365,10 @@ def _get_columns(self):
...
+            if self._column_widths_cache is None:
+                self._column_widths_cache = internal_sizes
```
YarShev, line 371:
```suggestion
                self._column_widths_cache = column_widths
```

modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -345,6 +350,10 @@ def _get_index(self):
         pandas.Index
             An index object containing the row labels.
         """
+        if self._index_cache is None:
+            self._index_cache, internal_sizes = self._compute_axis_labels(0)
+            if self._row_lengths_cache is None:
+                self._row_lengths_cache = internal_sizes
```
YarShev, line 356:
```suggestion
                self._row_lengths_cache = row_lengths
```

modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -423,14 +436,32 @@ def _compute_axis_labels(self, axis: int, partitions=None):
...
             Labels for the specified `axis`.
+        List of int
+            Size of partitions alongsize specified `axis`.
         """
         if partitions is None:
             partitions = self._partitions
-        new_index, _ = self._partition_mgr_cls.get_indices(axis, partitions)
-        return new_index
+        new_index, internal_idx = self._partition_mgr_cls.get_indices(axis, partitions)
+        return new_index, list(map(len, internal_idx))
```
YarShev, line 445:
```suggestion
        return new_index, internal_idx
```
Should we return external and internal labels since the function name says exactly about that? We will have many `list(map(len, internal_idx))` in different places though.

modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -423,14 +436,32 @@ def _compute_axis_labels(self, axis: int, partitions=None):
...
+        Parameters
+        ----------
+        compute_lengths : bool, default: True
+            Trigger the computations for partition sizes if they're not done already.
+        """
+        if not compute_lengths and (
+            self._index_cache is None
+            or self._columns_cache is None
+            or self._row_lengths_cache is None
+            or self._column_widths_cache is None
```
YarShev, line 460:
```suggestion
            self._index_cache is not None
            or self._columns_cache is not None
            or self._row_lengths_cache is not None
            or self._column_widths_cache is not None
```
Should this be so?

modin/core/dataframe/pandas/dataframe/dataframe.py
YarShev: Should this be moved closer to the constructor?

modin/core/dataframe/pandas/dataframe/dataframe.py
vnlitvinov: they're equal at this point, and current approach is shorter :)

modin/core/dataframe/pandas/dataframe/dataframe.py
vnlitvinov: No, the idea is expressed right - if we don't know some of the caches, and we're requested to _not_ compute we do not go further, this is to delay the blocking on partitions as much as possible.

modin/core/dataframe/pandas/dataframe/dataframe.py
vnlitvinov: This is the reason why I decided to implement this `list(map())` here - to avoid adding it everywhere else.
Maybe you have a better naming idea for the function?

modin/core/dataframe/pandas/dataframe/dataframe.py
YarShev: I understand they are equal at this point but we should probably unify the use of these things because there might be some postprocessing of these things (either `self._row_lengths_cache` (`self._column_widths_cache`) or `row_lengths` (`column_widths`), either both) and we can lose the right things for the check.

modin/core/dataframe/pandas/dataframe/dataframe.py
YarShev: We already have similar things in `get_indices` (AFAIRC introduced by you too 😄). We should probably do there the same as here.

As to the function name, maybe just `_compute_axis`?

mvashishtha (dismissed):

modin/core/dataframe/pandas/dataframe/dataframe.py
vnlitvinov: `get_indices()` is slightly different, living in a different layer and returning indices (as its name implies).
So I don't like the vagueness of `_compute_axis`, though, so what about `_compute_axis_labels_and_size` or `_compute_axis_labels_and_splits`?

YarShev: `_compute_axis_labels_and_splits` is fine to me. Another example would be `_compute_axis_labels_and_lengths`? as we use `lengths` word in many places. Do you think which is better?

modin/core/dataframe/pandas/dataframe/dataframe.py
YarShev: ping

modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -345,6 +350,10 @@ def _get_index(self):
         pandas.Index
             An index object containing the row labels.
         """
+        if self._index_cache is None:
+            self._index_cache, row_lengths = self._compute_axis_labels(0)
+            if self._row_lengths_cache is None:
```
YarShev, line 355: What if `row_lengths` gets changed?

modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -356,6 +365,10 @@ def _get_columns(self):
         pandas.Index
             An index object containing the column labels.
         """
+        if self._columns_cache is None:
+            self._columns_cache, column_widths = self._compute_axis_labels(1)
+            if self._column_widths_cache is None:
```
YarShev, line 370: What if widths_cache gets changed but `_column_widths_cache` is not None?

modin/core/dataframe/pandas/dataframe/dataframe.py
YarShev: Why do we need to use `self._index_cache` and `self._columns_cache` in the condition if the docstring for `compute_lengths` says about partition sizes?

modin/core/dataframe/pandas/dataframe/dataframe.py
vnlitvinov: hmm, postprocessing is a valid point... let me fix that in a sec

modin/core/dataframe/pandas/dataframe/dataframe.py
vnlitvinov: renamed to `_compute_axis_labels_and_lengths()`, this expresses the returns better (though it added a few lines due to some lines getting long and `black` reformatting)

modin/core/dataframe/pandas/dataframe/dataframe.py
vnlitvinov: - This technically should have an `assert row_lengths == self._row_lengths_cache`, but I felt it would be too much of a performance penalty

- This should actually almost never happen that we have valid (non-`None`) row lengths but no index

modin/core/dataframe/pandas/dataframe/dataframe.py
vnlitvinov: renamed the parameter to `compute_metadata` to align its naming with the `lazy_metadata_decorator()`

modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -177,49 +179,52 @@ def __constructor__(self):
...
+        self._index_cache = ensure_index(index) if index is not None else None
+        self._columns_cache = ensure_index(columns) if columns is not None else None
+        self._row_lengths_cache = row_lengths
+        self._column_widths_cache = column_widths
+        self._dtypes = dtypes
+
+        self._validate_axes_splits()
+        self._filter_empties(compute_metadata=False)
+
+    def _validate_axes_splits(self):
```
YarShev, line 198:
```suggestion
    def _validate_axes_lengths(self):
```
to be consistent 😄

modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -232,10 +237,12 @@ def _row_lengths(self):
...
         """
         if self._row_lengths_cache is None:
-            if len(self._partitions.T) > 0:
-                self._row_lengths_cache = [
-                    obj.length() for obj in self._partitions.T[0]
-                ]
+            if len(self._partitions) > 0:
+                index, self._row_lengths_cache = self._compute_axis_labels_and_lengths(
+                    0
+                )
```
YarShev, line 243:
```suggestion
                (
                    index,
                    self._row_lengths_cache
                ) = self._compute_axis_labels_and_lengths(0)
```
Is that nice for black? 😄

modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -423,14 +445,32 @@ def _compute_axis_labels(self, axis: int, partitions=None):
         -------
         pandas.Index
             Labels for the specified `axis`.
+        List of int
+            Size of partitions alongsize specified `axis`.
```
YarShev, line 449:
```suggestion
            Size of partitions alongside specified `axis`.
```

modin/core/dataframe/pandas/dataframe/dataframe.py
YarShev: I can't think up a case when we hit this branch. Let's leave as is.

modin/core/dataframe/pandas/dataframe/dataframe.py
YarShev: ping

YarShev:
@vnlitvinov, is there a case for now when we construct `PandasDataframe` with empty index or columns?

vnlitvinov:
> @vnlitvinov, is there a case for now when we construct `PandasDataframe` with empty index or columns?

Not yet, though I have another PR in my queue waiting for this one to be merged to improve `df[df.col < threshold]` query performance.

Also I'm guessing that some other queries like `df1.merge(df2)` could get a speedup as a separate PR when parallel shuffle and this one are merged - basically any operation which produces some dataframe which sizes aren't known before the operation is complete should be benefitting from this optimization.

YarShev:
> > @vnlitvinov, is there a case for now when we construct `PandasDataframe` with empty index or columns?
> 
> Not yet, though I have another PR in my queue waiting for this one to be merged to improve `df[df.col < threshold]` query performance.
> 
> Also I'm guessing that some other queries like `df1.merge(df2)` could get a speedup as a separate PR when parallel shuffle and this one are merged - basically any operation which produces some dataframe which sizes aren't known before the operation is complete should be benefitting from this optimization.

Got it, thanks! Let's resolve the rest of the comments and get this PR merged.

modin/core/dataframe/pandas/dataframe/dataframe.py
vnlitvinov: Renamed it to `__make_init_labels_args`... also I don't feel it should be up the constructor, it's a helper function which lives close to the code it's used into.

modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -177,49 +179,52 @@ def __constructor__(self):
...
+        self._columns_cache = ensure_index(columns) if columns is not None else None
+        self._row_lengths_cache = row_lengths
+        self._column_widths_cache = column_widths
+        self._dtypes = dtypes
+
+        self._validate_axes_lengths()
+        self._filter_empties(compute_metadata=False)
+
+    def _validate_axes_lengths(self):
+        """Validate that labels are split correctly if split is known."""
```
YarShev, line 199: The docstring should probably be changed.

YarShev:
There are also some CI jobs failed, please take a look.

modin/core/dataframe/pandas/dataframe/dataframe.py
vnlitvinov: Why is that?..

YarShev (approved):
LGTM!

vnlitvinov:
@prutskov another use case for this "lazy index" thing is it could help with ingest like `read_csv` - we no longer would have to wait for the reads to happen before returning from `pd.read_csv()` as we can postpone the computation of `df.index`.

mvashishtha:
@vnlitvinov I was thinking the exact same thing 😄

YarShev:
Merging the changes as CI failures do not relate to them. See more in #4745.

PR #4739, opened 2022-07-30
STY: rename view/mask to take_2d and take_2d_labels_or_positional

### LINKED ISSUE #4729: STY: rename QueryCompiler.view to something more consistent
QueryCompiler.view seems to behave more like PandasDataframe.mask (which in turn is completely different from pandas.DataFrame.mask).

I propose that both QueryCompiler.view and PandasDataframe.mask (and whatever other relevant methods are out there) be renamed to something like `take_2d` or possibly `take_or_slice_2d`

jbrockmendel:
- [x] Resolves #4729
- [ ] tests added and passing

### DIFF
--- a/modin/core/dataframe/base/dataframe/dataframe.py
+++ b/modin/core/dataframe/base/dataframe/dataframe.py
@@ -34,7 +34,7 @@ class ModinDataframe(ABC):
     """
 
     @abstractmethod
-    def mask(
+    def take_2d_labels_or_positional(
         self,
         row_labels: Optional[List[Hashable]] = None,
         row_positions: Optional[List[int]] = None,
--- a/modin/core/dataframe/pandas/dataframe/dataframe.py
+++ b/modin/core/dataframe/pandas/dataframe/dataframe.py
@@ -608,7 +608,7 @@ class PandasDataframe(ClassLogger):
             )
 
     @lazy_metadata_decorator(apply_axis=None)
-    def mask(
+    def take_2d_labels_or_positional(
         self,
         row_labels: Optional[List[Hashable]] = None,
         row_positions: Optional[List[int]] = None,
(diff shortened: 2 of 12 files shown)


RehanSD:
PR looks good to me, but this does change our algebra (mask is considered an algebra operator). Should we just reframe and say that take is the algebra operator @devin-petersohn? Take makes less sense as an algebra name to me since I think mask has intuition behind it, but take does make sense in this case since Modin frames are immutable, and it parallels pandas.

mvashishtha:
@RehanSD

> since I think mask has intuition behind it

In my view, "mask" does evoke "boolean masking", which is similar to what these functions do, but on the other hand these functions are very different from pandas `mask`, so I prefer the names here.

>  this does change our algebra

I think it's fine for academic terms for operators to be different from the names in implementation :)

RehanSD:
> @RehanSD
> 
> 
> 
> > since I think mask has intuition behind it
> 
> 
> 
> In my view, "mask" does evoke "boolean masking", which is similar to what these functions do, but on the other hand these functions are very different from pandas `mask`, so I prefer the names here.
> 
> 
> 
> >  this does change our algebra
> 
> 
> 
> I think it's fine for academic terms for operators to be different from the names in implementation :) 

Hmm that makes sense to me! I'll go ahead and put an approval on it!

jbrockmendel:
just rebased.  based on today's call i think everyone is on board?

mvashishtha (approved):

mvashishtha:
>  based on today's call i think everyone is on board?

Yep. 

@modin-project/modin-omnisci we need omnisci approval.

Garra1980 (approved):

devin-petersohn (approved):

PR #4767, opened 2022-08-03
FIX-#4099: Use mangled column names but keep the original when building frames from arrow

### LINKED ISSUE #4099: ``dataframe.sum()`` doesn`t work correctly at MODIN_STORAGE_FORMAT=omnisci
When I run:
```
import modin.pandas as pd
from modin.pandas.test.utils import test_data
import modin.config as cfg

data = test_data["float_nan_data"]
modin_df = pd.DataFrame(data)
modin_df.T.sum()
```
I have error:
```
TypeError: expected bytes, int found
```

AndreyPavlenko:
## What do these changes do?

The method Table.rename_columns() accepts a list of strings, but not numbers, thus, the columns must be converted to strings.

- [x] Resolves #4099 
- [x] tests added and passing

### DIFF
--- a/docs/release_notes/release_notes-0.16.0.rst
+++ b/docs/release_notes/release_notes-0.16.0.rst
@@ -33,7 +33,8 @@ Key Features and Updates
   * FIX-#4676: drain sub-virtual-partition call queues (#4695)
   * FIX-#4782: Exclude certain non-parquet files in `read_parquet` (#4783)    
   * FIX-#4808: Set dtypes correctly after column rename (#4809)
-  * FIX-#4811: Apply dataframe -> not_dataframe functions to virtual partitions (#4812)  
+  * FIX-#4811: Apply dataframe -> not_dataframe functions to virtual partitions (#4812)
+  * FIX-#4099: Use mangled column names but keep the original when building frames from arrow
 * Performance enhancements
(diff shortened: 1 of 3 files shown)


anmyachev:
@AndreyPavlenko thanks for contribution! Please also add a test and a release note into `docs\release_notes\release_notes-0.16.0.rst`.

dchigarev (commented):
@AndreyPavlenko thanks for the changes, it should indeed fix the problem. Although, to be sure about it it's better to add a test that verifies it. 

The fix looks general and may even be fixing the more general related issue (#3370), so I would suggest trying to add general testing for integer column names to see if we can close the #3370 with the fix, what do you think?

AndreyPavlenko:
> The fix looks general and may even be fixing the more general related issue (#3370), so I would suggest trying to add general testing for integer column names to see if we can close the #3370 with the fix, what do you think?

Sorry, but looking deeper, now I'm not sure that the fix is quite correct, because it changes the type of the column names.

ienkovich (approved):
LGTM!

YarShev (approved):

AndreyPavlenko:
I think, the issue #3370 is also resolved by this fix. The test test_set_index() covers this case.

YarShev:
@AndreyPavlenko, cool, closing that issue as well.

PR #4779, opened 2022-08-04
PERF: __getitem__

jbrockmendel:
This is in proof-of-concept phase at the moment.

Goals:
1) improve performance of `Series.__getitem__`, `DataFrame.iat`, etc.
2) Make `_get_dict_of_block_index` clearer, use some of the terminology/patterns used in pandas' BlockManager for the analogous lookups

Perf comparisons `iat` vs a hypothetical implementation currently parked at `iat2`, and `Series.__getitem__`
```
import numpy as np
import modin.pandas as pd

arr = np.random.randn(20_000, 50)
df = pd.DataFrame(arr)

x = 19000
y = 40

%timeit df.iat[x, y]
4.04 ms ± 526 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
1.55 ms ± 9.43 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)  # <- iat2
11.6 µs ± 417 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)  # <- pandas

ser = df[10]
_ = ser._query_compiler.to_pandas()  # not sure why this is needed; discussed below

%timeit ser[50]
3.69 ms ± 241 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)  # <- master
1.31 ms ± 23 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)  # <- PR
1.91 µs ± 43.1 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)  # <- pandas
```

Without the `ser._query_compiler.to_pandas()` call, `ser[50]` incorrectly returns `arr[50, 0]` instead of `arr[50, 10]`.  I have  no idea what's going on there.

### DIFF
--- a/modin/core/dataframe/pandas/dataframe/dataframe.py
+++ b/modin/core/dataframe/pandas/dataframe/dataframe.py
@@ -17,6 +17,7 @@ Module contains class PandasDataframe.
 PandasDataframe is a parent abstract class for any dataframe class
 for pandas storage format.
 """
+from pandas.util._decorators import cache_readonly
 from collections import OrderedDict
 import numpy as np
 import pandas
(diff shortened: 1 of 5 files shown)


pyrito:
@jbrockmendel I think the `to_pandas()` call flushes out the queue so any lazy operations will get evaluated.

jbrockmendel:
>  I think the to_pandas() call flushes out the queue so any lazy operations will get evaluated.

is there a maximally-cheap way to do this?

mvashishtha:
> is there a maximally-cheap way to do this?

@jbrockmendel what do you mean by "maximally cheap"? 

We do have #2814 tracking improvements to `to_pandas`.

jbrockmendel:
> what do you mean by "maximally cheap"?

is there a way to flush the queue that is cheaper than to_pandas?

mvashishtha:
> is there a way to flush the queue that is cheaper than to_pandas?

We usually use `BenchmarkMode.put(True)` for this purpose. That should drain call queues and make everything that's usually async finish immediately: 

https://github.com/modin-project/modin/blob/ef654c414364ca319812114324ad9d85e83ee68e/modin/core/dataframe/pandas/partitioning/partition_manager.py#L64-L75

But on the other hand, because we lose the ability to be async, it might not reflect practical performance accurately.

jbrockmendel:
I meant something to do inside Series._getitem to get the accurate results

mvashishtha:
> I meant something to do inside Series._getitem to get the accurate results

If `df` is made of `M by N` partitions, `df[10]` creates `M` new partitions object that point to the same physical data as the first column of `df`'s partitions. These `M` partitions each have something like `lambda x: x.iloc[:, 10]` on their call queue. My guess is that the remote code you've added is skipping the call queue, so you select from column 0 instead of column 10 of the original frame. `to_pandas` will flush the call queue so that `_data` of the partitions of `df[10]` contains only column 10. If you need to get a remote function result from a partition, you have to go through something like `apply`, which adds the function to the existing call queue.

ref. my comment from our earlier discussion https://github.com/modin-project/modin/issues/4729#issuecomment-1199858851

jbrockmendel:
So... what would be an edit to add to this PR in `Series._getitem` to get correct results?

modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -1131,6 +1132,74 @@ def numeric_columns(self, include_bool=True):
...
+        row_blkno = np.digitize(x, self._row_bins)
+        # TODO: searchsorted is marginally faster than digitize
+        col_blkno = np.digitize(y, self._col_bins)
+
+        part = self._partitions[row_blkno, col_blkno]
+
+        row_blkloc = self._get_row_blklocs(x, row_blkno, 0)
+        col_blkloc = self._get_row_blklocs(y, col_blkno, 1)
+
+        return part.getitem_iat(row_blkloc, col_blkloc)
```
mvashishtha, line 1201:
```suggestion
        return part.apply(lambda df: df.iat[row_blkloc, col_blkloc]).get()
```

I think this will work.

jbrockmendel: that does appear to work, but also almost-doubles the runtime of `ser[50]`.  confident its the best we can do?

PR #4792, opened 2022-08-06
PERF-#4713: Stop overriding the ray MacOS object store size limit.

### LINKED ISSUE #4713: PERF: Stop overriding Ray's 2 GiB object store size limit on mac
### System information
- **Computer model**: MacBook Pro (16-inch, 2019)
- **Memory**: 16 GB 2667 MHz DDR4
- **Processor**: 2.3 GHz 8-Core Intel Core i9
- **Modin version** (`modin.__version__`): cc713c5c9055f717ac891cd1fccbd08987e169cd
- **Python version**: 3.10.4

```python
import modin.pandas as pd
import numpy as np
import time

BenchmarkMode.put(True)

df = pd.DataFrame(np.random.randint(0, 100, size=(int(2**20), 2**8)))
start = time.time()
df.apply(lambda col: col + 1)
end = time.time()
print(f'apply took: {end - start}')
```

### Describe the problem
Ray limits the object store size on mac to 2 GiB, but Modin overrides this limit. In the script above, pandas takes 3.6 sec and modin takes 25 sec for the apply. If I instead initialize a default ray cluster with `ray.init()`, I get some spilling to disk, but the apply takes only 5.2 sec in Modin.

Spilling to disk is also bad (e.g. worse than pandas here), but it seems to be much better than using the slow object store.

mvashishtha:
## What do these changes do?

PERF-#4713: Stop overriding the ray MacOS object store size limit.

- [x] Resolves #4713
- [x] tests added and passing


modin/core/execution/ray/common/utils.py
```diff
@@ -186,6 +141,57 @@ def initialize_ray(
...
+        finally:
+            os.close(shm_fd)
+    else:
+        system_memory = virtual_memory
+    object_store_memory = int(0.6 * system_memory // 1e9 * 1e9)
+    # If the memory pool is smaller than 2GB, just use the default in ray.
+    if object_store_memory == 0:
+        object_store_memory = None
+    else:
+        # Versions of ray with MAC_DEGRADED_PERF_MMAP_SIZE_LIMIT don't allow us
```
vnlitvinov, line 180: I wonder if we should put something about different function behaviour on different platforms in the `Notes` section of the docstring... opinions?

vnlitvinov (dismissed):
Overall LGTM, left a few optional questions.

pyrito (dismissed):

modin/core/execution/ray/common/utils.py
```diff
@@ -186,6 +141,57 @@ def initialize_ray(
...
+                    + "increase size of /dev/shm with --shm-size in Docker. Also, you can set "
+                    + "the required memory size for each Ray worker in bytes to MODIN_MEMORY environment variable."
+                )
+        finally:
+            os.close(shm_fd)
+    else:
+        system_memory = virtual_memory
+    object_store_memory = int(0.6 * system_memory // 1e9 * 1e9)
+    # If the memory pool is smaller than 2GB, just use the default in ray.
+    if object_store_memory == 0:
```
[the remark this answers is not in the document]
mvashishtha: The limit is actually `1 / _OBJECT_STORE_TO_SYSTEM_MEMORY_RATIO` GB. I tried to clarify that in the docstring.

modin/core/execution/ray/common/utils.py
mvashishtha: I tried to clarify the behavior particular to linux and macs in the main docstring body.

modin/core/execution/ray/common/utils.py
```diff
@@ -186,6 +143,74 @@ def initialize_ray(
...
+        shm_fd = os.open("/dev/shm", os.O_RDONLY)
+        try:
+            shm_stats = os.fstatvfs(shm_fd)
+            system_memory = shm_stats.f_bsize * shm_stats.f_bavail
+            if system_memory / (virtual_memory / 2) < 0.99:
+                warnings.warn(
+                    f"The size of /dev/shm is too small ({system_memory} bytes). The required size "
+                    + f"at least half of RAM ({virtual_memory // 2} bytes). Please, delete files in /dev/shm or "
+                    + "increase size of /dev/shm with --shm-size in Docker. Also, you can set "
+                    + "the required memory size for each Ray worker in bytes to MODIN_MEMORY environment variable."
```
jeffreykennethli, line 187: nit: "Also, you can override the memory size for each Ray worker (in bytes) to the MODIN_MEMORY environment variable."

modin/core/execution/ray/common/utils.py
```diff
@@ -186,6 +143,74 @@ def initialize_ray(
...
+                )
+        finally:
+            os.close(shm_fd)
+    else:
+        system_memory = virtual_memory
+    object_store_memory = int(
+        _OBJECT_STORE_TO_SYSTEM_MEMORY_RATIO * system_memory // 1e9 * 1e9
```
jeffreykennethli, line 194: is this equivalent to `int(_OBJECT_STORE_TO_SYSTEM_MEMORY_RATIO * system_memory)`?

modin/core/execution/ray/common/utils.py
```diff
@@ -186,6 +143,74 @@ def initialize_ray(
...
+    )
+    if object_store_memory == 0:
+        object_store_memory = None
```
jeffreykennethli, line 197:
```suggestion
        return None
```

modin/core/execution/ray/common/utils.py
mvashishtha: No, the `// 1e9` rounds down to the nearest gigabyte. I tried to change the calculation to reflect that, and also to clarify the calculation in the docstring.

Note that we need the `int` because floor-dividing a float by an int gives a float.

jeffreykennethli: sg. ty for the clarification!

jeffreykennethli (approved):

modin/core/execution/ray/common/utils.py
```diff
@@ -186,6 +143,77 @@ def initialize_ray(
...
+        system_memory is psutil.virtual_memory().total
+      desired_memory = ((_OBJECT_STORE_TO_SYSTEM_MEMORY_RATIO * system_memory)
+                        rounded down to the nearest GB)
+      if desired_memory == 0:
+        return None
+      else if (on Mac and desired_memory > the Ray mac object store size
+               limit L):
+        return L
+      else:
+        return desired_memory
```
vnlitvinov, line 165: Isn't it basically copying the function code?

I was thinking more about stuff like
```
If set by ``Memory`` config variable, return that.
On Linux, take system memory from /dev/shm, on others assume total virtual memory.
On Mac, never return more than Ray-specified upper limit.
```
Other details could be seen from the function body itself.

mvashishtha: This is close to the function code, but the behavior of this function for callers (not just the internals) is quite complex, so I was trying to document the behavior. I thought the description I had here was much easier to read than the code, too.

But I'm okay with what you're suggesting. Done.

mvashishtha: I was thinking of rules like "A docstring should give enough information to write a call to the function without reading the function’s code." from the Google Python style guide: https://google.github.io/styleguide/pyguide.html#383-functions-and-methods

But I agree that that level of documentation does look awkward here.

pyrito (dismissed):

modin/core/execution/ray/common/utils.py
```diff
@@ -34,6 +35,8 @@
 )
 from modin.error_message import ErrorMessage
 
+_OBJECT_STORE_TO_SYSTEM_MEMORY_RATIO = 0.6
```
YarShev, line 38: Why is this ratio equal to 0.6?

modin/core/execution/ray/common/utils.py
```diff
@@ -98,53 +101,7 @@ def initialize_ray(
...
-                    + f"{mac_size_limit / 2 ** 30:.4} GiB. Ray by default does "
-                    + "not allow setting an object store size greater than "
-                    + "that. Modin is overriding that default limit because "
-                    + "it would rather have a larger, slower object store "
-                    + "than spill to disk more often. To override Modin's "
-                    + "behavior, you can initialize Ray yourself."
-                )
-                os.environ["RAY_ENABLE_MAC_LARGE_OBJECT_STORE"] = "1"
-
+            object_store_memory = _object_store_memory()
```
YarShev, line 104:
```suggestion
            object_store_memory = _get_object_store_memory()
```

modin/core/execution/ray/common/utils.py
```diff
@@ -186,6 +143,65 @@ def initialize_ray(
         NPartitions._put(num_cpus)
 
 
+def _object_store_memory() -> Optional[int]:
```
YarShev, line 146:
```suggestion
def _get_object_store_memory() -> Optional[int]:
```

YarShev (commented):
Left a couple of comments. Otherwise, LGTM.

modin/core/execution/ray/common/utils.py
mvashishtha: It's an arbitrary number that has been around for a long time. It goes back to 3542226076bd03473b524889959bdf3b52531ee8 from 2019/01/11.

YarShev: I was just wondering why we specified it equal to this number, i.e. is there any heuristic. If someone is aware of that, it would be good to know.

YarShev (approved):

PR #4807, opened 2022-08-10
FEAT-#4733: Support fastparquet as engine for `read_parquet`

### LINKED ISSUE #4733: Support fastparquet engine for `read_parquet`
Currently our implementation of `read_parquet` only supports PyArrow. Given that we have some more user demand, we should also support fastparquet as a library to read parquet files.

pyrito:
## What do these changes do?

This PR adds `fastparquet` as a supported engine for `read_parquet`. I took the liberty to abstract some details of the underlying engine in a way I thought best fit. Please drop suggestions for code cleanup. 

- [x] Resolves #4733 
- [x] tests added and passing

### DIFF
--- a/modin/core/io/column_stores/parquet_dispatcher.py
+++ b/modin/core/io/column_stores/parquet_dispatcher.py
@@ -30,10 +30,9 @@ from modin.core.io.column_stores.column_store_dispatcher import ColumnStoreDispa
 from modin.utils import import_optional_dependency, _inherit_docstrings
 
 
-class ColumnStoreDataset:  # noqa : PR01
+class ColumnStoreDataset:
     """
-    Base class that encapsulates Parquet engine-specific details for
-    a given Parquet dataset.
+    Base class that encapsulates Parquet engine-specific details.
 
     This class exposes a set of functions that are commonly used in the
     `read_parquet` implementation.
@@ -60,7 +59,7 @@ class ColumnStoreDataset:  # noqa : PR01
         List that contains the full paths of the parquet files in the dataset.
     """
 
-    def __init__(self, path, storage_options):
+    def __init__(self, path, storage_options):  # noqa : PR01
         self.path = path
         self.storage_options = storage_options
         self._fs_path = None
--- a/modin/core/io/column_stores/parquet_dispatcher.py
+++ b/modin/core/io/column_stores/parquet_dispatcher.py
@@ -563,7 +563,7 @@ class ParquetDispatcher(ColumnStoreDispatcher):
         path : str, path object or file-like object
             The filepath of the parquet file in local filesystem or hdfs.
         engine : str
-            Parquet library to use (only 'PyArrow' is supported for now).
+            Parquet library to use (either pyarrow or fastparquet).
         columns : list
             If not None, only these columns will be read from the file.
         **kwargs : dict
--- a/modin/core/io/column_stores/parquet_dispatcher.py
+++ b/modin/core/io/column_stores/parquet_dispatcher.py
@@ -22,7 +22,6 @@ from fsspec.core import split_protocol, url_to_fs
 from fsspec.registry import get_filesystem_class
 from fsspec.spec import AbstractBufferedFile
 import numpy as np
-from pandas.io.common import is_fsspec_url
 from packaging import version
 
 from modin.core.storage_formats.pandas.utils import compute_chunksize
@@ -33,13 +32,43 @@ from modin.core.io.column_stores.column_store_dispatcher import ColumnStoreDispa
 from modin.utils import import_optional_dependency, _inherit_docstrings
 
 
-class Dataset(ABC):
+class Dataset(ABC):  # noqa : PR01
     """
     Abstract class that encapsulates Parquet engine-specific details.
 
     This class exposes a set of functions that are commonly used in the
     `read_parquet` implementation.
+
+    Attributes
+    ----------
+    path : str, path object or file-like object
+        The filepath of the parquet file in local filesystem or hdfs.
+    storage_options : dict
+        Parameters for specific storage engine.
+    _fs_path : str, path object or file-like object
+        The filepath or handle of the parquet dataset specific to the
+        filesystem implementation. E.g. for `s3://test/example`, _fs
+        would be set to S3FileSystem and _fs_path would be `test/example`.
+    _fs : Filesystem
+        Filesystem object specific to the given parquet file/dataset.
+    dataset : ParquetDataset or ParquetFile
+        Underlying dataset implementation for PyArrow and fastparquet
+        respectively.
+    _row_groups_per_file : list
+        List that contains the number of row groups for each file in the
+        given parquet dataset.
+    _files : list
+        List that contains the full paths of the parquet files in the dataset.
     """
+
+    def __init__(self, path, storage_options):
+        self.path = path
+        self.storage_options = storage_options
+        self._fs_path = None
+        self._fs = None
+        self.dataset = self._init_dataset()
+        self._row_groups_per_file = None
+        self._files = None
 
     @abstractproperty
     def pandas_metadata(self):
@@ -61,6 +90,42 @@ class Dataset(ABC):
         """Return string representing what engine is being used."""
         pass
 
+    @property
+    def fs(self):
+        """
+        Return the filesystem object associated with the dataset path.
+
+        Returns
+        -------
+        filesystem
+            Filesystem object.
+        """
+        if self._fs is None:
+            if isinstance(self.path, AbstractBufferedFile):
+                self._fs = self.path.fs
+            else:
+                self._fs = get_filesystem_class(split_protocol(self.path)[0])(
+                    **self.storage_options
+                )
+        return self._fs
+
+    @property
+    def fs_path(self):
+        """
+        Return the filesystem-specific path or file handle.
+
+        Returns
+        -------
+        fs_path : str, path object or file-like object
+            String path specific to filesystem or a file handle.
+        """
+        if self._fs_path is None:
+            if isinstance(self.path, AbstractBufferedFile):
+                self._fs_path = self.path
+            else:
+                _, self._fs_path = url_to_fs(self.path, **self.storage_options)
+        return self._fs_path
+
     @abstractmethod
     def row_groups_per_file(self):
         """Return a list with the number of row groups per file."""
@@ -78,26 +143,6 @@ class Dataset(ABC):
         """
         pass
 
-    def _get_fs_and_fs_path(self):
-        """
-        Retrieve filesystem interface and filesystem-specific path.
-
-        For example, url_to_fs("s3://example/thing/this") would return
-        an S3FileSystem object and a fs_path of 'example/thing/this'.
-
-        Returns
-        -------
-        filesystem : Any
-            Protocol implementation of registry.
-        fs_path : list
-            Filesystem's specific path.
-        """
-        return (
-            url_to_fs(self.path, **self.storage_options)
-            if is_fsspec_url(self.path)
-            else (None, self.path)
-        )
-
     def _get_files(self, files):
         """
         Retrieve list of formatted file names in dataset path.
@@ -109,10 +154,8 @@ class Dataset(ABC):
 
         Returns
         -------
-        filesystem: Any
-            Protocol implementation of registry.
-        files: list
-            List of files from path.
+        fs_files : list
+            List of files from path with fs-protocol prepended.
         """
         # Older versions of fsspec doesn't support unstrip_protocol(). It
         # was only added relatively recently:
@@ -125,32 +168,24 @@ class Dataset(ABC):
             return f"{protos[0]}://{path}"
 
         if isinstance(self.path, AbstractBufferedFile):
-            return self.path.fs, [self.path]
-        fs = get_filesystem_class(split_protocol(self.path)[0])(**self.storage_options)
+            return [self.path]
         # version.parse() is expensive, so we can split this into two separate loops
         if version.parse(fsspec.__version__) < version.parse("2022.5.0"):
-            files = [_unstrip_protocol(fs.protocol, fpath) for fpath in files]
+            fs_files = [_unstrip_protocol(self.fs.protocol, fpath) for fpath in files]
         else:
-            files = [fs.unstrip_protocol(fpath) for fpath in files]
+            fs_files = [self.fs.unstrip_protocol(fpath) for fpath in files]
 
-        return fs, files
+        return fs_files
 
 
 @_inherit_docstrings(Dataset)
 class PyArrowDataset(Dataset):
-    def __init__(self, path, storage_options):
-        self.path = path
-        self.storage_options = storage_options
-        self.dataset = self._init_dataset()
-        self._row_groups_per_file = None
-        self.filesystem, self._files = self._get_files(self.dataset.files)
-
     def _init_dataset(self):  # noqa: GL08
         from pyarrow.parquet import ParquetDataset
 
-        fs, fs_path = self._get_fs_and_fs_path()
-        dataset = ParquetDataset(fs_path, filesystem=fs, use_legacy_dataset=False)
-        return dataset
+        return ParquetDataset(
+            self.fs_path, filesystem=self.fs, use_legacy_dataset=False
+        )
 
     @property
     def pandas_metadata(self):
@@ -172,7 +207,7 @@ class PyArrowDataset(Dataset):
             # Count up the total number of row groups across all files and
             # keep track of row groups per file to use later.
             for file in self.files:
-                with self.filesystem.open(file) as f:
+                with self.fs.open(file) as f:
                     row_groups = ParquetFile(f).num_row_groups
                     row_groups_per_file.append(row_groups)
             self._row_groups_per_file = row_groups_per_file
@@ -190,26 +225,15 @@ class PyArrowDataset(Dataset):
     ):
         from pyarrow.parquet import read_table
 
-        return read_table(
-            self.path, columns=columns, filesystem=self.filesystem
-        ).to_pandas()
+        return read_table(self.path, columns=columns, filesystem=self.fs).to_pandas()
 
 
 @_inherit_docstrings(Dataset)
 class FastParquetDataset(Dataset):
-    def __init__(self, path, storage_options):
-        self.path = path
-        self.storage_options = storage_options
-        self.dataset = self._init_dataset()
-        self._row_groups_per_file = None
-        self.filesystem, self._files = self._get_files(self._get_fastparquet_files())
-
     def _init_dataset(self):  # noqa: GL08
         from fastparquet import ParquetFile
 
-        fs, fs_path = self._get_fs_and_fs_path()
-        dataset = ParquetFile(fs_path, fs=fs)
-        return dataset
+        return ParquetFile(self.fs_path, fs=self.fs)
 
     @property
     def pandas_metadata(self):
@@ -233,7 +257,7 @@ class FastParquetDataset(Dataset):
             # Count up the total number of row groups across all files and
             # keep track of row groups per file to use later.
             for file in self.files:
-                with self.filesystem.open(file) as f:
+                with self.fs.open(file) as f:
                     row_groups = ParquetFile(f).info["row_groups"]
                     row_groups_per_file.append(row_groups)
             self._row_groups_per_file = row_groups_per_file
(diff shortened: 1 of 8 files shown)


vnlitvinov (changes requested):
@pyrito I haven't yet reviewed the changes, but, according to CI failures, you have to update the dependencies:
```
ImportError: Missing optional dependency 'fastparquet'. fastparquet is required for parquet support. Use pip or conda to install fastparquet.
```

modin/core/storage_formats/pandas/parsers.py
```diff
@@ -671,11 +699,9 @@ class PandasParquetParser(PandasParser):
 """,
     )
     def parse(files_for_parser, **kwargs):
-        from pyarrow.parquet import ParquetFile
-
         columns = kwargs.get("columns", None)
         storage_options = kwargs.pop("storage_options", {}) or {}
-
+        engine = kwargs.pop("engine")
```
jeffreykennethli, line 704: if "engine" is popped here, it won't be passed into [this following line](https://github.com/modin-project/modin/pull/4807/files#diff-2ad1882ffe66014d5097154a11168dde541dbf7d8fed468663b8098591912bf6L682)  

`pandas.read_parquet(files_for_parser, **kwargs)` for a "single_worker_read".

Why do we not pass `engine` here, if `pandas.read_parquet` supports that kwarg?

modin/core/storage_formats/pandas/parsers.py
```diff
@@ -663,6 +663,34 @@ class ParquetFileToRead(NamedTuple):
...
+                )
+                .to_pandas()
+            )
+        elif engine == "fastparquet":
+            from fastparquet import ParquetFile
+
+            return ParquetFile(f)[row_group_start:row_group_end].to_pandas(
+                columns=columns
+            )
+
```
jeffreykennethli, line 693: should we defensively raise an exception for the final `else` case for the engine type?

modin/core/storage_formats/pandas/parsers.py
pyrito: You are right. That's a good catch, I should just pass it along.

modin/core/storage_formats/pandas/parsers.py
pyrito: Probably a good call to do it here as well, but I did try to catch for it upstream off the bat.

modin/core/io/column_stores/parquet_dispatcher.py
```diff
@@ -28,23 +29,79 @@
...
+        """Return string representing what engine is being used."""
+        pass
+
+    @abstractmethod
+    def row_groups_per_file(self):
+        """Return a list with the number of row groups per file."""
+        pass
+
+    @abstractmethod
+    def read_table_as_pandas(self, columns):
```
mvashishtha, line 69: Could we call this `to_pandas_dataframe`? `table` was confusing to me.

modin/core/io/column_stores/parquet_dispatcher.py
```diff
@@ -28,23 +29,79 @@
...
         """
-        Retrieve filesystem interface and list of files from path.
+        Return a pandas dataframe with the given columns.
```
mvashishtha, line 71:
```suggestion
        Read the given columns as a pandas dataframe.
```

Maybe?

modin/core/io/column_stores/parquet_dispatcher.py
```diff
@@ -28,23 +29,79 @@
...
-        storage_options : dict
-            Parameters for specific storage engine.
+        columns : list
+            List of columns that should be read from file.
+        """
+        pass
+
+    def _get_fs_and_fs_path(self):
+        """
+        Retrieve filesystem interface and filesystem-specific path.
```
mvashishtha, line 82: "filesystem-specific path." is unclear to me. Maybe give an example?

modin/core/io/column_stores/parquet_dispatcher.py
```diff
@@ -63,73 +120,211 @@ def _unstrip_protocol(protocol, path):
...
+    def _init_dataset(self):  # noqa: GL08
+        from fastparquet import ParquetFile
+
+        fs, fs_path = self._get_fs_and_fs_path()
+        dataset = ParquetFile(fs_path, fs=fs)
+        return dataset
+
+    @property
+    def pandas_metadata(self):
+        import json
```
mvashishtha, line 212: Isn't this part of the python standard library? In that case we can move it to the top of the file.

modin/core/storage_formats/pandas/parsers.py
```diff
@@ -663,43 +663,66 @@ class ParquetFileToRead(NamedTuple):
 
 @doc(_doc_pandas_parser_class, data_type="PARQUET data")
 class PandasParquetParser(PandasParser):
+    @staticmethod
+    def read_row_group_chunk(
```
mvashishtha, line 667: Should this be made a private method?

mvashishtha (commented):
Looks very good overall. I have some minor comments.

modin/core/io/column_stores/parquet_dispatcher.py
pyrito: Yeah that's probably a better name.

modin/core/storage_formats/pandas/parsers.py
pyrito: Yeah it's an implementation detail so it should be.

modin/core/io/column_stores/parquet_dispatcher.py
```diff
@@ -63,73 +120,211 @@ def _unstrip_protocol(protocol, path):
...
-        Returns
-        -------
-        dataset : ParquetDataset
-            Dataset object of given Parquet file-path.
-        """
+    def _init_dataset(self):  # noqa: GL08
         from pyarrow.parquet import ParquetDataset
 
-        fs, fs_path = cls._get_fs_and_fs_path(path, storage_options)
+        fs, fs_path = self._get_fs_and_fs_path()
```
pyrito, line 147: @mvashishtha there are a few times where I repeat getting the file system from the path and I'm trying to reduce the amount of code duplication (see `_get_files`). I was a little lazy and kind of hacked a solution initially but I want to encapsulate these properly. I was wondering if you had any ideas?

modin/core/io/column_stores/parquet_dispatcher.py
```diff
@@ -63,73 +120,211 @@ def _unstrip_protocol(protocol, path):
...
+        # fastparquet doesn't have a nice method like PyArrow, so we
+        # have to copy some of their logic here while we work on getting
+        # an easier method to get a list of valid files.
+        # See: https://github.com/dask/fastparquet/issues/795
+        if isinstance(self.path, AbstractBufferedFile):
+            fs = self.path.fs
+        else:
+            fs = get_filesystem_class(split_protocol(self.path)[0])(
+                **self.storage_options
+            )
```
pyrito, line 257: @mvashishtha ditto comment from before

modin/core/io/column_stores/parquet_dispatcher.py
mvashishtha: I see one `get_filesystem_class` call in each parquet class, and each class calls the same `_get_fs_and_fs_path`, which calls `url_to_fs`. I think this level of repetition is ok. Alternatively we could have each class internally call `self.fs` and `self.fs_path`, both of which compute the filesystem and `fs_path` and cache them.

modin/core/io/column_stores/parquet_dispatcher.py
mvashishtha: I see. Let's discuss in the other thread.

mvashishtha (approved):
LGTM!

modin/core/io/column_stores/parquet_dispatcher.py
pyrito: Oh I see, that could work,  so make both `fs` and `fs_path` as properties and cache them after the first call?

modin/core/io/column_stores/parquet_dispatcher.py
```diff
@@ -428,18 +610,19 @@ def _read(cls, path, engine, columns, **kwargs):
...
         index_columns = (
-            dataset.schema.pandas_metadata.get("index_columns", [])
-            if dataset.schema.pandas_metadata
+            dataset.pandas_metadata.get("index_columns", [])
+            if dataset.pandas_metadata
             else []
         )
         # If we have columns as None, then we default to reading in all the columns
-        column_names = dataset.schema.names if not columns else columns
+        column_names = dataset.columns if not columns else columns
```
vnlitvinov, line 621:
```suggestion
        column_names = columns if columns else dataset.columns
```
I personally feel positive conditions read better

modin/core/io/column_stores/parquet_dispatcher.py
mvashishtha: Yeah, whenever someone calls `self.fs` or `self.fs_path`, compute both and cache them. But I think this is a minor improvement and the code as is is fine with me.

modin/core/storage_formats/pandas/parsers.py
```diff
@@ -663,43 +663,66 @@ class ParquetFileToRead(NamedTuple):
...
+
+            return (
+                ParquetFile(f)
+                .read_row_groups(
+                    list(
+                        range(
+                            row_group_start,
+                            row_group_end,
+                        )
+                    ),
```
vnlitvinov, line 681:
```suggestion
                        range(
                            row_group_start,
                            row_group_end,
                        )
```
can we omit `list()` call here? I think I asked this once already, but let's repeat the question

modin/core/io/column_stores/parquet_dispatcher.py
pyrito: Oh I actually really like that. That's clever, don't know why I didn't think of that 😃 . I think it's worth doing either way since I expect to get other comments for other style fixes.

modin/core/storage_formats/pandas/parsers.py
```diff
@@ -663,43 +663,66 @@ class ParquetFileToRead(NamedTuple):
...
+                    columns=columns,
+                    use_pandas_metadata=True,
+                )
+                .to_pandas()
+            )
+        elif engine == "fastparquet":
+            from fastparquet import ParquetFile
+
+            return ParquetFile(f)[row_group_start:row_group_end].to_pandas(
```
vnlitvinov, line 690: won't this read the entire dataset? or is `ParquetFile` lazy?

modin/core/storage_formats/pandas/parsers.py
```diff
@@ -663,43 +663,66 @@ class ParquetFileToRead(NamedTuple):
...
+                columns=columns
+            )
+        else:
+            # We shouldn't ever come to this case, so something went wrong
+            raise ValueError("engine must be one of 'pyarrow', 'fastparquet'")
```
vnlitvinov, line 695:
```suggestion

        # We shouldn't ever come to this case, so something went wrong
        raise ValueError(f"engine must be one of 'pyarrow', 'fastparquet', got: {engine}")
```

naren-ponder (approved):
LGTM

modin/core/io/column_stores/parquet_dispatcher.py
```diff
@@ -63,73 +124,209 @@ def _unstrip_protocol(protocol, path):
                     return path
             return f"{protos[0]}://{path}"
 
-        if isinstance(path, AbstractBufferedFile):
-            return path.fs, [path]
-        fs = get_filesystem_class(split_protocol(path)[0])(**storage_options)
-        dataset = cls._get_dataset(path, storage_options)
+        if isinstance(self.path, AbstractBufferedFile):
```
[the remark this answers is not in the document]
pyrito: It is set in the constructor.

modin/core/io/column_stores/parquet_dispatcher.py
```diff
@@ -28,23 +30,82 @@
 
 
 from modin.core.io.column_stores.column_store_dispatcher import ColumnStoreDispatcher
-from modin.utils import import_optional_dependency
+from modin.utils import import_optional_dependency, _inherit_docstrings
 
 
-class ParquetDispatcher(ColumnStoreDispatcher):
-    """Class handles utils for reading `.parquet` files."""
+class Dataset(ABC):
```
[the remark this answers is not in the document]
pyrito: `ParquetDataset` conflicts with pyarrow's version which is why I just went with Dataset. ColumnStoreDataset could work.

modin/core/io/column_stores/parquet_dispatcher.py
```diff
@@ -63,73 +124,209 @@ def _unstrip_protocol(protocol, path):
...
+
+
+@_inherit_docstrings(Dataset)
+class FastParquetDataset(Dataset):
+    def __init__(self, path, storage_options):
+        self.path = path
+        self.storage_options = storage_options
+        self.dataset = self._init_dataset()
+        self._row_groups_per_file = None
+        self.filesystem, self._files = self._get_files(self._get_fastparquet_files())
```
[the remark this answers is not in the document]
pyrito: This is a good idea.

modin/core/io/column_stores/parquet_dispatcher.py
```diff
@@ -63,73 +124,209 @@ def _unstrip_protocol(protocol, path):
...
+                + "support.\n"
+                + "Trying to import the above resulted in these errors:"
+                + f"{error_msgs}"
+            )
+        elif engine == "pyarrow":
+            return PyArrowDataset(path, storage_options)
+        elif engine == "fastparquet":
+            return FastParquetDataset(path, storage_options)
+        else:
+            raise ValueError("engine must be one of 'pyarrow', 'fastparquet'")
```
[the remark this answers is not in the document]
pyrito: I wanted to keep the error in concordance with pandas.

modin/core/storage_formats/pandas/parsers.py
pyrito: Yeah we can. Looks like it works without `list()`.

modin/core/storage_formats/pandas/parsers.py
pyrito: It won't read the entire dataset. I looked through the code base to confirm this as well. `ParquetFile` will initialize some metadata and then the slice for the row groups just sets some stuff internally and will actually be read in during the `to_pandas` call.

modin/core/io/column_stores/parquet_dispatcher.py
```diff
@@ -63,73 +124,209 @@ def _unstrip_protocol(protocol, path):
...
+        return "fastparquet"
+
+    def row_groups_per_file(self):
+        from fastparquet import ParquetFile
+
+        if self._row_groups_per_file is None:
+            row_groups_per_file = []
+            # Count up the total number of row groups across all files and
+            # keep track of row groups per file to use later.
+            for file in self.files:
```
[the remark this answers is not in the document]
pyrito: Should be fine now because I'm caching `self._files`

modin/core/io/column_stores/parquet_dispatcher.py
```diff
@@ -63,73 +124,209 @@ def _unstrip_protocol(protocol, path):
...
+
+    @property
+    def columns(self):
+        return self.dataset.schema.names
+
+    @property
+    def engine(self):
+        return "pyarrow"
+
+    def row_groups_per_file(self):
```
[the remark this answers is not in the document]
pyrito: Per our discussion on slack, seems like our doc checker is breaking right now because of @cache_readonly properties. We should address that issue in another PR. For now, I've added a TODO so that we can come back to this later.

mvashishtha (commented):
I have two minor comments.

modin/core/io/column_stores/parquet_dispatcher.py
```diff
@@ -51,48 +82,69 @@ def columns(self):
...
-        return (
-            url_to_fs(self.path, **self.storage_options)
-            if is_fsspec_url(self.path)
-            else (None, self.path)
-        )
+        if self._fs_path is None:
+            if isinstance(self.path, AbstractBufferedFile):
+                self._fs_path = self.path
+            else:
+                _, self._fs_path = url_to_fs(self.path, **self.storage_options)
```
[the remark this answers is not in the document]
pyrito: We would be setting `self._fs` twice then, right?

modin/core/io/column_stores/parquet_dispatcher.py
```diff
@@ -51,48 +82,69 @@ def columns(self):
...
+        -------
+        filesystem
+            Filesystem object.
         """
-        pass
+        if self._fs is None:
+            if isinstance(self.path, AbstractBufferedFile):
+                self._fs = self.path.fs
+            else:
+                self._fs = get_filesystem_class(split_protocol(self.path)[0])(
```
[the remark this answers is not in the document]
pyrito: I guess after looking at the implementation of `url_to_fs`, your suggestion would work as well, and is probably cleaner.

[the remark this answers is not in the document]
mvashishtha: I edited my suggestion! What i meant to say was that both methods should compute both `_fs` and `_fs_path`.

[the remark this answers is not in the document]
pyrito: I guess we could do that, but then I feel like we could try to consolidate this all into one method or something?

modin/core/io/column_stores/parquet_dispatcher.py
```diff
@@ -33,14 +32,45 @@
 from modin.utils import import_optional_dependency, _inherit_docstrings
 
 
-class Dataset(ABC):
+class Dataset(ABC):  # noqa : PR01
```
vnlitvinov, line 35:
```suggestion
class ColumnStoreDataset(ABC):  # noqa : PR01
```

vnlitvinov (approved):
LGTM with optional suggestion!

modin/core/io/column_stores/parquet_dispatcher.py
```diff
@@ -13,45 +13,152 @@
...
+        Underlying dataset implementation for PyArrow and fastparquet
+        respectively.
+    _row_groups_per_file : list
+        List that contains the number of row groups for each file in the
+        given parquet dataset.
+    _files : list
+        List that contains the full paths of the parquet files in the dataset.
+    """
+
+    def __init__(self, path, storage_options):
```
YarShev, line 65: An abstract class shouldn't be instantiated. We do use such a logic nowhere. I suggest using an ordinary python class instead of abc and raise NotImplemented error for the methods of this class that has to be overridden.

YarShev (changes requested):

modin/core/io/column_stores/parquet_dispatcher.py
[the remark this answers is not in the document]
mvashishtha: What were you thinking of consolidating? I think what I've suggested seems good. Within the class, `self._fs` means to get the filesystem, computing it with `url_to_fs` if needed. Likewise for `self._fs_path`. It just happens to be that it's sometimes convenient to compute both in a single function call.

modin/core/io/column_stores/parquet_dispatcher.py
pyrito: Yes you're right. Initially it made sense to make use of the abstract class but it's at a point where it doesn't make sense to do that anymore.

mvashishtha (dismissed):

mvashishtha (approved):

YarShev (commented):
Left one comment, other than that LGTM.

modin/core/io/column_stores/parquet_dispatcher.py
```diff
@@ -15,43 +15,144 @@
...
+    def columns(self):
+        """Return the list of columns in the dataset."""
+        raise NotImplementedError
+
+    @property
+    def engine(self):
+        """Return string representing what engine is being used."""
+        raise NotImplementedError
+
+    # TODO: make this cache_readonly after docstring inheritance is fixed.
```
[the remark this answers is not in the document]
pyrito: @YarShev I added this TODO on behalf of @vnlitvinov . When I tried to use `@cache_readonly`, the doc checker broke for some reason (see issue: https://github.com/modin-project/modin/issues/4821). The properties I've added the TODO for are the only ones that require some amount of computation to initially set the property value. This is not the case for the other properties.

YarShev (approved):
@pyrito, thank you for the changes.

PR #4820, opened 2022-08-15
FIX-#4818, PERF-#4825: Fix where by using the new n-ary operator

### LINKED ISSUE #4818: `where` function of DataFrame and Series works unexpectedly
### System information
- **OS Platform and Distribution**: ubuntu 20.04
- **Modin version**: 0.15.2
- **Ray version**: 1.12.1
- **Python version**: 3.8.11

**Modin**
```
>>> import modin.pandas as pd
>>> df1 = pd.DataFrame({'a': [pd.NA, 2, 3, 4]})
>>> df2 = pd.DataFrame({'a': [1, 2, 3, 4]})
>>> df1.where(df1.notna(), df2)
      a
0  <NA>
1     2
2     3
3     4

```

**Pandas**
```
>>> import pandas as pd
>>> df1 = pd.DataFrame({'a': [pd.NA, 2, 3, 4]})
>>> df2 = pd.DataFrame({'a': [1, 2, 3, 4]})
>>> df1.where(df1.notna(), df2)
      a
0     1
1     2
2     3
3     4
```

### Describe the problem
Modin `where` function works unexpectedly when the alternative value is DataFrame type. The same happens with Series

mvashishtha:
## What do these changes do?

- [x] Resolves #4818, #4825
- [x] tests added and passing

**Performance Results**

For three dataframes with same row and column labels:

<details>

<summary>Benchmark Script (ipython)</summary>

```python
import modin.pandas as pd
import numpy as np

BenchmarkMode.put(True)
df = pd.DataFrame(np.random.randint(0,100,(2**20,2**8)))
cond = df % 2 == 0
other = pd.DataFrame(np.random.randint(0,100,(2**20,2**8)))

%time result = df.where(cond, other)
```
</details>

Wall times (master at f7fd5595429e4edbd147af5c514312339f421813): 15.7 s, 15 s
new implementation times (this branch at c05ef5e56cf5547115aa08870cf89d4a8a47c6f2): 1.59 sec, 1.58 sec
Wall times (pandas 1.4.3): 6.43 s, 6.57 s

For three dataframes each with different row and column -labels:

<details>

<summary>Benchmark Script (ipython)</summary>

```python
import modin.pandas as pd
import numpy as np
import time
import ray

df = pd.DataFrame(np.random.randint(0,100,(2**20,2**8)))
cond = pd.DataFrame(np.random.choice(a=[False, True], size=(2**20, 2**8)), index=df.index[::-1], columns=df.columns[::-1])
other = pd.DataFrame(np.random.randint(0,100,(2**20,2**8)), index=df.index[-1:].append(df.index[:-1]), columns=df.columns[-1:].append(df.columns[:-1]))

# Benchmark mode artificially slows modin down because we have to block on
# each reindex. Instead, manually wait for the where to finish.
def get_timed_where_result():
  start = time.time()
  result = df.where(cond, other)
  all_futures = [x._data for x in result._query_compiler._modin_frame._partitions.flatten()]
  ray.wait(all_futures, num_returns=len(all_futures))
  print(f'where took {time.time() - start}')
  return result

result = get_timed_where_result()
```
</details>

master at f7fd5595429e4edbd147af5c514312339f421813 gives the wrong result.
times (this branch at c05ef5e56cf5547115aa08870cf89d4a8a47c6f2): 6.43 sec, 6.37 sec
times (pandas 1.4.3): 7.16 sec, 7.11 sec

**System information (Amazon EC2)**:

- AMI name: ubuntu/images/hvm-ssd/ubuntu-focal-20.04-amd64-server-20211129
- instance type: t2.2xlarge
- Ubuntu 20.04.3 LTS 
- 8 physical CPUs
- Memory: 32 GiB
- Python: 3.9.7


vnlitvinov (dismissed):
This changes the algebra rather a lot.

That said, I like the changes! Hopefully CI would like them, too.

modin/core/storage_formats/pandas/query_compiler.py
```diff
@@ -410,23 +410,13 @@ def where(self, cond, other, **kwargs):
             cond, type(self)
         ), "Must have the same QueryCompiler subclass to perform this operation"
         if isinstance(other, type(self)):
-            # Note: Currently we are doing this with two maps across the entire
-            # data. This can be done with a single map, but it will take a
-            # modification in the `BlockPartition` class.
-            # If this were in one pass it would be ~2x faster.
-            # TODO (devin-petersohn) rewrite this to take one pass.
```
pyrito, line 417: It looks like we are no longer having to call binary_op twice. Are we seeing any performance improvements?

mvashishtha: I think so. I put some benchmark numbers in the PR description.

pyrito: @mvashishtha I don't see the numbers?

pyrito:
@mvashishtha that's a pretty large speedup! How does pandas do?

pyrito (approved):

mvashishtha:
> that's a pretty large speedup! How does pandas do?

@pyrito I added wall times for pandas. It looks like on this particular benchmark Modin is now about 3x the performance of pandas.

vnlitvinov (dismissed):
LGTM!

I wonder if we should name the PR `PERF-` or `FEAT-`, rather than `FIX-`, though.

mvashishtha:
> I wonder if we should name the PR PERF- or FEAT-, rather than FIX-, though.

@vnlitvinov to be more precise, I created #4825 and referenced both the original issue and that one in the new commit description.

modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -2525,32 +2525,36 @@ def binary_op(self, op, right_frame, join_type="outer"):
...
-            row_lengths,
-            right_frame._column_widths,
-        )
+        new_right_frames = [
+            self.__constructor__(
+                right_parts,
+                joined_index,
+                right_frames[0].columns,
+                row_lengths,
+                right_frames[0]._column_widths,
```
vnlitvinov, line 2540: I've missed that earlier... there should be some kind of an assert that all right frames are equally column-ed (both in terms of which columns they have and how they are partitioned), otherwise this piece would break.

Or, probably better, change the loop to something like
```python
new_right_frames = [
    self.__constructor__(
        right_parts,
        joined_index,
        right_frame.columns,
        row_lengths,
        right_frame._column_widths,
    )
    for right_parts, right_frame in zip(list_of_right_parts, right_frames)
]
```

modin/core/dataframe/pandas/partitioning/partition_manager.py
```diff
@@ -1261,33 +1261,45 @@ def compute_part_size(indexer, remote_part, part_idx, axis):
...
         -------
         np.ndarray
             A NumPy array with new partitions.
         """
-        [part.drain_call_queue() for part in right.flatten()]
+        [
+            part.drain_call_queue()
+            for right_partitions in right
+            for part in right_partitions.flatten()
+        ]
```
vnlitvinov, line 1291: I think we should have a way to execute this in parallel somehow... cc @prutskov

modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -2507,16 +2507,16 @@ def _copartition(self, axis, other, how, sort, force_repartition=False):
         return (reindexed_frames[0], reindexed_frames[1:], joined_index, base_lengths)
 
     @lazy_metadata_decorator(apply_axis="both")
-    def binary_op(self, op, right_frame, join_type="outer"):
+    def nary_op(self, op, right_frames: list, join_type="outer"):
```
prutskov, line 2510: What does this name mean?

modin/core/dataframe/pandas/partitioning/partition_manager.py
prutskov: Do we really want to add some new API to run `drain_call_queue` in parallel?

`part.drain_call_queue` just runs remote calls asynchronously and it probably does not take a lot of time.

modin/core/storage_formats/pandas/query_compiler.py
```diff
@@ -435,10 +425,17 @@ def where_builder_second_pass(df, new_other, **kwargs):
...
                 return df.where(cond, other, **kwargs)
 
-            new_modin_frame = self._modin_frame.binary_op(
-                where_builder_series, cond._modin_frame, join_type="left"
+            new_modin_frame = self._modin_frame.nary_op(
+                where_builder_series, [cond._modin_frame], join_type="left"
             )
-        return self.__constructor__(new_modin_frame)
+        result = self.__constructor__(new_modin_frame)
+        # The binary operation may reorder rows or columns, but where requires
```
prutskov, line 432: Probably "binary" operation will confuse now, as it was renamed.

modin/core/storage_formats/pandas/query_compiler.py
```diff
@@ -435,10 +425,17 @@ def where_builder_second_pass(df, new_other, **kwargs):
...
+        # that we keep the original index and columns
+        if not result.index.equals(self.index):
+            result = result.reindex(axis=0, labels=self.index)
```
prutskov, line 435: `reindex` looks too costly. Could you make some perf measurements to cover this case?

modin/core/dataframe/pandas/dataframe/dataframe.py
mvashishtha: Done.

modin/core/storage_formats/pandas/query_compiler.py
mvashishtha: I passed `join=how` to the binary operation instead. When indexes are different, perf is not much good because we have to do 2 reindexes each for `other` and `cond` before we can even start the `where`. But that's a problem with binary operations in general. Anyway, perf in that case for the example I give is not worse than pandas.

modin/core/dataframe/pandas/partitioning/partition_manager.py
vnlitvinov: I was under false impression that `.drain_call_queue()` might wait on some intermediate results. I now see I was wrong.

modin/core/dataframe/pandas/dataframe/dataframe.py
vnlitvinov: should we name it `n_ary_op` to signify that naming?..

modin/core/storage_formats/pandas/query_compiler.py
vnlitvinov: I wonder if we can `_copartition` and `.reindex` at the same time in a single operation... would that be much faster?

modin/core/dataframe/pandas/dataframe/dataframe.py
mvashishtha: Yep, done.

modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -2516,16 +2516,16 @@ def _copartition(self, axis, other, how, sort, force_repartition=False):
         return (reindexed_frames[0], reindexed_frames[1:], joined_index, base_lengths)
 
     @lazy_metadata_decorator(apply_axis="both")
-    def binary_op(self, op, right_frame, join_type="outer"):
+    def n_ary_op(self, op, right_frames: list, join_type="outer"):
```
vnlitvinov, line 2519: I wonder how this change applies to non-pandas storage formats; also I wonder why I don't see any changes in dask-related execution?

modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -2534,32 +2534,36 @@ def binary_op(self, op, right_frame, join_type="outer"):
...
+        ) = new_left_frame._copartition(1, new_right_frames, join_type, sort=True)
 
         new_frame = (
             np.array([])
-            if len(left_parts) == 0 or len(right_parts[0]) == 0
-            else self._partition_mgr_cls.binary_operation(
-                left_parts, op, right_parts[0]
+            if len(left_parts) == 0
+            or any(len(right_parts) == 0 for right_parts in list_of_right_parts)
+            else self._partition_mgr_cls.nary_operation(
```
vnlitvinov, line 2565:
```suggestion
            else self._partition_mgr_cls.n_ary_operation(
```
?

modin/core/storage_formats/pandas/query_compiler.py
mvashishtha: This comment is outdated now that I took out the `reindex`.

modin/core/dataframe/pandas/partitioning/partition_manager.py
```diff
@@ -1263,33 +1263,45 @@ def compute_part_size(indexer, remote_part, part_idx, axis):
...
-        """
-        Apply a function that requires two ``PandasDataframe`` objects.
+    def n_ary_operation(cls, left, func, right: list):
+        r"""
+        Apply an n-ary operation to multiple ``PandasDataframe`` objects.
+
+        This method assumes that all the partitions of the dataframes in left
+        and right have the same dimensions. For each position i, j in each
+        dataframe's partitions, the result has a partition at (i, j) whose data
+        is func(let_partitions[i,j], \*each_right_partitions[i,j]).
```
vnlitvinov, line 1273:
```suggestion
        is func(left_partitions[i,j], \*each_right_partitions[i,j]).
```

modin/core/storage_formats/pandas/query_compiler.py
```diff
@@ -436,8 +426,8 @@ def where_builder_second_pass(df, new_other, **kwargs):
             def where_builder_series(df, cond):
                 return df.where(cond, other, **kwargs)
 
-            new_modin_frame = self._modin_frame.binary_op(
-                where_builder_series, cond._modin_frame, join_type="left"
+            new_modin_frame = self._modin_frame.nary_op(
```
vnlitvinov, line 429:
```suggestion
            new_modin_frame = self._modin_frame.n_ary_op(
```

modin/core/storage_formats/pandas/query_compiler.py
```diff
@@ -1888,8 +1878,8 @@ def fillna_builder(series, value_arg):
                         # objects (when `limit` parameter is absent) as it works on two `Series`.
                         return series.fillna(value=value_arg, **kwargs)
 
-                    new_modin_frame = self._modin_frame.binary_op(
-                        fillna_builder, value._modin_frame, join_type="left"
+                    new_modin_frame = self._modin_frame.nary_op(
```
vnlitvinov, line 1881:
```suggestion
                    new_modin_frame = self._modin_frame.n_ary_op(
```

modin/core/dataframe/pandas/dataframe/dataframe.py
mvashishtha: > I wonder how this change applies to non-pandas storage formats

Base format doesn't need to fix these bugs according to CI. I don't think we're testing this bug in other storage formats, but I always assume I don't have to worry about that.

The Dask partition manager uses the generic pandas partition manager method I modified here. Ray just wraps that method in a shim so it can apply the progress bar.

vnlitvinov: oh, okay then

Now I only worry if this changes the algebra for the "query compiler spec", and I think we need to either put a fallback `binary_op()` which would call `n_ary_op()` under the hood here, or rename `binary_op()` in other query compilers

cc @modin-project/modin-core @modin-project/modin-omnisci on thoughts

YarShev: I am okay with the renaming of binary op to n_ary op but we should probably change it everywhere.

mvashishtha: > Now I only worry if this changes the algebra for the "query compiler spec",

@vnlitvinov I don't understand this concern. This PR is changing the API of `PandasDataframe`, and that should only affect the pandas storage format. I thought we would only have to change the pandas query compiler. So we what exactly were you thinking of changing? Would I have to implement `nary_op` in other query compilers?

modin/core/storage_formats/pandas/query_compiler.py
```diff
@@ -411,23 +411,13 @@ def where(self, cond, other, **kwargs):
...
-
-            new_modin_frame = self._modin_frame.binary_op(
-                where_builder_second_pass, first_pass, join_type="left"
+            new_modin_frame = self._modin_frame.n_ary_op(
+                lambda df, cond, other: df.where(cond, other, **kwargs),
+                [
+                    cond._modin_frame,
+                    other._modin_frame,
+                ],
+                join_type=None,
```
YarShev, line 420: What does join_type=None mean?

modin/core/dataframe/pandas/dataframe/dataframe.py
vnlitvinov: I think we should maintain the same algebra logic (and hence interface level) for all query compilers.

I suppose implementing a wrapper `binary_op()` which would call `n_ary_op()` under the hood would suit me here just fine.

modin/pandas/test/dataframe/test_join_sort.py
```diff
@@ -529,16 +529,27 @@ def test_sort_values_with_string_index():
...
 
     pandas_result = pandas_df.where(pandas_cond_df, -pandas_df)
     modin_result = modin_df.where(modin_cond_df, -modin_df)
     assert all((to_pandas(modin_result) == pandas_result).all())
 
+    # Test that we choose the right values to replace when `other` == `True`
+    # everywhere.
+    other_data = np.full(shape=pandas_df.shape, fill_value=True)
+    pandas_other = pandas.DataFrame(other_data, columns=columns)
+    modin_other = pd.DataFrame(other_data, columns=columns)
```
prutskov, line 548: Use `create_test_dfs`

modin/pandas/test/dataframe/test_join_sort.py
```diff
@@ -529,16 +529,27 @@ def test_sort_values_with_string_index():
...
+    pandas_result = pandas_df.where(pandas_cond_df, pandas_other)
+    modin_result = modin_df.where(modin_cond_df, modin_other)
+    assert all((to_pandas(modin_result) == pandas_result).all())
```
prutskov, line 551: Probably, `df_equals` would be better to use

docs/release_notes/release_notes-0.16.0.rst
```diff
@@ -36,6 +36,8 @@ Key Features and Updates
   * FIX-#4811: Apply dataframe -> not_dataframe functions to virtual partitions (#4812)
   * FIX-#4099: Use mangled column names but keep the original when building frames from arrow (#4767)
   * FIX-#4838: Bump up modin-spreadsheet to latest master (#4839)
+  * FIX-#4099: Use mangled column names but keep the original when building frames from arrow
```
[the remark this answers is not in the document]
pyrito: I don't think this is needed anymore if you rebase @mvashishtha .

vnlitvinov:
If we're discarding older binary operators in favour of new N-ary, please rename all relevant `binary.py`-like named files and `Binary()`-like named classes.

modin/core/storage_formats/pandas/query_compiler.py
mvashishtha: It means we always use the left labels to align dataframes:

https://github.com/modin-project/modin/blob/bd326f1c4175102489f08d271a53cf374bd9125e/modin/core/dataframe/pandas/dataframe/dataframe.py#L1312-L1314

This is different from a `left` join because it uses the left index and columns exactly, whereas `left` join can change the order even if the binary operands have the same labels in different orders.

e.g. 

```python
import pandas as pd

df1 = pd.DataFrame([[2, 1]], columns=[1, 0])
df2 = pd.DataFrame([[2, 1]], columns=[0, 1])
# does not use df1 columns:
print((df1 + df2).columns)
# Uses df1 columns:
print(df1.where(df1 == 1, df2))
```

I added a comment explaining why we use `None`.

modin/pandas/test/dataframe/test_join_sort.py
```diff
@@ -529,16 +529,27 @@ def test_sort_values_with_string_index():
 
 
 def test_where():
+    columns = list("abcdefghij")
+
     frame_data = random_state.randn(100, 10)
-    pandas_df = pandas.DataFrame(frame_data, columns=list("abcdefghij"))
-    modin_df = pd.DataFrame(frame_data, columns=list("abcdefghij"))
+    pandas_df = pandas.DataFrame(frame_data, columns=columns)
+    modin_df = pd.DataFrame(frame_data, columns=columns)
```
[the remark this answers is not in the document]
mvashishtha: Done.

modin/pandas/test/dataframe/test_join_sort.py
mvashishtha: Done.

modin/pandas/test/dataframe/test_join_sort.py
mvashishtha: Done.

modin/core/dataframe/pandas/dataframe/dataframe.py
mvashishtha: We decided offline to keep `n_ary_op` because a lot of the logic in `Binary` and in in functions that use that class in the query compiler is particular to `n=2`.

vnlitvinov (approved):
LGTM!

pyrito (dismissed):
LGTM - excited to see this come in

RehanSD (approved):
LGTM

Issue #4692, opened 2022-07-20
BUG: transposing frame of virtual partitions raises TypeError: 'ray._raylet.ObjectRef' object is not callable


mvashishtha:
### System information
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: macOS Monterey
- **Modin version** (`modin.__version__`):  cc3bdb7e4bcd2f8305a5faa02d82036af22cc5df
- **Python version**: 3.10.4
- **Code we can use to reproduce**:

```python
import modin.pandas as pd
df_small = pd.DataFrame(list(range(1000)))
df = pd.concat([df_small] * 3)
print(df.T)
```

### Describe the problem
The above snippet works in pandas, but in Modin, I get a wall of errors that includes the following ray task error repeated many times:

<details>
<summary>Stack trace</summary>

```python
RayTaskError(TypeError): ray::_apply_func() (pid=94684, ip=127.0.0.1)
  At least one of the input arguments for this task could not be computed:
ray.exceptions.RayTaskError: ray::deploy_ray_func() (pid=94684, ip=127.0.0.1)
  At least one of the input arguments for this task could not be computed:
ray.exceptions.RayTaskError: ray::deploy_ray_func() (pid=94684, ip=127.0.0.1)
  File "/Users/maheshvashishtha/software_sources/modin/modin/core/execution/ray/implementations/pandas_on_ray/partitioning/virtual_partition.py", line 533, in deploy_ray_func
    result = func(*args, **kwargs)
  File "/Users/maheshvashishtha/software_sources/modin/modin/core/dataframe/pandas/partitioning/axis_partition.py", line 162, in deploy_axis_func
    result = func(dataframe, *kwargs.pop("args", ()), **kwargs)
  File "/Users/maheshvashishtha/software_sources/modin/modin/core/execution/ray/implementations/pandas_on_ray/partitioning/virtual_partition.py", line 419, in drain
    df = func(df, *args, **kwargs)
TypeError: 'ray._raylet.ObjectRef' object is not callable
```
</details>

### Solution

This is a bug with virtual partitions. Using non-virtual partitions (no concat followed by rebalance first) works in Modin:

```python
import modin.pandas as pd
values = list(range(1000)) * 3
df = pd.DataFrame(data=values, index=values)
print(df.T)
```

anmyachev:
Works on [master](https://github.com/modin-project/modin/commit/47b7cae3b6a60fbef33fa9e8319087bd539fb66f).

Issue #4691, opened 2022-07-20
BUG: Can't do binary operations between dataframes with virtual partitions


noloerino:
### System information
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: macOS Monterey 12.4 (Apple Silicon)
- **Modin version** (`modin.__version__`): latest master (cc3bdb)
- **Python version**: 3.9.13
- **Code we can use to reproduce**:
```
import ray
ray.init()
import modin.pandas as pd
s1 = 13
s2 = 13
df1 = pd.concat([pd.DataFrame([i]) for i in range(s1)])
df2 = pd.concat([pd.DataFrame([i]) for i in range(s2)])
print(df1 + df2)
```
If either `s1` or `s2` parameter is 12 or smaller, this error doesn't occur; the error also doesn't occur when `s1=14` and `s2=12` (I have not tried binary searching more thoroughly). The error also doesn't occur when the dataframes are constructed from a single list (such as `df1 = pd.DataFrame([i for i in range(13)])`).

### Describe the problem

On Ray, for dataframes with a certain number of partitions, attempting to add them together seems to cause something in the Modin codebase to try to treat a logical (column) partition as a physical (block) partition by accessing its `_data` field. In the given code, `df1._query_compiler._modin_frame._partitions.shape` is `(7, 1)`.

This may be related to an existing issue since it may be an issue with virtual partition construction, but I'm unsure what the precise root cause is.

### Source code / logs

<details>
<summary>Stack trace</summary>
<pre>
2022-07-20 14:53:39,547	INFO services.py:1456 -- View the Ray dashboard at http://127.0.0.1:8265
UserWarning: When using a pre-initialized Ray cluster, please ensure that the runtime env sets environment variable __MODIN_AUTOIMPORT_PANDAS__ to 1
UserWarning: Distributing <class 'list'> object. This may take some time.
Traceback (most recent call last):
  File "/Users/jhshi/code/modin/repros/new.py", line 8, in <module>
    print(df1 + df2)
  File "/Users/jhshi/code/modin/modin/logging/logger_decorator.py", line 128, in run_and_log
    return obj(*args, **kwargs)
  File "/Users/jhshi/code/modin/modin/pandas/dataframe.py", line 536, in add
    return self._binary_op(
  File "/Users/jhshi/code/modin/modin/logging/logger_decorator.py", line 128, in run_and_log
    return obj(*args, **kwargs)
[...]

Issue #4676, opened 2022-07-15
df.sort_values error ValueError: cannot reindex on an axis with duplicate labels


xiaogangzhu:
### System information
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: Linux Ubuntu 20.04
- **Modin version** (`modin.__version__`): 0.15.2
- **Python version**: 3.8
- **Code we can use to reproduce**:

### Describe the problem

after concat two df, I want to use df.sort_values() but it produces following error, which when I use pandas there is no error.
### Source code / logs

when I use it produce error
`pd.concat([df_proc, last_row_yesterday_copy]).sort_values(by=['msgtime', 'vin'])`

`Traceback (most recent call last):
  File "main.py", line 19, in <module>
    result,last_row_today,bad_vin_list = run_model(data,last_row_yesterday,today)
  File "/home/tstone10/PycharmProjects/isolation_resis/isolation_model.py", line 247, in run_model
    last_row_today = generate_last_row_today(df_proc,last_row_yesterday)
  File "/home/tstone10/PycharmProjects/isolation_resis/isolation_model.py", line 208, in generate_last_row_today
    last_row_today = pd.concat([last_row_today,last_row_yesterday]).sort_values(by='msgtime')
  File "/home/tstone10/anaconda3/envs/tf/lib/python3.8/site-packages/modin/logging/logger_metaclass.py", line 68, in log_wrap
    return method(*args, **kwargs)
  File "/home/tstone10/anaconda3/envs/tf/lib/python3.8/site-packages/modin/pandas/base.py", line 2889, in sort_values
    result = self._query_compiler.sort_rows_by_column_values(
  File "/home/tstone10/anaconda3/envs/tf/lib/python3.8/site-packages/modin/logging/logger_metaclass.py", line 68, in log_wrap
    return method(*args, **kwargs)
  File "/home/tstone10/anaconda3/envs/tf/lib/python3.8/site-packages/modin/core/storage_formats/pandas/query_compiler.py", line 3104, in sort_rows_by_column_values
    result = self.reset_index(drop=True).reindex(axis=0, labels=new_index2)
  File "/home/tstone10/anaconda3/envs/tf/lib/python3.8/site-packages/modin/logging/logger_metaclass.py", line 68, in log_wrap
    return method(*args, **kwargs)
[...]

PR #4829, opened 2022-08-17
REFACTOR-#4530: Unify access to physical data for any partition type

### LINKED ISSUE #4530: Standardize the `oid/future/data/...` property of the `<StorageFormat>On<Engine>DataframePartition` class

prutskov:
## What do these changes do?

This PR adds public API for block partition class to access to actual `._data`. Now, naming between axis/block partitions to access to physical data is aligned.

- [x] Resolves #4530 
- [x] tests added and passing

### DIFF
--- a/asv_bench/benchmarks/utils/common.py
+++ b/asv_bench/benchmarks/utils/common.py
@@ -504,19 +504,7 @@ def execute(
             return
 
         # compatibility with old Modin versions
-        blocks = [
-            block for partition in partitions for block in partition.list_of_blocks
-        ]
-        if ASV_USE_ENGINE == "ray":
-            from ray import wait
-
-            wait(blocks, num_returns=len(blocks))
-        elif ASV_USE_ENGINE == "dask":
-            from dask.distributed import wait
-
-            wait(blocks, return_when="ALL_COMPLETED")
-        elif ASV_USE_ENGINE == "python":
-            pass
+        df._query_compiler._modin_frame._partition_mgr_cls.wait_partitions(partitions)
 
     elif ASV_USE_IMPL == "pandas":
         pass
--- a/asv_bench/benchmarks/utils/common.py
+++ b/asv_bench/benchmarks/utils/common.py
@@ -29,7 +29,6 @@ from typing import Optional, Union
 from .compatibility import (
     ASV_USE_IMPL,
     ASV_DATASET_SIZE,
-    ASV_USE_ENGINE,
     ASV_USE_STORAGE_FORMAT,
 )
 from .data_shapes import RAND_LOW, RAND_HIGH
--- a/asv_bench/benchmarks/utils/common.py
+++ b/asv_bench/benchmarks/utils/common.py
@@ -504,7 +504,9 @@ def execute(
             return
 
         # compatibility with old Modin versions
-        blocks = [block for partition in partitions for block in partition.list_of_blocks]
+        blocks = [
+            block for partition in partitions for block in partition.list_of_blocks
+        ]
         if ASV_USE_ENGINE == "ray":
             from ray import wait
 
@@ -512,7 +514,7 @@ def execute(
         elif ASV_USE_ENGINE == "dask":
             from dask.distributed import wait
 
-            wait(blocks, return_when='ALL_COMPLETED')
+            wait(blocks, return_when="ALL_COMPLETED")
         elif ASV_USE_ENGINE == "python":
             pass
 
--- a/asv_bench/benchmarks/utils/common.py
+++ b/asv_bench/benchmarks/utils/common.py
@@ -504,14 +504,15 @@ def execute(
             return
 
         # compatibility with old Modin versions
+        blocks = [block for partition in partitions for block in partition.list_of_blocks]
         if ASV_USE_ENGINE == "ray":
             from ray import wait
 
-            all(map(lambda partition: wait([partition.list_of_blocks[0]]), partitions))
+            wait(blocks, num_returns=len(blocks))
         elif ASV_USE_ENGINE == "dask":
             from dask.distributed import wait
 
-            all(map(lambda partition: wait(partition.list_of_blocks[0]), partitions))
+            wait(blocks, return_when='ALL_COMPLETED')
         elif ASV_USE_ENGINE == "python":
             pass
 
--- a/asv_bench/benchmarks/utils/common.py
+++ b/asv_bench/benchmarks/utils/common.py
@@ -29,6 +29,7 @@ from typing import Optional, Union
 from .compatibility import (
     ASV_USE_IMPL,
     ASV_DATASET_SIZE,
+    ASV_USE_ENGINE,
     ASV_USE_STORAGE_FORMAT,
 )
 from .data_shapes import RAND_LOW, RAND_HIGH
@@ -503,7 +504,22 @@ def execute(
             return
 
         # compatibility with old Modin versions
-        df._query_compiler._modin_frame._partition_mgr_cls.wait_partitions(partitions)
+        all(
+            map(
+                lambda partition: partition.drain_call_queue() or True,
+                partitions,
+            )
+        )
+        if ASV_USE_ENGINE == "ray":
+            from ray import wait
+
+            all(map(lambda partition: wait([partition._data]), partitions))
+        elif ASV_USE_ENGINE == "dask":
+            from dask.distributed import wait
+
+            all(map(lambda partition: wait(partition._data), partitions))
+        elif ASV_USE_ENGINE == "python":
+            pass
 
     elif ASV_USE_IMPL == "pandas":
         pass
--- a/asv_bench/benchmarks/utils/common.py
+++ b/asv_bench/benchmarks/utils/common.py
@@ -504,20 +504,14 @@ def execute(
             return
 
         # compatibility with old Modin versions
-        all(
-            map(
-                lambda partition: partition.drain_call_queue() or True,
-                partitions,
-            )
-        )
         if ASV_USE_ENGINE == "ray":
             from ray import wait
 
-            all(map(lambda partition: wait([partition._data]), partitions))
+            all(map(lambda partition: wait([partition.list_of_blocks[0]]), partitions))
         elif ASV_USE_ENGINE == "dask":
             from dask.distributed import wait
 
-            all(map(lambda partition: wait(partition._data), partitions))
+            all(map(lambda partition: wait(partition.list_of_blocks[0]), partitions))
         elif ASV_USE_ENGINE == "python":
             pass
 
--- a/modin/core/dataframe/pandas/partitioning/partition.py
+++ b/modin/core/dataframe/pandas/partitioning/partition.py
@@ -54,12 +54,12 @@ class PandasDataframePartition(ABC):  # pragma: no cover
     @property
     def list_of_blocks(self):
         """
-        Get actual physical data holded by partition.
+        Get the list of physical partition objects that compose this partition.
 
         Returns
         -------
         list
-            List of objects, specific for concrete execution ([ray.ObjectRef] e.g.)
+            A list of physical partition objects (``ray.ObjectRef``, ``distributed.Future`` e.g.).
         """
         self.drain_call_queue()
         return [self._data]
--- a/modin/core/dataframe/pandas/partitioning/partition.py
+++ b/modin/core/dataframe/pandas/partitioning/partition.py
@@ -51,6 +51,11 @@ class PandasDataframePartition(ABC):  # pragma: no cover
         """
         pass
 
+    @property
+    def list_of_blocks(self):
+        self.drain_call_queue()
+        return [self._data]
+
     def apply(self, func, *args, **kwargs):
         """
         Apply a function to the object wrapped by this partition.
--- a/modin/core/dataframe/pandas/partitioning/partition.py
+++ b/modin/core/dataframe/pandas/partitioning/partition.py
@@ -53,6 +53,14 @@ class PandasDataframePartition(ABC):  # pragma: no cover
 
     @property
     def list_of_blocks(self):
+        """
+        Get actual physical data holded by partition.
+
+        Returns
+        -------
+        list
+            List of objects, specific for concrete execution ([ray.ObjectRef] e.g.)
+        """
         self.drain_call_queue()
         return [self._data]
 
--- a/modin/core/dataframe/pandas/partitioning/partition.py
+++ b/modin/core/dataframe/pandas/partitioning/partition.py
@@ -61,6 +61,8 @@ class PandasDataframePartition(ABC):  # pragma: no cover
         list
             A list of physical partition objects (``ray.ObjectRef``, ``distributed.Future`` e.g.).
         """
+        # Defer draining call queue until we get the partitions.
+        # TODO Look into draining call queue at the same time as the task
         self.drain_call_queue()
         return [self._data]
 
--- a/modin/core/dataframe/pandas/partitioning/partition_manager.py
+++ b/modin/core/dataframe/pandas/partitioning/partition_manager.py
@@ -1281,15 +1281,23 @@ class PandasDataframePartitionManager(ClassLogger, ABC):
         np.ndarray
             A NumPy array with new partitions.
         """
-        [part.drain_call_queue() for part in right.flatten()]
-
         func = cls.preprocess_func(func)
+
+        def get_right_block(row_idx, col_idx):
+            blocks = right[row_idx][col_idx].list_of_blocks
+            # TODO Resolve this assertion as a part of #4691, because the current implementation assumes
+            # that partition contains only 1 block.
+            assert (
+                len(blocks) == 1
+            ), f"Implementation assumes that partition contains only 1 block, but {len(blocks)} recieved."
+            return blocks[0]
+
         return np.array(
             [
                 [
                     part.apply(
                         func,
-                        right[row_idx][col_idx].list_of_blocks[0],
+                        get_right_block(row_idx, col_idx),
                     )
                     for col_idx, part in enumerate(left[row_idx])
                 ]
--- a/modin/core/dataframe/pandas/partitioning/partition_manager.py
+++ b/modin/core/dataframe/pandas/partitioning/partition_manager.py
@@ -1289,7 +1289,7 @@ class PandasDataframePartitionManager(ClassLogger, ABC):
                 [
                     part.apply(
                         func,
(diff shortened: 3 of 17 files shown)


modin/core/dataframe/pandas/partitioning/partition_manager.py
```diff
@@ -1289,7 +1289,7 @@ def binary_operation(cls, left, func, right):
                 [
                     part.apply(
                         func,
-                        right[row_idx][col_idx]._data,
+                        right[row_idx][col_idx].list_of_blocks[0],
```
mvashishtha, line 1292: Could you add an assertion that `len(list_of_blocks) == 1` for every right partition, and add a TODO linking to #4691? In case there are multiple blocks, I would much rather raise an assertion error here than try to do the multiply with a single block.

modin/core/execution/dask/implementations/pandas_on_dask/partitioning/partition_manager.py
```diff
@@ -49,7 +49,9 @@ def get_objects_from_partitions(cls, partitions):
         list
             The objects wrapped by `partitions`.
         """
-        return DaskWrapper.materialize([partition._data for partition in partitions])
+        return DaskWrapper.materialize(
```
mvashishtha, line 52: Could you add an assertion that `list_of_blocks` has length 1, and also put in the docstring that this function assumes that each partition contains a single block? If someone tries to get objects from partitions that have multiple blocks, we should fail fast instead of just picking up the first block from each partition.

modin/core/execution/dask/implementations/pandas_on_dask/partitioning/partition_manager.py
```diff
@@ -63,4 +65,7 @@ def wait_partitions(cls, partitions):
         partitions : np.ndarray
             NumPy array with ``PandasDataframePartition``-s.
         """
-        wait([partition._data for partition in partitions], return_when="ALL_COMPLETED")
+        wait(
+            [partition.list_of_blocks[0] for partition in partitions],
+            return_when="ALL_COMPLETED",
+        )
```
mvashishtha, line 71:
```suggestion
        wait(
            [block for partition in partitions for block in partition.list_of_blocks],
            return_when="ALL_COMPLETED",
        )
```

modin/core/execution/ray/generic/modin_aqp.py
```diff
@@ -51,7 +51,7 @@ def call_progress_bar(result_parts, line_no):
     except AttributeError:
         return
     pbar_id = str(cell_no) + "-" + str(line_no)
-    futures = [x._data for row in result_parts for x in row]
+    futures = [x.list_of_blocks[0] for row in result_parts for x in row]
```
mvashishtha, line 54:
```suggestion
    futures = [block for row in result_parts for partition in row for block in partition.list_of_blocks]
```

modin/core/execution/ray/implementations/pandas_on_ray/partitioning/partition_manager.py
```diff
@@ -107,7 +107,7 @@ def get_objects_from_partitions(cls, partitions):
         list
             The objects wrapped by `partitions`.
         """
-        return ray.get([partition._data for partition in partitions])
+        return ray.get([partition.list_of_blocks[0] for partition in partitions])
```
mvashishtha, line 110: Ditto from dask comment-- we should assert that each `list_of_blocks` has length 1 and put that assumption in the docstring.

modin/core/execution/ray/implementations/pandas_on_ray/partitioning/partition_manager.py
```diff
@@ -122,7 +122,8 @@ def wait_partitions(cls, partitions):
             NumPy array with ``PandasDataframePartition``-s.
         """
         ray.wait(
-            [partition._data for partition in partitions], num_returns=len(partitions)
+            [partition.list_of_blocks[0] for partition in partitions],
```
mvashishtha, line 125: We should wait on all the blocks. Probably defining `blocks = [block for partition in partitions for block in partition.list_of_blocks]` separately so we can get its length for `num_returns`.

modin/experimental/core/execution/ray/implementations/pyarrow_on_ray/partitioning/axis_partition.py
```diff
@@ -36,7 +36,7 @@ class PyarrowOnRayDataframeAxisPartition(BaseDataframeAxisPartition):
 
     def __init__(self, list_of_blocks):
         # Unwrap from PandasDataframePartition object for ease of use
-        self.list_of_blocks = [obj._data for obj in list_of_blocks]
+        self.list_of_blocks = [obj.list_of_blocks[0] for obj in list_of_blocks]
```
mvashishtha, line 39:
```suggestion
        self.list_of_blocks = [block for partition in list_of_blocks for block in partition.list_of_blocks]
```

I think you could initialize this class with a list including virtual partitions?

asv_bench/benchmarks/utils/common.py
```diff
@@ -504,20 +504,14 @@ def execute(
...
         if ASV_USE_ENGINE == "ray":
             from ray import wait
 
-            all(map(lambda partition: wait([partition._data]), partitions))
+            all(map(lambda partition: wait([partition.list_of_blocks[0]]), partitions))
         elif ASV_USE_ENGINE == "dask":
             from dask.distributed import wait
 
-            all(map(lambda partition: wait(partition._data), partitions))
+            all(map(lambda partition: wait(partition.list_of_blocks[0]), partitions))
```
mvashishtha, line 514: ditt about waiting for all blocks.

asv_bench/benchmarks/utils/common.py
```diff
@@ -504,20 +504,14 @@ def execute(
...
-            map(
-                lambda partition: partition.drain_call_queue() or True,
-                partitions,
-            )
-        )
         if ASV_USE_ENGINE == "ray":
             from ray import wait
 
-            all(map(lambda partition: wait([partition._data]), partitions))
+            all(map(lambda partition: wait([partition.list_of_blocks[0]]), partitions))
```
mvashishtha, line 510: We should wait for all blocks. Can we use the partition manager's `wait_partitions`?

modin/distributed/dataframe/pandas/partitions.py
```diff
@@ -58,12 +58,15 @@ def _unwrap_partitions():
             [p.drain_call_queue() for p in modin_frame._partitions.flatten()]
             if get_ip:
                 return [
-                    [(partition._ip_cache, partition._data) for partition in row]
+                    [
```
mvashishtha, line 61: Could you add an assert that each `list_of_blocks` has length 1? This method already checks below that we are using one of the block partition classes, which should only have one block, but I think it's safer to add an extra check.

mvashishtha (commented):
Thanks @prutskov ! I left comments about some cases where I think we need to take more care about assuming that `list_of_blocks` has only one block.

modin/core/dataframe/pandas/partitioning/partition_manager.py
prutskov: I agree, continuing of execution just hides problem. The check was added

modin/core/execution/dask/implementations/pandas_on_dask/partitioning/partition_manager.py
prutskov: fixed

modin/core/execution/ray/generic/modin_aqp.py
prutskov: fixed

modin/core/execution/ray/implementations/pandas_on_ray/partitioning/partition_manager.py
prutskov: fixed

modin/core/execution/dask/implementations/pandas_on_dask/partitioning/partition_manager.py
prutskov: The assertion and docstring were added

modin/distributed/dataframe/pandas/partitions.py
prutskov: The assertion was added

asv_bench/benchmarks/utils/common.py
prutskov: As I see, we already use internal functionality here like `df._query_compiler._modin_frame._execute()`, so I think using of `df._query_compiler._modin_frame._partition_mgr_cls.wait_partitions(partitions)` it's not against the rules ) 

Fixed

asv_bench/benchmarks/utils/common.py
prutskov: Use single way for ray/dask/python `ASV_USE_ENGINE` using the partition manager's `wait_partitions`

asv_bench/benchmarks/utils/common.py
prutskov: I think that variable `ASV_USE_ENGINE` isn't need anymore, because there is no any place where it's used. But I think we should remove the definition of this var from asv_bench/benchmarks/utils/compatibility.py in a separate PR

modin/experimental/core/execution/ray/implementations/pyarrow_on_ray/partitioning/axis_partition.py
prutskov: I'm not sure that it will be correct. In this case len(`self.list_of_blocks`) might be different. Are we breaking the correct partitioning of `PyarrowOnRayDataframeAxisPartition` here or not?

mvashishtha: I guess we should assume these are always "full axis" partitions that are made of other `PyarrowOnRayDataframeAxisPartition`. Let's add another assertion here that all `list_of_blocks` have length 1.

asv_bench/benchmarks/utils/common.py
```diff
@@ -504,22 +503,7 @@ def execute(
...
-            from ray import wait
-
-            all(map(lambda partition: wait([partition._data]), partitions))
-        elif ASV_USE_ENGINE == "dask":
-            from dask.distributed import wait
-
-            all(map(lambda partition: wait(partition._data), partitions))
-        elif ASV_USE_ENGINE == "python":
-            pass
+        df._query_compiler._modin_frame._partition_mgr_cls.wait_partitions(partitions)
```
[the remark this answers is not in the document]
prutskov: Does it mean that we can use `asv_bench` and `modin` of different versions? What is the need of this approach?
cc @anmyachev 

@mvashishtha Could you explain what's the need to have different approaches to check this?
1 - using `partition.wait`
2 - using `ray/distributed.wait([partition._data,...])`
3 - using `df._query_compiler._modin_frame._partition_mgr_cls.wait_partitions(partitions)`
4 - something new after API changes

Would it be better to use strict aligned version of `asv/bench` and `modin`?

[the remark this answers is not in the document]
anmyachev: The ASV has the ability to run the code of different modin versions. For example, you can compare `concat` function for the next Modin releases: 0.8 and 0.14. The code from `asv_bench` folder remains unchanged. Therefore, such code is necessary to maintain compatibility, but if we can decide that we will no longer check for performance on such and such an old release, then the corresponding compatibility layer can be removed.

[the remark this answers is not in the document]
anmyachev: yes, you are right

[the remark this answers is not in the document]
prutskov: Probably, need to define process of handling different API versions to not create a tree from `if-else`. I suggest to not include the changes in this file (leave current code as it is) in this PR, but create a new issue about fix of backward compatibility for asv_bench and discuss/implement in a separate PR with support of defined API versions.

@anmyachev, @mvashishtha waht do you think?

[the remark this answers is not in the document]
anmyachev: agree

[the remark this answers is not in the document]
prutskov: The issue is #4877

mvashishtha (approved):
LGTM @prutskov, thank you!

modin/core/dataframe/pandas/partitioning/partition.py
```diff
@@ -51,6 +51,21 @@ def get(self):
         """
         pass
 
+    @property
+    def list_of_blocks(self):
```
YarShev, line 55: Why did we decide to use `list_of_blocks` for a block partition? Doesn't it seem confusing?

mvashishtha: @prutskov, @vnlitvinov , and I thought it was more confusing to have `list_of_blocks` for non-full-axis virtual partitions, and a separate accessor like `block` or `data` for block partitions. non-full-axis virtual partitions and block partitions are also supposed to expose the same API.

In the long term, I want to consolidate the block and virtual partitions into a single class for each engine and remove the `full_axis` param from the axis partition constructor. We would always use the `num_splits` argument to `apply` explicitly to specify the number of splits. That will reduce code duplication and prevent bugs like the many we've had from not treating non-full-axis virtual partitions the same as full-axis virtual partitions, e.g. #4691, #4692, #4676, #4464. Using a common `list_of_blocks` for both partition class takes us toward that goal.

One alternative was to make all partition classes have a block accessor like `data()` that forces materialization of virtual partitions with multiple blocks, but this didn't seem necessary for any existing use case, and it seemed better to combine the blocks within the remote `apply` call for cases like binary operations, where we [use all the blocks](https://github.com/modin-project/modin/blob/eee5f435f68786778184c4886ff25d386fce0c4f/modin/core/dataframe/pandas/partitioning/partition_manager.py#L1288).

YarShev: > it was more confusing to have `list_of_blocks` for non-full-axis virtual partitions

I do not understand this. A virtual partition can encompass multiple block partitions. By calling `list_of_blocks` we would get the partitions.

> it was more confusing to have a separate accessor like block or data for block partitions

I do not understand this too. By calling partition.data we would get the data.

> non-full-axis virtual partitions and block partitions are also supposed to expose the same API.

Not sure that will lead to a nice design but will be happy to look at suggestions.

mvashishtha: I have below some inline responses to parts of your question. There was a request for comments (RFC) for this PR, but it was not public until today. [Here it is](https://www.notion.so/ponderdata/Unify-access-to-physical-data-of-partitions-93e06826ebe14937b3bc722c845947ee).

> I do not understand this too. By calling [partition.data](http://partition.data/) we would get the data.

If you can access the "updates" in the RFC link, you can scroll down to my comment starting with "still think list_of_blocks is better" and see the comment thread there. That section will hopefully explain more clearly the rationale for using list_of_blocks. Unfortunately you have to dig for the comment thread like this because we deleted the text that the comments were originally referring to.

> Not sure that will lead to a nice design but will be happy to look at suggestions.

Having non-full-axis virtual partitions and block partition expose the same API is not a proposal. It's already an assumption in the code that uses partitions, and it was so even prior to this PR. virtual partitions with `full_axis=False` are supposed to behave like block partitions, e.g. calling `apply` on them returns a single partition instead of a list of partitions.

> In the long term, I want to consolidate the block and virtual partitions into a single class 

I'll make sure to send a public proposal about this to you and other active modin maintainers before trying to implement it.

YarShev: Unfortunately, I can't access the RFC. If you could provide me with some brief comment, that would be great.

> I'll make sure to send a public proposal about this to you and other active modin maintainers before trying to implement it.

Thanks!

PR #4830, opened 2022-08-17
FIX-#4734: Handle `Series.apply` when return type is a `DataFrame`

### LINKED ISSUE #4734: BUG: passing a function returning a series to Series.apply raises KeyError: 'DataFrame'
### System information
- **Modin version** (`modin.__version__`): 05bf659b67d2b7153d8aaa004dd95252377eed51
- **Python version**: 3.10.4

```python
import modin.pandas as pd
import pandas

pd.Series(1).apply(lambda x: pandas.Series([x, x]))
```

### Describe the problem
In pandas the above script gives a dataframe:

```
   0  1
0  1  1
```

In modin I get an error from `result = globals()[return_type](query_compiler=result)`. The string `"DataFrame"` isn't in `globals()`.

### Source code / logs
<details>

```
UserWarning: Distributing <class 'int'> object. This may take some time.
Input In [3], in <cell line: 4>()
      1 import modin.pandas as pd
      2 import pandas
----> 4 pd.Series(1).apply(lambda x: pandas.Series([x, x]))

File ~/software_sources/modin/modin/logging/logger_decorator.py:128, in enable_logging.<locals>.decorator.<locals>.run_and_log(*args, **kwargs)
    113 """
    114 Compute function with logging if Modin logging is enabled.
    115
   (...)
    125 Any
    126 """
    127 if LogMode.get() == "disable":
--> 128     return obj(*args, **kwargs)
    130 logger = get_logger()
    131 logger_level = getattr(logger, log_level)

File ~/software_sources/modin/modin/_compat/pandas_api/latest/series.py:28, in LatestCompatibleSeries.apply(self, func, convert_dtype, args, **kwargs)
     27 def apply(self, func, convert_dtype=True, args=(), **kwargs):
---> 28     return self._apply(func, convert_dtype=convert_dtype, args=args, **kwargs)

File ~/software_sources/modin/modin/logging/logger_decorator.py:128, in enable_logging.<locals>.decorator.<locals>.run_and_log(*args, **kwargs)
    113 """
    114 Compute function with logging if Modin logging is enabled.
    115
   (...)
    125 Any
    126 """
    127 if LogMode.get() == "disable":
--> 128     return obj(*args, **kwargs)
    130 logger = get_logger()
    131 logger_level = getattr(logger, log_level)

File ~/software_sources/modin/modin/pandas/series.py:663, in Series._apply(self, func, convert_dtype, args, **kwargs)
    661         return result
    662 else:
--> 663     result = globals()[return_type](query_compiler=result)
    664     if result.name == self.index[0]:
    665         result.name = None

KeyError: 'DataFrame'
```

</details>

pyrito:

## What do these changes do?

This PR addresses a bug where `Series.apply` returns the incorrect value when the return type of the function is a `DataFrame`.

- [x] Resolves #4734 
- [x] tests added and passing

### DIFF
--- a/modin/core/storage_formats/base/query_compiler.py
+++ b/modin/core/storage_formats/base/query_compiler.py
@@ -2203,7 +2203,9 @@ class BaseQueryCompiler(ClassLogger, abc.ABC):
     # UDF (apply and agg) methods
     # There is a wide range of behaviors that are supported, so a lot of the
     # logic can get a bit convoluted.
-    def apply(self, func, axis, raw=False, result_type=None, *args, **kwargs):
+    def apply(
+        self, func, axis, raw=False, result_type=None, ret_series=False, *args, kwargs
+    ):
         """
         Apply passed function across given axis.
 
@@ -2225,9 +2227,12 @@ class BaseQueryCompiler(ClassLogger, abc.ABC):
             - "reduce": keep result into a single cell (opposite of "expand").
             - "broadcast": broadcast result to original data shape (overwrite the existing column/row with the function result).
             - None: use "expand" strategy if Series is returned, "reduce" otherwise.
+        ret_series : bool
+            Flag that specifies whether we have a func that returns a Series
+            object and are working on a Series object.
         *args : iterable
             Positional arguments to pass to `func`.
-        **kwargs : dict
+        kwargs : dict
             Keyword arguments to pass to `func`.
 
         Returns
@@ -2244,6 +2249,15 @@ class BaseQueryCompiler(ClassLogger, abc.ABC):
             - Each element is the result of execution of `func` against
               corresponding row/column.
         """
+        # There are some cases where a func to be applied returns a Series object and
+        # needs to be handled differently.
+        if self.is_series_like() and ret_series:
+            return SeriesDefault.register(pandas.Series.apply)(
+                self,
+                func=func,
+                *args,
+                **kwargs,
+            )
         return DataFrameDefault.register(pandas.DataFrame.apply)(
             self,
             func=func,
--- a/modin/core/storage_formats/base/query_compiler.py
+++ b/modin/core/storage_formats/base/query_compiler.py
@@ -2227,7 +2227,7 @@ class BaseQueryCompiler(ClassLogger, abc.ABC):
             - "reduce": keep result into a single cell (opposite of "expand").
             - "broadcast": broadcast result to original data shape (overwrite the existing column/row with the function result).
             - None: use "expand" strategy if Series is returned, "reduce" otherwise.
-        ret_series : bool
+        ret_series : bool, default: False
             Flag that specifies whether we have a func that returns a Series
             object and are working on a Series object.
         *args : iterable
--- a/modin/core/storage_formats/base/query_compiler.py
+++ b/modin/core/storage_formats/base/query_compiler.py
@@ -2203,9 +2203,7 @@ class BaseQueryCompiler(ClassLogger, abc.ABC):
     # UDF (apply and agg) methods
     # There is a wide range of behaviors that are supported, so a lot of the
     # logic can get a bit convoluted.
-    def apply(
-        self, func, axis, raw=False, result_type=None, ret_series=False, *args, kwargs
-    ):
+    def apply(self, func, axis, raw=False, result_type=None, *args, **kwargs):
         """
         Apply passed function across given axis.
 
@@ -2227,12 +2225,9 @@ class BaseQueryCompiler(ClassLogger, abc.ABC):
             - "reduce": keep result into a single cell (opposite of "expand").
             - "broadcast": broadcast result to original data shape (overwrite the existing column/row with the function result).
             - None: use "expand" strategy if Series is returned, "reduce" otherwise.
-        ret_series : bool, default: False
-            Flag that specifies whether we have a func that returns a Series
-            object and are working on a Series object.
         *args : iterable
             Positional arguments to pass to `func`.
-        kwargs : dict
+        **kwargs : dict
             Keyword arguments to pass to `func`.
 
         Returns
@@ -2249,21 +2244,38 @@ class BaseQueryCompiler(ClassLogger, abc.ABC):
             - Each element is the result of execution of `func` against
               corresponding row/column.
         """
-        # There are some cases where a func to be applied returns a Series object and
-        # needs to be handled differently.
-        if self.is_series_like() and ret_series:
-            return SeriesDefault.register(pandas.Series.apply)(
-                self,
-                func=func,
-                *args,
-                **kwargs,
-            )
         return DataFrameDefault.register(pandas.DataFrame.apply)(
             self,
             func=func,
             axis=axis,
             raw=raw,
             result_type=result_type,
+            *args,
+            **kwargs,
+        )
+
+    def apply_on_series(self, func, *args, **kwargs):
+        """
+        Apply passed function on underlying Series.
+
+        Parameters
+        ----------
+        func : callable(pandas.Series) -> scalar, str, list or dict of such
+            The function to apply to each row.
+        *args : iterable
+            Positional arguments to pass to `func`.
+        **kwargs : dict
+            Keyword arguments to pass to `func`.
+
+        Returns
+        -------
+        BaseQueryCompiler
+        """
+        assert self.is_series_like()
+
+        return SeriesDefault.register(pandas.Series.apply)(
+            self,
+            func=func,
             *args,
             **kwargs,
         )
--- a/modin/core/storage_formats/pandas/query_compiler.py
+++ b/modin/core/storage_formats/pandas/query_compiler.py
@@ -2381,13 +2381,15 @@ class PandasQueryCompiler(BaseQueryCompiler):
     # UDF (apply and agg) methods
     # There is a wide range of behaviors that are supported, so a lot of the
     # logic can get a bit convoluted.
-    def apply(self, func, axis, is_series, *args, **kwargs):
+    def apply(self, func, axis, *args, **kwargs):
         # if any of args contain modin object, we should
         # convert it to pandas
 
         # TODO(vkarthik): currently is_series only works for callable
         # functions that return Series objects. We should handle this
         # for other cases as well.
+        is_series = kwargs.pop("is_series", False)
+
         args = try_cast_to_pandas(args)
         kwargs = try_cast_to_pandas(kwargs)
         if isinstance(func, dict):
--- a/modin/core/storage_formats/pandas/query_compiler.py
+++ b/modin/core/storage_formats/pandas/query_compiler.py
@@ -2381,7 +2381,7 @@ class PandasQueryCompiler(BaseQueryCompiler):
     # UDF (apply and agg) methods
     # There is a wide range of behaviors that are supported, so a lot of the
     # logic can get a bit convoluted.
-    def apply(self, func, axis, ret_series=False, *args, kwargs):
+    def apply(self, func, axis, *args, **kwargs):
         # if any of args contain modin object, we should
         # convert it to pandas
         args = try_cast_to_pandas(args)
@@ -2391,7 +2391,18 @@ class PandasQueryCompiler(BaseQueryCompiler):
         elif is_list_like(func):
             return self._list_like_func(func, axis, *args, **kwargs)
         else:
-            return self._callable_func(func, axis, ret_series, *args, **kwargs)
+            return self._callable_func(func, axis, *args, **kwargs)
+
+    def apply_on_series(self, func, *args, **kwargs):
+        args = try_cast_to_pandas(args)
+        kwargs = try_cast_to_pandas(kwargs)
+
+        assert self.is_series_like()
+        return self.__constructor__(
+            self._modin_frame.apply_full_axis(
+                1, lambda df: df.squeeze(axis=1).apply(func, *args, **kwargs)
+            )
+        )
 
     def _dict_func(self, func, axis, *args, **kwargs):
         """
@@ -2470,7 +2481,7 @@ class PandasQueryCompiler(BaseQueryCompiler):
         )
         return self.__constructor__(new_modin_frame)
 
-    def _callable_func(self, func, axis, ret_series, *args, **kwargs):
+    def _callable_func(self, func, axis, *args, **kwargs):
         """
         Apply passed function to each row/column.
 
@@ -2481,9 +2492,6 @@ class PandasQueryCompiler(BaseQueryCompiler):
         axis : {0, 1}
             Target axis to apply function along. 0 means apply to columns,
             1 means apply to rows.
-        ret_series : bool
-            Flag that specifies whether we have a func that returns a Series
-            object and are working on a Series object.
         *args : args
             Arguments to pass to the specified function.
         **kwargs : kwargs
@@ -2498,16 +2506,11 @@ class PandasQueryCompiler(BaseQueryCompiler):
         if callable(func):
             func = wrap_udf_function(func)
 
-        def apply_func(df):
-            # Functions that return Series objects and are applied onto
-            # Series objects have different behaviors that need to be
-            # handled with the pandas.Series.apply method specifically.
-            if self.is_series_like() and ret_series:
-                return df.squeeze(axis=1).apply(func, *args, **kwargs)
-            else:
-                return df.apply(func, axis=axis, *args, **kwargs)
-
-        return self.__constructor__(self._modin_frame.apply_full_axis(axis, apply_func))
+        return self.__constructor__(
+            self._modin_frame.apply_full_axis(
+                axis, lambda df: df.apply(func, axis=axis, *args, **kwargs)
+            )
+        )
 
     # END UDF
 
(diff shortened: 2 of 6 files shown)


modin/core/storage_formats/pandas/query_compiler.py
```diff
@@ -2363,17 +2363,21 @@ def explode(self, column):
...
+        # for other cases as well.
         args = try_cast_to_pandas(args)
         kwargs = try_cast_to_pandas(kwargs)
         if isinstance(func, dict):
             return self._dict_func(func, axis, *args, **kwargs)
         elif is_list_like(func):
             return self._list_like_func(func, axis, *args, **kwargs)
         else:
-            return self._callable_func(func, axis, *args, **kwargs)
+            return self._callable_func(func, axis, is_series, *args, **kwargs)
```
pyrito, line 2380: I'm not sure if this is the best place to put this logic. Does anyone know where else we could put this information?

modin/pandas/series.py
```diff
@@ -660,19 +660,28 @@ def f(x):
...
+
+                # The return_type is only a DataFrame when we have a function
+                # return a Series object. This is a very particular case that
+                # has to be handled by the underlying pandas.Series apply
+                # function and not our default applymap call.
+                if return_type == "DataFrame":
+                    result = self._query_compiler.apply(f, axis=1, is_series=True)
+                else:
+                    result = self.map(f)._query_compiler
+        from .dataframe import DataFrame
```
vnlitvinov, line 672: let's import that under `if return_type == "DataFrame"`

vnlitvinov (changes requested):
This overall seems like a weird approach architecture-wise, maybe @devin-petersohn or @YarShev can tell what would be better in terms of enhancing query compilers?

modin/core/storage_formats/pandas/query_compiler.py
```diff
@@ -2376,17 +2376,21 @@ def explode(self, column):
     # UDF (apply and agg) methods
     # There is a wide range of behaviors that are supported, so a lot of the
     # logic can get a bit convoluted.
-    def apply(self, func, axis, *args, **kwargs):
+    def apply(self, func, axis, is_series=False, *args, **kwargs):
```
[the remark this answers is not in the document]
pyrito: Yeah, I was wondering if there is a better place to encapsulate this logic. It might be better to just hide this in the `kwargs` as well, but I'm not too sure. 

I'm not sure why there isn't a docstring here, but we can add one.

mvashishtha:
> This overall seems like a weird approach architecture-wise, maybe @devin-petersohn or @YarShev can tell what would be better in terms of enhancing query compilers?

@vnlitvinov @pyrito maybe a better approach is to add a query compiler function that `Map`s a user-defined `dataframe -> dataframe` function on each block. The API-layer code could then convert the scalar -> series function to a dataframe -> dataframe function as this PR is already doing, and pass the new function into the new query compiler function. The advantage of this approach is that we don't pollute the existing "row-wise or column-wise" `apply` function at the query compiler. A disadvantage is that the API layer code has to do a function transformation, but that's usually the job of the query compiler. We could call the new function `apply_to_blocks_as_series` or something like that and do the function transformation inside the query compiler as this PR already does.

modin/pandas/test/test_series.py
```diff
@@ -4705,3 +4705,30 @@ def func(val):
...
+
+
+@pytest.mark.parametrize("data", test_data_values, ids=test_data_keys)
+def test_apply_return_df_multiple_partitions(data):
+    modin_series, pandas_series = create_test_series(data)
+    modin_df = modin_series.apply(
+        lambda x: pandas.Series([x, x + 1, x + 2], index=["a", "b", "c"])
+    )
+    pandas_df = pandas_series.apply(
+        lambda x: pandas.Series([x, x + 1, x + 2], index=["a", "b", "c"])
```
RehanSD, line 4731: Should we also test when there will be multiple horizontal partitions too? I.e. increase the number of columns to make sure things get block partitioned appropriately and don't become long axis chunks?

modin/core/storage_formats/pandas/query_compiler.py
```diff
@@ -2379,14 +2379,20 @@ def explode(self, column):
     def apply(self, func, axis, *args, **kwargs):
```
RehanSD, line 2379: Potentially, instead of handling the logic in `_callable_func` do we want to create another method like `_callable_func_{on|with}_series` and call that from apply instead?

modin/core/storage_formats/pandas/query_compiler.py
```diff
@@ -2379,14 +2379,20 @@ def explode(self, column):
...
         # if any of args contain modin object, we should
         # convert it to pandas
+
+        # TODO(vkarthik): currently is_series only works for callable
+        # functions that return Series objects. We should handle this
```
RehanSD, line 2384: I'm not sure I follow? In the `_callable_func` docstring, `is_series` is documented as a flag that determines whether or not we're working with a Series, but here it seems to indicate that its a flag for when the function will return a Series? Perhaps we should update the docstring and this comment to something like `is_series` is a flag that indicates we are applying a function to a Series object that will result in a Dataframe?

RehanSD (commented):
Left some minor comments, but looks good to me!

RehanSD:
> better

We already have an API that does a blockwise map, if I'm understanding your comment correctly - the algebra layer map currently only does a blockwise map (eventually it will do a map along either an axis or block-wise and have an axis argument when we finish implementing the algebra). We could use that, but I don't think it would actually change any of the code functionally (although maybe it would be marginally faster since `map` is much simpler than `apply_full_axis`). That also feels hacky though, since there is an implicit assumption that under the hood, Series are partitioned only along one axis (which does make sense), but maybe we should document that that is what is happening in the code? (i.e. say we are using `map` here since a Series is only partitioned along one axis, so there is no need to do a row-wise apply when we already currently have row-wise partitions technically).

mvashishtha:
@RehanSD 

> We already have an API that does a blockwise map,

Yes, that's what I was proposing we use.

> We could use that, but I don't think it would actually change any of the code functionally

Yes. blockwise map made a little more sense to me because I didn't see a need to concatenate along the row axis. I'm not that concerned about whether this is a full-axis map or a blockwise map.

 > That also feels hacky though, since there is an implicit assumption that under the hood, Series are partitioned only along one axis

We should consider it valid for Series to have more than one column of block partitions. I think this should actually work if you have a series partitioned along both axes. It's just that one or more of the columns of partitions might consist of empty dataframes. The wrapped UDF should map those dataframes to empty dataframes anyway, so I don't think we'd have reason to be concerned.

modin/core/storage_formats/pandas/query_compiler.py
pyrito: That might be cleaner/more readable.

modin/pandas/test/test_series.py
mvashishtha: +1, but we can just do a single test case where the function result spans multiple horizontal partitions.

modin/core/storage_formats/pandas/query_compiler.py
pyrito: I remember why I called it `is_series` now. Technically the code I've added here should work for any case where we know we are trying to apply a function to a Series object. However, we should probably just use this code path for when we know what we are returning a Series object in the `func` being applied to an existing Series object. This will have to be documented clearly I think. There seems to be no way around this but we have to handle this weirdness from pandas.

modin/pandas/test/test_series.py
pyrito: After testing it seems like what's happening is that we are getting long axis chunks. The number of partitions isn't changing after doing an apply that creates lots of columns.

modin/core/storage_formats/base/query_compiler.py
```diff
@@ -2229,6 +2234,15 @@ def apply(self, func, axis, raw=False, result_type=None, *args, **kwargs):
             - Each element is the result of execution of `func` against
               corresponding row/column.
         """
+        # There are some cases where a func to be applied returns a Series object and
+        # needs to be handled differently.
+        if self.is_series_like() and ret_series:
```
vnlitvinov, line 2239: seeing things like this makes me think that original assumption that "there's no difference between Series and DataFrame on a low level" is completely false... we have quite a few edge cases already, and this one just adds to the heap

cc @devin-petersohn

devin-petersohn: pandas `Series.apply` should be an applymap, which should ideally be handled at the API layer. If the return type is a `modin.pandas.DataFrame`, ideally we should just convert to a `modin.pandas.DataFrame` at the API layer (`series.py`) and then do `apply(axis=1)`.

pyrito: @devin-petersohn So this case is actually a bit more nuanced than that. I think this is one of those cases where we **need** the `pandas.Series.apply` call instead of the DataFrame one because of an implementation detail.

See below:
```python
In [6]: pandas.Series(1).apply(lambda x: pandas.Series([x, x]))
   ...: 
   ...: 
Out[6]: 
   0  1
0  1  1

In [9]: pandas.DataFrame([1]).apply(lambda x: pandas.Series([x, x]), axis=1)
Out[9]: 
                              0                             1
0  0    1
Name: 0, dtype: int64  0    1
Name: 0, dtype: int64
```

YarShev: I think we can go with either of the both approaches for now. 1) Add a function to the QC where we use a Series as the object the function is going to be called on (like for [series_view](https://github.com/modin-project/modin/blob/84f638cc703beb550edecce715532c8e86fb7b25/modin/core/storage_formats/pandas/query_compiler.py#L1358)). 2) Add a parameter like `squeeze_self` to the current apply in the QC and in depend of the parameter we squeeze DataFrame or not in remote tasks (like for [dot](https://github.com/modin-project/modin/blob/84f638cc703beb550edecce715532c8e86fb7b25/modin/core/storage_formats/pandas/query_compiler.py#L1711)).

As a separate issue, we will need to define one of the approaches we will go with further. Of course, if we will not find a third way.

pyrito: @YarShev you can see in the PR I have gone with the first approach now. Please take a look.

pyrito: The second option is dicy because we need to change the function signature of `apply` and that breaks some things.

modin/core/storage_formats/pandas/query_compiler.py
```diff
@@ -2387,6 +2387,17 @@ def apply(self, func, axis, *args, **kwargs):
...
             return self._callable_func(func, axis, *args, **kwargs)
 
+    def apply_on_series(self, func, *args, **kwargs):
+        args = try_cast_to_pandas(args)
+        kwargs = try_cast_to_pandas(kwargs)
+
+        assert self.is_series_like()
+        return self.__constructor__(
+            self._modin_frame.apply_full_axis(
+                1, lambda df: df.squeeze(axis=1).apply(func, *args, **kwargs)
```
YarShev, line 2397: Why do we have to use `apply_full_axis` instead of `map` as it was before? That significantly slows down the things.

pyrito: I tried using map but things were breaking. I can try it again and look into it.

YarShev: Yes, please look into it.

pyrito: @YarShev we can't use `map` because I think the function doesn't assume that the dimensions of the DataFrame will not change. That's why I went with `apply_full_axis` here since we may need to recompute some metadata/indexing.

YarShev: So probably for now we only can go with `apply_full_axis`. That fixes the issue but the perf is not pretty good. As to the perf, we can thing of it as part of a separate issue.

YarShev (dismissed):

modin/core/storage_formats/pandas/query_compiler.py
```diff
@@ -2393,6 +2393,17 @@ def apply(self, func, axis, *args, **kwargs):
...
```
vnlitvinov, line 2403: @pyrito please add a comment here which is explaining why we picked `apply_full_axis` instead of `map`

modin/core/storage_formats/pandas/query_compiler.py
```diff
@@ -2495,10 +2506,11 @@ def _callable_func(self, func, axis, *args, **kwargs):
...
             func = wrap_udf_function(func)
 
-        new_modin_frame = self._modin_frame.apply_full_axis(
-            axis, lambda df: df.apply(func, axis=axis, *args, **kwargs)
+        return self.__constructor__(
+            self._modin_frame.apply_full_axis(
+                axis, lambda df: df.apply(func, axis=axis, *args, **kwargs)
+            )
         )
-        return self.__constructor__(new_modin_frame)
```
vnlitvinov, line 2501: this change seems unrelated to the PR itself, it's some small code refactoring, we should not be squeezing unrelated things in PRs

modin/pandas/series.py
```diff
@@ -671,19 +671,29 @@ def f(x):
...
+                # return a Series object. This is a very particular case that
+                # has to be handled by the underlying pandas.Series apply
+                # function and not our default applymap call.
+                if return_type == "DataFrame":
+                    result = self._query_compiler.apply_on_series(f)
+                else:
+                    result = self.map(f)._query_compiler
+
+        if return_type == "DataFrame":
+            from .dataframe import DataFrame
```
vnlitvinov, line 685: can we move this import at the top level of the file?

modin/pandas/test/test_series.py
```diff
@@ -4711,3 +4711,31 @@ def func(val):
...
+    pandas_df = pandas.Series(1).apply(lambda x: pandas.Series([x, x]))
+
+    df_equals(modin_df, pandas_df)
+
+    modin_df = pd.Series([2, 4, 6, 0, 1]).apply(lambda x: pandas.Series([x, x + 1]))
+    pandas_df = pandas.Series([2, 4, 6, 0, 1]).apply(
+        lambda x: pandas.Series([x, x + 1])
+    )
+
+    df_equals(modin_df, pandas_df)
```
vnlitvinov, line 4727: can we use `df_eval(modin_df, pandas_df, lambda df: df.apply(lambda x: pandas.Series([x, x])))` or smth similar here?

modin/pandas/test/test_series.py
```diff
@@ -4711,3 +4711,31 @@ def func(val):
...
+def test_apply_return_df_multiple_partitions(data, num_columns):
+    modin_series, pandas_series = create_test_series(data)
+    modin_df = modin_series.apply(
+        lambda x: pandas.Series([x + i for i in range(num_columns)])
+    )
+    pandas_df = pandas_series.apply(
+        lambda x: pandas.Series([x + i for i in range(num_columns)])
+    )
+
+    df_equals(modin_df, pandas_df)
```
vnlitvinov, line 4741: same on `df_eval()`

modin/pandas/series.py
pyrito: No, there is a problem with cyclic dependencies when I tried doing that. The pattern in this file is to import this when needed.

vnlitvinov (dismissed):
LGTM!

Though I wonder what happens for other backends?..

pyrito:
@vnlitvinov do you mean for other storage formats? At the very least the BaseQueryCompiler has implemented a version of `apply_on_series` if needed.

modin/pandas/test/test_series.py
```diff
@@ -4711,3 +4711,25 @@ def func(val):
...
+
+
+def test_apply_return_df():
+    modin_series, pandas_series = create_test_series([2, 4, 6, 0, 1])
+    eval_general(
+        modin_series,
+        pandas_series,
+        lambda series: series.apply(lambda x: pandas.Series([x, x + 1])),
+    )
+
```
mvashishtha, line 4723:
```suggestion
```

I don't see the need to test a small series in addition to a large series.

modin/pandas/test/test_series.py
```diff
@@ -4711,3 +4711,25 @@ def func(val):
...
+
+@pytest.mark.parametrize("data", test_data_values, ids=test_data_keys)
+@pytest.mark.parametrize("num_columns", [4, 8, 16, 32, 64])
+def test_apply_return_df_multiple_partitions(data, num_columns):
```
mvashishtha, line 4727:
```suggestion
def test_apply_return_df(data):
```

I don't see the value of testing multiple `num_columns`.

modin/pandas/test/test_series.py
```diff
@@ -4711,3 +4711,25 @@ def func(val):
...
+    modin_series, pandas_series = create_test_series(data)
+    eval_general(
+        modin_series,
+        pandas_series,
+        lambda series: series.apply(
+            lambda x: pandas.Series([x + i for i in range(num_columns)])
```
mvashishtha, line 4733:
```suggestion
            lambda x: pandas.Series([x + i for i in range(100)])
```

mvashishtha (commented):
I left minor comments about the tests.

modin/pandas/test/test_series.py
pyrito: I figured it would be better to test it with one partition and then with multiple partitions?

mvashishtha (approved):
LGTM! Thanks @pyrito

PR #4852, opened 2022-08-22
PERF-#4851: Compute `dtypes` for binary operations that can only return bool type and the right operand is not a Modin object

### LINKED ISSUE #4851: PERF: compute `dtypes` for binary operations that can only return bool type

anmyachev:

## What do these changes do?

- [x] Resolves #4851 
- [x] tests added and passing

### DIFF
--- a/modin/core/dataframe/algebra/default2pandas/default.py
+++ b/modin/core/dataframe/algebra/default2pandas/default.py
@@ -68,7 +68,7 @@ class DefaultMethod(Operator):
         if type(fn) == property:
             fn = cls.build_property_wrapper(fn)
 
-        def applyier(df, *args, dtypes=None, **kwargs):
+        def applyier(df, *args, **kwargs):
             """
             Apply target function to the casted to pandas frame.
 
@@ -76,6 +76,8 @@ class DefaultMethod(Operator):
             function under it and processes result so it is possible to create a valid
             query compiler from it.
             """
+            # pandas default implementation does not use argument
+            kwargs.pop("dtypes", None)
             df = cls.frame_wrapper(df)
             result = fn(df, *args, **kwargs)
 
--- a/modin/core/dataframe/algebra/default2pandas/default.py
+++ b/modin/core/dataframe/algebra/default2pandas/default.py
@@ -76,7 +76,7 @@ class DefaultMethod(Operator):
             function under it and processes result so it is possible to create a valid
             query compiler from it.
             """
-            # pandas default implementation does not use argument
+            # pandas default implementation doesn't know how to handle `dtypes` keyword argument
             kwargs.pop("dtypes", None)
             df = cls.frame_wrapper(df)
             result = fn(df, *args, **kwargs)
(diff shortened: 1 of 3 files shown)


modin/core/dataframe/algebra/default2pandas/default.py
```diff
@@ -68,7 +68,7 @@ def call(cls, func, obj_type=pandas.DataFrame, inplace=None, fn_name=None):
         if type(fn) == property:
             fn = cls.build_property_wrapper(fn)
 
-        def applyier(df, *args, **kwargs):
+        def applyier(df, *args, dtypes=None, **kwargs):
```
vnlitvinov, line 71: this is not used here, should we add an assert that it's always `None`?

vnlitvinov (commented):
LGTM with one comment!

modin/core/dataframe/algebra/default2pandas/default.py
anmyachev: There is a generic code that sets this argument even for the pure pandas case, so it's not always `None`.

RehanSD (commented):
I did some tracing, and it seems that the dtypes argument is largely ignored by these codepaths? It only seems to be used when broadcast is set to True for the algebra operator, but none of these functions (le, gt,...) set broadcast to True, and it defaults to False. Additionally, shouldn't we add dtypes as a named, necessary kwarg for base's binary_op as well?

anmyachev:
> I did some tracing, and it seems that the dtypes argument is largely ignored by these codepaths? It only seems to be used when broadcast is set to True for the algebra operator, but none of these functions (le, gt,...) set broadcast to True, and it defaults to False. Additionally, shouldn't we add dtypes as a named, necessary kwarg for base's binary_op as well?

Well, the changes now mostly cover cases where the right operand is not a Modin object, like just comparing with a scalar. https://github.com/modin-project/modin/blob/b0465fa4425247e7949881e93866c6b85bb10a1d/modin/core/dataframe/algebra/binary.py#L104

Yes, we should, I already opened the issue for that: https://github.com/modin-project/modin/issues/4856

anmyachev:
@RehanSD @vnlitvinov ready to merge if there is no concerns

modin/core/dataframe/algebra/default2pandas/default.py
YarShev: Is this dtypes used anywhere at all? Can you give an example?

anmyachev: It's used for example here: `df > 6` (code path will be through `map` function)

vnlitvinov:
> Well, the changes now mostly cover cases where the right operand is not a Modin object, like just comparing with a scalar.

This bit of the information seems very important to understand what's going on.
I would rather you put it at least to the opening message of this PR.

modin/core/dataframe/algebra/default2pandas/default.py
```diff
@@ -76,6 +76,8 @@ def applyier(df, *args, **kwargs):
             function under it and processes result so it is possible to create a valid
             query compiler from it.
             """
+            # pandas default implementation does not use argument
```
vnlitvinov, line 79:
```suggestion
            # pandas default implementation doesn't know how to handle `dtypes` keyword argument
```

YarShev (approved):

PR #4861, opened 2022-08-23
PERF-#4860: `deploy_axis_func` should be serialized only once

### LINKED ISSUE #4860: PERF: `PandasDataframeAxisPartition.deploy_axis_func` should be serialized only once
Place: https://github.com/modin-project/modin/blob/cfcdf081e197d35e1fd1526f5585e361793dc2fa/modin/core/execution/ray/implementations/pandas_on_ray/partitioning/virtual_partition.py#L191

Currently, this function is serialized on every call.

anmyachev:

## What do these changes do?

- [x] Resolves #4860 
- [x] tests added and passing

### DIFF
--- a/modin/core/execution/ray/implementations/pandas_on_ray/partitioning/virtual_partition.py
+++ b/modin/core/execution/ray/implementations/pandas_on_ray/partitioning/virtual_partition.py
@@ -24,7 +24,7 @@ from modin.core.execution.ray.common.utils import deserialize
 from .partition import PandasOnRayDataframePartition
 
 
-_DEPLOY_AXIS_FUNC = None
+_DEPLOY_AXIS_FUNC = ray.put(PandasDataframeAxisPartition.deploy_axis_func)
 
 
 class PandasOnRayDataframeVirtualPartition(PandasDataframeAxisPartition):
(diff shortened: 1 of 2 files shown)


RehanSD (commented):
LGTM - do we have any perf benches for this? One concern is that if we have multiple nodes there could be communication overhead since the function will be serialized and placed in one nodes object store, and the other nodes will have to request the function from that nodes object store every time.

modin/core/execution/ray/implementations/pandas_on_ray/partitioning/virtual_partition.py
```diff
@@ -24,6 +24,9 @@
 from .partition import PandasOnRayDataframePartition
 
 
+_DEPLOY_AXIS_FUNC = None
```
YarShev, line 27:
```suggestion
_DEPLOY_AXIS_FUNC = ray.put(PandasDataframeAxisPartition.deploy_axis_func)
```
maybe just so?

anmyachev:
> LGTM - do we have any perf benches for this? One concern is that if we have multiple nodes there could be communication overhead since the function will be serialized and placed in one nodes object store, and the other nodes will have to request the function from that nodes object store every time.

No, I see perf gain in my example. The machine that was used for the measurements has 112 cores. With 1000 calls to deploy_axis_func , I see a decrease in time of about 1 second. I don't test the function on several nodes, but doesn't Ray broadcast the function on all node (or at least cache it when first time called)?

YarShev (approved):

vnlitvinov:
> the other nodes will have to request the function from that nodes object store every time.

Other nodes should request it once and keep in their own object stores as long as there's free space.

That said, I think we should try to use this pattern more (and try to go away from functions defined inside other functions - this adds serialization overhead just as much as this one did).

PR #4874, opened 2022-08-24
 FIX-#4848: Fix rebalancing partitions when NPartitions == 1.

### LINKED ISSUE #4848: BUG: applying function to virtual partitions with NPartitions = 1 produces dataframe with 3 dimensional partitions
### System information
- **Modin version** (`modin.__version__`): 980f126dd941da534da10d6a5e919cf10410cae0
- **Python version**: 3.10.4

```python
import modin.pandas as pd

NPartitions.put(1)
df = pd.DataFrame([1])
df = pd.concat([df] * 100)
result = df.applymap(lambda x: x)
print(result._query_compiler._modin_frame._partitions.shape)
```

### Describe the problem

`df._query_compiler._modin_frame._partitions[0,0].apply(lambda x: x)` returns a list of 100 partitions instead of a single partition. Probably there is something wrong with how the virtual partition class is using the parent axis partition class's `apply`.

mvashishtha:
## What do these changes do?

 FIX-#4848: Fix rebalancing partitions when NPartitions == 1. Each partition after rebalancing partitions should always have `full_axis=False`.

- [x] Resolves #4848
- [x] tests added and passing

### DIFF
--- a/modin/core/dataframe/pandas/partitioning/partition_manager.py
+++ b/modin/core/dataframe/pandas/partitioning/partition_manager.py
@@ -1422,6 +1422,12 @@ class PandasDataframePartitionManager(ClassLogger, ABC):
                         for obj in partitions[stop]
                     ]
                     partition_size = ideal_partition_size
+                # The new virtual partitions are not `full_axis`, even if they
+                # happen to span all rows in the dataframe, because they are
+                # meant to be the final partitions of the dataframe. They've
+                # already been split up correctly along axis 0, but using the
+                # default full_axis=True would cause partition.apply() to split
+                # its result along axis 0.
                 new_partitions.append(
                     cls.column_partitions(
                         (partitions[start : stop + 1]), full_axis=False
--- a/modin/core/dataframe/pandas/partitioning/partition_manager.py
+++ b/modin/core/dataframe/pandas/partitioning/partition_manager.py
@@ -1429,9 +1429,7 @@ class PandasDataframePartitionManager(ClassLogger, ABC):
                 # default full_axis=True would cause partition.apply() to split
                 # its result along axis 0.
                 new_partitions.append(
-                    cls.column_partitions(
-                        partitions[start : stop + 1], full_axis=False
-                    )
+                    cls.column_partitions(partitions[start : stop + 1], full_axis=False)
                 )
                 start = stop + 1
             return np.array(new_partitions)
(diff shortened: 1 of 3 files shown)


modin/core/dataframe/pandas/partitioning/partition_manager.py
```diff
@@ -1424,8 +1424,7 @@ def rebalance_partitions(cls, partitions):
                     partition_size = ideal_partition_size
                 new_partitions.append(
                     cls.column_partitions(
-                        (partitions[start : stop + 1]),
-                        full_axis=partition_size == total_rows,
+                        (partitions[start : stop + 1]), full_axis=False
```
RehanSD, line 1427: Can we add a comment that we're setting it to False on purpose since setting it to True will cause the partition to be split again after the next apply, which doesn't make sense, since we just concatenated it?

RehanSD (changes requested):
Left a quick comment about documentation. Besides that, LGTM!

modin/core/dataframe/pandas/partitioning/partition_manager.py
mvashishtha: I added a comment that makes more sense to me.

RehanSD (approved):
LGTM!

modin/core/dataframe/pandas/partitioning/partition_manager.py
```diff
@@ -1422,10 +1422,15 @@ def rebalance_partitions(cls, partitions):
...
+                # happen to span all rows in the dataframe, because they are
+                # meant to be the final partitions of the dataframe. They've
+                # already been split up correctly along axis 0, but using the
+                # default full_axis=True would cause partition.apply() to split
+                # its result along axis 0.
                 new_partitions.append(
                     cls.column_partitions(
-                        (partitions[start : stop + 1]),
-                        full_axis=partition_size == total_rows,
+                        (partitions[start : stop + 1]), full_axis=False
```
vnlitvinov, line 1433:
```suggestion
                        partitions[start : stop + 1], full_axis=False
```
I don't see a point in these parenthesis

vnlitvinov (approved):
LGTM!

modin/core/dataframe/pandas/partitioning/partition_manager.py
```diff
@@ -1422,11 +1422,14 @@ def rebalance_partitions(cls, partitions):
                         for obj in partitions[stop]
                     ]
                     partition_size = ideal_partition_size
+                # The new virtual partitions are not `full_axis`, even if they
+                # happen to span all rows in the dataframe, because they are
+                # meant to be the final partitions of the dataframe. They've
+                # already been split up correctly along axis 0, but using the
+                # default full_axis=True would cause partition.apply() to split
```
RehanSD, line 1429:
```suggestion
                # allowing full_axis=True would cause partition.apply() to split
```

RehanSD (commented):
Left a quick comment on docs but LGTM!

RehanSD (approved):
LGTM!

YarShev (approved):

PR #4880, opened 2022-08-24
PERF-#4882: avoid reorder_labels

### LINKED ISSUE #4882: PERF: avoid reorder_labels in take_2d_label_or_positional
When profiling getitem-like calls, most of the time in take_2d_label_or_positional ends up being taken by reorder_labels.  AFAICT the current implementation sorts the indices, does a take with these, then re-sorts the result.  This should be doable with a single step.

jbrockmendel:
- [x] closes #4882 

```
import numpy as np
import modin.pandas as pd

arr = np.random.randn(1_000_000, 20)
df = pd.DataFrame(arr)

res = df[[4, 3]]

%timeit res = df[[4, 3]]
711 µs ± 4.12 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)  # <- PR
43.6 ms ± 1.81 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)  # <- master
```

TODO
- [ ] further optimizations to blklocs/blknos
- [ ] avoid reorder_labels in cudf_on_ray implemenation
- [ ] fix is_full_axis_mask
- [ ] deal with potential many-partitions issues (discussed briefly on call 8/23)


jbrockmendel:
Looks like this runs into a many-partitions issue.  @devin-petersohn on the call the other day it sounded like you had an idea to deal with this?

RehanSD:
> Looks like this runs into a many-partitions issue.  @devin-petersohn on the call the other day it sounded like you had an idea to deal with this?

Hi @jbrockmendel! By many-partitions issue, do you mean that too many partitions are created? If that's the issue, you could call the `rebalance_partitions` method of the partition manager on the partitions, which should balance them!

jbrockmendel:
@RehanSD i think we need to re-find new_row_lengths after rebalance_partitions.  suggestions on efficient way to do this?

mvashishtha:
> i think we need to re-find new_row_lengths after rebalance_partitions. suggestions on efficient way to do this?

@jbrockmendel it looks like draft PR #4893 is going to make `rebalance_partitions` return the new row lengths without doing any extra remote work. Will that work?

jbrockmendel:
sounds like it, yes

jbrockmendel:
@anmyachev tried to incorporate the rebalance_partition changes but it broke a bunch of new stuff.  suggestions?

anmyachev:
> @anmyachev tried to incorporate the rebalance_partition changes but it broke a bunch of new stuff. suggestions?

Perhaps https://github.com/modin-project/modin/issues/4914 will result in an issue with the insert operation.

anmyachev:
@jbrockmendel please rebase on master

modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -681,70 +683,66 @@ def take_2d_labels_or_positional(
         ):
             return self.copy()
 
+        import pandas._libs.internals as libinternals
+
         sorted_row_positions = sorted_col_positions = None
```
anmyachev, line 688: To pass flake8 checks.
```suggestion
```

modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -799,48 +746,70 @@ def _take_2d_positional(
...
-            sorted_row_positions = self._get_sorted_positions(row_positions)
-            # Get dict of row_parts as {row_index: row_internal_indices}
-            row_partitions_dict = self._get_dict_of_block_index(
-                0, sorted_row_positions, are_indices_sorted=True
-            )
-            new_row_lengths = self._get_new_lengths(row_partitions_dict, axis=0)
-            new_index, _ = self._get_new_index_obj(
-                row_positions, sorted_row_positions, axis=0
-            )
+            row_positions = np.asarray(row_positions, dtype=np.intp)
```
anmyachev, line 752: @jbrockmendel one of the current errors in CI is the loss of metadata, in particular `freq` field. This is because you are converting Range object to numpy array in this line.

modin/core/dataframe/pandas/partitioning/partition.py
```diff
@@ -187,27 +186,31 @@ def mask(self, row_labels, col_labels):
...
-                )
-            return (
-                hasattr(index, "__len__")
-                and isinstance(axis_length, int)
-                and len(index) == axis_length
-            )
+        # FIXME: for a `take` won't is_full_axis_mask be wrong if row_labels
+        #  is non-unique or not monotonic increasing.
+
+        # def is_full_axis_mask(index, axis_length):
```
anmyachev, line 192: It's not very clear to me why it's commented out. Can you add details or rephrase?

jbrockmendel:
I don't see a path here, closing.

PR #4883, opened 2022-08-24
REF: de-duplicate take_2d_labels_or_positional methods

jbrockmendel:
There's a ton in the cuDFOnRayDataframe method that looks like it should be the same as the parent method but just was never updated.  If I'm right about this, this is a textbook reason why the code should be shared.

Not bothering with opening a dummy issue or formatting the commit because I'm a cranky old man.


mvashishtha (approved):
LGTM @jbrockmendel , thanks for the cleanup.

anmyachev (approved):

PR #4898, opened 2022-08-29
FEAT-#4766: Support fsspec URLs in `read_csv` and `read_csv_glob`

### LINKED ISSUE #4766: FEAT: Add support for Distributed loading on GCS buckets
## Context
The company I work for, heavily uses pandas and Google Cloud Storage (GCS) buckets. We were interested in using Modin to improve performance of loading large datasets and improve data processing tasks. 

## Problem
However, it seems as though using GCS bucket urls for the `read_csv()` function of modin, is quite slow. The GCS bucket file patterns are `gs://`. Loading from local memory or S3 buckets works quite well, but the lack of support for GCS buckets results in Modin to default to pandas, which ends up being slower than using the vanilla pandas library!

## Feature Request
We are requesting support for GCS buckets for distributed loading so that we aren't limited to S3 or local file storage only especially since the company I work for, uses Google Suite products quite a lot and does not use Amazon offerings.

pyrito:

## What do these changes do?

This PR adds proper support for fsspec URLs that matches pandas behavior for `read_csv`. 

- [x] Resolves #4766 
- [x] tests added and passing

### DIFF
--- a/modin/core/io/text/csv_glob_dispatcher.py
+++ b/modin/core/io/text/csv_glob_dispatcher.py
@@ -24,13 +24,11 @@ import fsspec
 
 import pandas
 import pandas._libs.lib as lib
-from pandas.io.common import is_url
+from pandas.io.common import is_url, is_fsspec_url, stringify_path
 
 from modin.config import NPartitions
 from modin.core.io.file_dispatcher import OpenFile
 from modin.core.io.text.csv_dispatcher import CSVDispatcher
-
-_SUPPORTED_PROTOCOLS = {"s3", "S3"}
 
 
 class CSVGlobDispatcher(CSVDispatcher):
@@ -54,7 +52,7 @@ class CSVGlobDispatcher(CSVDispatcher):
             Query compiler with imported data for further processing.
         """
         # Ensures that the file is a string file path. Otherwise, default to pandas.
-        filepath_or_buffer = cls.get_path_or_buffer(filepath_or_buffer)
+        filepath_or_buffer = cls.get_path_or_buffer(stringify_path(filepath_or_buffer))
         if isinstance(filepath_or_buffer, str):
             # os.altsep == None on Linux
             is_folder = any(
@@ -331,17 +329,8 @@ class CSVGlobDispatcher(CSVDispatcher):
         bool
             True if the path is valid.
         """
-        if isinstance(file_path, str) and is_url(file_path):
-            raise NotImplementedError("`read_csv_glob` supports only s3-like paths.")
-
-        if (
-            not isinstance(file_path, str)
-            or fsspec.core.split_protocol(file_path)[0] not in _SUPPORTED_PROTOCOLS
-        ):
+        if not is_fsspec_url(file_path) and not is_url(file_path):
             return len(glob.glob(file_path)) > 0
-
-        # `file_path` may start with a capital letter, which isn't supported by `fsspec.core.url_to_fs` used below.
-        file_path = file_path[0].lower() + file_path[1:]
 
         from botocore.exceptions import (
             NoCredentialsError,
@@ -384,13 +373,10 @@ class CSVGlobDispatcher(CSVDispatcher):
         list
             List of strings of absolute file paths.
         """
-        if fsspec.core.split_protocol(file_path)[0] not in _SUPPORTED_PROTOCOLS:
+        if not is_fsspec_url(file_path) and not is_url(file_path):
             relative_paths = glob.glob(file_path)
             abs_paths = [os.path.abspath(path) for path in relative_paths]
             return abs_paths
-
-        # `file_path` may start with a capital letter, which isn't supported by `fsspec.core.url_to_fs` used below.
-        file_path = file_path[0].lower() + file_path[1:]
 
         from botocore.exceptions import (
             NoCredentialsError,
@@ -402,8 +388,8 @@ class CSVGlobDispatcher(CSVDispatcher):
             file_paths = fs_handle.glob(file_path)
             if len(file_paths) == 0 and not fs_handle.exists(file_path):
                 raise FileNotFoundError(f"Path <{file_path}> isn't available.")
-            s3_addresses = [f"s3://{path}" for path in file_paths]
-            return s3_addresses
+            fs_addresses = [fs_handle.unstrip_protocol(path) for path in file_paths]
+            return fs_addresses
 
         fs, _ = fsspec.core.url_to_fs(file_path)
         try:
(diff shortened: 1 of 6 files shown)


modin/core/io/text/csv_glob_dispatcher.py
```diff
@@ -331,18 +329,9 @@ def file_exists(cls, file_path: str, storage_options=None) -> bool:
         bool
             True if the path is valid.
         """
-        if isinstance(file_path, str) and is_url(file_path):
```
RehanSD, line 334: What if someone passes in a GitHub raw user content file - with an https irl prefix - would that work?

modin/experimental/pandas/test/test_io_exp.py
```diff
@@ -168,7 +168,7 @@ def test_read_single_csv_with_parse_dates(self, parse_dates):
     Engine.get() != "Ray", reason="Currently only support Ray engine for glob paths."
 )
 def test_read_multiple_csv_s3():
-    path = "S3://modin-datasets/testing/multiple_csv/test_data*.csv"
+    path = "s3://modin-datasets/testing/multiple_csv/test_data*.csv"
```
RehanSD, line 171: Why was this changed?

modin/pandas/test/test_io.py
```diff
@@ -915,6 +915,13 @@ def test_read_csv_categories(self):
             dtype={"one": "int64", "two": "category"},
         )
```
RehanSD, line 917: Shouldn't we add a test for glob as well?

RehanSD (commented):
Left some questions - great work!

modin/pandas/test/test_io.py
pyrito: We already have some glob tests for S3 that should test the new code paths but I can add one for gs as well.

modin/experimental/pandas/test/test_io_exp.py
pyrito: It is simpler for us to not deal with malformed protocols in the paths and this is what `fsspec` decided to do.

modin/core/io/text/csv_glob_dispatcher.py
pyrito: Yeah actually I think it shouldn't work for the glob case. I remember Alexey bringing this up, so I'll revert this back.

modin/experimental/pandas/test/test_io_exp.py
pyrito: For more context: https://github.com/fsspec/filesystem_spec/issues/1009

modin/pandas/test/test_io.py
```diff
@@ -915,6 +915,13 @@ def test_read_csv_categories(self):
...
 
+    def test_read_csv_gs(self):
```
mvashishtha, line 918:
```suggestion
    def test_read_csv_google_cloud_storage(self):
```

modin/experimental/pandas/test/test_io_exp.py
```diff
@@ -167,8 +167,11 @@ def test_read_single_csv_with_parse_dates(self, parse_dates):
 @pytest.mark.skipif(
     Engine.get() != "Ray", reason="Currently only support Ray engine for glob paths."
 )
-def test_read_multiple_csv_s3():
-    path = "S3://modin-datasets/testing/multiple_csv/test_data*.csv"
+def test_read_multiple_csv_cloud_store():
+    paths = [
```
mvashishtha, line 171: Can make these test parameters?

mvashishtha (commented):
Overall LGTM. I have two minor comments.

mvashishtha (dismissed):
LGTM!

modin/core/io/text/text_file_dispatcher.py
```diff
@@ -963,6 +964,7 @@ def _read(cls, filepath_or_buffer, **kwargs):
         new_query_compiler : BaseQueryCompiler
             Query compiler with imported data for further processing.
         """
+        filepath_or_buffer = stringify_path(filepath_or_buffer)
```
anmyachev, line 967: Why is this needed?

pyrito: @anmyachev it's a nice utility function used by pandas. Sometimes users pass in Path-like objects and we need to attempt to stringify the path. If it ends up being just a buffer, this method will handle that for us as well.

anmyachev: Doesn't this method do part of the job as our functions `get_path` and `get_path_or_buffer`? In this case, we could refuse to use our functions.

pyrito: Not for get_path since that function expects a string to be passed in. Unsure about the latter

anmyachev: It seems to me the code before worked with Path-like types without `stringify_path`, didn't it?

pyrito: Yes but now that I think about it, it might be better to use stringify_path in place of both get_path and get_path_or_buffer.

pyrito: @anmyachev reading the warning in `get_path_or_buffer` makes me wonder if we should leave our current implementation as is?

anmyachev: Yes, `stringify_path` cannot replace `get_path_or_buffer`.

anmyachev: We can try to replace `get_path` with `stringify_path` in a separate pull request if needed.

pyrito: @anmyachev yes, I reverted things back for now. Please take a look and approve if it looks good!

modin/core/io/text/csv_glob_dispatcher.py
```diff
@@ -54,7 +52,7 @@ def _read(cls, filepath_or_buffer, **kwargs):
             Query compiler with imported data for further processing.
         """
         # Ensures that the file is a string file path. Otherwise, default to pandas.
-        filepath_or_buffer = cls.get_path_or_buffer(filepath_or_buffer)
+        filepath_or_buffer = cls.get_path_or_buffer(stringify_path(filepath_or_buffer))
```
anmyachev, line 55: `stringify_path` is needed here because `get_path` isn't called?

pyrito: `get_path` in the glob file reader returns a list of files, so `stringify_path` makes sense here.

mvashishtha (approved):

anmyachev (approved):

PR #4910, opened 2022-08-31
FIX-#4022: Fixed empty data frame with index

### LINKED ISSUE #4022: ``dataframe.loc()`` doesn`t work correctly at MODIN_STORAGE_FORMAT=omnisci
When I run:
```
import modin.pandas as pd
import modin.config as cfg

modin_df = pd.DataFrame(index=range(5))
modin_df.loc[1]
```
I have error:
```
KeyError: 1
```

AndreyPavlenko:
## What do these changes do?
Put index into columns if there are no columns but only index.

- [x] Resolves #4022 
- [x] tests added and passing


modin/pandas/indexing.py
```diff
@@ -653,7 +653,7 @@ def __getitem__(self, key):
         --------
         pandas.DataFrame.loc
         """
-        if self.df.empty:
+        if len(self.df.index) == 0 and self.df.empty:
```
ienkovich, line 656: Why do you need these changes?

AndreyPavlenko: If the frame has only index, the following lambda fails with KeyError.

ienkovich: Is the problem specific to OmniSci backend or to all backends? Probably, the problem is in backend implementation of `empty`?

AndreyPavlenko: Actually, I could find only one implementation of this method - https://github.com/modin-project/modin/blob/master/modin/pandas/dataframe.py#L325

ienkovich: I don't understand why making a condition for default to pandas more strict might resolve any issue. We should be able to always be able to default to pandas.

AndreyPavlenko: You are right. Actually, the problem is not in default_to_pandas() but in loc[], which is called from lambda. Pandas returns an empty Series for df.loc[1], but modin with omnisci backend fails with KeyError: 'Column F_1 does not exist in schema'. This issue should be fixed instead.

ienkovich (dismissed):
LGTM!

YarShev (approved):

PR #4911, opened 2022-08-31
PERF-#4870: Avoid index materialization in `__getattribute__` and `__getitem__`

### LINKED ISSUE #4870: PERF: try to avoid index materialization in `BasePandasDataset.__getattribute__`
Place: https://github.com/modin-project/modin/blob/cfcdf081e197d35e1fd1526f5585e361793dc2fa/modin/pandas/base.py#L3415

For optimization purpose we use `self.empty` check that trigger index materialization. It is most likely possible to do this check only when the index is already materialized. To do this, we can check the corresponding cache.

anmyachev:

## What do these changes do?

- [x] Resolves #4870 
- [x] tests added and passing

### DIFF
--- a/modin/core/storage_formats/base/query_compiler.py
+++ b/modin/core/storage_formats/base/query_compiler.py
@@ -98,6 +98,13 @@ class BaseQueryCompiler(ClassLogger, abc.ABC):
     This class defines common query compilers API, most of the methods
     are already implemented and defaulting to pandas.
 
+    Attributes
+    ----------
+    lazy_execution : bool
+        Whether underlying execution engine is designed to be executed in a lazy mode only.
+        If True, such QueryCompiler will be handled differently at the front-end in order
+        to reduce execution triggering as much as possible.
+
     Notes
     -----
     See the Abstract Methods and Fields section immediately below this
@@ -130,20 +137,7 @@ class BaseQueryCompiler(ClassLogger, abc.ABC):
     # some of these abstract methods, but for the sake of generality they are
     # treated differently.
 
-    @property
-    def lazy_execution(self):
-        """
-        Whether underlying execution engine is designed to be executed in a lazy mode only.
-
-        If True, such QueryCompiler will be handled differently at the front-end in order
-        to reduce execution triggering as much as possible.
-
-        Returns
-        -------
-        bool
-        """
-        frame = self._modin_frame
-        return frame._index_cache is None or frame._columns_cache is None
+    lazy_execution = False
 
     # Metadata modification abstract methods
     def add_prefix(self, prefix, axis=1):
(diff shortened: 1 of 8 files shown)


anmyachev:
@vnlitvinov could you review this PR?

modin/core/storage_formats/base/query_compiler.py
```diff
@@ -137,7 +130,20 @@ def default_to_pandas(self, pandas_op, *args, **kwargs):
...
     # treated differently.
 
-    lazy_execution = False
+    @property
+    def lazy_execution(self):
+        """
+        Whether underlying execution engine is designed to be executed in a lazy mode only.
+
+        If True, such QueryCompiler will be handled differently at the front-end in order
+        to reduce execution triggering as much as possible.
```
vnlitvinov, line 139: This docstring is no longer true compared to what we're seeing in the implementation; plus I would rather see this change in PandasQueryCompiler, not BaseQueryCompiler (and leave this as a plain attribute in other places).

vnlitvinov (changes requested):

modin/core/storage_formats/pandas/query_compiler.py
```diff
@@ -255,6 +255,21 @@ def default_to_pandas(self, pandas_op, *args, **kwargs):
         else:
             return result
 
+    @property
+    def lazy_execution(self):
+        """
+        Whether underlying execution engine is executed in a lazy mode.
+
+        If True, such QueryCompiler will be handled differently at the front-end in order
+        to reduce execution triggering as much as possible.
```
vnlitvinov, line 264:
```suggestion
        Whether underlying Modin frame should be executed in a lazy mode.

        If True, such QueryCompiler will be handled differently at the front-end in order
        to reduce triggering the computation as much as possible.
```

vnlitvinov (commented):
Good general idea, I'd like to fix the wording tho.

anmyachev:
@vnlitvinov let's merge the PR

vnlitvinov (approved):

anmyachev:
@mvashishtha could you review?

mvashishtha:
@anmyachev it looks like the changes to `modin/pandas/base.py` are gone? I just see the PR adding `lazy_execution` but not using it anywhere.

anmyachev:
> @anmyachev it looks like the changes to `modin/pandas/base.py` are gone? I just see the PR adding `lazy_execution` but not using it anywhere.

It's already used [here](https://github.com/modin-project/modin/blob/5991580cc8ac79c038d6dcde8b02d12feda06109/modin/pandas/base.py#L3174) and [here](https://github.com/modin-project/modin/blob/5991580cc8ac79c038d6dcde8b02d12feda06109/modin/pandas/base.py#L3419).

I just update `lazy_execution` implementation itself.

mvashishtha (approved):

Issue #4845, opened 2022-08-19
BUG: KeyError occuring in `__getitem_bool` if dataframe has only one row


pyrito:
### System information
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: macOS 12.2.1
- **Modin version** (`modin.__version__`): 0.15.2
- **Python version**: 3.9.12
- **Code we can use to reproduce**:
```python
import os
import regex as re
import modin.pandas as pd

# Where server.log has simple arbitrary strings
fp = "server.log"
data_lines = []

with open(fp, "rt", encoding="utf8") as file:
    lines = file.readlines()
    # data_lines += [
    #     line for line in lines
    #     if "POST" in line
    # ]
    data_lines += lines

# Processing
df = pd.DataFrame({"src": data_lines})
df.src = df.src.astype("string")

# This appears to fail
df = df[df.src.str.contains("POST")]
print(df)
```

### Describe the problem
I got this error while trying to reproduce: https://stackoverflow.com/questions/73395456/string-methods-fail-with-modin-but-same-work-with-pandas, which, funnily enough, is facing a different error.

### Source code / logs
<details>
<summary>Error logs</summary>

```python
File ~/opt/anaconda3/envs/modin/lib/python3.9/site-packages/IPython/core/formatters.py:707, in PlainTextFormatter.__call__(self, obj)
    700 stream = StringIO()
    701 printer = pretty.RepresentationPrinter(stream, self.verbose,
    702     self.max_width, self.newline,
    703     max_seq_length=self.max_seq_length,
    704     singleton_pprinters=self.singleton_printers,
    705     type_pprinters=self.type_printers,
    706     deferred_pprinters=self.deferred_printers)
--> 707 printer.pretty(obj)
    708 printer.flush()
    709 return stream.getvalue()

File ~/opt/anaconda3/envs/modin/lib/python3.9/site-packages/IPython/lib/pretty.py:410, in RepresentationPrinter.pretty(self, obj)
    407                         return meth(obj, self, cycle)
    408                 if cls is not object \
    409                         and callable(cls.__dict__.get('__repr__')):
--> 410                     return _repr_pprint(obj, self, cycle)
    412     return _default_pprint(obj, self, cycle)
    413 finally:

File ~/opt/anaconda3/envs/modin/lib/python3.9/site-packages/IPython/lib/pretty.py:778, in _repr_pprint(obj, p, cycle)
    776 """A pprint that just redirects to the normal repr function."""
[...]

PR #4917, opened 2022-09-01
FIX-#4845: Fixed KeyError from `__getitem_bool` for single row dataframes

Billy2551:

## What do these changes do?

Adds check in `__getitem_bool` for 1 dimensional objects and corresponding test. Resolves KeyError issue with trying to index a dataframe at a boolean value `df[True]`.

- [ ] Resolves #? 
- [ ] tests added and passing


modin/core/storage_formats/pandas/query_compiler.py
```diff
@@ -2152,7 +2152,7 @@ def applyier(df, internal_indices, other=[], internal_other_indices=[]):
 
     # __getitem__ methods
     __getitem_bool = Binary.register(
-        lambda df, r: df[r], join_type="left", labels="drop"
+        lambda df, r: df[r] if r.size > 1 else df[[r]], join_type="left", labels="drop"
```
pyrito, line 2155: Just for clarification, but basically what was happening is that we had `r` being a list of bools: `[False, True, False, ...]`? And if we had only one item then the array would get squeezed to a scalar?

modin/pandas/test/dataframe/test_indexing.py
```diff
@@ -2111,3 +2111,17 @@ def call_from_frame(df):
...
+def test__getitem_bool_single_row_dataframe():
+    # This test case comes from
+    # https://github.com/modin-project/modin/issues/4845
+    df_modin = pd.DataFrame({"src": ["foo"]})
+    df_modin.src = df_modin.src.astype("string")
+    df_modin = df_modin[df_modin.src.str.contains("POST")]
+
+    df_pandas = pandas.DataFrame({"src": ["foo"]})
+    df_pandas.src = df_pandas.src.astype("string")
+    df_pandas = df_pandas[df_pandas.src.str.contains("POST")]
```
pyrito, line 2125: You could try rewriting some parts of this test to be a little cleaner:
```python
modin_df, pandas_df = create_test_dfs({"src" : ["foo"]})

def func(df):
    df.src = df.src.astype("string")
    df = df[df.src.str.contains("test")]
    return df

eval_general(modin_df, pandas_df, func)
```

pyrito (commented):
Could you add release notes for this PR please?

modin/pandas/test/dataframe/test_indexing.py
pyrito: Could you also check that this test fails before your change? It may be helpful to add a short comment here as well to explain a little on why we are checking things this way.

modin/pandas/test/dataframe/test_indexing.py
```diff
@@ -2111,3 +2111,17 @@ def call_from_frame(df):
             return pd.MultiIndex.from_frame(df, sortorder)
 
     eval_general(modin_df, pandas_df, call_from_frame, comparator=assert_index_equal)
+
+
+def test__getitem_bool_single_row_dataframe():
```
mvashishtha, line 2116: I think this test came from the original example in #4845. It will be clearer if we can trim the reproducer down as much as possible. I don't think we need the `contains` or `astype`. Can we do something like 

```python
df = pd.DataFrame([1])
result = df[pd.Series(True)]
```

and compare to doing the same for pandas? We should also check that the smaller test case still reproduces the bug.

pyrito: I was just thinking about how we could trim down the test. I think your suggestion is better.

modin/pandas/test/dataframe/test_indexing.py
pyrito: Ignore my suggestion, please go with: https://github.com/modin-project/modin/pull/4917#discussion_r961015430

modin/pandas/test/dataframe/test_indexing.py
pyrito: We could probably do:
```python
modin_df, pandas_df = create_test_dfs([1])
eval_general(modin_df, pandas_df, lambda df: df[pd.Series(True)])
```

Billy2551: I'm not sure why, but when I run this reproduce in ipython, it reproduces the same error. However, when I write it as a test, the test always passes.

When I use the original reproducer, the test passes/fails as expected.

modin/core/storage_formats/pandas/query_compiler.py
vnlitvinov: this would make a nice in-line comment, pls add

vnlitvinov: this would make a nice in-line comment, pls add

modin/pandas/test/dataframe/test_indexing.py
vnlitvinov: You should not be running `eval_general()` like that, as pandas won't accept Modin Series (yours `pd.Series()`) here.

vnlitvinov (commented):
@Billy2551 nice debugging and fixing!

Please format the PR per template and address the comments, overall looks nice.

modin/pandas/test/dataframe/test_indexing.py
vnlitvinov: I wonder if we can do smth like 
```python
eval_general(pd, pandas, lambda lib: lib.DataFrame([1])[lib.Series([True])])
```

pyrito:
@Billy2551 there might be a merge conflict, so I would pull from upstream and rebase your changes.

pyrito (approved):
LGTM!

vnlitvinov (approved):
LGTM, thanks for clearing this up!

Issue #4907, opened 2022-08-31
BUG: Appending string to column returns in different order


pyrito:
### System information
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: macOS 12.2.1
- **Modin version** (`modin.__version__`): 0.15.0+101.g14edd1ca
- **Python version**: 3.9.12
- **Code we can use to reproduce**:
```python
import modin.pandas as pd
import pandas
from modin.pandas.test.utils import df_equals

pdf = pandas.DataFrame({'a' : ['a','b','c','d']})
mdf = pd.DataFrame({'a' : ['a','b','c','d']})

pdf['a'] = 'random_string' + pdf['a']
mdf['a'] = 'random_string' + mdf['a']

df_equals(pdf, mdf)
```
### Describe the problem

Appending a string to a `str` column in a Modin DataFrame returns values in a different order.

### Source code / logs

<details>
<summary> Error logs </summary>

```python
---------------------------------------------------------------------------
AssertionError                            Traceback (most recent call last)
Input In [18], in <cell line: 11>()
      8 pdf['a'] = 'random_string' + pdf['a']
      9 mdf['a'] = 'random_string' + mdf['a']
---> 11 df_equals(pdf, mdf)

File ~/Documents/modin/modin/pandas/test/utils.py:589, in df_equals(df1, df2)
    586     assert_empty_frame_equal(df1, df2)
    588 if isinstance(df1, pandas.DataFrame) and isinstance(df2, pandas.DataFrame):
--> 589     assert_frame_equal(
    590         df1,
    591         df2,
    592         check_dtype=False,
    593         check_datetimelike_compat=True,
    594         check_index_type=False,
    595         check_column_type=False,
    596         check_categorical=False,
    597     )
    598     df_categories_equals(df1, df2)
    599 elif isinstance(df1, pandas.Index) and isinstance(df2, pandas.Index):

    [... skipping hidden 2 frame]

File ~/opt/anaconda3/envs/modin/lib/python3.9/site-packages/pandas/_libs/testing.pyx:52, in pandas._libs.testing.assert_almost_equal()

File ~/opt/anaconda3/envs/modin/lib/python3.9/site-packages/pandas/_libs/testing.pyx:167, in pandas._libs.testing.assert_almost_equal()

File ~/opt/anaconda3/envs/modin/lib/python3.9/site-packages/pandas/_testing/asserters.py:682, in raise_assert_detail(obj, message, left, right, diff, index_values)
[...]

PR #4932, opened 2022-09-06
FEAT-#4931: Create a query compiler that can connect to a service

### LINKED ISSUE #4931: Create a Query Compiler that can connect to a service

devin-petersohn:

## What do these changes do?

- [x] Resolves #4931 
- [ ] tests added and passing


vnlitvinov:
@devin-petersohn CI is red

modin/core/execution/client/io.py
```diff
@@ -0,0 +1,120 @@
...
+        -------
+        ClientQueryCompiler
+            Query compiler with CSV data read in.
+        """
+        if isinstance(filepath_or_buffer, str):
+            filepath_or_buffer = fsspec.open(filepath_or_buffer).full_name
+            if filepath_or_buffer.startswith("file://"):
+                # We will do this so that the backend can know whether this
+                # is a path or a URL.
+                filepath_or_buffer = filepath_or_buffer[7:]
```
pyrito, line 74: You can save `fsspec.open(filepath_or_buffer)` and call `path` to get the location within the schema. That is probably the cleaner solution here.

pyrito: Instead of dealing with this here, why not have the server handle `fsspec` paths?

modin/core/execution/client/query_compiler.py
```diff
@@ -0,0 +1,794 @@
...
+
+    def mul(self, other, **kwargs):
+        if isinstance(other, ClientQueryCompiler):
+            other = other._id
+            is_qc = True
+        else:
+            is_qc = False
+        return self.__constructor__(self._service.mul(self._id, other, is_qc, **kwargs))
+
+    def rmul(self, other, **kwargs):
```
mvashishtha, line 441: base query compiler doesn't implement `rmul`. I think this hasn't been an issue because I don't know of any types that implement non-commutative `*`. So we wouldn't get bugs like #4907. Do we need `rmul`?

modin/config/envvars.py
```diff
@@ -131,17 +131,18 @@ def _get_default(cls) -> str:
             pass
         else:
             return "Native"
-        raise ImportError(
-            "Please refer to installation documentation page to install an engine"
-        )
+
+        # If we can't import any other engines we should go ahead and default to Python being
+        # the default backend engine.
+        return "Python"
```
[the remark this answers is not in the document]
pyrito: When we set the Engine, we default to the default implementation of `put` which tries to get the default value for the Engine. Without this change an ImportError would get thrown since our test environments wouldn't have Dask or Ray installed.

modin/core/execution/client/query_compiler.py
```diff
@@ -0,0 +1,794 @@
...
+        else:
+            is_qc = False
+        return self.__constructor__(self._service.getitem_array(self._id, key, is_qc))
+
+    def getitem_column_array(self, key, numeric=False):
+        return self.__constructor__(
+            self._service.getitem_column_array(self._id, key, numeric)
+        )
+
+    def getitem_row_labels_array(self, labels):
```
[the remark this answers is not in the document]
mvashishtha: I added a `numeric` param to `getitem_row_array`

modin/core/execution/client/query_compiler.py
```diff
@@ -0,0 +1,794 @@
...
+        )
+
+    def setitem(self, axis, key, value):
+        if isinstance(value, ClientQueryCompiler):
+            value = value._id
+            is_qc = True
+        else:
+            is_qc = False
+        return self.__constructor__(
+            self._service.setitem(self._id, axis, key, value, is_qc)
```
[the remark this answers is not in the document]
mvashishtha: I changed my mind-- just one method is ok.

modin/core/execution/client/query_compiler.py
```diff
@@ -0,0 +1,794 @@
...
+        return self.__constructor__(
+            self._service.pivot(self._id, index, columns, values)
+        )
+
+    def get_dummies(self, columns, **kwargs):
+        return self.__constructor__(
+            self._service.get_dummies(self._id, columns, **kwargs)
+        )
+
+    def view(self, index=None, columns=None):
```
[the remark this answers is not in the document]
mvashishtha: Nope, just take_2d.

modin/core/execution/client/query_compiler.py
```diff
@@ -0,0 +1,794 @@
...
+    def concat(self, axis, other, **kwargs):
+        if is_list_like(other):
+            other = [o._id for o in other]
+        else:
+            other = [other._id]
+        return self.__constructor__(
+            self._service.concat(self._id, axis, other, **kwargs)
+        )
+
+    def eq(self, other, **kwargs):
```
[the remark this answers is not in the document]
mvashishtha: Done.

modin/core/execution/client/query_compiler.py
mvashishtha: Implemented `rmul`.

modin/core/execution/client/query_compiler.py
```diff
@@ -0,0 +1,794 @@
...
+    def idxmin(self, **kwargs):
+        return self.__constructor__(self._service.idxmin(self._id, **kwargs))
+
+    def idxmax(self, **kwargs):
+        return self.__constructor__(self._service.idxmax(self._id, **kwargs))
+
+    def query(self, expr, **kwargs):
+        is_variable = False
+        variable_list = []
+        for k, v in tokenize_string(expr):
```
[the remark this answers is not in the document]
mvashishtha: I think it does-- and it happens at the API layer. I removed this code.

modin/core/execution/client/io.py
```diff
@@ -0,0 +1,123 @@
...
+        **kwargs : dict
+            Parameters of ``read_csv`` function.
+
+        Returns
+        -------
+        ClientQueryCompiler
+            Query compiler with CSV data read in.
+        """
+        if isinstance(filepath_or_buffer, str):
+            filepath_or_buffer = fsspec.open(filepath_or_buffer).full_name
```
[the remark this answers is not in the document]
mvashishtha: Never mind, I think I see why we're doing this.

modin/core/execution/client/io.py
mvashishtha: @pyrito for local paths, we don't want to include the `file://`, but for non-local paths like `s3://` we do want the prefix

modin/core/execution/client/container.py
```diff
@@ -0,0 +1,742 @@
...
+        Generate an ID for a new query compiler.
+
+        Returns
+        -------
+        UUID
+            The generated ID.
+        """
+        id = uuid4()
+        while id in self._qc:
+            id = uuid4()
```
pyrito, line 53: Basically keep retrying until we ensure we have a unique ID? Might be helpful to have a comment for this 😄

modin/core/execution/client/container.py
```diff
@@ -0,0 +1,742 @@
...
+            The ID of the query compiler to convert.
+
+        Returns
+        -------
+        pandas.DataFrame
+            The converted DataFrame.
+        """
+        return self._qc[id].to_pandas()
+
+    class DefaultToPandasResult(NamedTuple):
```
pyrito, line 89: Why not just use the `NamedTuple` directly? It will require way less lines of code.

modin/core/storage_formats/base/query_compiler.py
```diff
@@ -1740,20 +1748,22 @@ def describe(self, **kwargs):  # noqa: PR02
     # data in the same place.
 
     @doc_utils.doc_cum_agg(method="sum", refer_to="cumsum")
-    def cumsum(self, fold_axis, **kwargs):  # noqa: PR02
-        return DataFrameDefault.register(pandas.DataFrame.cumsum)(self, **kwargs)
+    def cumsum(self, fold_axis, *args, **kwargs):  # noqa: PR02
```
pyrito, line 1751: Why did `*args` need to get added to all the reduction methods?

modin/core/storage_formats/base/query_compiler.py
```diff
@@ -2143,14 +2153,15 @@ def get_column(df, key):
...
-    def getitem_row_array(self, key):
+    def getitem_row_array(self, key: List[Hashable], numeric: bool = False):
         """
         Get row data for target indices.
 
         Parameters
         ----------
         key : list-like
             Numeric indices of the rows to pick.
+        numeric : bool, default: False
```
pyrito, line 2164: We should explain what this parameter means and why it's needed.

modin/core/storage_formats/pandas/query_compiler.py
```diff
@@ -2348,7 +2353,11 @@ def dropna(self, **kwargs):
             )
         )
 
-    def drop(self, index=None, columns=None):
+    def drop(self, index=None, columns=None, errors: str = "raise"):
+        # `errors` parameter needs to be part of the function signature because
+        # other query compilers may not take care of error handling at the API
+        # layer. This query compiler assumes there won't be any errors due to
+        # invald keys.
```
pyrito, line 2360:
```suggestion
        # invalid keys.
```

modin/experimental/core/storage_formats/hdk/query_compiler.py
```diff
@@ -555,7 +555,11 @@ def concat(self, axis, other, **kwargs):
         )
         return self.__constructor__(new_modin_frame)
 
-    def drop(self, index=None, columns=None):
+    def drop(self, index=None, columns=None, errors: str = "raise"):
+        # `errors` parameter needs to be part of the function signature because
+        # other query compilers may not take care of error handling at the API
+        # layer. This query compiler assumes there won't be any errors due to
+        # invald keys.
```
pyrito, line 562:
```suggestion
        # invalid keys.
```

modin/core/execution/client/container.py
```diff
@@ -0,0 +1,742 @@
...
+
+        Returns
+        -------
+        DefaultToPandasResult
+            The result of the operation. The result is a query compiler ID if
+            and only if the result of the pandas operation is a new
+            query compiler.
+        """
+        result = self._qc[id].default_to_pandas(pandas_op, *args, **kwargs)
+        result_is_qc_id = isinstance(result, self._query_compiler_class)
```
pyrito, line 132: Is `result_is_qc_id` the best name for when we are checking if the types match? The documentation says that the result is a qc ID iff the result is a new query compiler. Doesn't this mean we need to check and see if the qcs are the same? Or do they just have to be the same class?

modin/core/execution/client/container.py
```diff
@@ -0,0 +1,742 @@
...
+    """
+
+    def forwarding_method(self, id, by_is_qc, by, *args, **kwargs):
+        if by_is_qc:
+            by = self._qc[by]
+        new_id = self._generate_id()
+        self._qc[new_id] = getattr(self._qc[id], method_name)(by, *args, **kwargs)
+        return new_id
+
+    setattr(ForwardingQueryCompilerContainer, method_name, forwarding_method)
```
pyrito, line 556: Just wanted to say that this is a super clever way of forwarding methods using `getattr` and `setattr`.

modin/core/execution/client/io.py
```diff
@@ -0,0 +1,142 @@
...
+        self.query_compiler_cls
+            Query compiler with CSV data read in.
+        """
+        if isinstance(filepath_or_buffer, str):
+            filepath_or_buffer = fsspec.open(filepath_or_buffer).full_name
+            file_protocol = "file://"
+            if filepath_or_buffer.startswith(file_protocol):
+                # We will do this so that the backend can know whether this
+                # is a path or a URL.
+                filepath_or_buffer = filepath_or_buffer[len(file_protocol) :]
```
pyrito, line 74: Is there anyway we can do this in a less janky manner? I think we can use `fsspec` to parse things out using `split_protocol`, right?

modin/core/execution/client/query_compiler.py
```diff
@@ -0,0 +1,451 @@
...
+        ID of this query compiler.
+    """
+
+    lazy_execution: bool = True
+
+    def __init__(self, id: UUID):
+        self._id = id
+
+    @classmethod
+    def set_server_connection(cls, conn: Any):
```
pyrito, line 46: Why is this a class method? Once we set the server connection we want every instance of a new ClientQueryCompiler to connect to this specific service? Is there ever a case where we don't want this? (E.g. cases where we have multiple services perhaps)

modin/pandas/base.py
```diff
@@ -149,9 +149,7 @@ def _build_repr_df(self, num_rows, num_cols):
             A pandas dataset with `num_rows` or fewer rows and `num_cols` or fewer columns.
         """
         # Fast track for empty dataframe.
-        if len(self.index) == 0 or (
-            hasattr(self, "columns") and len(self.columns) == 0
-        ):
+        if len(self.index) == 0 or (len(self._query_compiler.columns) == 0):
```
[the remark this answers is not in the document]
pyrito: Ditto

modin/pandas/base.py
```diff
@@ -3172,8 +3171,16 @@ def __getitem__(self, key):
...
+        if not self._query_compiler.lazy_execution:
+            if len(self) == 0:
+                return self._default_to_pandas("__getitem__", key)
+            # fastpath for common case
+            if isinstance(key, str) and key in self._query_compiler.columns:
+                return self._getitem(key)
+            elif is_list_like(key) and all(
+                k in self._query_compiler.columns for k in key
+            ):
+                return self._getitem(key)
```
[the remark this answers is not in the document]
pyrito: I think this is just fast tracking common cases, but we should probably give a good reason as to why we're doing this via a comment or something.

pyrito (commented):
This is looking really good @mvashishtha . Thanks for all the hard work on this. I've added a couple of small nit fixes and asked a couple questions.

modin/pandas/base.py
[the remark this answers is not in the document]
devin-petersohn: This change should not be in this PR. Honestly I thought this was a part of the original codebase

modin/conftest.py
```diff
@@ -319,11 +360,12 @@ def pytest_configure(config):
...
-    if execution == BASE_EXECUTION_NAME:
-        set_base_execution(BASE_EXECUTION_NAME)
-        config.addinivalue_line(
-            "filterwarnings", "default:.*defaulting to pandas.*:UserWarning"
-        )
+    if execution == "BaseOnPython":
+        set_base_on_python_execution()
+        config.addinivalue_line("filterwarnings", default_to_pandas_ignore_string)
+    elif execution == "Client":
+        config.addinivalue_line("filterwarnings", default_to_pandas_ignore_string)
```
helmeleegy, line 367: Is it necessary to reverse the order of lines (367, 368) compared to (364, 365)?

modin/core/storage_formats/base/query_compiler.py
```diff
@@ -2143,14 +2153,17 @@ def get_column(df, key):
 
         return DataFrameDefault.register(get_column)(self, key=key)
 
-    def getitem_row_array(self, key):
+    def getitem_row_array(self, key: List[Hashable], numeric: bool = False):
```
helmeleegy, line 2156: I'm not sure where `numeric=False` is used? It seems that the old behavior has always been equivalent to `numeric=True`, and the same for the new behavior of all occurrences of `getitem_row_array`. Also, why is the default set to `False` rather than `True` which just by looking at this PR seems more commonly used?

modin/pandas/base.py
```diff
@@ -1247,7 +1246,7 @@ def drop(
                         axes[axis] = None
 
         new_query_compiler = self._query_compiler.drop(
-            index=axes["index"], columns=axes["columns"]
+            index=axes["index"], columns=axes["columns"], errors=errors
```
helmeleegy, line 1249: Can we also pass the `errors` parameter to `astype` similar to `drop`?

modin/pandas/series_utils.py
```diff
@@ -331,7 +331,7 @@ def partition(self, sep=" ", expand=True):
             )
 
     def repeat(self, repeats):
-        return self._default_to_pandas(pandas.Series.str.repeat, repeats)
+        return Series(query_compiler=self._query_compiler.str_repeat(repeats))
```
helmeleegy, line 334: Thanks for doing this! What about the other cases `_default_to_pandas` is called in this file? Are there cases, beyond this file, where the decision to default to pandas is made before entering the query compiler layer?

modin/pandas/test/dataframe/test_default.py
```diff
@@ -132,7 +132,12 @@ def test_to_numpy(data):
 @pytest.mark.parametrize("data", test_data_values, ids=test_data_keys)
 def test_partition_to_numpy(data):
     frame = pd.DataFrame(data)
-    for partition in frame._query_compiler._modin_frame._partitions.flatten().tolist():
+    qc = frame._query_compiler
+    if get_current_execution() == "Client":
+        modin_frame = qc._service._qc[qc._id]._modin_frame
+    else:
+        modin_frame = qc._modin_frame
+    for partition in modin_frame._partitions.flatten().tolist():
```
helmeleegy, line 140: Is it expected that the client query compiler would have partitions?

modin/pandas/test/dataframe/test_map_metadata.py
```diff
@@ -708,12 +713,36 @@ def test_drop():
     df_equals(modin_simple.drop([0, 1, 3], axis=0), simple.loc[[2], :])
     df_equals(modin_simple.drop([0, 3], axis="index"), simple.loc[[1, 2], :])
 
-    pytest.raises(ValueError, modin_simple.drop, 5)
-    pytest.raises(ValueError, modin_simple.drop, "C", 1)
-    pytest.raises(ValueError, modin_simple.drop, [1, 5])
-    pytest.raises(ValueError, modin_simple.drop, ["A", "C"], 1)
+    # TODO(https://github.com/modin-project/modin/issues/5163): raise a
+    # KeyError like pandas when the label is not found when lazy_execution is
+    # off. Also use df_equals instead of
```
helmeleegy, line 718: Complete sentence?

modin/core/execution/client/container.py
helmeleegy: +1. It saves a lot of lines of code!

helmeleegy:
LGTM overall! Thanks, Mahesh for all the hard work!

modin/conftest.py
mvashishtha: Nope, I made them match.

modin/pandas/base.py
mvashishtha: We could, but I don't know of a need to do that on the service side yet. In the case of `drop` here, we bypass the usual error checking at the API layer in line 1232 for the typical client query compiler, so we have to pass `errors` down.

modin/pandas/series_utils.py
mvashishtha: I think we should fix the other cases in this file and other places in another PR. Maybe someone added this case because the pandas and base query compilers already had `str_repeat`, but we just weren't using it.

modin/pandas/test/dataframe/test_default.py
mvashishtha: The client query compiler won't have partitions, but here we peak deep into its internals to get partitions-- specifically we get partitions from the `BaseQueryCompiler` in the container in the client query compiler.

modin/pandas/test/dataframe/test_map_metadata.py
mvashishtha: I really don't know what I was thinking of. Deleted.

modin/pandas/base.py
mvashishtha: It turns out there is a need for `errors` on the service side, so I've made the API layer pass `errors` down.

modin/core/storage_formats/base/query_compiler.py
mvashishtha: Devin's branch initially used `getitem_row_labels_array().getitem_column_array()` to do indexing with `loc`. I introduced the `numeric` parameter to `getitem_row_array` so we'd use `numeric=True` instead of the new method `getitem_row_labels_array`. Ultimately we decided we should use the new `take_2d_labels` instead of either `getitem_row_labels_array` or `getitem_row_array(numeric=True)`. `getitem_row_array` should now be as before.

modin/experimental/core/storage_formats/hdk/query_compiler.py
```diff
@@ -687,7 +691,7 @@ def reset_index(self, **kwargs):
             self._modin_frame.reset_index(drop), shape_hint=shape_hint
         )
 
-    def astype(self, col_dtypes, **kwargs):
+    def astype(self, col_dtypes, errors: str):
```
anmyachev, line 694: Why has it changed?

mvashishtha: The way the new client query compiler converts types, we can't tell whether things will error until we actually try to cast types within the query compiler. So the query compiler has to know what do with errors. The other query compilers, including base, pandas, and HDK, can ignore the `errors`. I did the same with `drop`.

modin/core/execution/client/container.py
```diff
@@ -0,0 +1,743 @@
...
+"""Module contains ``ForwardingQueryCompilerContainer`` class."""
+
+import numpy as np
+import pandas
+from typing import Any, NamedTuple, Optional, Union
+from uuid import UUID, uuid4
+
+from modin.core.storage_formats.base.query_compiler import BaseQueryCompiler
+
+
```
AndreyPavlenko, line 23: Is it supposed to have multiple, concurrent clients? If so, then, this class seems not thread safe.

modin/pandas/__init__.py
```diff
@@ -129,6 +130,11 @@ def _update_engine(publisher: Parameter):
     else:
         is_hdk = False
 
+    if Engine.get() == "Client":
```
AndreyPavlenko, line 133: I think, this block could be moved to the very beginning of the method.

modin/pandas/test/dataframe/test_default.py
```diff
@@ -132,7 +132,12 @@ def test_to_numpy(data):
 @pytest.mark.parametrize("data", test_data_values, ids=test_data_keys)
 def test_partition_to_numpy(data):
     frame = pd.DataFrame(data)
-    for partition in frame._query_compiler._modin_frame._partitions.flatten().tolist():
+    qc = frame._query_compiler
+    if get_current_execution() == "Client":
```
AndreyPavlenko, line 136: Similar code is duplicated multiple times. What about having the _modin_frame property in ClientQueryCompiler?
Something like:

    @property
    def _modin_frame(self):
        return self._service._qc[self._id]._modin_frame

modin/core/execution/client/io.py
```diff
@@ -0,0 +1,142 @@
...
+
+        Parameters
+        ----------
+        conn : Any
+            Connection object that is implementation specific.
+        """
+        cls._data_conn = conn
+
+    @classmethod
+    def read_csv(cls, filepath_or_buffer, **kwargs):
```
AndreyPavlenko, line 54: If my understanding is correct, this is a network service, that can read arbitrary file in the file system. It seems like a security issue. I think, absolute file paths must be restricted here. Only paths, relative to the user defined directories must be allowed.

modin/core/execution/client/container.py
mvashishtha: Modin users are supposed to only have one client.

modin/pandas/test/dataframe/test_default.py
mvashishtha: I think this code only appears twice. Also, the client query compiler will not necessarily be backed by a query compiler that has a `_modin_frame`. I prefer leaving the code as is.

modin/pandas/__init__.py
mvashishtha: Done.

modin/core/execution/client/io.py
devin-petersohn: @AndreyPavlenko that's a good question to raise. This implementation doesn't directly imply a network service in my opinion (any query compiler can be used in the service model). Restrictions on what types of files used is perhaps out of scope for this PR, but something we should potentially consider in the future.

AndreyPavlenko: Sure, I don't insist on implementing it in this PR. I think, some configuration capability could be added, to restrict the arbitrary file reading.

mvashishtha (approved):

modin/core/io/file_dispatcher.py
```diff
@@ -170,7 +170,8 @@ def read(cls, *args, **kwargs):
                     t: dtypes[t]
                     for t in dtypes.index
                     if isinstance(dtypes[t], kernel_lib.CategoricalDtype)
-                }
+                },
+                kwargs.get("errors", "raise"),
```
vnlitvinov, line 174: this feels like a sneak fix for something unrelated to the PR

modin/core/storage_formats/base/query_compiler.py
```diff
@@ -529,6 +529,14 @@ def mod(self, other, **kwargs):  # noqa: PR02
...
         return BinaryDefault.register(pandas.DataFrame.mul)(self, other=other, **kwargs)
 
+    @doc_utils.doc_binary_method(
+        operation="multiplication", sign="*", self_on_right=True
+    )
+    def rmul(self, other, **kwargs):  # noqa: PR02
+        return BinaryDefault.register(pandas.DataFrame.rmul)(
+            self, other=other, **kwargs
+        )
+
```
vnlitvinov, line 539: sneak feature, unrelated to the PR?

modin/core/storage_formats/base/query_compiler.py
```diff
@@ -1408,24 +1416,24 @@ def stack(self, level, dropna):
...
+            Control raising of exceptions on invalid data for provided dtype.
 
         Returns
         -------
         BaseQueryCompiler
             New QueryCompiler with updated dtypes.
         """
         return DataFrameDefault.register(pandas.DataFrame.astype)(
-            self, dtype=col_dtypes, **kwargs
+            self, dtype=col_dtypes, errors=errors
```
vnlitvinov, line 1436: sneak fix 🙃 here and below

modin/core/storage_formats/base/query_compiler.py
```diff
@@ -3106,6 +3118,40 @@ def applyer(df):
 
         return DataFrameDefault.register(applyer)(self)
 
+    def take_2d_labels(
+        self,
+        index,
+        columns,
+    ):
```
vnlitvinov, line 3125: why is this function added in this PR?

modin/experimental/core/storage_formats/hdk/query_compiler.py
vnlitvinov: I would rather one would extract this to a separate PR

vnlitvinov (commented):
I went through some of the files, still have to cover others, but I would rather publish part of feedback earlier.

modin/pandas/series_utils.py
vnlitvinov: I would say we _shouldn't_ fix that particular issue in this PR, as it's unrelated to introducing a client QC

vnlitvinov:
@devin-petersohn @mvashishtha do we still need this PR? I think it can be closed for now.

PR #4956, opened 2022-09-12
enhance security of os.system cmd input to prevent command injection

gellanyhassan0:
am fix security bug of os.system in line 6 modin/ci/teamcity/build-docker.py to prevent command injection

"Python possesses many mechanisms to invoke an external executable. However, doing so may present a security issue if appropriate care is not taken to sanitize any user provided or variable input."

https://cwe.mitre.org/data/definitions/78.html


ci/teamcity/build-docker.py
```diff
@@ -3,7 +3,7 @@
 
 
 def execute_command(cmd):
-    status = os.system(cmd)
+    status = os.system('{}.format(cmd)')
```
anmyachev, line 6: @gellanyhassan0 thanks for the PR! Could you please explain how `format` function protects against command injection?

Also what is `ci/teamcity/.build-docker.py.swp` for?

noloerino: I would guess that `.swp` was generated when editing the file in vim; it's used to restore the editor in case of a crash and shouldn't be committed (we might want to add the extension to `.gitignore`.

I think if we really wanted to err on the side of safety here, we should forgo `os.system` in favor of a [`subprocess`](https://docs.python.org/3/library/subprocess.html) call with `shell=False` to explicitly prevent command injection, though I don't know if this would require any other changes to checking the return code.

gellanyhassan0: 1st - am sorry for nano issue (swp) 
2nd - Applications that execute operating system commands or execute commands that interact with the underlying system should neutralize any externally-provided values used in those commands. Failure to do so could allow an attacker to include input that executes unintended commands or exposes sensitive data.

in final my be wrong in os.system('{}.format(cmd)') but can replace with batter  by os.popen(cmd) # Noncompliant

PR #4971, opened 2022-09-14
REFACTOR-#4970: Rewrite base implementations of a partition' `width/length`

### LINKED ISSUE #4970: REFACTOR: rewrite base implementations of a partition' `width/length`
These functions can be reused for `pandas_on_python` execution.

anmyachev:

## What do these changes do?

- [x] Resolves #4970 
- [x] tests added and passing


modin/core/dataframe/pandas/partitioning/partition.py
```diff
@@ -298,10 +298,7 @@ def length(self):
             The length of the object.
         """
         if self._length_cache is None:
-            cls = type(self)
-            func = cls._length_extraction_fn()
-            preprocessed_func = cls.preprocess_func(func)
-            self._length_cache = self.apply(preprocessed_func)
+            self._length_cache = self.apply(self._length_extraction_fn()).get()
```
pyrito, line 301: Ha, this is interesting. Doesn't `self._length_extraction_fn()` effectively do `type(self)._length_extraction_fn()`?

modin/core/execution/python/implementations/pandas_on_python/partitioning/partition.py
```diff
@@ -63,7 +63,7 @@ def get(self):
         Since this object is a simple wrapper, just return the copy of data.
         """
         self.drain_call_queue()
-        return self._data.copy()
+        return self._data.copy() if hasattr(self._data, "copy") else self._data
```
pyrito, line 66: In what case would `data` not have the `copy` attribute?

modin/core/dataframe/pandas/partitioning/partition.py
anmyachev: Almost, applying to an object, not a class, the search space should be slightly larger, but this does not bother us in this case. Therefore, I do not think that we need to do an additional action to calculate the class.

pyrito (approved):

anmyachev:
@mvashishtha could you take a look?

mvashishtha (approved):

anmyachev:
@mvashishtha we can merge now, I think

modin/core/dataframe/pandas/partitioning/partition.py
YarShev, line 301: If some execution relies on this implementation of `length`/`width` and at the same time has `preprocess_func`, which is useful for perf, then we may have perf penalties here.

anmyachev: If I understand correctly, it helps if we apply this function several times in a row, but this is not the case. Also, if it's really needed, it's better to do it in the `apply` function itself.

YarShev: > If I understand correctly, it helps if we apply this function several times in a row, but this is not the case. Also, if it's really needed, it's better to do it in the `apply` function itself.

for the current engines, yes

> Also, if it's really needed, it's better to do it in the apply function itself.

not sure if it is better in the apply

anyway, let's leave as is for now

PR #4988, opened 2022-09-16
Port subset of pandas asvs

jbrockmendel:
Probably needs to be trimmed further, open to suggestions as to where.


asv_bench/benchmarks/pandas/arithmetic.py
```diff
@@ -0,0 +1,379 @@
...
+except ImportError:
+    pass
+
+
+class IntFrameWithScalar:
+    params = [
+        [np.float64, np.int64],
+        [2, 3.0, np.int32(4), np.float64(5)],
+        [
+            operator.add,
```
Garra1980, line 28: do we need all of them, really? I think we should try to pick minimum number of operations that could represent some particular piece of Modin's code

asv_bench/benchmarks/pandas/io/sas.py
```diff
@@ -0,0 +1,30 @@
...
+            "tests",
+            "io",
+            "sas",
+            "data",
+            file,
+        ]
+        self.f = os.path.join(*paths)
+
+    def time_read_sas(self, format):
+        read_sas(self.f, format=format)
```
Garra1980, line 30: not sure this is important one

asv_bench/benchmarks/pandas/io/style.py
```diff
@@ -0,0 +1,93 @@
+import numpy as np
+
+from modin.pandas import (
+    DataFrame,
+    IndexSlice,
+)
+
+
+class Render:
```
Garra1980, line 9: is this really frequently used?

asv_bench/benchmarks/pandas/io/csv.py
```diff
@@ -0,0 +1,573 @@
...
+        self.df_custom_index_then_head.to_csv(self.fname)
+
+
+class StringIORewind:
+    def data(self, stringio_object):
+        stringio_object.seek(0)
+        return stringio_object
+
+
+class ReadCSVDInferDatetimeFormat(StringIORewind):
```
Garra1980, line 181: Don't think we need all of these. We have some for readcsv in current asv runs btw

asv_bench/benchmarks/pandas/gil.py
```diff
@@ -0,0 +1,185 @@
...
+                thread.start()
+            for thread in threads:
+                thread.join()
+
+        return inner
+
+    return wrapper
+
+
+class ParallelGroupbyMethods:
```
Garra1980, line 82: do we really need this in Modin testing?

asv_bench/benchmarks/pandas/frame_methods.py
```diff
@@ -0,0 +1,558 @@
...
+
+class SortValues:
+
+    params = [True, False]
+    param_names = ["ascending"]
+
+    def setup(self, ascending):
+        self.df = DataFrame(np.random.randn(1000000, 2), columns=list("AB"))
+
+    def time_frame_sort_values(self, ascending):
```
Garra1980, line 447: thought we have it in asv already

asv_bench/benchmarks/pandas/strings.py
```diff
@@ -0,0 +1,288 @@
...
+        Series(self.series_cat_arr, dtype=dtype)
+
+    def time_cat_frame_construction(self, dtype):
+        DataFrame(self.frame_cat_arr, dtype=dtype)
+
+    def peakmem_cat_frame_construction(self, dtype):
+        DataFrame(self.frame_cat_arr, dtype=dtype)
+
+
+class Methods(Dtypes):
```
Garra1980, line 64: Is it worth to measure at all?

asv_bench/benchmarks/pandas/io/sas.py
YarShev: agree, we are just defaulting to pandas for read_sas for now

asv_bench/benchmarks/pandas/arithmetic.py
```diff
@@ -0,0 +1,379 @@
...
+            operator.truediv,
+            operator.floordiv,
+            operator.pow,
+            operator.mod,
+            operator.eq,
+            operator.ne,
+            operator.gt,
+            operator.ge,
+            operator.lt,
+            operator.le,
```
YarShev, line 40:
```suggestion
            operator.mul,
```
I think we can leave 1 or 2 operations because they go through the same code path.

asv_bench/benchmarks/pandas/arithmetic.py
```diff
@@ -0,0 +1,379 @@
...
+        arr = np.random.randn(20000, 100)
+        self.df = DataFrame(arr.astype(dtype))
+
+    def time_frame_op_with_scalar(self, dtype, scalar, op):
+        op(self.df, scalar)
+
+
+class OpWithFillValue:
+    def setup(self):
+        # GH#31300
```
YarShev, line 55: Should this be removed?

asv_bench/benchmarks/pandas/arithmetic.py
```diff
@@ -0,0 +1,379 @@
...
+            "lt",
+            "le",
+            "ge",
+            "gt",
+            "add",
+            "sub",
+            "truediv",
+            "floordiv",
+            "mul",
+            "pow",
```
YarShev, line 84:
```suggestion
            "add",
            "mul",
```
I think we can leave 1 or 2 operations because they go through the same code path.

asv_bench/benchmarks/pandas/arithmetic.py
```diff
@@ -0,0 +1,379 @@
...
+    def time_frame_op_with_series_axis1(self, opname):
+        getattr(operator, opname)(self.df, self.ser)
+
+
+class FrameWithFrameWide:
+    # Many-columns, mixed dtypes
+
+    params = [
+        [
+            # GH#32779 has discussion of which operators are included here
```
YarShev, line 109: Should this be removed?

asv_bench/benchmarks/pandas/arithmetic.py
```diff
@@ -0,0 +1,379 @@
...
+            operator.add,
+            operator.floordiv,
+            operator.gt,
```
YarShev, line 112: I think we can leave 1 or 2 operations because they go through the same code path.

asv_bench/benchmarks/pandas/arithmetic.py
```diff
@@ -0,0 +1,379 @@
...
+        # we choose dtypes so as to make the blocks
+        #  a) not perfectly match between right and left
+        #  b) appreciably bigger than single columns
+        n_rows, n_cols = shape
+
+        if op is operator.floordiv:
+            # floordiv is much slower than the other operations -> use less data
+            n_rows = n_rows // 10
+
+        # construct dataframe with 2 blocks
```
YarShev, line 134: We should probably remove such comments as this is not applicable to Modin. We will have the number of partitions that is equal to `NPartitions.get()`.

asv_bench/benchmarks/pandas/arithmetic.py
```diff
@@ -0,0 +1,379 @@
...
+        op(self.left, self.right)
+
+    def time_op_same_blocks(self, op, shape):
+        # blocks (and dtypes) are aligned
+        op(self.left, self.left)
+
+
+
+
+class Ops2:
```
YarShev, line 167: The comments above are applicable to all the changes below.

anmyachev:
@jbrockmendel The current implementation assumes that if there are changes in the pull request in `asv_bench` folder, then benchmarks are run on a small dataset and on different configurations: modin on ray, just pandas, modin on omnisci, to check for correctness (in case some combination of parameters is not checked during unit testing, there is some redundancy that can be abandoned). However, if we increase the number of benchmarks so much, then this approach stops working due to the amount of time it will take to run all the benchmarks completely on all configurations.

To start, I would like to ask how this works in pandas. If someone adds new benchmarks, will they be tested in the CI of the pull request and will the rest be launched at the same time?

asv_bench/benchmarks/pandas/strftime.py
```diff
@@ -0,0 +1,64 @@
+import numpy as np
+
+import modin.pandas as pd
+from pandas import offsets
+
+
+class DatetimeStrftime:
```
Garra1980, line 7: Don't think we need these

asv_bench/benchmarks/pandas/stat_ops.py
```diff
@@ -0,0 +1,144 @@
+import numpy as np
+
+import modin.pandas as pd
+
+ops = ["mean", "sum", "median", "std", "skew", "kurt", "mad", "prod", "sem", "var"]
```
Garra1980, line 5: Probably we can pick up not all of these ops

asv_bench/benchmarks/pandas/sparse.py
```diff
@@ -0,0 +1,75 @@
+import numpy as np
```
Garra1980, line 1: I'd say we can drop these as not very popular

asv_bench/benchmarks/pandas/join_merge.py
```diff
@@ -0,0 +1,294 @@
...
+    Series,
+    concat,
+    date_range,
+    merge,
+)
+
+import pandas._testing as tm
+
+
+class Concat:
```
Garra1980, line 17: We have concat already I believe

asv_bench/benchmarks/pandas/frame_ctor.py
```diff
@@ -0,0 +1,225 @@
+import numpy as np
+
+import modin.pandas as pd
+from modin.pandas import (
```
Garra1980, line 4: We may not want to test this separately given Modin lazy execution abilities, @anmyachev thoughts?

asv_bench/benchmarks/pandas/frame_methods.py
```diff
@@ -0,0 +1,530 @@
...
+
+
+class MemoryUsage:
+    def setup(self):
+        self.df = DataFrame(np.random.randn(100000, 2), columns=list("AB"))
+        self.df2 = self.df.copy()
+        self.df2["A"] = self.df2["A"].astype("object")
+
+    def time_memory_usage(self):
+        self.df.memory_usage(deep=True)
```
Garra1980, line 524: Hard to imagine how Modin can parallelize this one

asv_bench/benchmarks/pandas/frame_methods.py
```diff
@@ -0,0 +1,530 @@
+import string
```
Garra1980, line 1: this is too big and some of functions already in testing - @jbrockmendel please make sure we don't have duplicates

asv_bench/benchmarks/pandas/indexing.py
```diff
@@ -0,0 +1,406 @@
+"""
```
Garra1980, line 1: we definitely need to reduce this - I think we have something tested already and we don't need such large number fof cases, let's decide what we really want to leave - eg multiindexing, what else is valuable?

asv_bench/benchmarks/pandas/join_merge.py
```diff
@@ -0,0 +1,294 @@
+import string
```
Garra1980, line 1: Let's reduce these as well. I believe we don't really need all cases here

asv_bench/benchmarks/pandas/arithmetic.py
```diff
@@ -0,0 +1,266 @@
...
+    offsets.MonthBegin(),
+    offsets.DateOffset(months=2, days=2),
+    offsets.BusinessDay(),
+    offsets.SemiMonthEnd(),
+    offsets.SemiMonthBegin(),
+]
+offsets = non_apply + other_offsets
+
+
+class OffsetArrayArithmetic:
```
Garra1980, line 221: How is this operation important?

asv_bench/benchmarks/pandas/dtypes.py
```diff
@@ -0,0 +1,86 @@
...
+    string_dtypes,
+)
+
+_numpy_dtypes = [
+    np.dtype(dtype) for dtype in (numeric_dtypes + datetime_dtypes + string_dtypes)
+]
+_dtypes = _numpy_dtypes + extension_dtypes
+
+
+class SelectDtypes:
```
Garra1980, line 22: Is this worth to benchmark for Modin?

asv_bench/benchmarks/pandas/series_methods.py
```diff
@@ -0,0 +1,278 @@
...
+
+    def time_any(self, N, case, dtype):
+        self.s.any()
+
+
+class NanOps:
+
+    params = [
+        [
+            "var",
```
Garra1980, line 235: We test some of these functions already in another place, right? Do we really need to test them again?

asv_bench/benchmarks/pandas/frame_ctor.py
anmyachev: I'm not sure if the current constructors are lazily executed. Therefore, it is probably interesting to see an Modin' overhead in the construction of objects.

asv_bench/benchmarks/pandas/frame_methods.py
jbrockmendel: easy to imagine how it could be painfully slow

Garra1980: Probably yes, but anyway how is this particular case important compare to others, that is the question

asv_bench/benchmarks/pandas/dtypes.py
jbrockmendel: i think its harmless

asv_bench/benchmarks/pandas/frame_methods.py
jbrockmendel: done

asv_bench/benchmarks/pandas/arithmetic.py
jbrockmendel: removed

asv_bench/benchmarks/pandas/frame_ctor.py
Garra1980: Do you think we need all of these cases?

asv_bench/benchmarks/pandas/groupby.py
```diff
@@ -0,0 +1,994 @@
+from functools import partial
```
Garra1980, line 1: Size of this file scares me - we definitely don't want to benchmark everything in this file. Let's pick only useful cases

asv_bench/benchmarks/pandas/frame_ctor.py
```diff
@@ -0,0 +1,225 @@
...
+class FromSeries:
+    def setup(self):
+        mi = MultiIndex.from_product([range(100), range(100)])
+        self.s = Series(np.random.randn(10000), index=mi)
+
+    def time_mi_series(self):
+        DataFrame(self.s)
+
+
+class FromDictwithTimestamp:
```
anmyachev, line 76: We do not have a special implementation, when such a data type is used in a dictionary, it can be deleted.

jbrockmendel: Why would the lack of a special implementation matter?

anmyachev: My guess is that we don't want to have a universal test suite for pandas and modin to begin with, but rather a sufficient test suite for all modin mechanisms. So the modin mechanism for processing dictionaries does not depend on the data types inside.

asv_bench/benchmarks/pandas/frame_methods.py
YarShev: We can remove this because it is not an important case and we are defaulting to pandas here.

asv_bench/benchmarks/pandas/arithmetic.py
```diff
@@ -0,0 +1,207 @@
...
+
+
+class IntFrameWithScalar:
+    params = [
+        [np.float64, np.int64],
+        [2, 3.0, np.int32(4), np.float64(5)],
+        [
+            # subset of operators used in the corresponding pandas asv;
+            #  include floordiv in particular bc it goes through
+            #  its own code-path.
```
YarShev, line 22:
```suggestion
            # include floordiv in particular bc it goes through
            # its own code-path.
```
Do you mean floordiv goes through its own code-path in pandas?

asv_bench/benchmarks/pandas/arithmetic.py
```diff
@@ -0,0 +1,207 @@
...
+        if op is operator.floordiv:
+            # floordiv is much slower than the other operations -> use less data
+            n_rows = n_rows // 10
+
+        # construct dataframe with 2 blocks
+        arr1 = np.random.randn(n_rows, n_cols // 2).astype("f8")
+        arr2 = np.random.randn(n_rows, n_cols // 2).astype("f4")
+        df = pd.concat([DataFrame(arr1), DataFrame(arr2)], axis=1, ignore_index=True)
+        # should already be the case, but just to be sure
+        df._consolidate_inplace()
```
YarShev, line 73: There is no this operation in Modin.

asv_bench/benchmarks/pandas/timeseries.py
```diff
@@ -0,0 +1,139 @@
...
+        self.series.dt.day_name()
+
+    def time_dt_accessor_time(self, tz):
+        self.series.dt.time
+
+    def time_dt_accessor_date(self, tz):
+        self.series.dt.date
+
+    def time_dt_accessor_year(self, tz):
+        self.series.dt.year
```
anmyachev, line 136: All these operations are implemented through the same mechanism in Modin - [simple mapping.]( https://github.com/modin-project/modin/blob/d6d503ac7c3028d871c34d9e99e925ddb0746df6/modin/core/storage_formats/pandas/query_compiler.py#L1463)
So we can test only `dt.time` for now.

anmyachev:
A small summary for further integration of this set of benchmarks into Modin:
* We will also need to make sure that the time that ASV will show for the benchmark is correct. So we need to make sure that the function under test actually performed the calculations. We have `execute` function for the purpose.

* It will also be necessary to check that after preparing the data in `setup` function, there are no pending calculations left that will affect the final time.

* In addition, it is necessary that benchmarks can be run using pure pandas. This behavior is currently controlled by the `MODIN_ASV_USE_IMPL` environment variable. The numbers obtained without comparison with the pandas do not make much sense.

asv_bench/benchmarks/pandas/arithmetic.py
jbrockmendel: yes

jbrockmendel:
> A small summary for further integration of this set of benchmarks into Modin:

@anmyachev pls go ahead and push any relevant changes to this branch

asv_bench/benchmarks/pandas/join_merge.py
jbrockmendel: trimmed many, pls take a look

asv_bench/benchmarks/pandas/series_methods.py
jbrockmendel: cut these in half.

asv_bench/benchmarks/pandas/groupby.py
jbrockmendel: trimmed a bunch of these. pls take a look

asv_bench/benchmarks/pandas/reshape.py
```diff
@@ -0,0 +1,114 @@
...
+        self.df2 = self.df.iloc[:-1]
+
+    def time_full_product(self, dtype):
+        self.df.unstack()
+
+    def time_without_last_row(self, dtype):
+        self.df2.unstack()
+
+
+class Explode:
```
Garra1980, line 101: we have this already i believe

asv_bench/benchmarks/pandas/frame_methods.py
```diff
@@ -0,0 +1,292 @@
...
+        nrows = 10000
+        data = np.random.randn(nrows, 10)
+        arrays = np.tile(np.random.randn(3, nrows // 100), 100)
+        idx = MultiIndex.from_arrays(arrays)
+        self.df3 = DataFrame(data, index=idx)
+        self.df4 = DataFrame(data, index=np.random.randn(nrows))
+        self.df_tall = DataFrame(np.random.randn(nrows, 10))
+        self.df_wide = DataFrame(np.random.randn(10, nrows))
+
+    def time_html_repr_trunc_mi(self):
```
Garra1980, line 99: what exactly do these do? I feel like we don't want to benchmark them, at least repr_html

asv_bench/benchmarks/pandas/reshape.py
jbrockmendel: removed

asv_bench/benchmarks/pandas/frame_methods.py
jbrockmendel: removed

asv_bench/benchmarks/pandas/timeseries.py
```diff
@@ -0,0 +1,28 @@
...
+from modin.pandas import (
+    DataFrame,
+    Series,
+    date_range,
+    period_range,
+    timedelta_range,
+)
+
+
+class ResetIndex:
```
anmyachev, line 15: We already have benchmarks for `reset_index`. Can be safely removed.

jbrockmendel: removed.

PR #4997, opened 2022-09-19
FIX-#4996: Evaluate BenchmarkMode at each function call.

### LINKED ISSUE #4996: BUG: Benchmark Mode is evaluated only once when each function is defined
### Modin version checks

### Reproducible Example

```python
import time
import modin.pandas as pd

df = pd.DataFrame(list(range(32)))
BenchmarkMode.put(True)
%time result = df.applymap(lambda col: time.sleep(0.3))
%time print(result)
```

### Issue Description

wait_computations_if_benchmark_mode decides [when each function is defined](https://github.com/modin-project/modin/blob/2660aa33b79c79af828bce3b8f5c6573bc4d7c56/modin/core/dataframe/pandas/partitioning/partition_manager.py#L54) whether benchmark mode is on. Consequently, it's not possible to set benchmark mode after the methods have been defined. In the example above, turning benchmark mode on has no effect, even though the user turns benchmark mode on before the `applymap`. The `applymap` finishes in 2.39 ms because it's still async, and then `print` finishes in 13 seconds because it blocks on the `applymap`.

### Expected Behavior

Users should be able to set benchmark mode on or off at any point in the main process.

### Error Logs

N/A

### Installed Versions

<details>

INSTALLED VERSIONS
------------------
commit           : f727c045729b969de0e4a876e14209cde77166b9
python           : 3.10.4.final.0
python-bits      : 64
OS               : Darwin
OS-release       : 21.5.0
Version          : Darwin Kernel Version 21.5.0: Tue Apr 26 21:08:22 PDT 2022; root:xnu-8020.121.3~4/RELEASE_X86_64
machine          : x86_64
processor        : i386
byteorder        : little
LC_ALL           : None
LANG             : en_US.UTF-8
LOCALE           : en_US.UTF-8

Modin dependencies
------------------
modin            : 0.15.0+134.gf727c0457
ray              : 2.0.0
dask             : 2022.7.1
distributed      : 2022.7.1
omniscidbe       : None

pandas dependencies
-------------------
pandas           : 1.4.4
numpy            : 1.23.2
pytz             : 2022.2.1
dateutil         : 2.8.2
setuptools       : 61.2.0
pip              : 22.2.2
Cython           : None
pytest           : 7.1.2
hypothesis       : None
sphinx           : 5.1.1
blosc            : None
feather          : 0.4.1
xlsxwriter       : None
lxml.etree       : 4.9.1
html5lib         : None
pymysql          : None
psycopg2         : 2.9.3
jinja2           : 3.1.2
IPython          : 8.4.0
pandas_datareader: None
bs4              : 4.11.1
bottleneck       : None
brotli           : 1.0.9
fastparquet      : 0.8.1
fsspec           : 2022.7.1
gcsfs            : None
markupsafe       : 2.1.1
matplotlib       : 3.5.2
numba            : None
numexpr          : 2.8.3
odfpy            : None
openpyxl         : 3.0.10
pandas_gbq       : 0.17.7
pyarrow          : 8.0.0
pyreadstat       : None
pyxlsb           : None
s3fs             : 2022.7.1
scipy            : 1.9.0
snappy           : None
sqlalchemy       : 1.4.39
tables           : 3.7.0
tabulate         : None
xarray           : 2022.6.0
xlrd             : 2.0.1
xlwt             : None
zstandard        : None

</details>

mvashishtha:

## What do these changes do?

FIX-#4996: Evalute BenchmarkMode at each function call.

- [x] Resolves #4996
- [x] tests added and passing

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -421,7 +421,6 @@ jobs:
       - name: Install HDF5
         run: sudo apt update && sudo apt install -y libhdf5-dev
       - run: pytest modin/test/storage_formats/omnisci/test_internals.py
-      - run: MODIN_BENCHMARK_MODE=True pytest modin/pandas/test/internals/test_benchmark_mode.py
       - run: pytest modin/experimental/core/execution/native/implementations/omnisci_on_native/test/test_dataframe.py
       - run: pytest modin/pandas/test/test_io.py::TestCsv --verbose
       - run: pytest modin/test/interchange/dataframe_protocol/test_general.py
--- a/docs/release_notes/release_notes-0.16.0.rst
+++ b/docs/release_notes/release_notes-0.16.0.rst
@@ -50,6 +50,7 @@ Key Features and Updates
   * FIX-#4734: Handle Series.apply when return type is a DataFrame (#4830)
   * FIX-#4983: Set `frac` to `None` in _sample when `n=0` (#4984)
   * FIX-#4993: Return `_default_to_pandas` in `df.attrs` (#4995)
+  * FIX-#4996: Evaluate BenchmarkMode at each function call (#4997)
 * Performance enhancements
   * PERF-#4182: Add cell-wise execution for binary ops, fix bin ops for empty dataframes (#4391)
   * PERF-#4288: Improve perf of `groupby.mean` for narrow data (#4591)
--- a/modin/core/dataframe/pandas/partitioning/partition_manager.py
+++ b/modin/core/dataframe/pandas/partitioning/partition_manager.py
@@ -813,7 +813,6 @@ class PandasDataframePartitionManager(ClassLogger, ABC):
 
     @classmethod
     def wait_partitions(cls, partitions):
-        print("calling python wait_partitions")
         """
         Wait on the objects wrapped by `partitions`, without materializing them.
 
--- a/modin/core/dataframe/pandas/partitioning/partition_manager.py
+++ b/modin/core/dataframe/pandas/partitioning/partition_manager.py
@@ -51,16 +51,16 @@ def wait_computations_if_benchmark_mode(func):
     -----
     `func` should return NumPy array with partitions.
     """
-    if BenchmarkMode.get():
 
-        @wraps(func)
-        def wait(cls, *args, **kwargs):
-            """Wait for computation results."""
-            result = func(cls, *args, **kwargs)
-            if isinstance(result, tuple):
-                partitions = result[0]
-            else:
-                partitions = result
+    @wraps(func)
+    def wait(cls, *args, **kwargs):
+        """Wait for computation results."""
+        result = func(cls, *args, **kwargs)
+        if isinstance(result, tuple):
+            partitions = result[0]
+        else:
+            partitions = result
+        if BenchmarkMode.get():
             # When partitions have a deferred call queue, calling
             # partition.wait() on each partition serially will serially kick
             # off each deferred computation and wait for each partition to
@@ -73,10 +73,9 @@ def wait_computations_if_benchmark_mode(func):
             # (We can't just add a `cls` argument to this `wait` function, since doing so
             # seems to be incompatible with the way function decorators work)
             cls.wait_partitions(partitions.flatten())
-            return result
+        return result
 
-        return wait
-    return func
+    return wait
 
 
 class PandasDataframePartitionManager(ClassLogger, ABC):
@@ -814,6 +813,7 @@ class PandasDataframePartitionManager(ClassLogger, ABC):
 
     @classmethod
     def wait_partitions(cls, partitions):
+        print("calling python wait_partitions")
         """
         Wait on the objects wrapped by `partitions`, without materializing them.
 
--- a/modin/core/dataframe/pandas/partitioning/partition_manager.py
+++ b/modin/core/dataframe/pandas/partitioning/partition_manager.py
@@ -56,11 +56,11 @@ def wait_computations_if_benchmark_mode(func):
     def wait(cls, *args, **kwargs):
         """Wait for computation results."""
         result = func(cls, *args, **kwargs)
-        if isinstance(result, tuple):
-            partitions = result[0]
-        else:
-            partitions = result
         if BenchmarkMode.get():
+            if isinstance(result, tuple):
+                partitions = result[0]
+            else:
+                partitions = result
             # When partitions have a deferred call queue, calling
             # partition.wait() on each partition serially will serially kick
             # off each deferred computation and wait for each partition to
--- a/modin/core/dataframe/pandas/partitioning/partition_manager.py
+++ b/modin/core/dataframe/pandas/partitioning/partition_manager.py
@@ -70,8 +70,6 @@ def wait_computations_if_benchmark_mode(func):
             [part.drain_call_queue() for part in partitions.flatten()]
             # The partition manager invokes the relevant .wait() method under
             # the hood, which should wait in parallel for all computations to finish
-            # (We can't just add a `cls` argument to this `wait` function, since doing so
-            # seems to be incompatible with the way function decorators work)
             cls.wait_partitions(partitions.flatten())
         return result
 
--- a/modin/pandas/test/internals/test_benchmark_mode.py
+++ b/modin/pandas/test/internals/test_benchmark_mode.py
@@ -16,7 +16,6 @@ import unittest.mock as mock
 import modin.pandas as pd
 from modin.pandas.test.utils import test_data_values
 from modin.config import BenchmarkMode, Engine
-from modin.test.test_utils import warns_that_defaulting_to_pandas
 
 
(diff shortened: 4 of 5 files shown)


docs/release_notes/release_notes-0.16.0.rst
```diff
@@ -49,6 +49,7 @@ Key Features and Updates
   * FIX-#4845: Fix KeyError from `__getitem_bool` for single row dataframes (#4845)
   * FIX-#4734: Handle Series.apply when return type is a DataFrame (#4830)
   * FIX-#4983: Set `frac` to `None` in _sample when `n=0` (#4984)
+  * FIX-#4996: Evalute BenchmarkMode at each function call (#4997)
```
anmyachev, line 52:
```suggestion
  * FIX-#4996: Evaluate BenchmarkMode at each function call (#4997)
```

anmyachev:
@mvashishtha could you add a test?

anmyachev:
> @anmyachev I can test that switching benchmark mode on or off doesn't break anything, but do you have any ideas about how to test whether benchmark mode is actually making code async? I can't think of any way to do that that doesn't require some kind of infinite loop in a remote task, and that's not good testing practice.

A sign of asynchronous code is the absence of a `wait_partitions` call. If instead of this function we make a mock object, then it will be possible to check whether this function was called.

modin/core/dataframe/pandas/partitioning/partition_manager.py
```diff
@@ -51,16 +51,16 @@ def wait_computations_if_benchmark_mode(func):
     -----
     `func` should return NumPy array with partitions.
     """
-    if BenchmarkMode.get():
-
-        @wraps(func)
```
pyrito, line 56: If BenchmarkMode is enabled here, shouldn't we be blocking? Am I missing something here?

mvashishtha: Resolved offline. We only evaluate the benchmark mode once, when we run `wait_computations_if_benchmark_mode`.

mvashishtha (commented):
@anmyachev I used mocks to check that we are calling `wait_partitions`.

modin/pandas/test/internals/test_benchmark_mode.py
```diff
@@ -11,21 +11,57 @@
...
     # On Omnisci storage, transpose() defaults to Pandas.
-    with (
-        warns_that_defaulting_to_pandas()
-        if StorageFormat.get() == "Omnisci"
-        else nullcontext()
-    ):
+    with warns_that_defaulting_to_pandas(), mock.patch(wait_method) as wait:
         pd.DataFrame(test_data_values[0]).mean()
+
+    wait.assert_called()
```
pyrito, line 50: This is cool! Shouldn't this be called within the context manager?

modin/pandas/test/internals/test_benchmark_mode.py
```diff
@@ -11,21 +11,57 @@
...
+
+
+def test_turn_off():
+    df = pd.DataFrame([0])
+    BenchmarkMode.put(False)
+    with mock.patch(wait_method) as wait:
+        df.dropna()
+    wait.assert_not_called()
```
pyrito, line 58: ditto

modin/pandas/test/internals/test_benchmark_mode.py
```diff
@@ -11,21 +11,57 @@
...
+
+
+def test_turn_on():
+    BenchmarkMode.put(False)
+    df = pd.DataFrame([0])
+    BenchmarkMode.put(True)
+    with mock.patch(wait_method) as wait:
+        df.dropna()
+    wait.assert_called()
```
pyrito, line 67: ditto

pyrito (commented):
Just a few Qs

modin/pandas/test/internals/test_benchmark_mode.py
mvashishtha: `unittest.mock` documentation seems to [show](https://docs.python.org/3/library/unittest.mock.html#quick-guide) doing an `assert_called_once_with` assertion outside the context manager. I think we should only use the context manager to wrap the code that  strictly needs it, that is to say the code that will indirectly call the mocked method.

modin/core/dataframe/pandas/partitioning/partition_manager.py
```diff
@@ -51,16 +51,16 @@ def wait_computations_if_benchmark_mode(func):
...
-        def wait(cls, *args, **kwargs):
-            """Wait for computation results."""
-            result = func(cls, *args, **kwargs)
-            if isinstance(result, tuple):
-                partitions = result[0]
-            else:
-                partitions = result
+
+    @wraps(func)
```
vnlitvinov, line 55: Ideally this should be done in a subscribe way - the whole logic of configuration objects was constructed to allow subscribing to changes and changing the code behaviour only once per configuration change (in this case when benchmark mode goes from on to off or vise versa).

That said, I don't have a good suggestion on how to change this particular piece...

vnlitvinov (changes requested):
Tests look good, let's try to lose less performance in benchmark mode off, though.

.github/workflows/ci.yml
```diff
@@ -421,7 +421,6 @@ jobs:
       - name: Install HDF5
         run: sudo apt update && sudo apt install -y libhdf5-dev
       - run: pytest modin/test/storage_formats/omnisci/test_internals.py
-      - run: MODIN_BENCHMARK_MODE=True pytest modin/pandas/test/internals/test_benchmark_mode.py
```
anmyachev, line 424: Why isn't this being tested anymore?

anmyachev: This is apparently because although it inherits the pandas partition manager, it does not use methods that are wrapped in the `wait_computations_if_benchmark_mode` decorator. I wonder if he uses at least something from the inherited class. If not, then it might be better to inherit from a more basic object. OK for now.

anmyachev (approved):
LGTM!

pyrito (approved):

vnlitvinov (approved):
LGTM, thanks!

modin/pandas/test/internals/test_benchmark_mode.py
```diff
@@ -11,21 +11,55 @@
...
+        + "PandasOnDaskDataframePartitionManager.wait_partitions"
+    )
+else:
+    wait_method = (
+        "modin.core.dataframe.pandas.partitioning."
+        + "partition_manager.PandasDataframePartitionManager.wait_partitions"
+    )
+
+
+def test_from_environment_variable():
```
YarShev, line 43: I would suggest to make this test depending on a parameter named `value`, which get a value from the env var in `pytest.mark.parametrize` decorator. Because if someone runs the test locally without setting the env var, he gets the test failed, which might be confusing, especially for newbies.

modin/pandas/test/internals/test_benchmark_mode.py
```diff
@@ -11,21 +11,55 @@
...
-        if StorageFormat.get() == "Omnisci"
-        else nullcontext()
-    ):
+    with mock.patch(wait_method) as wait:
         pd.DataFrame(test_data_values[0]).mean()
+
+    wait.assert_called()
+
+
+def test_turn_off():
```
YarShev, line 51: I would combine both of the tests below into one.

mvashishtha: I very much prefer that tests test only one behavior (see e.g. [here](https://stackoverflow.com/a/61868/17554722) for why). I think many Modin unit tests are too long. When a single test case has many assertions within it that may fail, it's hard to tell from the unit test report what exactly is going wrong. Ideally we should be able to use the name of each failed test to identify one incorrect behavior.

modin/pandas/test/internals/test_benchmark_mode.py
mvashishtha: That's a problem that predates this PR. I'm not a newbie, but I have also been surprised to find this test failing for the reason you describe. We can open a separate issue for this.

YarShev: Yes, create one please.

modin/pandas/test/internals/test_benchmark_mode.py
YarShev: Ok, let's leave as is.

YarShev: That said, we can do two tests: 1) Test for the class config setting, 2) Test for the env var setting. I think it would be okay.

modin/pandas/test/internals/test_benchmark_mode.py
mvashishtha: I filed #5008.

YarShev: thanks

PR #5013, opened 2022-09-21
REFACTOR-#5012: Add mypy checks for singleton files in base modin directory

### LINKED ISSUE #5012: REFACTOR: Add mypy checks for singleton files in base modin directory
Before adding annotations to modin.core, I think it would be best to first add type annotations to singleton files (modin/_version.py, modin/db_conn.py, modin/error_message.py, and modin/utils.py) to help types further propagate throughout the codebase.

noloerino:
## What do these changes do?
This PR adds type annotations to `modin/{_version,db_conn,error_message,utils}.py` in anticipation of adding more annotations to the rest of the codebase.

A few notes:
- In cases where the same variable is reassigned to a different type (like assigning `s = [s]` to convert a string to list of strings), mypy's preferred solution is to create a new variable for the new type.
- mypy supports [structural subtyping](https://mypy.readthedocs.io/en/stable/protocols.html), which lets us define what are essentially placeholder classes to perform `isinstance` checks on in cases where we would otherwise use `hasattr` for duck-typed function calls/field accesses.
- In some situations where the origin of a property is not obvious, or is hard to convey to the type checker via annotations, I threw in a `# type: ignore` annotation on the offending line.

- [x] Resolves #5012
- [x] tests added and passing

### DIFF
--- a/modin/_version.py
+++ b/modin/_version.py
@@ -112,14 +112,11 @@ def run_command(
         if verbose:
             print("unable to find command, tried %s" % (commands,))
         return None, None
-    stdout: str
-    stdout = p.communicate()[0].strip()  # type: ignore
-    if sys.version_info[0] >= 3:
-        stdout = stdout.decode()  # type: ignore
+    stdout = p.communicate()[0].strip().decode()
     if p.returncode != 0:
         if verbose:
-            print("unable to run %s (error)" % dispcmd)
-            print("stdout was %s" % stdout)
+            print(f"unable to run {dispcmd} (error)")
+            print(f"stdout was {stdout}")
         return None, p.returncode
     return stdout, p.returncode
 
--- a/modin/_version.py
+++ b/modin/_version.py
@@ -14,9 +14,12 @@ import os
 import re
 import subprocess
 import sys
+from typing import Any, Callable, Dict, Optional, List, Tuple
+
+VcsPieces = Dict[str, Any]
 
 
-def get_keywords():
+def get_keywords() -> Dict[str, str]:
     """Get the keywords needed to look up the version information."""
     # these strings will be replaced by git during git-archive.
     # setup.py/versioneer.py will grep for the variable names, so they must
@@ -32,8 +35,15 @@ def get_keywords():
 class VersioneerConfig:
     """Container for Versioneer configuration parameters."""
 
+    VCS: str
+    style: str
+    tag_prefix: str
+    parentdir_prefix: str
+    versionfile_source: str
+    verbose: bool
 
-def get_config():
+
+def get_config() -> VersioneerConfig:
     """Create, populate and return the VersioneerConfig() object."""
     # these strings are filled in when 'setup.py versioneer' creates
     # _version.py
@@ -51,14 +61,13 @@ class NotThisMethod(Exception):
     """Exception raised if a method is not valid for the current scenario."""
 
 
-LONG_VERSION_PY = {}
-HANDLERS = {}
+HANDLERS: Dict[str, Dict[str, Callable]] = {}
 
 
-def register_vcs_handler(vcs, method):  # decorator
+def register_vcs_handler(vcs: str, method: str) -> Callable:  # decorator
     """Decorator to mark a method as the handler for a particular VCS."""
 
-    def decorate(f):
+    def decorate(f: Callable) -> Callable:
         """Store f in HANDLERS[vcs][method]."""
         if vcs not in HANDLERS:
             HANDLERS[vcs] = {}
@@ -68,7 +77,14 @@ def register_vcs_handler(vcs, method):  # decorator
     return decorate
 
 
-def run_command(commands, args, cwd=None, verbose=False, hide_stderr=False, env=None):
+def run_command(
+    commands: List[str],
+    args: List[Any],
+    cwd: Optional[str] = None,
+    verbose: bool = False,
+    hide_stderr: bool = False,
+    env: Optional[Dict[str, str]] = None,
+) -> Tuple[Optional[str], Optional[int]]:
     """Call the given command(s)."""
     assert isinstance(commands, list)
     p = None
@@ -86,7 +102,7 @@ def run_command(commands, args, cwd=None, verbose=False, hide_stderr=False, env=
             break
         except EnvironmentError:
             e = sys.exc_info()[1]
-            if e.errno == errno.ENOENT:
+            if e.errno == errno.ENOENT:  # type: ignore
                 continue
             if verbose:
                 print("unable to run %s" % dispcmd)
@@ -96,9 +112,10 @@ def run_command(commands, args, cwd=None, verbose=False, hide_stderr=False, env=
         if verbose:
             print("unable to find command, tried %s" % (commands,))
         return None, None
-    stdout = p.communicate()[0].strip()
+    stdout: str
+    stdout = p.communicate()[0].strip()  # type: ignore
     if sys.version_info[0] >= 3:
-        stdout = stdout.decode()
+        stdout = stdout.decode()  # type: ignore
     if p.returncode != 0:
         if verbose:
             print("unable to run %s (error)" % dispcmd)
@@ -107,7 +124,7 @@ def run_command(commands, args, cwd=None, verbose=False, hide_stderr=False, env=
     return stdout, p.returncode
 
 
-def versions_from_parentdir(parentdir_prefix, root, verbose):
+def versions_from_parentdir(parentdir_prefix: str, root: str, verbose: bool) -> Dict:
     """Try to determine the version from the parent directory name.
 
     Source tarballs conventionally unpack into a directory that includes both
@@ -139,7 +156,7 @@ def versions_from_parentdir(parentdir_prefix, root, verbose):
 
 
 @register_vcs_handler("git", "get_keywords")
-def git_get_keywords(versionfile_abs):
+def git_get_keywords(versionfile_abs: str) -> Dict:
     """Extract version information from the given file."""
     # the code embedded in _version.py can just fetch the value of these
     # keywords. When used from setup.py, we don't want to import _version.py,
@@ -168,7 +185,9 @@ def git_get_keywords(versionfile_abs):
 
 
 @register_vcs_handler("git", "keywords")
-def git_versions_from_keywords(keywords, tag_prefix, verbose):
+def git_versions_from_keywords(
+    keywords: Dict[str, str], tag_prefix: str, verbose: bool
+) -> Dict:
     """Get version information from git keywords."""
     if not keywords:
         raise NotThisMethod("no keywords at all, weird")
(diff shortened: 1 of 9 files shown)


modin/utils.py
```diff
@@ -258,30 +297,31 @@ def _replace_doc(
     target_doc = target_obj.__doc__ or ""
     overwrite = overwrite or not target_doc
     doc = source_doc if overwrite else target_doc
-    apilink = [apilink] if not isinstance(apilink, list) and apilink else apilink
+    apilink_l: Optional[List[str]]
+    apilink_l = [apilink] if not isinstance(apilink, list) and apilink else apilink  # type: ignore
```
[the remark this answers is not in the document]
noloerino: The logic for the isinstance check is a little weird:
if (apilink is not a list) and (apilink is truth-y): apilink_l = apilink
else: apilink = [apilink]

The code as written means that if `apilink` is `None`, it would fall into the second case, meaning`apilink_l = None`. This is checked by an if statement later before the variable is read.

The `type: ignore` is there because without it, the "else" branch can be taken if `apilink` is the empty string, making the type of the expression `Union[List[str], str, None]`. I'm not sure if there's a cleaner way to rewrite this logic to prevent that.

modin/utils.py
```diff
@@ -258,30 +297,31 @@ def _replace_doc(
...
 
     if parent_cls and not attr_name:
         if isinstance(target_obj, property):
-            attr_name = target_obj.fget.__name__
+            attr_name = target_obj.fget.__name__  # type: ignore
```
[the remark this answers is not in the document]
noloerino: `modin/utils.py:305:25: error: Item "None" of "Optional[Callable[[Any], Any]]" has no attribute "__name__"  [union-attr]`

It seems like `fget` is an optional field of `property`: https://docs.python.org/3/library/functions.html#property

modin/utils.py
```diff
@@ -258,30 +297,31 @@ def _replace_doc(
...
         elif isinstance(target_obj, (staticmethod, classmethod)):
             attr_name = target_obj.__func__.__name__
         else:
-            attr_name = target_obj.__name__
+            attr_name = target_obj.__name__  # type: ignore
```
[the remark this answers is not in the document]
noloerino: `modin/utils.py:309:25: error: "object" has no attribute "__name__"; maybe "__ne__" or "__new__"?  [attr-defined]`

This branch seems to assume that `target_obj` is `callable`, so maybe it would be better to add an assertion to that effect here.

modin/utils.py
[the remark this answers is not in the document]
pyrito: What if you made it `Optional[Union[List[str], str]]` ?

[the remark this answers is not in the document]
noloerino: The code later on expects a list when iterating over `apilink_l` (though now that I think about it, since strings also act like lists it might typecheck fine). I'll fix this by moving the declaration of `apilink_l` further down so it can't be None.

modin/utils.py
```diff
@@ -15,25 +15,59 @@
...
+from typing import (
+    Any,
+    Callable,
+    Dict,
+    List,
+    Mapping,
+    Optional,
+    Union,
+    TypeVar,
+    Protocol,
```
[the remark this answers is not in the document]
noloerino: Fixed by importing from `typing_extensions` for earlier versions (prior to 3.8).

modin/error_message.py
```diff
@@ -11,18 +11,19 @@
...
 from modin.logging import get_logger
 from modin.utils import get_current_execution
 
 
 class ErrorMessage(object):
     # Only print full ``default to pandas`` warning one time.
-    printed_default_to_pandas = False
-    printed_warnings = set()
+    printed_default_to_pandas: bool = False
+    printed_warnings: Set[int] = set()  # Set of hashes of printed warnings
```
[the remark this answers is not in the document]
noloerino: mypy complains if I don't set it for the `set()`, but I've now removed the annotation for the boolean.

modin/utils.py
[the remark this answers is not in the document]
noloerino: It seems like not all `callable`s have this field: https://stackoverflow.com/questions/33576932/do-all-callables-have-name
I'll leave this ignore in for now.

modin/_version.py
```diff
@@ -86,7 +102,7 @@ def run_command(commands, args, cwd=None, verbose=False, hide_stderr=False, env=
             break
         except EnvironmentError:
             e = sys.exc_info()[1]
-            if e.errno == errno.ENOENT:
+            if e.errno == errno.ENOENT:  # type: ignore
```
mvashishtha, line 105: Why ignore?

modin/_version.py
```diff
@@ -96,9 +112,10 @@ def run_command(commands, args, cwd=None, verbose=False, hide_stderr=False, env=
         if verbose:
             print("unable to find command, tried %s" % (commands,))
         return None, None
-    stdout = p.communicate()[0].strip()
+    stdout: str
+    stdout = p.communicate()[0].strip()  # type: ignore
```
mvashishtha, line 116: Why ignore?

modin/_version.py
```diff
@@ -96,9 +112,10 @@ def run_command(commands, args, cwd=None, verbose=False, hide_stderr=False, env=
...
     if sys.version_info[0] >= 3:
-        stdout = stdout.decode()
+        stdout = stdout.decode()  # type: ignore
```
mvashishtha, line 118: Why ignore?

modin/utils.py
```diff
@@ -258,30 +299,30 @@ def _replace_doc(
...
         source_doc.strip()
         and apilink
         and "pandas API documentation for " not in target_doc
         and (not (attr_name or "").startswith("_"))
     ):
-        links = [None] * len(apilink)
-        for i, link in enumerate(apilink):
+        apilink_l = [apilink] if not isinstance(apilink, list) and apilink else apilink
+        links = []
+        for i, link in enumerate(apilink_l):
```
mvashishtha, line 319: Is this loop using `i` any more?

modin/utils.py
```diff
@@ -296,19 +337,25 @@ def _replace_doc(
...
+        assert attr_name is not None
         setattr(
             parent_cls,
             attr_name,
             property(target_obj.fget, target_obj.fset, target_obj.fdel, doc),
         )
     else:
         if overwrite:
-            target_obj.__doc_inherited__ = True
+            target_obj.__doc_inherited__ = True  # type: ignore
```
mvashishtha, line 349: Why ignore?

modin/utils.py
```diff
@@ -342,21 +389,21 @@ def _inherit_docstrings(parent, excluded=[], overwrite_existing=False, apilink=N
...
 
-    def decorator(cls_or_func):
+    def decorator(cls_or_func: Fn) -> Fn:
         if parent not in excluded:
             _replace_doc(parent, cls_or_func, overwrite_existing, apilink)
 
         if not isinstance(cls_or_func, types.FunctionType):
             seen = set()
-            for base in cls_or_func.__mro__:
+            for base in cls_or_func.__mro__:  # type: ignore
```
mvashishtha, line 406: Why ignore?

modin/utils.py
```diff
@@ -472,7 +525,7 @@ def try_cast_to_pandas(obj, squeeze=False):
     return obj
 
 
-def wrap_into_list(*args, skipna=True):
+def wrap_into_list(*args: Any, skipna: bool = True) -> List[Any]:
```
mvashishtha, line 528: Should `*args*` be a `Sequence` as in other methods?

mvashishtha (commented):
Thanks @noloerino. Left minor comments and some questions about ignores.

modin/_version.py
noloerino: Using `sys.exc_info()` loses type information, and it seems like `errno` isn't a guaranteed field.

modin/_version.py
noloerino: `stdout` obtained from `Popen.communicate` is a bytes object in Python 3+ (and I think a string in Python 2). The type of this variable is going from bytes to string, which should probably be fixed by adding a new variable instead.

Since we don't support Python 2, should we just remove this if statement?

modin/_version.py
noloerino: This ignore shouldn't be here; I didn't realize `bytes` also had a `strip` method (`stdout` should be annotated as `bytes` instead).

modin/utils.py
noloerino: Looks like it's not. I'll remove the `enumerate`.

modin/utils.py
noloerino: Without this, we run into `modin/utils.py:349:13: error: "object" has no attribute "__doc_inherited__"  [attr-defined]`

modin/utils.py
noloerino: `__mro__` is only defined for classes, not functions, and is gated here by the `isinstance(cls_or_func, types.FunctionType)` check. The type annotation isn't strict enough for mypy to recognize that the variable must be a class in this branch, and I'm not sure what the correct annotation would be to get `__mro__` as a property.

modin/utils.py
noloerino: mypy wants `*args` and `**kwargs` annotated with the type of the elements, rather than the type of the container (https://mypy.readthedocs.io/en/stable/getting_started.html#more-function-signatures).

modin/_version.py
mvashishtha: Yep, I think we can remove it.

noloerino:
I've made the `type: ignore` directives suppress specific warnings (`union-attr` includes the case where a variable can be `None`, and `attr-defined` is for cases where a field may not be defined).

mvashishtha (approved):
> I've made the type: ignore directives suppress specific warnings (union-attr includes the case where a variable can be None, and attr-defined is for cases where a field may not be defined).

@noloerino great! thank you.

mypy.ini
```diff
@@ -17,5 +17,11 @@ disallow_any_generics=False
...
-; files = **/modin/*.py
-files = **/modin/config/*.py, **/modin/logging/*.py
+; files = modin/*.py
+files = 
+        modin/config/*.py,
+        modin/logging/*.py,
+        modin/_version.py,
+        modin/db_conn.py,
+        modin/error_message.py,
+        modin/utils.py
```
pyrito, line 27: Any reason why we're skipping `apply_license_header.py` so that we can just do `modin/*.py` here?

noloerino: I only annotated files that get imported elsewhere in the codebaes, and `modin/conftest.py` shows a lot of errors for missing annotations. I can make mypy ignore conftest.py and add simple annotations for main.py and __init__.py to make it easier.

mvashishtha (approved):

mvashishtha:
@noloerino thank you!

Issue #4966, opened 2022-09-13
BUG: calling pd.to_timedelta on modin series gives TimedeltaIndex instead of Series


mvashishtha:
### System information
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: macOS Monterey
- **Modin version** (`modin.__version__`): b09dd424e95830e6a19a631741eb8381b9020ee5
- **Python version**: 3.10.4
- **Code we can use to reproduce**:

```python
import modin.pandas as pd
import pandas

duration = pd.Series([1])
print(pd.to_timedelta(duration), '\n\n')

duration_pandas = pandas.Series([1])
print(pandas.to_timedelta(duration_pandas))
```

### Describe the problem
pandas documents the correct return types here: https://pandas.pydata.org/docs/reference/api/pandas.to_timedelta.html

If input is a series, result should be another series.

pyrito:
@mvashishtha this is an interesting bug. It seems like we just default to the pandas implementation which obviously doesn't know how to handle Modin series. It looks like we need to implement our own version of `to_timedelta` and if the input it a Series we change the dtype of the df? Do we need to do any explicit casting on each element or is it enough to just change the dtype?

mvashishtha:
#5028 fixed this bug. Thanks @Billy2551

PR #5022, opened 2022-09-23
FIX-#4966: Fix to_timedelta to return Series instead of TimedeltaIndex

Billy2551:
## What do these changes do?

Added `to_timedelta` implementation to `general.py` instead of importing and using pandas version, which previously returned the wrong output when provided a Modin Series.

Added test to confirm functionality of `to_timedelta` on Modin Series.

- [x] Resolves #? 
- [x] tests added and passing

### DIFF
--- a/modin/pandas/general.py
+++ b/modin/pandas/general.py
@@ -759,3 +759,11 @@ def _determine_name(objs: Iterable[BaseQueryCompiler], axis: Union[int, str]):
         return list(names[0]) if is_list_like(names[0]) else [names[0]]
     else:
         return None
+
+
+def to_timedelta(arg, unit=None, errors="raise"):
+    if isinstance(arg, Series):
+        arg = arg._to_pandas()
+    result = pandas.to_timedelta(arg, unit=unit, errors=errors)
+    if isinstance(result, pandas.Series):
+        return Series(result)
--- a/README.md
+++ b/README.md
@@ -204,6 +204,7 @@ to be modular so we can plug in different components as they develop and improve
 - [10-min Quickstart Guide](https://modin.readthedocs.io/en/latest/getting_started/quickstart.html)
 - [Examples and Tutorials](https://modin.readthedocs.io/en/latest/getting_started/examples.html)
 - [Videos and Blogposts](https://modin.readthedocs.io/en/latest/getting_started/examples.html#talks-podcasts)
+- [Benchmarking Modin](https://modin.readthedocs.io/en/latest/usage_guide/benchmarking.html)
 
 #### Modin Community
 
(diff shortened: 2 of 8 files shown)


modin/pandas/general.py
```diff
@@ -759,3 +759,11 @@ def get_names(obj):
...
     else:
         return None
+
+
+def to_timedelta(arg, unit=None, errors="raise"):
+    if isinstance(arg, Series):
+        arg = arg._to_pandas()
+    result = pandas.to_timedelta(arg, unit=unit, errors=errors)
+    if isinstance(result, pandas.Series):
+        return Series(result)
```
vnlitvinov, line 769: What would happen if `result` is _not_ a `pandas.Series`?.. this function would return `None` then, which is weird.
It feels like it should be an assert instead, something like
```suggestion
    assert isinstance(result, pandas.Series), f"Expected pandas.Series, got {type(result)}"
    return Series(result)
```

vnlitvinov (changes requested):

modin/pandas/general.py
```diff
@@ -759,3 +759,11 @@ def get_names(obj):
         return list(names[0]) if is_list_like(names[0]) else [names[0]]
     else:
         return None
+
+
+def to_timedelta(arg, unit=None, errors="raise"):
+    if isinstance(arg, Series):
+        arg = arg._to_pandas()
```
mvashishtha, line 766: I think we can implement this case, where the arg is a `Series`, as a `map` without materializing the `arg`. Something like `s._query_compiler._modin_frame.map(lambda s: pandas.to_timedelta(s.squeeze(1)).to_frame()).to_pandas()` seem to work. Note the `squeeze` within the partition function because `pandas.to_timedelta` can take a series but can't take a dataframe, and the `to_frame` because (I think) the partitions have to have dataframes internally and can't have series. You may have to add a new query compiler function for this.

pyrito: @mvashishtha that's right, the partitions need to have dataframes internally. We don't treat Series separately. We just treat Series a special case of DataFrames, so we may need to have another query compiler function for this.

modin/pandas/general.py
Billy2551: I believe the output type is dependent on the input type. From pandas documentation:
```
If parsing succeeded. Return type depends on input:
list-like: TimedeltaIndex of timedelta64 dtype
Series: Series of timedelta64 dtype
scalar: Timedelta
```

pyrito (commented):
It also looks like you didn't sign your first commit. You might need to do an interactive rebase and amend a previous commit.

Billy2551:
I messed up the branch by merging instead of rebasing. Will be moving changes to a different PR.

PR #5028, opened 2022-09-23
FIX-#4966: Fix `to_timedelta` to return Series instead of TimedeltaIndex

Billy2551:

## What do these changes do?

Added implementation for `to_timedelta` instead of using pandas version. Previously returned incorrect output when using a Modin Series as input.

Added tests for `to_timedelta` with Modin Series.

- [x] Resolves #? 
- [x] tests added and passing

### DIFF
--- a/docs/release_notes/release_notes-0.16.0.rst
+++ b/docs/release_notes/release_notes-0.16.0.rst
@@ -55,7 +55,7 @@ Key Features and Updates
   * FIX-#4996: Evaluate BenchmarkMode at each function call (#4997)
   * FIX-#4022: Fixed empty data frame with index (#4910)
   * FIX-#4090: Fixed check if the index is trivial (#4936)
-  * FIX-#4966: Fix `to_timedelta` to return Series instead of TimedeltaIndex (#4966)
+  * FIX-#4966: Fix `to_timedelta` to return Series instead of TimedeltaIndex (#5028)
 * Performance enhancements
   * PERF-#4182: Add cell-wise execution for binary ops, fix bin ops for empty dataframes (#4391)
   * PERF-#4288: Improve perf of `groupby.mean` for narrow data (#4591)
--- a/docs/release_notes/release_notes-0.16.0.rst
+++ b/docs/release_notes/release_notes-0.16.0.rst
@@ -55,6 +55,7 @@ Key Features and Updates
   * FIX-#4996: Evaluate BenchmarkMode at each function call (#4997)
   * FIX-#4022: Fixed empty data frame with index (#4910)
   * FIX-#4090: Fixed check if the index is trivial (#4936)
+  * FIX-#4966: Fix `to_timedelta` to return Series instead of TimedeltaIndex (#4966)
 * Performance enhancements
   * PERF-#4182: Add cell-wise execution for binary ops, fix bin ops for empty dataframes (#4391)
   * PERF-#4288: Improve perf of `groupby.mean` for narrow data (#4591)
--- a/modin/core/storage_formats/base/query_compiler.py
+++ b/modin/core/storage_formats/base/query_compiler.py
@@ -1322,7 +1322,7 @@ class BaseQueryCompiler(ClassLogger, abc.ABC):
 
     @doc_utils.add_one_column_warning
     @doc_utils.add_refer_to("to_timedelta")
-    def to_timedelta(self, unit="ns", errors="raise", *args, **kwargs):  # noqa: PR02
+    def to_timedelta(self, unit="ns", errors="raise"):  # noqa: PR02
         """
         Convert argument to timedelta.
 
@@ -1331,10 +1331,6 @@ class BaseQueryCompiler(ClassLogger, abc.ABC):
         unit : str, default: "ns"
             Denotes the unit of the arg for numeric arg. Defaults to "ns".
         errors : {"ignore", "raise", "coerce"}, default: "raise"
-        *args : iterable
-            Serves the compatibility purpose. Does not affect the result.
-        **kwargs : dict
-            Serves the compatibility purpose. Does not affect the result.
 
         Returns
         -------
@@ -1342,7 +1338,7 @@ class BaseQueryCompiler(ClassLogger, abc.ABC):
             New QueryCompiler with converted to timedelta values.
         """
         return SeriesDefault.register(pandas.to_timedelta)(
-            self, unit=unit, errors=errors, *args, **kwargs
+            self, unit=unit, errors=errors
         )
 
     # FIXME: get rid of `**kwargs` parameter (Modin issue #3108).
--- a/modin/core/storage_formats/base/query_compiler.py
+++ b/modin/core/storage_formats/base/query_compiler.py
@@ -1322,7 +1322,7 @@ class BaseQueryCompiler(ClassLogger, abc.ABC):
 
     @doc_utils.add_one_column_warning
     @doc_utils.add_refer_to("to_timedelta")
-    def to_timedelta(self, *args, **kwargs):  # noqa: PR02
+    def to_timedelta(self, unit="ns", errors="raise", *args, **kwargs):  # noqa: PR02
         """
         Convert argument to timedelta.
 
@@ -1341,7 +1341,9 @@ class BaseQueryCompiler(ClassLogger, abc.ABC):
         BaseQueryCompiler
             New QueryCompiler with converted to timedelta values.
         """
-        return SeriesDefault.register(pandas.to_timedelta)(self, *args, **kwargs)
+        return SeriesDefault.register(pandas.to_timedelta)(
+            self, unit=unit, errors=errors, *args, **kwargs
+        )
 
     # FIXME: get rid of `**kwargs` parameter (Modin issue #3108).
     @doc_utils.add_one_column_warning
--- a/modin/core/storage_formats/base/query_compiler.py
+++ b/modin/core/storage_formats/base/query_compiler.py
@@ -1328,9 +1328,9 @@ class BaseQueryCompiler(ClassLogger, abc.ABC):
 
         Parameters
         ----------
-        unit : str
+        unit : str, default: "ns"
             Denotes the unit of the arg for numeric arg. Defaults to "ns".
-        errors : {"ignore", "raise", "coerce"}
+        errors : {"ignore", "raise", "coerce"}, default: "raise"
         *args : iterable
             Serves the compatibility purpose. Does not affect the result.
         **kwargs : dict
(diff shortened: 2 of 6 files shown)


modin/core/storage_formats/pandas/query_compiler.py
```diff
@@ -1381,6 +1381,11 @@ def stack(self, level, dropna):
             pandas.to_numeric(df.squeeze(axis=1), *args, **kwargs)
         )
     )
+    to_timedelta = Map.register(
```
mvashishtha, line 1384: I think we can pass `dtypes="timedelta64[ns]"` because the pandas docs say that the output is "Series of timedelta64 dtype". I can't find any `timedelta64` type besides `timedelta64[ns]`, so I guess that's the type to pass. Setting dtypes ahead of time means that we don't have to figure them out from the partitions later.

modin/pandas/test/test_general.py
```diff
@@ -801,3 +802,9 @@ def test_empty_dataframe():
 def test_empty_series():
     s = pd.Series([])
     pd.to_numeric(s)
+
+
+def test_to_timedelta():
+    # This test case comes from
+    # https://github.com/modin-project/modin/issues/4966
+    eval_general(pd, pandas, lambda lib: lib.to_timedelta(lib.Series([1])))
```
mvashishtha, line 810: We should test float and int data, at least, with enough data to fill a few partitions. `test_data` in the utils has exactly those types. We can adapt the code here to parametrize by the data type-- maybe we don't need a method as complicated as `create_test_series`. https://github.com/modin-project/modin/blob/0a2c0de4451f7e2e8f337a9478d7595473aa348e/modin/pandas/test/test_series.py#L210-L212

pyrito (commented):
@Billy2551 looks like there are some documentation errors. Could you take a look at them?

mvashishtha:
> looks like there are some documentation errors. Could you take a look at them?

@Billy2551 I recommend running locally the docstyle command from the [failed CI job](https://github.com/modin-project/modin/actions/runs/3115012026/jobs/5051480045): 

```shell
  python scripts/doc_checker.py --add-ignore=D101,D102,D103,D105 --disable-numpydoc \
    modin/pandas/dataframe.py modin/pandas/series.py \
    modin/pandas/groupby.py \
    modin/pandas/series_utils.py modin/pandas/general.py \
    modin/pandas/plotting.py modin/pandas/utils.py \
    modin/pandas/iterator.py modin/pandas/indexing.py \
```

modin/core/storage_formats/base/query_compiler.py
```diff
@@ -1320,6 +1320,29 @@ def to_numeric(self, *args, **kwargs):  # noqa: PR02
...
+    @doc_utils.add_refer_to("to_timedelta")
+    def to_timedelta(self, *args, **kwargs):  # noqa: PR02
+        """
+        Convert argument to timedelta.
+
+        Parameters
+        ----------
+        unit : str
+            Denotes the unit of the arg for numeric arg. Defaults to "ns".
+        errors : {"ignore", "raise", "coerce"}
```
vnlitvinov, line 1333: I don't see those arguments in the function signature, it should error out in the docs

modin/core/storage_formats/pandas/query_compiler.py
```diff
@@ -1381,6 +1381,12 @@ def stack(self, level, dropna):
...
+        lambda s, *args, **kwargs: pandas.to_timedelta(
+            s.squeeze(1), *args, **kwargs
```
vnlitvinov, line 1386:
```suggestion
            s.squeeze(axis=1), *args, **kwargs
```
is that right? I would prefer using named arguments for better readability

modin/pandas/general.py
```diff
@@ -759,3 +759,15 @@ def get_names(obj):
...
+@_inherit_docstrings(pandas.to_datetime, apilink="pandas.to_timedelta")
+@enable_logging
+def to_timedelta(arg, unit=None, errors="raise"):  # noqa: PR01, RT01, D200
+    """
+    Convert argument to timedelta.
+    """
+    if isinstance(arg, Series):
+        query_compiler = arg._query_compiler.to_timedelta(unit=unit, errors=errors)
+        return Series(query_compiler=query_compiler)
+    return pandas.to_timedelta(arg, unit=unit, errors=errors)
```
vnlitvinov, line 773: this feels really weird - is there a reason why we should return a pandas object from this method?.. I would prefer to return Modin Series here at all cases

Billy2551: Fixed to convert pandas Series results to modin Series

modin/core/storage_formats/pandas/query_compiler.py
Billy2551: Fixed.

modin/core/storage_formats/base/query_compiler.py
Billy2551: Fixed.

docs/release_notes/release_notes-0.16.0.rst
```diff
@@ -53,6 +53,7 @@ Key Features and Updates
   * FIX-#4597: Refactor Partition handling of func, args, kwargs (#4715)
   * FIX-#4996: Evaluate BenchmarkMode at each function call (#4997)
   * FIX-#4022: Fixed empty data frame with index (#4910)
+  * FIX-#4966: Fix `to_timedelta` to return Series instead of TimedeltaIndex (#4966)
```
pyrito, line 56:
```suggestion
  * FIX-#4966: Fix `to_timedelta` to return Series instead of TimedeltaIndex (#5028 )
```

modin/core/storage_formats/base/query_compiler.py
```diff
@@ -1320,6 +1320,31 @@ def to_numeric(self, *args, **kwargs):  # noqa: PR02
...
+
+        Parameters
+        ----------
+        unit : str, default: "ns"
+            Denotes the unit of the arg for numeric arg. Defaults to "ns".
+        errors : {"ignore", "raise", "coerce"}, default: "raise"
+        *args : iterable
+            Serves the compatibility purpose. Does not affect the result.
+        **kwargs : dict
+            Serves the compatibility purpose. Does not affect the result.
```
mvashishtha, line 1337:
```suggestion
    def to_timedelta(self, unit="ns", errors="raise"):
        """
        Convert argument to timedelta.

        Parameters
        ----------
        unit : str, default: "ns"
            Denotes the unit of the arg for numeric arg. Defaults to "ns".
        errors : {"ignore", "raise", "coerce"}, default: "raise"
```

I don't think pandas supports `*args` or `**kwargs`

modin/core/storage_formats/base/query_compiler.py
```diff
@@ -1320,6 +1320,31 @@ def to_numeric(self, *args, **kwargs):  # noqa: PR02
...
+
+        Returns
+        -------
+        BaseQueryCompiler
+            New QueryCompiler with converted to timedelta values.
+        """
+        return SeriesDefault.register(pandas.to_timedelta)(
+            self, unit=unit, errors=errors, *args, **kwargs
+        )
```
mvashishtha, line 1346:
```suggestion
        return SeriesDefault.register(pandas.to_timedelta)(self, unit=unit, errors=errors)
```

modin/pandas/general.py
```diff
@@ -759,3 +759,18 @@ def get_names(obj):
...
+    """
+    Convert argument to timedelta.
+    """
+    if isinstance(arg, Series):
+        query_compiler = arg._query_compiler.to_timedelta(unit=unit, errors=errors)
+        return Series(query_compiler=query_compiler)
+    result = pandas.to_timedelta(arg, unit=unit, errors=errors)
+    if isinstance(result, pandas.Series):
+        return Series(result)
+    return result
```
mvashishtha, line 776:
```suggestion
    return pandas.to_timedelta(arg, unit=unit, errors=errors)
```

```suggestion
```

According to the documentation, return type is series if and only if the input is a series. So I don't think we could hit this case.

modin/pandas/test/test_general.py
```diff
@@ -801,3 +801,22 @@ def test_empty_dataframe():
 def test_empty_series():
     s = pd.Series([])
     pd.to_numeric(s)
+
+
+@pytest.mark.parametrize("data", test_data_values, ids=test_data_keys)
+def test_to_timedelta(data):
```
mvashishtha, line 807: Could you add another test case where we pass:

- a list-like of ints, like `[1]`
- a list-like of invalid strings, like `['a']`. This should raise `ValueError`.
- a valid scalar, like `1`
- an invalid scalar, like '"a"'

We can do something like `@pytest.mark.parametrize('arg', [[1, 2], ["a"], 1, "a"], ids=["list_of_ints", "list_of_invalid_strings", "scalar", "invalid_scalar"])`

mvashishtha (dismissed):
LGTM @Billy2551 ! Thank you.

pyrito (approved):

modin/pandas/general.py
vnlitvinov, line 773: This is again returning some pandas object, not Modin, is that really what we want?

I think our stance on API is we mimic pandas, but we use our own types everywhere.

modin/pandas/test/test_general.py
```diff
@@ -801,3 +802,31 @@ def test_empty_dataframe():
...
+    else:
+        modin_series = pd.Series(data)
+        pandas_series = pandas.Series(data)
+
+    modin_series = pd.to_timedelta(modin_series)
+    pandas_series = pandas.to_timedelta(pandas_series)
+    df_equals(
+        modin_series.to_frame(name="timedelta"),
+        pandas_series.to_frame(name="timedelta"),
+    )
```
vnlitvinov, line 832: let's use `eval_general()` here, too - it takes care of comparing exceptions if they raise (plus it correctly waits for async Modin calls as they only raise exceptions when we ask for the result):
```suggestion
    def make_frame(lib):
        series = lib.Series(
            next(iter(data.values())) if isinstance(data, dict) else data
        )
        return lib.to_timedelta(series).to_frame(name="timedelta")

    eval_general(pd, pandas, make_frame)

```

vnlitvinov (changes requested):

mvashishtha (dismissed):
@Billy2551 I have a minor comment, but LGTM once @vnlitvinov 's comments are resolved.

mvashishtha (approved):
Thanks @Billy2551

pyrito (approved):

vnlitvinov (approved):
LGTM, thanks @Billy2551!

YarShev (approved):

Issue #5029, opened 2022-09-23
PERF: don't use `_compute_axis_labels_and_lengths` for computing `_row_lengths/_column_widths`


anmyachev:
It will be much faster here to call `length/width` functions on partitions.

I can think of an arbitrarily large multi-index that needs to be serialized/deserialized instead of serializing/deserializing just ints.

PR #5030, opened 2022-09-23
PERF-#5029: Don't use `_compute_axis_labels_and_lengths` for computing `_row_lengths/_column_widths`

anmyachev:

## What do these changes do?

- [ ] Resolves #? 
- [ ] tests added and passing


mvashishtha:
@anmyachev could you give an example where this is useful?

anmyachev:
> @anmyachev could you give an example where this is useful?

I see about twice the speedup relative to the master (comparing best times).
```python
import modin.pandas as pd
import numpy as np
from time import time

arr = np.random.rand(1000000, 10)
index = pd.MultiIndex.from_product(
    [np.random.randint(100, size=125000), ["a", "b", "c", "d"], ["c0", "c1"]]
)
for _ in range(5):
    df = pd.DataFrame(arr, index=index)
    mask = pd.Series([True] * df.shape[0], index=index)
    df = df[mask]
    start = time()
    _ = df.reset_index(drop=True)
    print(f"reset_index: {time()-start}")

```

YarShev:
@anmyachev, I would suggest to check this in a pipeline, i.e. multiple oprations chain, to see if other operations do not slow down.

anmyachev:
> @anmyachev, I would suggest to check this in a pipeline, i.e. multiple oprations chain, to see if other operations do not slow down.

@YarShev this speeds up an internal workload I'm working on, but since I can't show the source, I had to make a synthetic example.

mvashishtha:
@anmyachev I am concerned about the case where partitions shapes and the index are both unknown. Currently, we would do a single remote call to get both index and shapes, but after this PR, we will have to do two separate remote calls: one to get the shapes, then another to get the index. The benefit of this change is, as you mentioned, that we can avoid computing the index in some cases where we'll never need it.

It's hard for me to say without examples which way is better.

Am I missing something? cc @YarShev

YarShev:
@mvashishtha, yes, we are on the same page with you.

vnlitvinov:
I don't get how this speeds things up. To compute the size of a partition (the `partition.length()` call) you'd need to have that partition be ready, don't you? And, if it is ready, getting only its size or its size and index should be costing you the same.

anmyachev:
> I don't get how this speeds things up. To compute the size of a partition (the `partition.length()` call) you'd need to have that partition be ready, don't you? And, if it is ready, getting only its size or its size and index should be costing you the same.

Not exactly the same. We also have to take into account the amount of data being transferred (much more in the case of an index), as well as the time spent on the `append` function, which is called in `get_indices` function.

anmyachev:
@mvashishtha @YarShev @vnlitvinov  I also test the example with 112 cores (on master and `issue5029` branch):

```python
import modin.pandas as pd
import numpy as np
from time import time

arr = np.random.rand(1000000, 10)
index = pd.MultiIndex.from_product(
    [np.random.randint(100, size=125000), ["a", "b", "c", "d"], ["c0", "c1"]]
)
for _ in range(5):
    df = pd.DataFrame(arr, index=index)
    mask = pd.Series([True] * df.shape[0], index=index)
    df = df[mask]
    start = time()
    df._query_compiler._modin_frame.row_lengths
    df._query_compiler.index
    print(f"length+index: {time()-start}\n")
```

The time difference is not noticeable.

vnlitvinov:
> The time difference is not noticeable.

So, I still would like to see performance proof before going and removing some pieces of code introduced for potential optimization (your current change trades off performance gains on suuuuuuuuuuper-huge frame versus performance losses on everything else, so I'd really like to look at the numbers proving that this change actually accomplishes something).

anmyachev:
> > The time difference is not noticeable.
> 
> So, I still would like to see performance proof before going and removing some pieces of code introduced for potential optimization (your current change trades off performance gains on suuuuuuuuuuper-huge frame versus performance losses on everything else, so I'd really like to look at the numbers proving that this change actually accomplishes something).


I've already given two examples, one of which (with `reset_index`) shows the increase perf (0.25 sec vs 0.01 sec), and one that shows that all other cases slow down by a negligible amount of time (0.25 sec vs 0.26 sec). 

Do you have a suggestion for which case to measure next?

vnlitvinov:
Can you show a small table how your benchmarks scale? As of right now, you've showed 1M rows only.

I would like to request a 1M/10M/100M and, if possible, something like 8/16/32/64 cores, too.
I now see the point that you made about not transferring an index if we don't need that, but it is still interesting to see how the things scale.

anmyachev:
> Can you show a small table how your benchmarks scale? As of right now, you've showed 1M rows only.
> 
> I would like to request a 1M/10M/100M and, if possible, something like 8/16/32/64 cores, too. I now see the point that you made about not transferring an index if we don't need that, but it is still interesting to see how the things scale.

Ran the benchmark (with `reset_index`) several times on the master just for comparison.

image

anmyachev:
> > Can you show a small table how your benchmarks scale? As of right now, you've showed 1M rows only.
> > I would like to request a 1M/10M/100M and, if possible, something like 8/16/32/64 cores, too. I now see the point that you made about not transferring an index if we don't need that, but it is still interesting to see how the things scale.
> 
> Ran the benchmark (with `reset_index`) several times on the master just for comparison (all times in sec).
> 
> image



**UPD**
image

anmyachev:
@vnlitvinov I have added the results for another example, please review

dchigarev:
The problem that causes this PR to be controversial is that it tries to optimize a local case while modifying the global one.

The idea of the PR is to optimize one certain case: do not trigger indices extraction from partitions in the `.reset_index(drop=True)` that happens here: https://github.com/modin-project/modin/blob/6440684f3fd3eea477f83dcf77f8d66103b9df95/modin/core/storage_formats/pandas/query_compiler.py#L645-L646
So maybe we could change the implementation of how `.row_lengths` are computed only for this specific case so one wouldn't argue that PR slow-downed a common case?

It's an open question on how to do it pretty. For now, I came up with a dedicated private method for computing row/column frame lengths that can be parametrized of whether or not to compute index caches:
```python
def _axis_lengths(self, axis, populate_index_cache=True):
    ...
```
We may then call this private method with the respective parameter from `QC.reset_index` in order to recieve row lengths without extra computations for indices.
```diff
@@ -646,7 +646,13 @@ def reset_index(self, **kwargs):
- pandas.RangeIndex(sum(new_self._modin_frame.row_lengths))
+ pandas.RangeIndex(sum(new_self._modin_frame._axis_lengths(axis=0, populate_index_cache=False)))
```
I don't really like this private method call in the QC, but maybe we could find some other ways to do the same.

I did a separate branch with a draft of that idea here, feel free to comment: https://github.com/dchigarev/modin/commit/e13f1528f55529452248bb676854bbae7bd6227a

YarShev:
I think [the table](https://github.com/modin-project/modin/pull/5030#issuecomment-1273441959) above proves the necessity of having the changes merged. If no one has any comments, I will go ahead and merge the changes today evening.

docs/release_notes/release_notes-0.16.0.rst
```diff
@@ -84,6 +84,7 @@ Key Features and Updates
   * PERF-#4870: Avoid index materialization in `__getattribute__` and `__getitem__` (4911)
   * PERF-#4886: Use lazy index and columns evaluation in `query` method (#4887)
   * PERF-#4866: `iloc` function that used in `partition.mask` should be serialized only once (#4901)
+  * PERF-#5029: Don't use `_compute_axis_labels_and_lengths` for computing `_row_lengths/_column_widths` (#5030)
```
YarShev, line 87:
```suggestion
```

YarShev (approved):

Issue #5034, opened 2022-09-26
BUG: `df.get()` with list throws unhashable type error


pyrito:
### Modin version checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the latest released version of Modin.

- [ ] I have confirmed this bug exists on the main branch of Modin. (In order to do this you can follow [this guide](https://modin.readthedocs.io/en/stable/getting_started/installation.html#installing-from-the-github-master-branch).)

### Reproducible Example

```python
import modin.pandas as pd
df = pd.DataFrame({"col0": [0,1], "col1":[1,2]})
pd.get(["col0"]) # This throws an error
```

### Issue Description

Trying to call `get` with a list of column names throws an "unhashable type" error. 

### Expected Behavior

Output should return a DataFrame with the specified columns.

### Error Logs

<details>

```python-traceback

TypeError                                 Traceback (most recent call last)
Input In [7], in <cell line: 1>()
----> 1 df.get(["col0"])

File ~/opt/anaconda3/envs/modin/lib/python3.9/site-packages/modin/logging/logger_metaclass.py:68, in logger_class_wrapper.<locals>.log_wrap(*args, **kwargs)
     66     logger.info(f"END::PANDAS-API::{class_name}.{method_name}")
     67     return result
---> 68 return method(*args, **kwargs)

File ~/opt/anaconda3/envs/modin/lib/python3.9/site-packages/modin/pandas/base.py:1646, in BasePandasDataset.get(self, key, default)
   1642 def get(self, key, default=None):  # noqa: PR01, RT01, D200
   1643     """
   1644     Get item from object for given key.
   1645     """
-> 1646     if key in self.keys():
   1647         return self.__getitem__(key)
   1648     else:

File ~/opt/anaconda3/envs/modin/lib/python3.9/site-packages/pandas/core/indexes/base.py:5022, in Index.__contains__(self, key)
   4987 def __contains__(self, key: Any) -> bool:
   4988     """
   4989     Return a boolean indicating whether the provided key is in the index.
   4990 
   (...)
   5020     False
   5021     """
-> 5022     hash(key)
   5023     try:
   5024         return key in self._engine

TypeError: unhashable type: 'list'

```

</details>

### Installed Versions

<details>

INSTALLED VERSIONS
------------------
commit           : 138a95401473cd96d5d7456d897b9d087bc45b93
python           : 3.9.12.final.0
python-bits      : 64
OS               : Darwin
[...]

PR #5035, opened 2022-09-26
FIX-#5034: Handle lists in `df.get()`

pyrito:

## What do these changes do?

Matching pandas behavior here.

- [ ] Resolves #? 
- [ ] tests added and passing


modin/pandas/base.py
```diff
@@ -1527,9 +1527,9 @@ def get(self, key, default=None):  # noqa: PR01, RT01, D200
         """
         Get item from object for given key.
         """
-        if key in self.keys():
+        try:
             return self.__getitem__(key)
-        else:
+        except (KeyError, ValueError, IndexError):
```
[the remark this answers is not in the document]
pyrito: I followed: https://github.com/pandas-dev/pandas/blob/v1.5.0/pandas/core/generic.py#L4268-L4318

It's ugly, but I can't think of a better way, especially for pandas concordance.

modin/pandas/test/test_general.py
```diff
@@ -801,3 +803,9 @@ def test_empty_dataframe():
 def test_empty_series():
     s = pd.Series([])
     pd.to_numeric(s)
+
+
+def test_get():
```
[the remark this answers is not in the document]
pyrito: @mvashishtha could you elaborate here? Not sure what you mean 😄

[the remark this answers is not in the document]
mvashishtha: I would like to do something like `@pytest.mark.parametrize("key",  [["col1"], "col1", "col2"], ids=['valid_list_of_string', 'valid_string', invalid_string'])`

modin/pandas/test/test_general.py
```diff
@@ -801,3 +803,13 @@ def test_empty_dataframe():
 def test_empty_series():
     s = pd.Series([])
     pd.to_numeric(s)
+
+
+@pytest.mark.parametrize(
+    "key",
+    [["col1"], "col1", "col2"],
+    ids=["valid_list_of_string", "valid_string", "invalid_string"],
+)
```
mvashishtha, line 812:
```suggestion
@pytest.mark.parametrize(
    "key",
    [["col1"], "col1", "col2", ["col2"], ["col1", "col2"], (), []],
    ids=["valid_list_of_string", "valid_string", "invalid_string", "invalid_list", "list_of_valid_and_invalid_string", "tuple", "empty_list"],
)
@pytest.mark.parametrize("default,pass_default", [(None, False), (3, True)])
```

modin/pandas/test/test_general.py
```diff
@@ -801,3 +803,13 @@ def test_empty_dataframe():
...
+def test_get(key):
+    modin_df, pandas_df = create_test_dfs({"col0": [0, 1]})
+    eval_general(modin_df, pandas_df, lambda df: df.get(key))
```
mvashishtha, line 815:
```suggestion
def test_get(key, default, pass_default):
    modin_df, pandas_df = create_test_dfs({"col0": [0, 1]})
    if pass_default:
      default_kwargs = {'default': default}
    else:
      default_kwargs = {}
    eval_general(modin_df, pandas_df, lambda df: df.get(key, **default_kwargs))
```

modin/pandas/base.py
```diff
@@ -1527,10 +1527,14 @@ def get(self, key, default=None):  # noqa: PR01, RT01, D200
         """
         Get item from object for given key.
         """
-        if key in self.keys():
-            return self.__getitem__(key)
-        else:
-            return default
+
+        def _get(key, list_like):
```
mvashishtha, line 1531: I'd rather not have this inner function if we can avoid it. Why not use the try-except that pandas uses?

pyrito: There was a CI test failure deep in `take_2d_positional` for a different exception when I just passed in a non-list key.

pyrito: No - it was only our implementation.

pyrito: Waiting on implementation for: https://github.com/modin-project/modin/issues/5042

modin/pandas/base.py
```diff
@@ -1527,10 +1527,14 @@ def get(self, key, default=None):  # noqa: PR01, RT01, D200
...
+            if all(k in self.keys() for k in key):
+                return self.__getitem__(key if list_like else key[0])
```
dchigarev, line 1533:
```suggestion
                return self.__getitem__(key)
```
Why do we need this `if is_list_like` check? I believe `.__getitem__` handles lists pretty well:
```python
import modin.pandas as pd

df = pd.DataFrame({"col0": [1, 2, 3]})
print(df.__getitem__(["col0"]))
```

pyrito: @dchigarev thanks for the review. Yes, the check is actually not needed. After @mvashishtha merged in https://github.com/modin-project/modin/pull/5048, I should be able to write this in a cleaner way.

modin/pandas/test/test_general.py
```diff
@@ -801,3 +803,13 @@ def test_empty_dataframe():
...
     pd.to_numeric(s)
+
+
+@pytest.mark.parametrize(
+    "key",
+    [["col1"], "col1", "col2"],
+    ids=["valid_list_of_string", "valid_string", "invalid_string"],
+)
+def test_get(key):
+    modin_df, pandas_df = create_test_dfs({"col0": [0, 1]})
```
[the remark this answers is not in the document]
pyrito: Thanks! Good catch. I've made changes accordingly.

modin/pandas/base.py
pyrito: Changes should be made now.

mvashishtha (dismissed):
Thanks @pyrito

modin/pandas/base.py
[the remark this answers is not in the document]
vnlitvinov: should we follow that code even more and write `return self[key]`?

vnlitvinov (commented):
LGTM!

vnlitvinov (approved):

PR #5047, opened 2022-09-27
port indexing, reindex asvs from pandas

jbrockmendel:
## What do these changes do?

- [ ] Resolves #? 
- [ ] tests added and passing

### DIFF
--- a/asv_bench/benchmarks/pandas/indexing.py
+++ b/asv_bench/benchmarks/pandas/indexing.py
@@ -1,0 +1,296 @@
+"""
+These benchmarks are for Series and DataFrame indexing methods.  For the
+lower-level methods directly on Index and subclasses, see index_object.py,
+indexing_engine.py, and index_cached.py
+"""
+import warnings
+
+import numpy as np
+
+from modin.pandas import (
+    DataFrame,
+    Float64Index,
+    Int64Index,
+    MultiIndex,
+    Series,
+    UInt64Index,
+    concat,
+    date_range,
+    period_range,
+)
+
+import pandas._testing as tm
+
+
+class NumericSeriesIndexing:
+
+    params = [
+        (Int64Index, UInt64Index, Float64Index),
+        ("unique_monotonic_inc", "nonunique_monotonic_inc"),
+    ]
+    param_names = ["index_dtype", "index_structure"]
+
+    def setup(self, index, index_structure):
+        N = 10**6
+        indices = {
+            "unique_monotonic_inc": index(range(N)),
+            "nonunique_monotonic_inc": index(
+                list(range(55)) + [54] + list(range(55, N - 1))
+            ),
+        }
+        self.data = Series(np.random.rand(N), index=indices[index_structure])
+        self.array = np.arange(10000)
+        self.array_list = self.array.tolist()
+
+    def time_getitem_scalar(self, index, index_structure):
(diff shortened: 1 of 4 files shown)


asv_bench/benchmarks/pandas/indexing.py
```diff
@@ -0,0 +1,296 @@
...
+
+
+class AssignTimeseriesIndex:
+    def setup(self):
+        N = 100000
+        idx = date_range("1/1/2000", periods=N, freq="H")
+        self.df = DataFrame(np.random.randn(N, 1), columns=["A"], index=idx)
+
+    def time_frame_assign_timeseries_index(self):
+        self.df["date"] = self.df.index
```
Garra1980, line 263: Is this somehow particularly important to track?

asv_bench/benchmarks/pandas/indexing.py
```diff
@@ -0,0 +1,296 @@
...
+    UInt64Index,
+    concat,
+    date_range,
+    period_range,
+)
+
+import pandas._testing as tm
+
+
+class NumericSeriesIndexing:
```
Garra1980, line 25: As far as I understand we have some of these already, let's leave only really valuable ones

asv_bench/benchmarks/pandas/indexing.py
```diff
@@ -0,0 +1,296 @@
...
+        indexes = {
+            "int": Int64Index(np.arange(N)),
+            "datetime": date_range("2011-01-01", freq="S", periods=N),
+        }
+        index = indexes[index]
+        self.s = Series(np.random.rand(N), index=index)
+        self.indexer = np.random.randint(0, N, size=N)
+
+    def time_take(self, index):
+        self.s.take(self.indexer)
```
Garra1980, line 138: Let's drop this if Modin implementation uses loc/iloc under the hood

asv_bench/benchmarks/pandas/indexing.py
```diff
@@ -0,0 +1,296 @@
...
+        self.data.loc[[800000]]
+
+    def time_loc_scalar(self, index, index_structure):
+        self.data.loc[800000]
+
+    def time_loc_slice(self, index, index_structure):
+        self.data.loc[:800000]
+
+
+class DataFrameNumericIndexing:
```
Garra1980, line 85: Same as comment for NumericSeriesIndexing

asv_bench/benchmarks/pandas/indexing.py
jbrockmendel: In essentially every case the answer is that "pandas implemented this for a reason"

asv_bench/benchmarks/pandas/indexing.py
jbrockmendel: Please be more specific.

Garra1980: My concern is only to avoid benchmarking duplication

Garra1980: I let @YarShev and @anmyachev decide whether we can/should skip some cases

YarShev: I would suggest to integrate new tests directly into our asv benchmarking system so we can see the difference and new added cases.

PR #5049, opened 2022-09-28
FIX-#4691: Fix binary operations between virtual partitions.

### LINKED ISSUE #4691: BUG: Can't do binary operations between dataframes with virtual partitions

mvashishtha:

## What do these changes do?

FIX-#4691: Fix binary operations between virtual partitions.

- [x] Resolves #4691
- [x] tests added and passing

### DIFF
--- a/modin/core/dataframe/pandas/partitioning/partition_manager.py
+++ b/modin/core/dataframe/pandas/partitioning/partition_manager.py
@@ -1288,13 +1288,27 @@ class PandasDataframePartitionManager(ClassLogger, ABC):
         func = cls.preprocess_func(func)
 
         def get_right_block(right_partitions, row_idx, col_idx):
-            blocks = right_partitions[row_idx][col_idx].list_of_blocks
-            # TODO Resolve this assertion as a part of #4691, because the current implementation assumes
-            # that partition contains only 1 block.
-            assert (
-                len(blocks) == 1
-            ), f"Implementation assumes that partition contains only 1 block, but {len(blocks)} recieved."
-            return blocks[0]
+            partition = right_partitions[row_idx][col_idx]
+            blocks = partition.list_of_blocks
+            """
+            NOTE:
+            Currently we do one remote call per right virtual partition to
+            materialize the partitions' blocks, then another remote call to do
+            the n_ary operation. we could get better performance if we
+            assembled the other partition within the remote `apply` call, by
+            passing the partition in as `other_axis_partition`. However,
+            passing `other_axis_partition` requires some extra care that would
+            complicate the code quite a bit:
+            - block partitions don't know how to deal with `other_axis_partition`
+            - the right axis partition's axis could be different from the axis
+              of the corresponding left partition
+            - there can be multiple other_axis_partition because this is an n-ary
+              operation and n can be > 2.
+            So for now just do the materialization in a separate remote step.
+            """
+            if len(blocks) > 1:
+                partition.force_materialization()
+            return partition.list_of_blocks[0]
 
         return np.array(
             [
--- a/modin/core/dataframe/pandas/partitioning/partition_manager.py
+++ b/modin/core/dataframe/pandas/partitioning/partition_manager.py
@@ -1308,6 +1308,7 @@ class PandasDataframePartitionManager(ClassLogger, ABC):
             """
             if len(blocks) > 1:
                 partition.force_materialization()
+            assert len(partition.list_of_blocks) == 1
             return partition.list_of_blocks[0]
 
         return np.array(
--- a/modin/pandas/test/dataframe/test_binary.py
+++ b/modin/pandas/test/dataframe/test_binary.py
@@ -16,6 +16,9 @@ import pandas
 import matplotlib
 import modin.pandas as pd
 
+from modin.core.dataframe.pandas.partitioning.axis_partition import (
+    PandasDataframeAxisPartition,
+)
 from modin.pandas.test.utils import (
     df_equals,
     test_data_values,
@@ -25,7 +28,7 @@ from modin.pandas.test.utils import (
     create_test_dfs,
     default_to_pandas_ignore_string,
 )
-from modin.config import NPartitions
+from modin.config import Engine, NPartitions
 from modin.test.test_utils import warns_that_defaulting_to_pandas
 
 NPartitions.put(4)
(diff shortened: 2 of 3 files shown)


noloerino:
Thanks for fixing this. Since this only affects the case where a binary operator RHS has more than one partition, the only cases where there's extra materialization should be in cases where modin would previously error, right?

mvashishtha:
>  the only cases where there's extra materialization should be in cases where modin would previously error, right?

@noloerino I think that's right. Also we can reuse the materialized data in future operations that use the RHS.

modin/pandas/test/dataframe/test_binary.py
```diff
@@ -151,6 +154,32 @@ def test_comparison(data, op, other):
     )
 
 
+@pytest.mark.skipif(
+    Engine.get() not in ("Ray", "Dask"),
+    reason="Modin on this engine doesn't create virtual partitions.",
+)
+@pytest.mark.parametrize(
+    "left_virtual,right_virtual", [(True, False), (False, True), (True, True)]
+)
```
vnlitvinov, line 163: Maybe we should just parametrize `left_virtual` and `right_virtual` independently? That way we would also check both non-virtual left and right, which might be an extra test but still nice to have

modin/core/dataframe/pandas/partitioning/partition_manager.py
```diff
@@ -1288,13 +1288,27 @@ def n_ary_operation(cls, left, func, right: list):
...
+            - block partitions don't know how to deal with `other_axis_partition`
+            - the right axis partition's axis could be different from the axis
+              of the corresponding left partition
+            - there can be multiple other_axis_partition because this is an n-ary
+              operation and n can be > 2.
+            So for now just do the materialization in a separate remote step.
+            """
+            if len(blocks) > 1:
+                partition.force_materialization()
+            return partition.list_of_blocks[0]
```
vnlitvinov, line 1311: Should we move the `assert len(partition.list_of_clocks) == 1` here instead of just deleting it?

modin/pandas/test/dataframe/test_binary.py
mvashishtha: I think we shouldn't check non-virtual left and right. The ordinary binary operations test should definitely cover that, and I don't want to add a redundant test.

vnlitvinov (approved):
LGTM!

modin/core/dataframe/pandas/partitioning/partition_manager.py
```diff
@@ -1288,13 +1288,28 @@ def n_ary_operation(cls, left, func, right: list):
...
-            # TODO Resolve this assertion as a part of #4691, because the current implementation assumes
-            # that partition contains only 1 block.
-            assert (
-                len(blocks) == 1
-            ), f"Implementation assumes that partition contains only 1 block, but {len(blocks)} recieved."
-            return blocks[0]
+            partition = right_partitions[row_idx][col_idx]
+            blocks = partition.list_of_blocks
+            """
+            NOTE:
```
YarShev, line 1294: Shouldn't we do the aligning of two modin dfs in `_copartition` method?

mvashishtha: I wouldn't refer to this step as "aligning." `_copartition` aligns the partitions so that multiple modin dataframes have the same number of partitions and all the partitions at one coordinate have the same indices along certain axes. This method is solving a different problem of getting a single physical block for binary operations. Also, some methods like `concat` will call `_copartition` but don't need the resulting partitions to have a single block each.

vnlitvinov (approved):
LGTM (re-approving after master merge)

YarShev (approved):

PR #5021, opened 2022-09-22
BENCH: add some cases for `join` and `merge` ops from pandas

jbrockmendel:
Broken off from #4988

Closes #5111

PR #5084, opened 2022-10-04
TEST-#5083: Add `merge` op with categorical data

### LINKED ISSUE #5083: BENCH: add merge op with categorical data

anmyachev:

## What do these changes do?

- [x] Resolves #5083 
- [ ] tests added and passing

### DIFF
--- a/asv_bench/benchmarks/benchmarks.py
+++ b/asv_bench/benchmarks/benchmarks.py
@@ -148,6 +148,45 @@ class TimeMerge:
         )
 
 
+class TimeMergeCategoricals:
+    param_names = ["shapes", "data_type"]
+    params = [
+        get_benchmark_shapes("MergeCategoricals"),
+        ["object", "category"],
+    ]
+
+    def setup(self, shapes, data_type):
+        assert len(shapes) == 2
+        assert shapes[1] == 2
+        size = (shapes[0],)
+        self.left = IMPL.DataFrame(
+            {
+                "X": np.random.choice(range(0, 10), size=size),
+                "Y": np.random.choice(["one", "two", "three"], size=size),
+            }
+        )
+
+        self.right = IMPL.DataFrame(
+            {
+                "X": np.random.choice(range(0, 10), size=size),
+                "Z": np.random.choice(["jjj", "kkk", "sss"], size=size),
+            }
+        )
+
+        if data_type == "category":
+            self.left = self.left.assign(
+                Y=self.left["Y"].astype("category")
+            )
+            execute(self.left)
+            self.right = self.right.assign(
+                Z=self.right["Z"].astype("category")
+            )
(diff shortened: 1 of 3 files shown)


asv_bench/benchmarks/utils/data_shapes.py
```diff
@@ -166,6 +166,9 @@
     ),
 ]
 DEFAULT_CONFIG = {}
+DEFAULT_CONFIG["MergeCategoricals"] = (
+    [[10_000, 2]] if ASV_DATASET_SIZE == "big" else [[1_000, 2]]
```
YarShev, line 170: Why don't we want to use binary_op data size as for TimeMerge? Also, why only 2 colums, not 10, for example?

asv_bench/benchmarks/benchmarks.py
```diff
@@ -148,6 +148,41 @@ def time_merge(self, shapes, how, sort):
         )
 
 
+class TimeMergeCategoricals:
+    param_names = ["shapes", "data_type"]
```
YarShev, line 152: What about "how", "sort"?

asv_bench/benchmarks/benchmarks.py
```diff
@@ -148,6 +148,41 @@ def time_merge(self, shapes, how, sort):
...
+    params = [
+        get_benchmark_shapes("MergeCategoricals"),
+        ["object", "category"],
+    ]
+
+    def setup(self, shapes, data_type):
+        assert len(shapes) == 2
+        assert shapes[1] == 2
```
anmyachev, line 160: We cannot specify an arbitrary shape for this benchmark.

asv_bench/benchmarks/utils/data_shapes.py
anmyachev: As for the size, it turned out to be not very convenient and fast to fit each operation to the same size. In order to save time, we can not do this. This size is enough to try to optimize this function. I tested this on my local machine. Maybe we can only increase the number of rows, since the local machine has 8 cores, and the benchmark machine has 44 cores. For example: `[100_000, 2]`.

To vary the number of columns, we will need to write a special function that generates a dataframe with the required data type. This is more complicated than just transferring the creation of the dataframe as is. Again, this size is enough to try to optimize this function I think.

asv_bench/benchmarks/benchmarks.py
anmyachev: I have an assumption that it is enough to test these parameters in the rest `merge` benchmarks. This seems to be a special case with categories.

anmyachev: https://github.com/modin-project/modin/pull/5021/files#r987893073

YarShev (approved):

Issue #5099, opened 2022-10-06
BUG: groupby with timestamp in `by` raises KeyError


mvashishtha:
### Modin version checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the latest released version of Modin.

- [X] I have confirmed this bug exists on the main branch of Modin. (In order to do this you can follow [this guide](https://modin.readthedocs.io/en/stable/getting_started/installation.html#installing-from-the-github-master-branch).)

### Reproducible Example

```python
import modin.pandas as pd

df = pd.DataFrame({'timestamp': [pd.to_datetime(1490195805, unit='s')], 'numeric': [0]})
print(df.groupby('timestamp').mean())
```

### Issue Description

I get KeyError in Modin, but the groupby works in pandas.

### Expected Behavior

pandas gives output like

```
                     numeric
timestamp
2017-03-22 15:16:45      0.0
```

### Error Logs

<details>

```python-traceback

UserWarning: Distributing <class 'dict'> object. This may take some time.
pd.---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
File ~/opt/anaconda3/envs/modin-dev/lib/python3.10/site-packages/pandas/core/indexes/base.py:3800, in Index.get_loc(self, key, method, tolerance)
   3799 try:
-> 3800     return self._engine.get_loc(casted_key)
   3801 except KeyError as err:

File ~/opt/anaconda3/envs/modin-dev/lib/python3.10/site-packages/pandas/_libs/index.pyx:138, in pandas._libs.index.IndexEngine.get_loc()

File ~/opt/anaconda3/envs/modin-dev/lib/python3.10/site-packages/pandas/_libs/index.pyx:165, in pandas._libs.index.IndexEngine.get_loc()

File pandas/_libs/hashtable_class_helper.pxi:5745, in pandas._libs.hashtable.PyObjectHashTable.get_item()

File pandas/_libs/hashtable_class_helper.pxi:5753, in pandas._libs.hashtable.PyObjectHashTable.get_item()

KeyError: 'timestamp'

The above exception was the direct cause of the following exception:

KeyError                                  Traceback (most recent call last)
Input In [5], in <cell line: 4>()
      1 import modin.pandas as pd
      3 df = pd.DataFrame({'timestamp': [pd.to_datetime(1490195805, unit='s')], 'numeric': [0]})
----> 4 print(df.groupby('timestamp').mean())

[...]

PR #5140, opened 2022-10-19
FIX-#5099: Fix PandasQueryCompiler.groupby_mean with timestamp in by

Billy2551:

## What do these changes do?
Added check to see if timestamp columns are in `by` before casting to int64 to calculate mean. Timestamp columns in `by` are filtered out and kept as timestamps.

Added test to validate fix.

- [ ] Resolves #? 
- [ ] tests added and passing

### DIFF
--- a/modin/core/storage_formats/pandas/query_compiler.py
+++ b/modin/core/storage_formats/pandas/query_compiler.py
@@ -2533,12 +2533,24 @@ class PandasQueryCompiler(BaseQueryCompiler):
     groupby_sum = GroupByReduce.register("sum")
 
     def groupby_mean(self, by, axis, groupby_kwargs, agg_args, agg_kwargs, drop=False):
+        by_columns = by
+        if isinstance(by, BaseQueryCompiler):
+            by_columns = by.columns
+        elif isinstance(by, list):
+            if isinstance(by[0], BaseQueryCompiler):
+                by_columns = []
+                for by_qc in by:
+                    by_columns.extend(by_qc.columns)
+            else:
+                by_columns = by
+        elif not is_list_like(by):
+            by_columns = [by]
         numeric_only = agg_kwargs.get("numeric_only", False)
         datetime_cols = (
             {
                 col: dtype
                 for col, dtype in zip(self.dtypes.index, self.dtypes)
-                if is_datetime64_any_dtype(dtype)
+                if is_datetime64_any_dtype(dtype) and col not in by_columns
             }
             if not numeric_only
             else dict()
--- a/modin/pandas/test/test_groupby.py
+++ b/modin/pandas/test/test_groupby.py
@@ -2165,3 +2165,10 @@ def test_groupby_with_frozenlist():
     pandas_df = pandas_df.set_index(["a", "b"])
     modin_df = from_pandas(pandas_df)
     eval_general(modin_df, pandas_df, lambda df: df.groupby(df.index.names).count())
+
+
+def test_mean_with_datetime():
+    data = {'timestamp': [pd.to_datetime(1490195805, unit='s')], 'numeric': [0]}
+    pandas_df = pandas.DataFrame(data)
+    modin_df = pd.DataFrame(data)
+    eval_general(modin_df, pandas_df, lambda df: df.groupby("timestamp").mean())


modin/pandas/test/test_groupby.py
```diff
@@ -2165,3 +2165,10 @@ def test_groupby_with_frozenlist():
     pandas_df = pandas_df.set_index(["a", "b"])
     modin_df = from_pandas(pandas_df)
     eval_general(modin_df, pandas_df, lambda df: df.groupby(df.index.names).count())
+
+
+def test_mean_with_datetime():
+    data = {"timestamp": [pd.to_datetime(1490195805, unit="s")], "numeric": [0]}
+    pandas_df = pandas.DataFrame(data)
+    modin_df = pd.DataFrame(data)
```
pyrito, line 2173:
```suggestion
    modin_df, pandas_df = create_test_dfs(data)
```

modin/core/storage_formats/pandas/query_compiler.py
```diff
@@ -2533,12 +2533,24 @@ def _callable_func(self, func, axis, *args, **kwargs):
     groupby_sum = GroupByReduce.register("sum")
 
     def groupby_mean(self, by, axis, groupby_kwargs, agg_args, agg_kwargs, drop=False):
+        by_columns = by
+        if isinstance(by, BaseQueryCompiler):
+            by_columns = by.columns
+        elif isinstance(by, list):
+            if isinstance(by[0], BaseQueryCompiler):
```
pyrito, line 2540: why do we only check if the first element of `by`?

mvashishtha: +1, we should add test coverage for that case

modin/core/storage_formats/pandas/query_compiler.py
```diff
@@ -2533,12 +2533,24 @@ def _callable_func(self, func, axis, *args, **kwargs):
     groupby_sum = GroupByReduce.register("sum")
 
     def groupby_mean(self, by, axis, groupby_kwargs, agg_args, agg_kwargs, drop=False):
+        by_columns = by
+        if isinstance(by, BaseQueryCompiler):
```
pyrito, line 2537: is it possible for `by` to be a normal pandas objects?

modin/core/storage_formats/pandas/query_compiler.py
```diff
@@ -2533,12 +2533,24 @@ def _callable_func(self, func, axis, *args, **kwargs):
     groupby_sum = GroupByReduce.register("sum")
 
     def groupby_mean(self, by, axis, groupby_kwargs, agg_args, agg_kwargs, drop=False):
+        by_columns = by
```
mvashishtha, line 2536: Can we reuse any of the logic in `groupby_agg`, which maybe is doing something similar? https://github.com/modin-project/modin/blob/08ec2e1397cf48fd26cb2a69fb378bb2f38c5d16/modin/core/storage_formats/pandas/query_compiler.py#L2809-L2827

If we can, we should write a common utility function.

modin/core/storage_formats/pandas/query_compiler.py
```diff
@@ -2533,12 +2533,24 @@ def _callable_func(self, func, axis, *args, **kwargs):
...
+        if isinstance(by, BaseQueryCompiler):
+            by_columns = by.columns
+        elif isinstance(by, list):
```
mvashishtha, line 2539: We should add test coverage for these two cases

modin/core/storage_formats/pandas/query_compiler.py
Billy2551: Checks all items in by now

modin/core/storage_formats/pandas/query_compiler.py
Billy2551: Done

modin/core/storage_formats/pandas/query_compiler.py
Billy2551: Done

modin/pandas/test/test_groupby.py
```diff
@@ -2165,3 +2165,22 @@ def test_groupby_with_frozenlist():
...
     modin_df = from_pandas(pandas_df)
     eval_general(modin_df, pandas_df, lambda df: df.groupby(df.index.names).count())
+
+
+@pytest.mark.parametrize(
+    "by_func",
+    [
+        lambda df: df.columns[0],
+        lambda df: list(df.columns[:-1]),
+        lambda df: [df.columns[0], df.iloc[:, 1].copy()],
```
mvashishtha, line 2175: Isn't column 1 out of bounds in the `iloc`?

mvashishtha, line 2175: Why do we need the `copy`?

Billy2551: Column `df.iloc[:, 1]` is the same thing as `df["timestamp1"]`. I will use `df["timestamp1"]` to make it clearer.

Billy2551: Whoops, we don't. Output doesn't change with or without it.

modin/core/storage_formats/pandas/query_compiler.py
mvashishtha: We normally don't allow that kind of mixing pandas and modin objects. However, it looks like this case defaults to pandas later in this function:

```python
import modin.pandas as pd

df = pd.DataFrame([[1, 2], [3, 4]])
print(df.groupby(0).mean())
print(df.groupby(df._to_pandas()[0]).mean())
```

... but it's also possibly buggy (results for pandas by vs modin by are different). Anyway I don't think we need to worry about it here.

pyrito (commented):
LGTM - just one minor nit

mvashishtha (approved):
@Billy2551 this looks great with the refactor!

Issue #4637, opened 2022-07-01
PERF: read_parquet incorrectly thinks datasets are column-partitioned when they contain non-partition folders


mvashishtha:
### System information
- **Modin version** (`modin.__version__`): 86d361078fbad03143bf02ce212ffa8af4d416b5
- **Python version**: 3.9.12
- - **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: macOS Monterey 12.4

Parquet datasets are arbitrarily deep directories of data files that can include column partitioning (ref. [here](https://spark.apache.org/docs/2.4.0/sql-data-sources-parquet.html#partition-discovery)). In column partitioning, a directory like `col1=3` only has data where column 1 is equal to 3. Modin tries to default to pandas for datasets with column partitioning, but it assumes almost any directory name is a column partitioning directory. For example, the following defaults to pandas:

```python
import pandas
import os

import modin.pandas as pd

os.makedirs("/tmp/level0/level1", exist_ok=True)
pandas.DataFrame([1], columns=['col0']).to_parquet("/tmp/level0/level1/df.parquet")
pd.read_parquet("/tmp/level0/")
```

Instead of the check [here](https://github.com/modin-project/modin/blob/5ca6d40f0ab9e6fede0dd5e77a415231a394fa90/modin/core/io/column_stores/parquet_dispatcher.py#L67) we should maybe check against a regex like `'.+=.+'`.

PR #5192, opened 2022-11-04
FIX-#4636: allows `read_parquet` to detect column partitioning in non-local filesystems

### LINKED ISSUE #4636: BUG: read_parquet can't detect column partitioning in non-local filesystems
### System information
- **Modin version** (`modin.__version__`): 86d361078fbad03143bf02ce212ffa8af4d416b5

Parquet datasets are arbitrarily deep directories of data files that can include column partitioning (ref. [here](https://spark.apache.org/docs/2.4.0/sql-data-sources-parquet.html#partition-discovery)). In column partitioning, a directory like `col1=3` only has data where column 1 is equal to 3. Modin tries to default to pandas for datasets with column partitioning, but it [assumes](https://github.com/modin-project/modin/blob/5ca6d40f0ab9e6fede0dd5e77a415231a394fa90/modin/core/io/column_stores/parquet_dispatcher.py#L59) that column partitioning can only happen when the path to read is a **local** directory. The directory could, for example, be on s3.

I think that nothing is functionally wrong when we fail to detect column partitioning. Modin can still read the data correctly. It's just not defaulting to pandas even though we expect it to.

Billy2551:

## What do these changes do?

Added support for `read_parquet` on column partitioned parquet files in non-local filesystems. Includes tests on non-local S3 parquet files.

- [x] Resolves #4636 
- [x] tests added and passing

### DIFF
--- a/modin/core/io/column_stores/parquet_dispatcher.py
+++ b/modin/core/io/column_stores/parquet_dispatcher.py
@@ -609,18 +609,24 @@ class ParquetDispatcher(ColumnStoreDispatcher):
         )
         from modin.pandas.io import PQ_INDEX_REGEX
 
-        if isinstance(path, str) and os.path.isdir(path):
+        if isinstance(path, str):
+            if os.path.isdir(path):
+                path_generator = os.walk(path)
+            else:
+                storage_options = kwargs.get("storage_options")
+                fs, fs_path = url_to_fs(path, **storage_options)
+                path_generator = fs.walk(fs_path)
             partitioned_columns = set()
             # We do a tree walk of the path directory because partitioned
             # parquet directories have a unique column at each directory level.
             # Thus, we can use os.walk(), which does a dfs search, to walk
             # through the different columns that the data is partitioned on
-            for (_, dir_names, files) in os.walk(path):
+            for (_, dir_names, files) in path_generator:
                 if dir_names:
                     partitioned_columns.add(dir_names[0].split("=")[0])
                 if files:
                     # Metadata files, git files, .DSStore
-                    if files[0][0] == ".":
+                    if len(files[0]) > 0 and files[0][0] == ".":
                         continue
                     break
             partitioned_columns = list(partitioned_columns)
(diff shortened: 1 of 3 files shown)


modin/core/io/column_stores/parquet_dispatcher.py
```diff
@@ -609,18 +609,24 @@ def _read(cls, path, engine, columns, **kwargs):
         )
         from modin.pandas.io import PQ_INDEX_REGEX
 
-        if isinstance(path, str) and os.path.isdir(path):
+        if isinstance(path, str):
+            if os.path.isdir(path):
+                path_generator = os.walk(path)
+            else:
+                storage_options = kwargs.get("storage_options")
+                fs, fs_path = url_to_fs(path, **storage_options)
```
anmyachev, line 617: `storage_options` variable can be `None`, so this construction breaks several tests in CI.

modin/core/io/column_stores/parquet_dispatcher.py
```diff
@@ -609,18 +609,27 @@ def _read(cls, path, engine, columns, **kwargs):
...
             # Thus, we can use os.walk(), which does a dfs search, to walk
             # through the different columns that the data is partitioned on
-            for (_, dir_names, files) in os.walk(path):
+            for (_, dir_names, files) in path_generator:
                 if dir_names:
                     partitioned_columns.add(dir_names[0].split("=")[0])
                 if files:
                     # Metadata files, git files, .DSStore
-                    if files[0][0] == ".":
+                    if len(files[0]) > 0 and files[0][0] == ".":
```
mvashishtha, line 632: Do you know why we only check the first file in the directory? i.e. `files[0]`?

modin/core/io/column_stores/parquet_dispatcher.py
Billy2551: Done.

modin/core/io/column_stores/parquet_dispatcher.py
mvashishtha: If I understand the current logic correctly, it’s “if there’s any directory within path where the first file [and apparently the files don’t come in any guaranteed order] starts with ‘.’, then stop looking for partitioned columns”. I think is buggy in a few ways, e.g. see #4637. For now could you just add a TODO to fix the logic here?

pyrito (approved):

mvashishtha (approved):
Thanks @Billy2551 for the fix!

PR #5254, opened 2022-11-23
FEAT-#5253: Upgrade pandas to 1.5.2

### LINKED ISSUE #5253: Upgrade pandas to 1.5.2

YarShev:

## What do these changes do?

- [x] Resolves #5253  
- [x] tests passing

### DIFF
--- a/environment-dev.yml
+++ b/environment-dev.yml
@@ -2,7 +2,7 @@ name: modin
 channels:
   - conda-forge
 dependencies:
-  - pandas==1.5.1
+  - pandas==1.5.2
   - numpy>=1.18.5
   - pyarrow>=4.0.1
   - dask>=2.22.0
--- a/.github/workflows/ci-notebooks.yml
+++ b/.github/workflows/ci-notebooks.yml
@@ -44,7 +44,9 @@ jobs:
           environment-file: requirements/env_hdk.yml
           python-version: 3.8
           channel-priority: strict
-          use-only-tar-bz2: true # IMPORTANT: This needs to be set for caching to work properly!
+          # we set use-only-tar-bz2 to false in order for conda to properly find new packages to be installed
+          # for more info see https://github.com/conda-incubator/setup-miniconda/issues/264
+          use-only-tar-bz2: false
         if: matrix.execution == 'hdk_on_native'
       - name: Cache datasets
         uses: actions/cache@v2
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -192,7 +192,9 @@ jobs:
           environment-file: environment-dev.yml
           python-version: 3.8
           channel-priority: strict
-          use-only-tar-bz2: true # IMPORTANT: This needs to be set for caching to work properly!
+          # we set use-only-tar-bz2 to false in order for conda to properly find new packages to be installed
+          # for more info see https://github.com/conda-incubator/setup-miniconda/issues/264
+          use-only-tar-bz2: false
       - name: Conda environment
         run: |
           conda info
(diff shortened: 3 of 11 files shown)


environment-dev.yml
```diff
@@ -2,7 +2,7 @@ name: modin
 channels:
   - conda-forge
 dependencies:
-  - pandas==1.5.1
+  # - pandas==1.5.2
```
YarShev, line 5: conda hasn't picked up yet the release.

mvashishtha: Can we merge the PR like this, or should we wait until conda has picked it up?

YarShev: It seems that is a conda issue in GH actions. I can install pandas 1.5.2 locally with conda. I encountered the same problem in #5059 when installing unidist-mpi 0.2.1. I think it is okay to install pandas from PyPI in dev env files for now but we should definetely figure out why conda can't install latest versions of pandas and unidist in GH actions.

mvashishtha (dismissed):

environment-dev.yml
vnlitvinov: IMHO this should be investigated instead of playing an ostrich shoving the head beneath some sand.

environment-dev.yml
```diff
@@ -42,6 +41,9 @@ dependencies:
   # for release script
   - pygit2
   - pip:
+      # TODO(https://github.com/modin-project/modin/issues/5256): Move pandas back
+      # to installation with conda
+      - pandas==1.5.2
```
vnlitvinov, line 46: This feels like _plain wrong_. This should be investigated (for example, we won't be able to make a conda-forge release doing things like that).

vnlitvinov (changes requested):

environment-dev.yml
YarShev: I do not understand what you mean by "plain wrong". We have already been installing Ray, for instance, from PyPI in environment-dev.yml. I do not see anything bad if we install pandas from PyPI for now in order not to stop making the release in PyPI. While we are preparing the release for PyPI, we can investigate the issue in backgroung and when we come to the release in conda-forge, the issue might already be resolved. Created https://github.com/conda-incubator/setup-miniconda/issues/264. So I propose to merge the changes as is.

vnlitvinov: I believe the fact that we mix things from `pip` and `conda` is causing us issues already - https://github.com/modin-project/modin/issues/5208 (probable cause explained around https://github.com/conda-forge/ray-packages-feedstock/pull/78#issuecomment-1316556694).

Plus it means we're not testing what our typical end user is installing, I think most customers use either `pip` or `conda` but not a semi-random mix.

YarShev: I agree that we shouldn't mix things from `pip` and `conda` but we have been doing that for a quite a long time (e.g., Ray). The fact we currenlty install pandas from `conda` and Ray from `pip` doesn't guarantee the correctness of Modin as we don't test this. Here is the similar case. My proposals are

1. Merge the changes as is.
2. Wait for a response in https://github.com/conda-incubator/setup-miniconda/issues/264 or https://github.com/pandas-dev/pandas/issues/49894.

If you have anything else, you're welcome.

I tried to use `channels` option for `conda-incubator/setup-miniconda@v2` in GH Actions to specify `conda-forge` manually and it didn't work.
I tried to remove pandas from the file and install it manually with `conda install pandas=.1.5.2 -c conda-forge` in GH Actions. It didn't work too.

YarShev: I returned back pandas to installation with conda. See more about the fix in https://github.com/conda-incubator/setup-miniconda/issues/264.

vnlitvinov (approved):
Cool, awesome you've figured it our @YarShev!

PR #5270, opened 2022-11-25
FIX-#5284: move ray, redis, tqdm, xgboost packages from pip to conda deps

### LINKED ISSUE #5284: TEST: move ray, redis, tqdm, xgboost packages from pip to conda deps
This should make it easier to find problems with Ray 2.1.0.
See details here: https://github.com/modin-project/modin/issues/5208

anmyachev:

## What do these changes do?

- [x] Resolves #5284 
- [x] tests added and passing

### DIFF
--- a/environment-dev.yml
+++ b/environment-dev.yml
@@ -5,6 +5,8 @@ dependencies:
   - pandas==1.5.2
   - numpy>=1.18.5
   - pyarrow>=4.0.1
+  - ray-core>=1.4.0
+  - redis-py>=3.5.0,<4.0.0
   - dask>=2.22.0
   - distributed>=2.22.0
   - fsspec
@@ -39,7 +41,6 @@ dependencies:
   - mypy<0.990
   - pandas-stubs
   - fastparquet
-  - grpcio<=1.48.2
   # for release script
   - pygit2
   - pip:
@@ -50,7 +51,6 @@ dependencies:
       - git+https://github.com/modin-project/modin-spreadsheet.git@49ffd89f683f54c311867d602c55443fb11bf2a5
       - tqdm
       - git+https://github.com/airspeed-velocity/asv.git@ef016e233cb9a0b19d517135104f49e0a3c380e9
-      - ray[default]>=1.13.0
       - connectorx>=0.2.6a4
       - black
       # TODO: remove when flake8 5.x stabilizes and appears in both pip and conda-forge; see GH-#4745
--- a/environment-dev.yml
+++ b/environment-dev.yml
@@ -6,6 +6,7 @@ dependencies:
   - numpy>=1.18.5
   - pyarrow>=4.0.1
   - ray-default>=1.4.0
+  - redis-py>=3.5.0,<4.0.0
   - dask>=2.22.0
   - distributed>=2.22.0
   - fsspec
--- a/environment-dev.yml
+++ b/environment-dev.yml
@@ -53,7 +53,7 @@ dependencies:
       - ray[default]>=1.4.0
       - connectorx>=0.2.6a4
       # TODO: remove when resolving GH#4398
-      - redis>=3.5.0
+      - redis>=3.5.0,<4.0.0
       - black
       # TODO: remove when flake8 5.x stabilizes and appears in both pip and conda-forge; see GH-#4745
       - flake8<5
--- a/environment-dev.yml
+++ b/environment-dev.yml
@@ -6,7 +6,6 @@ dependencies:
   - numpy>=1.18.5
   - pyarrow>=4.0.1
   - ray-default>=1.4.0
-  - redis-py>=3.5.0,<4.0.0
   - dask>=2.22.0
   - distributed>=2.22.0
   - fsspec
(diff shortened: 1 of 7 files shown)


environment-dev.yml
```diff
@@ -5,6 +5,11 @@ dependencies:
   - pandas==1.5.2
   - numpy>=1.18.5
   - pyarrow>=4.0.1
+  - ray-default>=1.4.0,<2.1.0
+  # This will allow us to install ray==2.0.1 instead of ray==1.12.1;
```
anmyachev, line 9: A little more detail on why limiting protobuf version allows us to install a later version of ray-default:
- ray-default of 1.12.1 and 2.1.0 versions hasn't upper bound for protobuf.
- ray-default of 1.13.0 and 2.0.1 versions has "<4.0.0" upper bound for protobuf.

environment-dev.yml
```diff
@@ -5,6 +5,11 @@ dependencies:
...
+  # Apparently the priority is to install a later version of protobuf, instead of ray
+  - protobuf>=3.15.3,<4.0.0
```
anmyachev, line 11: This is a temporary pin, I suppose after we support the new version of ray (2.1.0), this can be removed.

environment-dev.yml
anmyachev: I suggest, if possible, not to interfere with the installation using conda and pip, since there may be situations when there is a difference in the distributed files of the same version, for example: https://github.com/conda-forge/ray-packages-feedstock/pull/78#issuecomment-1312572638.

environment-dev.yml
```diff
@@ -41,18 +46,16 @@ dependencies:
...
   - pip:
-      # TODO(https://github.com/modin-project/modin/issues/5194): Uncap xgboost
-      # when we use collective instead of rabit.
-      - xgboost>=1.7.1,<2.0.0
       # Fixes breaking ipywidgets changes, but didn't release yet.
       - git+https://github.com/modin-project/modin-spreadsheet.git@49ffd89f683f54c311867d602c55443fb11bf2a5
-      - tqdm
       - git+https://github.com/airspeed-velocity/asv.git@ef016e233cb9a0b19d517135104f49e0a3c380e9
-      - ray[default]>=1.4.0,<2.1.0
+      # no conda package for windows
```
vnlitvinov, line 57: what does this comment do?

vnlitvinov (commented):
I wonder if this also fixes https://github.com/modin-project/modin/issues/5208? I have hopes it does... I mean if `<2.1.0` restriction is lifted

environment-dev.yml
anmyachev: I tried to install this package using conda, however this package is only available for linux and macOS. https://anaconda.org/conda-forge/connectorx/files

anmyachev:
Interesting notes:
* `conda install -c conda-forge grpcio==1.46.3` will try to install `libprotobuf==3.20.2`
* `conda install -c conda-forge grpcio==1.46.3 libprotobuf` will try to install `libprotobuf==3.21.9` (however in both cases grpcio has restriction: `libprotobuf >=3.20.1,<3.21.0a0`)

YarShev:
Two jobs (`ci / test-compat-ubuntu (engine ray, python 3.6)` and `ci / test-compat-win (engine ray, python 3.6)`) were cancelled. I re-ran those.

anmyachev:
@YarShev @vnlitvinov CI is green, I would merge this before trying to unpin ray. What do you think?

environment-dev.yml
Garra1980: So we will remove protobuf restriction along with allowing  ray 2.1.0?

anmyachev: That's the plan.

Or I can remove this restriction and we will temporarily use not the newest ray.

Garra1980: we don't use it already (>=1.4.0,<2.1.0) or I miss something? :)

anmyachev: I mean restriction for protobuf. 

Due to the fact that conda gives priority to the `protobuff` and ray does not support the newest versions of the `protobuff`, a newest `protobuff` and an old `ray` are installed, since `ray` has no restrictions on the `protobuff` (for its own internal reasons).

If we want to remove the restriction on ray right now, then we need to add a fix from https://github.com/modin-project/modin/pull/5283. The fix will include some restrictions for `grpcio` (it's a dependency of Ray).
Do not mind doing it right now, but it seemed to me that these are tasks that are better logically divided into different pull requests.

environment-dev.yml
```diff
@@ -5,6 +5,11 @@ dependencies:
   - pandas==1.5.2
   - numpy>=1.18.5
   - pyarrow>=4.0.1
+  - ray-default>=1.4.0,<2.1.0
+  # This will allow us to install ray==2.0.1 instead of ray==1.12.1;
+  # Apparently the priority is to install a later version of protobuf, instead of ray
```
YarShev, line 10:
```suggestion
  # Apparently, the priority is to install a later version of protobuf, instead of ray
```

environment-dev.yml
YarShev, line 57:
```suggestion
      # no conda package for windows so we install it with pip
```

YarShev (approved):

environment-dev.yml
anmyachev: https://github.com/sfu-db/connector-x/issues/413

PR #5371, opened 2022-12-07
REFACTOR-#5370: Move merge_asof implementation to base query compiler.

### LINKED ISSUE #5370: REFACTOR: Add merge_asof query compiler function

mvashishtha:
Some modin implementations might not want to use the current implementation, which is at the API layer and partially defaults to pandas.

## What do these changes do?

- [x] Resolves #5370
- [x] tests added and passing

### DIFF
--- a/modin/core/storage_formats/base/query_compiler.py
+++ b/modin/core/storage_formats/base/query_compiler.py
@@ -38,8 +38,9 @@ from modin.utils import MODIN_UNNAMED_SERIES_LABEL
 from pandas.core.dtypes.common import is_scalar
 import pandas.core.resample
 import pandas
+from pandas._typing import IndexLabel, Suffixes
 import numpy as np
-from typing import List, Hashable
+from typing import List, Hashable, Optional
 
 
 def _get_axis(axis):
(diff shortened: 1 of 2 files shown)


modin/core/storage_formats/base/query_compiler.py
```diff
@@ -856,6 +857,142 @@ def merge(self, right, **kwargs):  # noqa: PR02
...
+            or not (right_by is None or is_scalar(right_by))
+            # The implementation below assumes that the right index is unique
+            # because it uses merge_asof to map each position in the merged
+            # index to the label of the one right row that should be merged
+            # at that row position.
+            or not right.index.is_unique
+        ):
+            return self.default_to_pandas(
+                default_merge_asof,
+                right,
```
anmyachev, line 913: Why don't you pass `on` and `by` parameters here?

modin/core/storage_formats/base/query_compiler.py
```diff
@@ -856,6 +857,142 @@ def merge(self, right, **kwargs):  # noqa: PR02
...
+        ----------
+        key : Any
+            Column label.
+
+        Returns
+        -------
+        pandas.Series
+        """
+        result = self.getitem_array([key]).to_pandas().squeeze(axis=1)
+        assert isinstance(result, pandas.Series)
```
anmyachev, line 874: I assume we shouldn't use `assert` in any place except tests.

modin/core/storage_formats/base/query_compiler.py
```diff
@@ -856,6 +857,142 @@ def merge(self, right, **kwargs):  # noqa: PR02
...
+            or not (left_by is None or is_scalar(left_by))
+            or not (right_by is None or is_scalar(right_by))
+            # The implementation below assumes that the right index is unique
+            # because it uses merge_asof to map each position in the merged
+            # index to the label of the one right row that should be merged
+            # at that row position.
+            or not right.index.is_unique
+        ):
+            return self.default_to_pandas(
+                default_merge_asof,
```
vnlitvinov, line 912: why cannot we just pass `pandas.merge_asof` without defining an intermediary function? I don't see a point

modin/core/storage_formats/base/query_compiler.py
mvashishtha: The query compiler doesn't receive `on` or `by` params. I didn't think that those needed to be part of the query compiler interface because the API layer should translate them to `left_on`, `right_on`, `left_by`, and `right_by`.

modin/core/storage_formats/base/query_compiler.py
mvashishtha: I think that's a controversial position :) In my view, it's better to fail fast when we rely on assumptions that fail. Here we assume we get a series because 1)  `left_by`, `left_on`, `right_by`, `right_on` are all scalars if they're not None 2) they don't represent duplicate column names. 

I think 1) is wrong because we don't check this for `left_on` and `right_on`, which can be lists, and 2) will break with an uninformative `AttributeError: 'DataFrame' object has no attribute 'dtype'` in pandas

But these bugs are out of scope for this PR.

modin/core/storage_formats/base/query_compiler.py
mvashishtha: Done.

modin/core/storage_formats/base/query_compiler.py
mvashishtha: Okay, now the code raises `RuntimeError`.

mvashishtha:
@anmyachev @vnlitvinov could you please take a look?

vnlitvinov (approved):
LGTM!

mvashishtha:
@anmyachev @vnlitvinov ready for merge

mvashishtha:
@anmyachev @vnlitvinov ready for merge

anmyachev (approved):

PR #5390, opened 2022-12-08
FIX-#0000: check old versions of dependencies

anmyachev:

## What do these changes do?

- [ ] Resolves #? 
- [ ] tests added and passing

### DIFF
--- a/environment-dev.yml
+++ b/environment-dev.yml
@@ -5,7 +5,7 @@ dependencies:
   - pandas==1.5.2
   - numpy>=1.18.5
   - pyarrow>=4.0.1
-  - ray-default>=1.4.0
+  - ray-default==1.8.0
   # workaround for https://github.com/conda/conda/issues/11744
   - grpcio!=1.45.*
   - grpcio!=1.46.*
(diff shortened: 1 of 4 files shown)


YarShev (approved):
Fine by me!

environment-dev.yml
```diff
@@ -4,12 +4,11 @@ channels:
 dependencies:
   - pandas==1.5.2
   - numpy>=1.18.5
-  - pyarrow>=4.0.1
-  - ray-default>=1.4.0
+  - ray-default>=1.13.0
+  - pyarrow
```
vnlitvinov, line 8: why remove minimal requirement on pyarrow?

vnlitvinov (approved):
I'm not _entirely_ sold on Ray 1.13 - it was released less than half a year ago..
However, moving to a better defined dependency is probably worth it...

environment-dev.yml
anmyachev: I'm not sure. At the moment, we can keep at least the lower bounds of dependency versions for different engines consistent. However, with each new engine added, this will become more and more difficult. Perhaps at some point, some engines will become completely incompatible...

PR #5486, opened 2022-12-19
FEAT-#3044: add support for `api.extensions.register_*` functions

### LINKED ISSUE #3044: Add support for registering a custom accessor to allow user to extend DataFrames/Series
- pandas supports adding custom methods directly onto a DataFrame/Series. [pandas docs](https://pandas.pydata.org/pandas-docs/stable/development/extending.html#registering-custom-accessors)
- This pandas feature allows users to extend a pandas DataFrame/Series, and add their own functionality. 
  - Two examples of libraries that use this are [geopands](https://github.com/geopandas/geopandas) and [woodwork](https://github.com/alteryx/woodwork)
- It would be cool to also have the ability to register custom accessor in a modin DataFrame/Series
```python
import modin.pandas as pd

@pd.api.extensions.register_dataframe_accessor("geo")
class GeoAccessor:
    def __init__(self, pandas_obj):
        self._validate(pandas_obj)
        self._obj = pandas_obj

    @staticmethod
    def _validate(obj):
        # verify there is a column latitude and a column longitude
        if "latitude" not in obj.columns or "longitude" not in obj.columns:
            raise AttributeError("Must have 'latitude' and 'longitude'.")

    @property
    def center(self):
        # return the geographic center point of this DataFrame
        lat = self._obj.latitude
        lon = self._obj.longitude
        return (float(lon.mean()), float(lat.mean()))

    def plot(self):
        # plot this array's data on a map, e.g., using Cartopy
        pass
```
```python
In [1]: ds = pd.DataFrame({"longitude": np.linspace(0, 10),
   ...:                    "latitude": np.linspace(0, 20)})
In [2]: ds.geo.center
Out[2]: (5.0, 10.0)
In [3]: ds.geo.plot()  # plots data on a map
```

anmyachev:

## What do these changes do?

- [x] Resolves #3044 
- [ ] tests added and passing

### DIFF
--- a/modin/pandas/api/extensions/__init__.py
+++ b/modin/pandas/api/extensions/__init__.py
@@ -1,3 +1,16 @@
+# Licensed to Modin Development Team under one or more contributor license agreements.
+# See the NOTICE file distributed with this work for additional information regarding
+# copyright ownership.  The Modin Development Team licenses this file to you under the
+# Apache License, Version 2.0 (the "License"); you may not use this file except in
+# compliance with the License.  You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software distributed under
+# the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific language
+# governing permissions and limitations under the License.
+
 """
 Public API for extending modin objects.
 """
--- a/modin/pandas/api/extensions/__init__.py
+++ b/modin/pandas/api/extensions/__init__.py
@@ -1,0 +1,24 @@
+"""
+Public API for extending modin objects.
+"""
+
+from pandas.core.accessor import _register_accessor
+from pandas.api.extensions import *  # noqa: F401, F403
+
+
+def register_dataframe_accessor(name):
+    from modin.pandas import DataFrame
+
+    return _register_accessor(name, DataFrame)
+
+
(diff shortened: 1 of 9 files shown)


modin/pandas/api/extensions/__init__.py
```diff
@@ -0,0 +1,37 @@
...
+"""
+
+from pandas.core.accessor import _register_accessor
+from pandas.api.extensions import *  # noqa: F401, F403
+
+
+def register_dataframe_accessor(name):
+    from modin.pandas import DataFrame
+
+    return _register_accessor(name, DataFrame)
```
anmyachev, line 25: In pandas the function looks like:

```python
def register_dataframe_accessor(name):
    from pandas import DataFrame

    return _register_accessor(name, DataFrame)
```

Therefore, when called, it modify pandas objects, not modin. So we can't just import them unchanged.

modin/pandas/series.py
```diff
@@ -81,6 +81,7 @@ class Series(BasePandasDataset):
 
     _pandas_class = pandas.Series
     __array_priority__ = pandas.Series.__array_priority__
+    _accessors = {"dt", "cat", "str", "sparse"}
```
anmyachev, line 84: @jbrockmendel in order to reuse `_register_accessor` function from pandas, it's needed to add `_accessors` in Modin objects. 

Can you tell me what this attribute is for and maybe you can also tell me how to properly integrate it into Modin?

jbrockmendel: its used to control what shows up in `__dir__` via _dir_deletions and _dir_additions

PR #5494, opened 2022-12-22
PERF-#5182: Precompute dtypes when performing binary operations in certain cases

### LINKED ISSUE #5182: [PERF] compute `dtypes` when performing binary operations
It should be possible and beneficial to compute `dtypes` when doing binary operations without remote calls so other subsequent operations, which require `dypes` to be known, do not perform the remote calls to compute it.

Here are some examples when we can compute/pass known dtypes to the core dataframe.

1. `left_df + right_df`, leftleft_df.dtypes and right_df.dtypes are known and equal to each other
2. `left_df + 1`, left_df.dtypes is known, dtype of 1 is int.

arunjose696:
## What do these changes do?

Added logic to precompute dtypes before add operation was performed on two dataframes.

Benchmarking code

```
import modin.pandas as pd
import numpy as np
import time
rows=1000
cols=100000
df = DataFrame(np.ones((rows, cols)),dtype=np.int32)
df1=df.copy()

start_time = time.time()
df_sum = df1 / df
df_sum.astype('int32')
print("--- %s seconds ---" % (time.time() - start_time))
```

For master branch

**--- 20.01868987083435 seconds ---**

For the current change

**--- 4.792952299118042 seconds ---**

For boolean operators  (Replacing `df_sum = df1 / df` with below line )

`df_sum = df1 > df
`

For Master:
**--- 21.683951139450073 seconds ---**

For current

**--- 5.464660882949829 seconds ---**

- [x] Resolves #5182 ? 
- [x] tests added and passing


modin/core/dataframe/algebra/binary.py
```diff
@@ -94,11 +94,21 @@ def caller(
...
                     )
                 else:
+                    if other.dtypes is not None and query_compiler.dtypes is not None:
+                        if func.__name__ == "add":
+                            dtypes = pandas.Series(
+                                map(
+                                    lambda x: np.find_common_type([*x], []),
+                                    zip(other.dtypes, query_compiler.dtypes),
+                                )
+                            )
```
YarShev, line 104: 1. Which binary operations do not follow this logic for computing resultant dtypes?
2. Why should we use `np.find_common_type` instead of `pandas.core.dtypes.cast.find_common_type`?

arunjose696: 1)Binary operations such as divide(returns a float even if both operands are int), modulus etc might not follow this logic. I will collate a list of operators.
2)Tried to use pandas.core.dtypes.cast.find_common_type directly on dtypes of both series but this was not working in 1.5.2 version of pandas 

```
pd.core.dtypes.cast.find_common_type([df1.dtypes,df3.dtypes]) #error in pandas 1.5.2
pd.core.dtypes.cast.find_common_type([df1.dtypes,df3.dtypes]) #works for pandas 1.3.x
```
So decided to go with np.find_common_type as dtypes were used one at a time, I will change this to pd.core.dtypes.cast.find_common_type if this is the preffered way.

arunjose696: Surprising to me most binary operators even including modulus and floordivision seems to follow the pd.core.dtypes.cast.find_common_type so i have included  these too in the PR. Below are few binary operations which are different.
1)division
2)boolean operators such as equals not equals, > , >=, < ,<= etc

modin/utils.py
```diff
@@ -740,3 +740,31 @@ def show_versions(as_json: Union[str, bool] = False) -> None:
             print(f"\n{name} dependencies\n{'-' * (len(name) + 13)}")
             for k, v in d.items():
                 print(f"{k:<{maxlen}}: {v}")
+
+
+def int_to_float32(dtype: np.dtype) -> np.dtype:
```
noloerino, line 745: Should this be `int_to_float64`?

modin/utils.py
```diff
@@ -740,3 +740,31 @@ def show_versions(as_json: Union[str, bool] = False) -> None:
...
+    """
+    Check if a datatype is a variant of integer.
+
+    If dtype is integer function returns float32 datatype if not returns the
```
noloerino, line 749:
```suggestion
    If dtype is integer function returns float64 datatype if not returns the
```

modin/core/dataframe/algebra/binary.py
noloerino: What scenarios is `find_common_type` failing in, and what error message do you get? Modin uses this method internally in a few places, so I'm a little surprised there aren't existing bugs for it.

modin/core/dataframe/algebra/binary.py
```diff
@@ -94,11 +95,28 @@ def caller(
...
+                    if other.dtypes is not None and query_compiler.dtypes is not None:
+                        if func.__name__ in boolean_operators:
+                            dtypes = pandas.Series([bool] * len(other.dtypes))
+                        else:
+                            dtypes = pandas.Series(
+                                map(
+                                    lambda x: pandas.core.dtypes.cast.find_common_type(
+                                        [*x]
+                                    ),
+                                    zip(other.dtypes, query_compiler.dtypes),
```
noloerino, line 108: I believe this would create an incorrect result in the case where `other` and `query_compiler` have different dimensions. For example:
```
import pandas as pd
df1 = pd.DataFrame({"a": [1, 2], "b": [2, 3]})  # dtypes: a: int64, b: int64
df2 = pd.DataFrame({"a": [1, 2], "b": [2, 3], "c": [4, 5]}) # dtypes a: int64, b: int64, c: int64
df1 + df2 # Result: {"a": [2, 4], "b": [4, 6], "c": [NaN, NaN]}
```
Since `zip` cuts off at the length of the shorter of its two arguments, the resulting dtypes array in your code would be an element short.

modin/core/dataframe/algebra/binary.py
```diff
@@ -94,11 +95,28 @@ def caller(
                         )
                     )
                 else:
+                    boolean_operators = ["eq", "ge", "gt", "le", "lt", "ne"]
```
noloerino, line 98: I'm a little wary of hardcoding the list of operators in this fashion--though it's not likely to change, it does risk bugs in case there's an edge case for a binary operator that we haven't checked.

Can you add a flag to `Binary.caller` that controls this behavior instead? Something like
```
def caller(query_compiler, [etc...], dtypes=None, precompute_dtypes=False, result_dtype_is_bool=False):
```

where the [corresponding locations in the query compiler](https://github.com/modin-project/modin/blob/882433c46ea59df9850b11b4c6088467a95a1a8f/modin/core/storage_formats/pandas/query_compiler.py#L378) would then look something like this:
```
add = Binary.register(pandas.DataFrame.add, precompute_dtypes=True)
...
eq = Binary.register(pandas.DataFrame.eq, result_dtype_is_bool=True)
```

This per-operator configuration is easier to make changes for, compared to hardcoding them in a list here.

noloerino (changes requested):

modin/core/dataframe/algebra/binary.py
arunjose696: I could see `find_common_type` failing  when  passing 2 series objects as arguments (as mentioned in my previous post). The current code with zip works for pd.core.dtypes.cast.find_common_type as well.

> 1)Binary operations such as divide(returns a float even if both operands are int), modulus etc might not follow this logic. I will collate a list of operators. 2)Tried to use pandas.core.dtypes.cast.find_common_type directly on dtypes of both operand dataframes but this was not working in 1.5.2 version of pandas
> 
> ```
> pd.core.dtypes.cast.find_common_type([df1.dtypes,df3.dtypes]) #error in pandas 1.5.2
> pd.core.dtypes.cast.find_common_type([df1.dtypes,df3.dtypes]) #works for pandas 1.3.x
> ```
> 
> So decided to go with np.find_common_type as dtypes were anyways used one at a time, I can change this to pd.core.dtypes.cast.find_common_type if this is the preffered way.




This is the error message.

**ValueError: The truth value of a Series is ambiguous. Use a.empty, a.bool(), a.item(), a.any() or a.all().**

### Full traceback.

_---------------------------------------------------------------------------
<ipython-input-3-a67402f7da40> in <module>
----> 1 pd.core.dtypes.cast.find_common_type([df1.dtypes,df3.dtypes])
      2 #pd.core.dtypes.cast.find_common_type([other.dtypes, query_compiler.dtypes])

2 frames
/usr/local/lib/python3.8/dist-packages/pandas/core/dtypes/cast.py in find_common_type(types)
   1611     # workaround for find_common_type([np.dtype('datetime64[ns]')] * 2)
   1612     # => object
-> 1613     if lib.dtypes_all_equal(list(types)):
   1614         return first
   1615 

/usr/local/lib/python3.8/dist-packages/pandas/_libs/lib.pyx in pandas._libs.lib.dtypes_all_equal()

/usr/local/lib/python3.8/dist-packages/pandas/core/generic.py in __nonzero__(self)
   1525     @final
   1526     def __nonzero__(self) -> NoReturn:
-> 1527         raise ValueError(
   1528             f"The truth value of a {type(self).__name__} is ambiguous. "
   1529             "Use a.empty, a.bool(), a.item(), a.any() or a.all()."_

ValueError: The truth value of a Series is ambiguous. Use a.empty, a.bool(), a.item(), a.any() or a.all().

modin/utils.py
arunjose696: Yes this was a typo.

modin/core/dataframe/algebra/binary.py
arunjose696: Yes this is an issue, Another corner case would be when the column names are different.

import pandas as pd
df1 = pd.DataFrame({"a": [1, 2], "b": [2, 3]})  # dtypes: a: int64, b: int64
df2 = pd.DataFrame({"1": [1, 2], "2": [2, 3]}) # dtypes a: int64, b: int64, c: int64
df1 + df2 # Result: {"a": [NaN, NaN], "b": [NaN, NaN], "1": [NaN, NaN], "2":[NaN, NaN]}

modin/core/dataframe/algebra/binary.py
YarShev: @noloerino, do you think if it is better to precompute dtypes at the query compiler layer? Then we could avoid this hardcoding, as well as branching with checks for a function name.

arunjose696: @YarShev @noloerino I have changed the logic for precomputing dtypes to query compiler as suggested by @YarShev . Currently I have added logic only for add function. 
Would this be a better option considering the code would be repeated(maybe as a function call) once for each binary function(~30 times)?

noloerino: I think the actual code that does the precomputing should be in the operator `caller` method to avoid repetition, but configured in the query compiler with flags in the manner I suggested. That would avoid both the repetition and the hardcoding.

YarShev: I still don't feel that the precomputing should be in the operator `caller` method. The flags `precompute_dtypes` and `result_dtype_is_bool` seem to me contradicting a bit (Say when `precompute_dtypes is False` and `result_dtype_is_bool is True`. In addition, we still would have the branching in depend on a concrete binary function (check for `truediv` is already presented in this PR).

> Would this be a better option considering the code would be repeated(maybe as a function call) once for each binary function(~30 times)?

I think a function call once for each binary function doesn't seem horrible.

Considering that we have already been passing dtypes from the qc layer in to map operator, for instance, as well as this text from `Type Rules by Operator` section of https://people.eecs.berkeley.edu/~totemtang/paper/Modin.pdf,

```
For operators that allow a UDF/SDF as input, the output types can
be specified by the developer (i.e., “Specified” in Table 1), or left
unspecified. Suppose a developer wants to perform a map with a
SDF that returns TRUE for non-NULL values, and FALSE otherwise.
Since all columns in the output are known to be boolean, this information can be provided by the developer implementing the SDF
to the map upfront to avoid costly type inference. Alternatively,
when the types of the output dataframe after a map is not known, it
ends up being UNSPECIFIED for every column.
```

I think the precomputing dtypes should  happend at the qc.

cc @modin-project/modin-core, other thoughts?

YarShev: @noloerino suggestion implemented in this PR, my suggestion in #5525.

@modin-project/modin-core, do you think which approach we should go with?

dchigarev: > text from Type Rules by Operator section

Good reference, although I think it's more related to actual dataframe algebra defined in a low-level dataframe object. Our "algebra" module is more like wrappers around actual operators, thus I think it's acceptable here to pass instructions on how to get new dtypes instead of already computed ones.

Regarding where to put the inferring logic: I believe it would be better to encapsulate it in the Binary operator. Since the logic tends to be only binary ops related it's then unnecessary to bloat general QC with operator-specific functionality.

As for configuring the type-inferring logic, I would propose something like this. A parameter for `.register` that either takes a resulted type (bool for logical ops, float for `truediv`, ...) or an "infer" keyword, indicating automatic type inferring.
```python
dtypes : {int, float, bool, object, 'infer', None}, default: None
    Specifies a dtype for the result of the binary operation. If 'infer' - try to
    automatically infer the resulted type based on the operand's types. If ``None``
    - do not infer new dtypes (they will be computed manually once accessed).
```
If you don't like mixing actual types and 'infer' in a single parameter we could either reformat the parameter a bit:
```python
how_compute_dtypes : enum.HOW_INFER_TYPE = {INT, FLOAT, BOOL, NOHOW, AUTO}
```
or split them into two separate parameters:
```python
def register(..., dtypes=None, auto_dtypes_inferring=False): # 'dtypes' is ignored if 'auto_dtypes_inferring=True' was specified
```

YarShev: In order not to bloat QC module we wanted to have kernels.py file where we would define respective kernels and would just import those into QC to keep it clean (I do not remember the exact issue for this). We already have the QC bloated enough with some kernels. Also, we may want to compute dtypes for other operators as well and it might be complicated to do that at the algebra layer as here. Each operator is not aware of a function to be executed by it so that might be complex to infer dtypes at this layer. In addition, I am afraid of possible multiple branching in depend on a concrete function in every operator. I still feel we should compute dtypes at the QC layer. @dchigarev, thoughts/comments?

I also want to listen to thoughts from someone from ponder team. @mvashishtha, @RehanSD, @vnlitvinov, @devin-petersohn

devin-petersohn: The way I typically think about dtype handling is that it should be controlled by the algebra layer (what we called the Modin Dataframe layer previously), but the QC layer will need to give hints or suggestions.

We have logical operators and physical operators. The algebra are logical, they call the physical operators from the partition manager.

YarShev: The query compiler is supposed to have knowledge of the compute kernels and in-memory format of the data. It gives hints to the underlying layer (Modin Dataframe layer). We have the structure isolated from each other but the code may become too complicated (for instance, like here the code of the algebra layer may become complicated). And so we will have to expand the knowlodge of a concrete layer in future if we want the code to be clean. These are my thoughts regarding this. So, let's move forward with the suggestion above and see how it goes. We can always refactor the code. @arunjose696, please push this forward with @dchigarev suggestion. I think it looks clearer than adding `precompute_dtypes` and `result_dtype_is_bool` flags.

> As for configuring the type-inferring logic, I would propose something like this. A parameter for .register that either takes a resulted type (bool for logical ops, float for truediv, ...) or an "infer" keyword, indicating automatic type inferring.

```python
dtypes : {int, float, bool, object, 'infer', None}, default: None
    Specifies a dtype for the result of the binary operation. If 'infer' - try to
    automatically infer the resulted type based on the operand's types. If ``None``
    - do not infer new dtypes (they will be computed manually once accessed).
```

vnlitvinov: I would say that dtypes _inference_ should probably be handled by QC (inner layer), but dtypes _knowledge_ seems to be tied more to outer layer, i.e. closer to the real data structures (QC should be more abstract).

Hence I think we should define dtypes at the algebra layer.

As for the interface, I propose to introduce two `object()` instances and use them as values so as to not mix strings and `None`-s, like
```python
InferDtypes = object()
ComputeOnAccess = object()
# skip
def register(..., dtypes=...):
    # skip
    if dtypes is InferDtypes:
        # infer them
    elif dtypes is ComputeOnAccess:
        # do not infer
    else:
        # dtypes were specified explicitly
```

This works much like `pandas._libs.lib.NoDefault` and is super-explicit.

arunjose696: I prefer to go with a enumerator parameter how_compute_dtypes as in @dchigarev  suggestion as mixing actual types and 'infer'  might be confusing.

> ```
> 
> If you don't like mixing actual types and 'infer' in a single parameter we could either reformat the parameter a bit:
> 
> ```python
> how_compute_dtypes : enum.HOW_INFER_TYPE = {INT, FLOAT, BOOL, NOHOW, AUTO}
> ```
>

YarShev: I am okay with @dchigarev suggestion I pointed, as well as @vnlitvinov suggestion but I wouldn't introduce `ComputeOnAccess = object()` for now because we use `None` for dtypes everywhere to compute it on access.

modin/core/dataframe/algebra/binary.py
```diff
@@ -96,13 +160,23 @@ def caller(
                         )
                     )
                 else:
+                    if other.dtypes is not None and query_compiler.dtypes is not None:
```
YarShev, line 163:
```suggestion
                    if other._modin_frame._dtypes is not None and query_compiler._modin_frame._dtypes is not None:
```

YarShev: Please also fix other places where .dtypes is accessed.

modin/core/dataframe/algebra/binary.py
RehanSD: I like the suggestion @arunjose696 and @YarShev provide - the algebra layer is agnostic to what specific functions are being called (e.g. if we have a map whose function is to convert the dtype of all columns to float), while the query compiler is often the one specifying what these kernels are (in this case the function that converts all values to float), so it makes sense for the QC to tell the algebra layer hints about the dtype (either the specific types that will result, or to infer the type (if it is necessary, e.g. during query composition) or that we don't know what the resultant types will be, and we don't need to defer them - the "NOHOW" suggestion as I understand).

modin/core/dataframe/algebra/binary.py
arunjose696: Changed .dtypes to  ._modin_frame._dtypes to be aligned with the code.

modin/core/dataframe/algebra/binary.py
```diff
@@ -19,11 +19,95 @@
...
+    if dtype in np.sctypes["int"] + np.sctypes["uint"]:
+        return np.dtype(np.float64)
+    else:
+        return dtype
+
+
+# To precompute datatypes for binary operations which follow pandas find_common_type
+def compute_dtypes_common_cast(first, second) -> np.dtype:
+    """
+    Precompute data types for those binary operations by finding common type between operands.
```
YarShev, line 53:
```suggestion
    Precompute data types for binary operations by finding common type between operands.
```

modin/core/dataframe/algebra/binary.py
```diff
@@ -19,11 +19,95 @@
...
+
+    Parameters
+    ----------
+    first : PandasQueryCompiler
+        Operand dataframe for which the binary operation would be performed later.
```
YarShev, line 58:
```suggestion
        First operand for which the binary operation would be performed later.
```

modin/core/dataframe/algebra/binary.py
```diff
@@ -19,11 +19,95 @@
...
+    second : PandasQueryCompiler
+        Operand dataframe for which the binary operation would be performed later.
```
YarShev, line 60:
```suggestion
        Second operand for which the binary operation would be performed later.
```

modin/core/dataframe/algebra/binary.py
```diff
@@ -96,13 +187,29 @@ def caller(
                         )
                     )
                 else:
+                    if (
+                        other._modin_frame._dtypes is not None
+                        and query_compiler._modin_frame._dtypes is not None
+                        and other.is_series_like is False
```
YarShev, line 193: Why do we need the third condition? Isn't broadcast=False enough?

modin/core/dataframe/algebra/binary.py
```diff
@@ -19,11 +19,95 @@
...
+    first : PandasQueryCompiler
+        Operand dataframe for which the binary operation would be performed later.
+    second : PandasQueryCompiler
+        Operand dataframe for which the binary operation would be performed later.
+
+    Returns
+    -------
+    dtypes
+        The pandas series with precomputed dtypes.
+    """
```
YarShev, line 66: Please also add Notes section and put a comment there that dtypes of the operands are supposed to be known.

modin/core/dataframe/algebra/binary.py
```diff
@@ -19,11 +19,95 @@
...
+    dtypes_first = dict(zip(first.columns, first._modin_frame._dtypes))
+    dtypes_second = dict(zip(second.columns, second._modin_frame._dtypes))
+    columns_first = set(first.columns)
+    columns_second = set(second.columns)
+    common_columns = columns_first.intersection(columns_second)
+    mismatch_columns = columns_first.union(columns_second) - common_columns
+    # If one columns dont match the result of the non matching column would be nan.
```
YarShev, line 73:
```suggestion
    # If at least one column doesn't match, the result of the non matching column would be nan.
```

modin/core/dataframe/algebra/binary.py
```diff
@@ -36,6 +120,13 @@ def register(cls, func, join_type="outer", labels="replace"):
         labels : {"keep", "replace", "drop"}, default: "replace"
             Whether keep labels from left Modin DataFrame, replace them with labels
             from joined DataFrame or drop altogether to make them be computed lazily later.
+        how_compute_dtypes : {"common_cast", "float","bool", None}, default: None
+            How dtypes should be computed.
+            If "common_cast" casts to common dtype of operand columns.
+            If "float" performs type casting by finding common dtype, if the common dtype is any of the
+            integer types perform type casting to float. Used in case of truediv.
+            If "bool" dtypes would be a boolean series with same size as that of operands.
+            If ``None`` do not infer new dtypes (they will be computed manually once accessed).
```
YarShev, line 129:
```suggestion
        infer_dtypes : {"common_cast", "float","bool", None}, default: None
            How dtypes should be inferred.
                * If "common_cast", casts to common dtype of operand columns.
                * If "float", performs type casting by finding common dtype.
                   If the common dtype is any of the integer types, perform type casting to float.
                   Used in case of truediv.
                * If "bool", dtypes would be a boolean series with same size as that of operands.
                * If ``None``, do not infer new dtypes (they will be computed manually once accessed).
```

modin/core/storage_formats/pandas/query_compiler.py
```diff
@@ -375,34 +375,42 @@ def to_numpy(self, **kwargs):
...
+        pandas.DataFrame.combine, how_compute_dtypes="common_cast"
+    )
+    combine_first = Binary.register(
+        pandas.DataFrame.combine_first, how_compute_dtypes="bool"
+    )
+    eq = Binary.register(pandas.DataFrame.eq, how_compute_dtypes="bool")
+    floordiv = Binary.register(
+        pandas.DataFrame.floordiv, how_compute_dtypes="common_cast"
+    )
+    ge = Binary.register(pandas.DataFrame.ge, how_compute_dtypes="bool")
```
YarShev, line 389: Please check the API layer methods where we pass dtypes in to the qc and remove those arguments. We should consolidate dtypes inferreng logic in one place. As an example, https://github.com/modin-project/modin/blob/843b0cf02df9ff76f32cc75749203762633ffb27/modin/pandas/base.py#L1409.

modin/core/dataframe/algebra/binary.py
```diff
@@ -19,11 +19,95 @@
 from .operator import Operator
 
 
+def int_to_float64(dtype: np.dtype) -> np.dtype:
```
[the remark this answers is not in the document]
YarShev: or maybe_coerce_int_to_float64, or maybe_int_to_float64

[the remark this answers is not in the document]
arunjose696: Done

modin/core/storage_formats/pandas/query_compiler.py
arunjose696: Done.

modin/core/dataframe/algebra/binary.py
```diff
@@ -19,11 +19,98 @@
...
+    dtypes = pandas.concat(
+        [
+            dtypes,
+            pandas.Series(
+                [nan_dtype] * (len(mismatch_columns)),
+                index=mismatch_columns,
+            ),
+        ]
+    )
+    dtypes = dtypes.sort_index()
```
YarShev, line 99: Why do we have to sort index?

modin/core/dataframe/algebra/binary.py
```diff
@@ -36,6 +123,14 @@ def register(cls, func, join_type="outer", labels="replace"):
         labels : {"keep", "replace", "drop"}, default: "replace"
             Whether keep labels from left Modin DataFrame, replace them with labels
             from joined DataFrame or drop altogether to make them be computed lazily later.
+        how_compute_dtypes : {"common_cast", "float","bool", None}, default: None
+            How dtypes should be inferred.
+                * If "common_cast", casts to common dtype of operand columns.
+                * If "float", performs type casting by finding common dtype.
+                   If the common dtype is any of the integer types, perform type casting to float.
+                   Used in case of truediv.
```
YarShev, line 131:
```suggestion
                * If "float", performs type casting by finding common dtype.
                  If the common dtype is any of the integer types, perform type casting to float.
                  Used in case of truediv.
```

modin/core/dataframe/algebra/binary.py
```diff
@@ -19,11 +19,98 @@
...
     """Builder class for Binary operator."""
 
     @classmethod
-    def register(cls, func, join_type="outer", labels="replace"):
+    def register(
+        cls,
+        func,
+        join_type="outer",
+        labels="replace",
+        how_compute_dtypes=None,
```
YarShev, line 112:
```suggestion
        infer_dtypes=None,
```
I suggest renaming this to `infer_dtypes`.

YarShev:
@arunjose696, some jobs falied. Please take a look.

modin/core/dataframe/algebra/binary.py
```diff
@@ -19,11 +19,98 @@
...
+    -------
+    dtypes
+        The pandas series with precomputed dtypes.
+
+    Notes
+    -----
+    The dtypes of the operands are supposed to be known.
+    """
+    dtypes_first = dict(zip(first.columns, first._modin_frame._dtypes))
+    dtypes_second = dict(zip(second.columns, second._modin_frame._dtypes))
```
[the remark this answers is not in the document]
arunjose696: I prefer to use .to_dict() method as in first suggestion. In the second approach again a list would have to be created to iterate over the col names with `dtypes_first.keys()`, which makes them take up same space at end,
I feel the .to_dict() method to be more readable.

modin/core/dataframe/algebra/binary.py
arunjose696: Result obtained when performing a binary operation has the combined indices sorted

eg

```
import pandas as pd
df1 = pd.DataFrame({"a": [1, 2],"b":[3,4],"d":[3,4]},dtype=np.float128)  # dtypes: a: int64, b: int64
df2 = pd.DataFrame({"b": [1, 2], "a": [2., 3], "c": [4, 5]},dtype=np.float128) # dtypes a: int64, b: int64, c: int64
df=df1*df2
df.dtypes 
```

Results in 

a    float128
b    float128
c     float64
d     float64
dtype: object

modin/core/dataframe/algebra/binary.py
arunjose696: renamed to infer_dtypes.

modin/core/dataframe/algebra/binary.py
arunjose696: done.

.github/workflows/ci.yml
```diff
@@ -627,7 +627,7 @@ jobs:
       - run: ./.github/workflows/sql_server/set_up_sql_server.sh
       - run: mpiexec -n 1 python -m pytest modin/pandas/test/test_io.py --verbose
       - run: mpiexec -n 1 python -m pytest modin/experimental/pandas/test/test_io_exp.py
-      - run: pip install "dfsql>=0.4.2" "pyparsing<=2.4.7" && mpiexec -n 1 python -m pytest modin/experimental/sql/test/test_sql.py
+      - run: pip install "dfsql>=0.4.2" "pyparsing" && mpiexec -n 1 python -m pytest modin/experimental/sql/test/test_sql.py
```
YarShev, line 630:
```suggestion
      - run: pip install "dfsql>=0.4.2" "pyparsing<=2.4.7" && mpiexec -n 1 python -m pytest modin/experimental/sql/test/test_sql.py
```
Let's revert the change as it doesn't relate to the issue.

.github/workflows/ci.yml
```diff
@@ -720,7 +720,7 @@ jobs:
       - run: ./.github/workflows/sql_server/set_up_sql_server.sh
       - run: python -m pytest modin/pandas/test/test_io.py --verbose
       - run: python -m pytest modin/experimental/pandas/test/test_io_exp.py
-      - run: pip install "dfsql>=0.4.2" "pyparsing<=2.4.7" && python -m pytest modin/experimental/sql/test/test_sql.py
+      - run: pip install "dfsql>=0.4.2" "pyparsing" && python -m pytest modin/experimental/sql/test/test_sql.py
```
YarShev, line 723:
```suggestion
      - run: pip install "dfsql>=0.4.2" "pyparsing<=2.4.7" && python -m pytest modin/experimental/sql/test/test_sql.py
```

modin/core/dataframe/algebra/binary.py
```diff
@@ -36,6 +151,14 @@ def register(cls, func, join_type="outer", labels="replace"):
         labels : {"keep", "replace", "drop"}, default: "replace"
             Whether keep labels from left Modin DataFrame, replace them with labels
             from joined DataFrame or drop altogether to make them be computed lazily later.
+        infer_dtypes : {"common_cast", "float","bool", None}, default: None
```
YarShev, line 154:
```suggestion
        infer_dtypes : {"common_cast", "float", "bool", None}, default: None
```

modin/core/dataframe/algebra/binary.py
```diff
@@ -96,13 +219,26 @@ def caller(
...
+                            dtypes = dtypes.apply(coerce_int_to_float64)
                     return query_compiler.__constructor__(
                         query_compiler._modin_frame.n_ary_op(
                             lambda x, y: func(x, y, *args, **kwargs),
                             [other._modin_frame],
                             join_type=join_type,
+                            dtypes=dtypes,
                         )
                     )
+
```
YarShev, line 241:
```suggestion
```

YarShev (commented):
Left a few comments. Otherwise, LGTM!

YarShev (approved):
@arunjose696, thanks! Great job.

PR #5547, opened 2023-01-17
PERF-#5575: Don't trigger axes computation in `reset_index` function

### LINKED ISSUE #5575: Don't trigger axes computation in `reset_index` function

anmyachev:

## What do these changes do?

- [x] Resolves #5575 
- [x] tests added and passing

### DIFF
--- a/modin/core/dataframe/pandas/dataframe/dataframe.py
+++ b/modin/core/dataframe/pandas/dataframe/dataframe.py
@@ -2684,13 +2684,13 @@ class PandasDataframe(ClassLogger):
         apply_func_args = None
         if pass_cols_to_partitions:
             apply_func_args = (
-                self._columns_cache
+                [self._columns_cache]
                 if self._columns_cache is not None
                 else self._partition_mgr_cls.get_indices(
                     axis=1,
                     partitions=self._partitions,
                     materialize=False,
-                )[1],
+                )[1]
             )
 
         new_partitions = self._partition_mgr_cls.broadcast_axis_partitions(
--- a/modin/core/dataframe/pandas/dataframe/dataframe.py
+++ b/modin/core/dataframe/pandas/dataframe/dataframe.py
@@ -2192,6 +2192,7 @@ class PandasDataframe(ClassLogger):
         dtypes=None,
         keep_partitioning=True,
         sync_labels=True,
+        pass_axis_lengths_to_partitions=False,
     ):
         """
         Perform a function across an entire axis.
@@ -2219,6 +2220,9 @@ class PandasDataframe(ClassLogger):
             Synchronize external indexes (`new_index`, `new_columns`) with internal indexes.
             This could be used when you're certain that the indices in partitions are equal to
             the provided hints in order to save time on syncing them.
+        pass_axis_lengths_to_partitions : bool, default: False
+            Whether pass partition lengths along `axis ^ 1` to the kernel `func`.
+            Note that `func` must be able to obtain `df, *axis_lengths`.
 
         Returns
         -------
@@ -2238,6 +2242,7 @@ class PandasDataframe(ClassLogger):
             other=None,
             keep_partitioning=keep_partitioning,
             sync_labels=sync_labels,
+            pass_axis_lengths_to_partitions=pass_axis_lengths_to_partitions,
         )
 
     @lazy_metadata_decorator(apply_axis="both")
--- a/modin/core/dataframe/pandas/dataframe/dataframe.py
+++ b/modin/core/dataframe/pandas/dataframe/dataframe.py
@@ -2681,6 +2681,18 @@ class PandasDataframe(ClassLogger):
                 axis ^ 1, numeric_indices
             ).keys()
 
+        apply_func_args = None
+        if pass_cols_to_partitions:
+            apply_func_args = (
+                self._columns_cache
+                if self._columns_cache is not None
+                else self._partition_mgr_cls.get_indices(
+                    axis=1,
+                    partitions=self._partitions,
+                    materialize=False,
+                )[1],
+            )
+
         new_partitions = self._partition_mgr_cls.broadcast_axis_partitions(
             axis=axis,
             left=self._partitions,
@@ -2689,7 +2701,7 @@ class PandasDataframe(ClassLogger):
             apply_indices=apply_indices,
             enumerate_partitions=enumerate_partitions,
             keep_partitioning=keep_partitioning,
-            pass_cols_to_partitions=pass_cols_to_partitions,
+            apply_func_args=apply_func_args,
         )
         kw = {"row_lengths": None, "column_widths": None}
         if dtypes == "copy":
--- a/modin/core/dataframe/pandas/dataframe/dataframe.py
+++ b/modin/core/dataframe/pandas/dataframe/dataframe.py
@@ -2663,7 +2663,7 @@ class PandasDataframe(ClassLogger):
             the provided hints in order to save time on syncing them.
         pass_cols_to_partitions : bool, default: False
             Whether pass columns into applied `func` or not.
-            Note that `func` must be able to obtain `*columns` arg.
+            Note that `func` must be able to obtain `df, *columns`.
 
         Returns
         -------
--- a/modin/core/dataframe/pandas/dataframe/dataframe.py
+++ b/modin/core/dataframe/pandas/dataframe/dataframe.py
@@ -2626,7 +2626,7 @@ class PandasDataframe(ClassLogger):
         dtypes=None,
         keep_partitioning=True,
         sync_labels=True,
-        pass_cols_to_partitions=False,
+        pass_axis_lengths_to_partitions=False,
     ):
         """
         Broadcast partitions of `other` Modin DataFrame and apply a function along full axis.
@@ -2661,9 +2661,9 @@ class PandasDataframe(ClassLogger):
             Synchronize external indexes (`new_index`, `new_columns`) with internal indexes.
             This could be used when you're certain that the indices in partitions are equal to
             the provided hints in order to save time on syncing them.
-        pass_cols_to_partitions : bool, default: False
+        pass_axis_lengths_to_partitions : bool, default: False
             Whether pass columns into applied `func` or not.
-            Note that `func` must be able to obtain `df, *columns`.
+            Note that `func` must be able to obtain `df, *axis_lengths`.
 
         Returns
         -------
@@ -2682,16 +2682,21 @@ class PandasDataframe(ClassLogger):
             ).keys()
 
         apply_func_args = None
-        if pass_cols_to_partitions:
-            apply_func_args = (
-                [self._columns_cache]
-                if self._columns_cache is not None
-                else self._partition_mgr_cls.get_indices(
-                    axis=1,
-                    partitions=self._partitions,
-                    materialize=False,
-                )[1]
-            )
+        if pass_axis_lengths_to_partitions:
+            if axis == 0:
+                apply_func_args = (
+                    self._column_widths_cache
+                    if self._column_widths_cache is not None
+                    else [part.width(materialize=False) for part in self._partitions[0]]
+                )
+            else:
+                apply_func_args = (
+                    self._row_lengths_cache
+                    if self._row_lengths_cache is not None
+                    else [
+                        part.length(materialize=False) for part in self._partitions.T[0]
+                    ]
+                )
 
         new_partitions = self._partition_mgr_cls.broadcast_axis_partitions(
             axis=axis,
--- a/modin/core/dataframe/pandas/dataframe/dataframe.py
+++ b/modin/core/dataframe/pandas/dataframe/dataframe.py
@@ -2626,7 +2626,7 @@ class PandasDataframe(ClassLogger):
         dtypes=None,
         keep_partitioning=True,
         sync_labels=True,
-        give_columns=False,
+        pass_cols_to_partitions=False,
     ):
         """
         Broadcast partitions of `other` Modin DataFrame and apply a function along full axis.
@@ -2661,6 +2661,9 @@ class PandasDataframe(ClassLogger):
             Synchronize external indexes (`new_index`, `new_columns`) with internal indexes.
             This could be used when you're certain that the indices in partitions are equal to
             the provided hints in order to save time on syncing them.
+        pass_cols_to_partitions : bool, default: False
+            Whether pass columns into applied `func` or not.
+            Note that `func` must be able to obtain `*columns` arg.
 
         Returns
         -------
@@ -2686,7 +2689,7 @@ class PandasDataframe(ClassLogger):
             apply_indices=apply_indices,
             enumerate_partitions=enumerate_partitions,
             keep_partitioning=keep_partitioning,
-            give_columns=give_columns,
+            pass_cols_to_partitions=pass_cols_to_partitions,
         )
         # Index objects for new object creation. This is shorter than if..else
         kw = self.__make_init_labels_args(new_partitions, new_index, new_columns)
--- a/modin/core/dataframe/pandas/dataframe/dataframe.py
+++ b/modin/core/dataframe/pandas/dataframe/dataframe.py
@@ -2662,7 +2662,7 @@ class PandasDataframe(ClassLogger):
             This could be used when you're certain that the indices in partitions are equal to
             the provided hints in order to save time on syncing them.
         pass_axis_lengths_to_partitions : bool, default: False
-            Whether pass columns into applied `func` or not.
+            Whether pass partition lengths along `axis ^ 1` to the kernel `func`.
             Note that `func` must be able to obtain `df, *axis_lengths`.
 
         Returns
--- a/modin/core/dataframe/pandas/dataframe/dataframe.py
+++ b/modin/core/dataframe/pandas/dataframe/dataframe.py
@@ -2619,6 +2619,7 @@ class PandasDataframe(ClassLogger):
         enumerate_partitions=False,
         dtypes=None,
         keep_partitioning=True,
+        give_columns=False,
     ):
         """
         Broadcast partitions of `other` Modin DataFrame and apply a function along full axis.
@@ -2674,6 +2675,7 @@ class PandasDataframe(ClassLogger):
             apply_indices=apply_indices,
             enumerate_partitions=enumerate_partitions,
             keep_partitioning=keep_partitioning,
+            give_columns=give_columns,
         )
         # Index objects for new object creation. This is shorter than if..else
         kw = self.__make_init_labels_args(new_partitions, new_index, new_columns)
--- a/modin/core/dataframe/pandas/partitioning/partition_manager.py
+++ b/modin/core/dataframe/pandas/partitioning/partition_manager.py
@@ -441,6 +441,14 @@ class PandasDataframePartitionManager(ClassLogger, ABC):
         else:
             num_splits = NPartitions.get()
         preprocessed_map_func = cls.preprocess_func(apply_func)
+
+        give_columns = kwargs.pop("give_columns", False)
+        if give_columns:
+            index_func = lambda df: df.columns  # noqa: E731
+            func = cls.preprocess_func(index_func)
+            new_idx = [idx.apply(func) for idx in left[0]]
+            columns_idx = [index._data for index in new_idx]
+
         left_partitions = cls.axis_partition(left, axis)
         right_partitions = None if right is None else cls.axis_partition(right, axis)
         # For mapping across the entire axis, we don't maintain partitioning because we
@@ -462,6 +470,7 @@ class PandasDataframePartitionManager(ClassLogger, ABC):
             [
                 left_partitions[i].apply(
                     preprocessed_map_func,
+                    *(columns_idx if give_columns else []),
                     **kw,
                     **({"partition_idx": idx} if enumerate_partitions else {}),
                     **kwargs,
--- a/modin/core/dataframe/pandas/partitioning/partition_manager.py
+++ b/modin/core/dataframe/pandas/partitioning/partition_manager.py
@@ -445,7 +445,6 @@ class PandasDataframePartitionManager(ClassLogger, ABC):
         else:
             num_splits = NPartitions.get()
         preprocessed_map_func = cls.preprocess_func(apply_func)
-
         left_partitions = cls.axis_partition(left, axis)
         right_partitions = None if right is None else cls.axis_partition(right, axis)
         # For mapping across the entire axis, we don't maintain partitioning because we
--- a/modin/core/dataframe/pandas/partitioning/partition_manager.py
+++ b/modin/core/dataframe/pandas/partitioning/partition_manager.py
@@ -444,10 +444,10 @@ class PandasDataframePartitionManager(ClassLogger, ABC):
 
         give_columns = kwargs.pop("give_columns", False)
         if give_columns:
-            index_func = lambda df: df.columns  # noqa: E731
-            func = cls.preprocess_func(index_func)
-            new_idx = [idx.apply(func) for idx in left[0]]
-            columns_idx = [index._data for index in new_idx]
+            column_func = lambda df: df.columns  # noqa: E731
+            func = cls.preprocess_func(column_func)
+            partition_cols = [part.apply(func) for part in left[0]]
+            column_refs = [col.list_of_blocks[0] for col in partition_cols]
 
         left_partitions = cls.axis_partition(left, axis)
         right_partitions = None if right is None else cls.axis_partition(right, axis)
@@ -470,7 +470,7 @@ class PandasDataframePartitionManager(ClassLogger, ABC):
             [
                 left_partitions[i].apply(
                     preprocessed_map_func,
-                    *(columns_idx if give_columns else []),
+                    *(column_refs if give_columns else []),
                     **kw,
                     **({"partition_idx": idx} if enumerate_partitions else {}),
                     **kwargs,
--- a/modin/core/dataframe/pandas/partitioning/partition_manager.py
+++ b/modin/core/dataframe/pandas/partitioning/partition_manager.py
@@ -398,7 +398,7 @@ class PandasDataframePartitionManager(ClassLogger, ABC):
         apply_indices=None,
         enumerate_partitions=False,
         lengths=None,
-        pass_cols_to_partitions=False,
+        apply_func_args=None,
         **kwargs,
     ):
         """
@@ -424,9 +424,9 @@ class PandasDataframePartitionManager(ClassLogger, ABC):
             Note that `apply_func` must be able to accept `partition_idx` kwarg.
         lengths : list of ints, default: None
             The list of lengths to shuffle the object.
-        pass_cols_to_partitions : bool, default: False
-            Whether pass columns into applied `func` or not.
-            Note that `func` must be able to obtain `*columns` arg.
+        apply_func_args : bool, optional
+            Whether pass extra args to `func` or not.
+            Note that `func` must be able to obtain `df, *args`.
         **kwargs : dict
             Additional options that could be used by different engines.
 
@@ -445,12 +445,6 @@ class PandasDataframePartitionManager(ClassLogger, ABC):
         else:
             num_splits = NPartitions.get()
         preprocessed_map_func = cls.preprocess_func(apply_func)
-
-        if pass_cols_to_partitions:
-            column_func = lambda df: df.columns  # noqa: E731
-            func = cls.preprocess_func(column_func)
-            partition_cols = [part.apply(func) for part in left[0]]
-            column_refs = [col.list_of_blocks[0] for col in partition_cols]
 
         left_partitions = cls.axis_partition(left, axis)
         right_partitions = None if right is None else cls.axis_partition(right, axis)
@@ -473,7 +467,7 @@ class PandasDataframePartitionManager(ClassLogger, ABC):
             [
                 left_partitions[i].apply(
                     preprocessed_map_func,
-                    *(column_refs if pass_cols_to_partitions else []),
+                    *(apply_func_args if apply_func_args else []),
                     **kw,
                     **({"partition_idx": idx} if enumerate_partitions else {}),
                     **kwargs,
@@ -865,7 +859,7 @@ class PandasDataframePartitionManager(ClassLogger, ABC):
             partition.wait()
 
     @classmethod
-    def get_indices(cls, axis, partitions, index_func=None):
+    def get_indices(cls, axis, partitions, index_func=None, materialize=True):
         """
         Get the internal indices stored in the partitions.
 
@@ -877,13 +871,16 @@ class PandasDataframePartitionManager(ClassLogger, ABC):
             NumPy array with PandasDataframePartition's.
         index_func : callable, default: None
             The function to be used to extract the indices.
+        materialize : bool, default: True
+            Whether materialize indices or not.
 
         Returns
         -------
-        pandas.Index
-            A pandas Index object.
-        list of pandas.Index
+        pandas.Index or None
+            A pandas Index object. None if `materialize==False`.
+        list of pandas.Index or list of futures
             The list of internal indices for each partition.
+            The list of futures if `materialize==False`.
 
         Notes
         -----
@@ -898,6 +895,9 @@ class PandasDataframePartitionManager(ClassLogger, ABC):
         target = partitions.T if axis == 0 else partitions
         if len(target):
             new_idx = [idx.apply(func) for idx in target[0]]
+            if not materialize:
+                refs = [part_of_index.list_of_blocks[0] for part_of_index in new_idx]
+                return None, refs
             new_idx = cls.get_objects_from_partitions(new_idx)
         else:
             new_idx = [pandas.Index([])]
--- a/modin/core/dataframe/pandas/partitioning/partition_manager.py
+++ b/modin/core/dataframe/pandas/partitioning/partition_manager.py
@@ -398,6 +398,7 @@ class PandasDataframePartitionManager(ClassLogger, ABC):
         apply_indices=None,
         enumerate_partitions=False,
         lengths=None,
+        pass_cols_to_partitions=False,
         **kwargs,
     ):
         """
@@ -423,6 +424,9 @@ class PandasDataframePartitionManager(ClassLogger, ABC):
             Note that `apply_func` must be able to accept `partition_idx` kwarg.
         lengths : list of ints, default: None
             The list of lengths to shuffle the object.
+        pass_cols_to_partitions : bool, default: False
+            Whether pass columns into applied `func` or not.
+            Note that `func` must be able to obtain `*columns` arg.
         **kwargs : dict
             Additional options that could be used by different engines.
 
@@ -442,8 +446,7 @@ class PandasDataframePartitionManager(ClassLogger, ABC):
             num_splits = NPartitions.get()
         preprocessed_map_func = cls.preprocess_func(apply_func)
 
-        give_columns = kwargs.pop("give_columns", False)
-        if give_columns:
+        if pass_cols_to_partitions:
             column_func = lambda df: df.columns  # noqa: E731
             func = cls.preprocess_func(column_func)
             partition_cols = [part.apply(func) for part in left[0]]
@@ -470,7 +473,7 @@ class PandasDataframePartitionManager(ClassLogger, ABC):
             [
                 left_partitions[i].apply(
                     preprocessed_map_func,
-                    *(column_refs if give_columns else []),
+                    *(column_refs if pass_cols_to_partitions else []),
                     **kw,
                     **({"partition_idx": idx} if enumerate_partitions else {}),
                     **kwargs,
--- a/modin/core/dataframe/pandas/partitioning/partition_manager.py
+++ b/modin/core/dataframe/pandas/partitioning/partition_manager.py
@@ -878,7 +878,7 @@ class PandasDataframePartitionManager(ClassLogger, ABC):
             A pandas Index object. None if `materialize==False`.
         list of pandas.Index or list of futures
             The list of internal indices for each partition.
-            The list of futures if `materialize==False`.
+            The list of futures if `materialize==False` and `partitions` is not empty.
 
         Notes
         -----
--- a/modin/core/dataframe/pandas/partitioning/partition_manager.py
+++ b/modin/core/dataframe/pandas/partitioning/partition_manager.py
@@ -857,7 +857,7 @@ class PandasDataframePartitionManager(ClassLogger, ABC):
             partition.wait()
 
     @classmethod
-    def get_indices(cls, axis, partitions, index_func=None, materialize=True):
+    def get_indices(cls, axis, partitions, index_func=None):
         """
         Get the internal indices stored in the partitions.
 
@@ -869,16 +869,13 @@ class PandasDataframePartitionManager(ClassLogger, ABC):
             NumPy array with PandasDataframePartition's.
         index_func : callable, default: None
             The function to be used to extract the indices.
-        materialize : bool, default: True
-            Whether materialize indices or not.
 
         Returns
         -------
-        pandas.Index or None
-            A pandas Index object. None if `materialize==False`.
-        list of pandas.Index or list of futures
+        pandas.Index
+            A pandas Index object.
+        list of pandas.Index
             The list of internal indices for each partition.
-            The list of futures if `materialize==False` and `partitions` is not empty.
 
         Notes
         -----
@@ -893,9 +890,6 @@ class PandasDataframePartitionManager(ClassLogger, ABC):
         target = partitions.T if axis == 0 else partitions
         if len(target):
             new_idx = [idx.apply(func) for idx in target[0]]
-            if not materialize:
-                refs = [part_of_index.list_of_blocks[0] for part_of_index in new_idx]
-                return None, refs
             new_idx = cls.get_objects_from_partitions(new_idx)
         else:
             new_idx = [pandas.Index([])]
--- a/modin/core/dataframe/pandas/partitioning/partition_manager.py
+++ b/modin/core/dataframe/pandas/partitioning/partition_manager.py
@@ -424,9 +424,8 @@ class PandasDataframePartitionManager(ClassLogger, ABC):
             Note that `apply_func` must be able to accept `partition_idx` kwarg.
         lengths : list of ints, default: None
             The list of lengths to shuffle the object.
-        apply_func_args : bool, optional
-            Whether pass extra args to `func` or not.
-            Note that `func` must be able to obtain `df, *args`.
+        apply_func_args : list-like, optional
+            Positional arguments to pass to the `func`.
         **kwargs : dict
             Additional options that could be used by different engines.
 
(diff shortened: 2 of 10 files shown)


modin/core/storage_formats/pandas/query_compiler.py
```diff
@@ -526,6 +526,12 @@ def reindex(self, axis, labels, **kwargs):
         return self.__constructor__(new_modin_frame)
 
     def reset_index(self, **kwargs):
+        if self._modin_frame._index_cache is None:
+            new_self = Map.register(
+                lambda df, *args, **kwargs: df.reset_index(**kwargs)
+            )(self, **kwargs)
```
dchigarev, line 532: We have to apply this function across the whole column axis, right? `new_self = self._modin_frame.apply_full_axis(axis=0, lambda df: df.reset_index(**kwargs)` Otherwise we'll have an incorrect index in partitions:
```
  a b
1 . .
2 . .    <- part1
------
3 . .    <- part2
4 . .

If we use 'Map' to reset indices:
  a b
0 . .
1 . .    <- part1
------
0 . .    <- part2
1 . .

When it should be:
  a b
0 . .
1 . .    <- part1
------
2 . .    <- part2
3 . .
```

modin/pandas/base.py
```diff
@@ -2393,7 +2393,8 @@ def reset_index(
         # insert a dropped index into a DataFrame if these columns already
         # exist.
         if (
-            not drop
+            False  # have to avoid this check somehow
+            and not drop
             and not self._query_compiler.has_multiindex()
```
dchigarev, line 2398: As here `has_multiindex()` decides whether to access the index we may have a flag in the query compiler indicating if the index is MI. I bet most of the preceding functions are able to predict whether they produce MI or not and set an appropriate flag.

modin/core/storage_formats/pandas/query_compiler.py
anmyachev: You are right, but it will become significantly slower.

modin/pandas/base.py
```diff
@@ -2389,24 +2389,14 @@ def reset_index(
...
         inplace = validate_bool_kwarg(inplace, "inplace")
-        # Error checking for matching pandas. Pandas does not allow you to
-        # insert a dropped index into a DataFrame if these columns already
-        # exist.
-        if (
-            not drop
-            and not self._query_compiler.has_multiindex()
-            and all(n in self.columns for n in ["level_0", "index"])
-        ):
-            raise ValueError("cannot insert level_0, already exists")
```
anmyachev, line 2400: This check has been moved to the query compiler level.

modin/core/storage_formats/pandas/query_compiler.py
```diff
@@ -538,6 +538,45 @@ def reindex(self, axis, labels, **kwargs):
         return self.__constructor__(new_modin_frame)
 
     def reset_index(self, **kwargs):
+        if self._modin_frame._index_cache is None:
+
+            def _reset(df, *columns_idx, partition_idx):
+                _kw = dict(kwargs)
+                if len(columns_idx) > 1 and partition_idx == 0:
```
dchigarev, line 545: could you please provide with an example of what's going on here?

I don't understand why we can't just do:
```python
return df.reset_index(drop=drop or partition_idx != 0)
```

modin/pandas/base.py
anmyachev: Because, depending on the implementation, we may not want to trigger the calculation of indexes (in the main process) that are checked in the `has_multiindex` function. HDK engine, for example, has its own implementation of this check.

modin/core/storage_formats/pandas/query_compiler.py
anmyachev: If the index does not have a name, but there is a column named `index`, then pandas gives `level_0` name to the new column. That is, knowledge of the full axis of the columns is required. Perhaps there is something similar for a multi-index. This implementation helps us get rid of the implementation details of this function in pandas. 

P.S. we have a test for the case: `test_reset_index`.

modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -2626,6 +2626,7 @@ def broadcast_apply_full_axis(
         dtypes=None,
         keep_partitioning=True,
         sync_labels=True,
+        give_columns=False,
```
YarShev, line 2629:
```suggestion
        pass_cols_to_partitions=False,
```
\+ docstring

modin/core/dataframe/pandas/partitioning/partition_manager.py
```diff
@@ -441,6 +441,14 @@ def broadcast_axis_partitions(
         else:
             num_splits = NPartitions.get()
         preprocessed_map_func = cls.preprocess_func(apply_func)
+
+        give_columns = kwargs.pop("give_columns", False)
+        if give_columns:
+            index_func = lambda df: df.columns  # noqa: E731
+            func = cls.preprocess_func(index_func)
+            new_idx = [idx.apply(func) for idx in left[0]]
+            columns_idx = [index._data for index in new_idx]
```
YarShev, line 450:
```suggestion
        if give_columns:
            column_func = lambda df: df.columns  # noqa: E731
            func = cls.preprocess_func(column_func)
            partition_cols = [part.apply(func) for part in left[0]]
            column_refs = [col.list_of_blocks[0] for col in partition_cols]
```

modin/core/storage_formats/pandas/query_compiler.py
```diff
@@ -538,6 +538,45 @@ def reindex(self, axis, labels, **kwargs):
         return self.__constructor__(new_modin_frame)
 
     def reset_index(self, **kwargs):
+        if self._modin_frame._index_cache is None:
+
+            def _reset(df, *columns_idx, partition_idx):
```
YarShev, line 543:
```suggestion
            def _reset(df, *columns, partition_idx=None):
```

modin/core/storage_formats/pandas/query_compiler.py
```diff
@@ -538,6 +538,45 @@ def reindex(self, axis, labels, **kwargs):
...
+                        .columns
+                    )
+                    len_new_cols = len(cols) - len(idx)
+
+                if partition_idx != 0:
+                    _kw["drop"] = True
+                res = df.reset_index(**_kw)
+                if len(columns_idx) > 1 and partition_idx == 0:
+                    res.columns = cols[: len(columns_idx[0]) + len_new_cols]
+                return res
```
YarShev, line 559:
```suggestion
                    old_cols = columns[0].append(columns[1:])
                    new_cols = (
                        pandas.DataFrame(index=df.index, columns=old_cols)
                        .reset_index(**kwargs)
                        .columns
                    )
                    cols_diff_len = len(new_cols) - len(old_cols)

                if partition_idx != 0:
                    _kw["drop"] = True
                result = df.reset_index(**_kw)
                if len(columns) > 1 and partition_idx == 0:
                    result.columns = new_cols[: len(columns[0]) + cols_diff_len]
                return result
```

modin/core/storage_formats/pandas/query_compiler.py
```diff
@@ -538,6 +538,45 @@ def reindex(self, axis, labels, **kwargs):
...
+
+        # Error checking for matching pandas. Pandas does not allow you to
+        # insert a dropped index into a DataFrame if these columns already
+        # exist.
+        if (
+            not kwargs["drop"]
+            and not self.has_multiindex()
+            and all(n in self.columns for n in ["level_0", "index"])
+        ):
+            raise ValueError("cannot insert level_0, already exists")
```
YarShev, line 579: Haven't we tested it yet?

YarShev (commented):
Left a few comments. Also, I would like to see some perf measurements.

modin/core/storage_formats/pandas/query_compiler.py
anmyachev: It's better to require `partition_idx` parameter to avoid errors.

modin/core/storage_formats/pandas/query_compiler.py
anmyachev: We tested it, but the check was not at the level of query compiler, but higher.

YarShev: Why doesn't codecov see we check this branch?

modin/pandas/test/dataframe/test_indexing.py
```diff
@@ -1534,18 +1551,31 @@ def test_reset_index_with_named_index(index_levels_names_max_levels):
     )
     modin_df.index.name = pandas_df.index.name = index_name
     df_equals(modin_df, pandas_df)
+    if test_async_reset_index:
+        # The change in index is not automatically handled by Modin.
+        modin_df._query_compiler._modin_frame._propagate_index_objs(axis=0)
```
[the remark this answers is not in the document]
YarShev: Don't we want to call `_propagate_index_objs` even if we change index.name?

modin/core/storage_formats/pandas/query_compiler.py
anmyachev: Why? If the parameter is no longer passed, then we will see it: `TypeError: _reset() missing 1 required keyword-only argument: 'partition_idx'`.

What problem are you trying to solve by making this parameter optional?

modin/core/dataframe/pandas/partitioning/partition_manager.py
```diff
@@ -462,6 +470,7 @@ def broadcast_axis_partitions(
             [
                 left_partitions[i].apply(
                     preprocessed_map_func,
+                    *(columns_idx if give_columns else []),
```
anmyachev, line 473:
```suggestion
                    *(column_refs if give_columns else []),4
```

modin/pandas/test/dataframe/test_indexing.py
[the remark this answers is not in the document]
anmyachev: This is how it works now. Or do you mean that regardless of the value of `test_async_reset_index` variable?

modin/core/storage_formats/pandas/query_compiler.py
anmyachev, line 543:
```suggestion
            def _reset(df, *columns, partition_idx):
```

modin/core/dataframe/pandas/partitioning/partition_manager.py
```diff
@@ -441,6 +445,13 @@ def broadcast_axis_partitions(
         else:
             num_splits = NPartitions.get()
         preprocessed_map_func = cls.preprocess_func(apply_func)
+
+        if pass_cols_to_partitions:
+            column_func = lambda df: df.columns  # noqa: E731
+            func = cls.preprocess_func(column_func)
+            partition_cols = [part.apply(func) for part in left[0]]
+            column_refs = [col.list_of_blocks[0] for col in partition_cols]
```
dchigarev, line 453: Maybe it would be better to handle this parameter at the `modin_frame` level and then just pass the `column_refs` as a `func_args` to the partition manager (yeah, we would still have to introduce a new parameter to `part_mgr.broadcast_apply(...func_args...)` but it would have more general appliance).

If we handle the `pass_cols_to_partitions` at the higher level we can easily dispatch on whether to compute refs or get the values from `._columns_cache` if we have some.

modin/core/storage_formats/pandas/query_compiler.py
anmyachev: I don't see this problem anymore.

modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -2660,6 +2661,9 @@ def broadcast_apply_full_axis(
             Synchronize external indexes (`new_index`, `new_columns`) with internal indexes.
             This could be used when you're certain that the indices in partitions are equal to
             the provided hints in order to save time on syncing them.
+        pass_cols_to_partitions : bool, default: False
+            Whether pass columns into applied `func` or not.
+            Note that `func` must be able to obtain `*columns` arg.
```
anmyachev, line 2666:
```suggestion
            Note that `func` must be able to obtain `df, *columns`.
```

modin/core/dataframe/pandas/partitioning/partition_manager.py
```diff
@@ -441,6 +445,7 @@ def broadcast_axis_partitions(
         else:
             num_splits = NPartitions.get()
         preprocessed_map_func = cls.preprocess_func(apply_func)
+
```
anmyachev, line 448:
```suggestion
```

modin/core/dataframe/pandas/partitioning/partition_manager.py
anmyachev: @dchigarev I raised the processing of this parameter to the dataframe level. Please review.

dchigarev (commented):
overall looks good!

The only thing I'm worried a bit is that you moved the pandas-specific validity check to the query compiler from the high-level API level. It's an upsetting trend that we're constantly moving pandas-depending stuff to the query compiler thus losing the layers' separatability. 

An initial sense of `modin.pandas` layer was sanitizing inputs/preprocessing parameters/catching errors thus being able to pass to the QC much more standardized and clean from error inputs. Now it seems that the role of matching pandas behavior goes more and more to the backend's responsibility which is quite sad and certainly doesn't scale well with the philosophy of having various amounts of execution backends in modin. Hope we would find a way to gently delegate these error checkings to the backend without making them know a lot about pandas and not losing in performance at the same time. But of course, it's a topic for a separate discussion.

modin/pandas/test/dataframe/test_indexing.py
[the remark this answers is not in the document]
YarShev: I mean in general. If we called `_propagate_index_objs` only in the tests regardless of the value of `test_async_reset_index` variable, we would hide possible errors outside of the tests. Does `_propagate_index_objs` really introduce significant slowdown for further operations when doing `modin_df.index.name = index_name`?

modin/core/storage_formats/pandas/query_compiler.py
YarShev: If we start to pass `partition_idx` in to remote functions as part of *args at the core dataframe layer, that will break this code. Since currently we pass `partition_idx` in to remote functions as part of **kwargs, I would prefer to have the named argument in the function definition.

modin/core/storage_formats/pandas/query_compiler.py
YarShev: Why, I am still seeing.

```
modin/core/storage_formats/pandas/query_compiler.py#L582

Added line #L582 was not covered by tests
```

modin/pandas/test/dataframe/test_indexing.py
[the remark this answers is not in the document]
anmyachev: I can only say that this will especially affect dataframes with complex data types, for which zero copy serialization will not work. Impact research should be done as part of the work with https://github.com/modin-project/modin/issues/3941, I guess :)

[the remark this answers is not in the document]
dchigarev: I would like to see the issue to be mentioned in this in-code comment then, explaining that this is a hack to workaround the known problem

modin/core/storage_formats/pandas/query_compiler.py
anmyachev: I tried your suggestion on artificial examples. Don't you think that the first option has more expected behavior than the second option?

```python
# current definition of `_reset` function (example)
>>> def test(*args, idx):
...   print(args, "\n", idx)

>>> test(1,2,3, idx=2)
(1, 2, 3)
 2

>>> test(1,2,3, 2)
TypeError: test() missing 1 required keyword-only argument: 'idx'

# your suggestion about definition of `_reset` function (example)
>>> def test(*args, idx=None):
...   print(args, "\n", idx)

>>> test(1,2,3, idx=2)
(1, 2, 3)
 2

>>> test(1,2,3, 2)
(1, 2, 3, 2)
 None
```

modin/core/storage_formats/pandas/query_compiler.py
```diff
@@ -541,6 +541,45 @@ def reindex(self, axis, labels, **kwargs):
...
+                    new_cols = (
+                        pandas.DataFrame(index=df.index, columns=old_cols)
+                        .reset_index(**kwargs)
+                        .columns
+                    )
+                    cols_diff_len = len(new_cols) - len(old_cols)
+
+                if partition_idx != 0:
+                    _kw["drop"] = True
+                result = df.reset_index(**_kw)
```
anmyachev, line 559:
```suggestion
                    kw["drop"] = True
                result = df.reset_index(**kw)
```

modin/core/storage_formats/pandas/query_compiler.py
```diff
@@ -541,6 +541,45 @@ def reindex(self, axis, labels, **kwargs):
...
+                if len(columns) > 0 and partition_idx == 0:
+                    result.columns = new_cols[: len(df.columns) + cols_diff_len]
```
anmyachev, line 561:
```suggestion
                    result.columns = new_cols[: len(df.columns) + num_inserted_cols]
```

modin/core/storage_formats/pandas/query_compiler.py
anmyachev: This check does not work correctly. It does not take into account that the index may have a name other than `None`.

I will create a issue for this after we merge this pull request to include its new location in the issue description.

P.S. Apparently, it was the exception that was not tested before, or something changed, but before my pull request.

anmyachev:
> overall looks good!
> 
> The only thing I'm worried a bit is that you moved the pandas-specific validity check to the query compiler from the high-level API level. It's an upsetting trend that we're constantly moving pandas-depending stuff to the query compiler thus losing the layers' separatability.
> 
> An initial sense of `modin.pandas` layer was sanitizing inputs/preprocessing parameters/catching errors thus being able to pass to the QC much more standardized and clean from error inputs. Now it seems that the role of matching pandas behavior goes more and more to the backend's responsibility which is quite sad and certainly doesn't scale well with the philosophy of having various amounts of execution backends in modin. Hope we would find a way to gently delegate these error checkings to the backend without making them know a lot about pandas and not losing in performance at the same time. But of course, it's a topic for a separate discussion.

The main reason is that the frontend code is not designed for asynchronous execution.

In Modin we already have the following code:

```python
# avoid materializing columns in lazy mode. the query compiler
# will handle errors where dtype dict includes keys that are not
# in columns.
if (
    not self._query_compiler.lazy_execution
    and not set(dtype.keys()).issubset(set(self._query_compiler.columns))
    and errors == "raise"
):
    raise KeyError(
        "Only a column name can be used for the key in "
        + "a dtype mappings argument."
    )
```

Perhaps I can do the same for the check. However, in the future, this will lead to the fact that there will be at least 2 code branches for each action: asynchronous and synchronous.

dchigarev:
> Perhaps I can do the same for the check.

I'm okay with leaving things as they are in this PR for now. Just wanted to express my soul pain regarding this trend :D

modin/pandas/base.py
```diff
@@ -2394,19 +2394,19 @@ def reset_index(
         # exist.
         if (
             not drop
+            and not self._query_compiler.lazy_execution
             and not self._query_compiler.has_multiindex()
             and all(n in self.columns for n in ["level_0", "index"])
         ):
             raise ValueError("cannot insert level_0, already exists")
-        else:
```
anmyachev, line 2401: No need to use `else` statement here.

anmyachev:
A full axis implementation of `reset_index` can bring slowdowns even if the execution is asynchronous. We should use the map operator. To do this, I will also need to transfer index lengths asynchronously in order to correctly create `RangeIndex` in each partition. Found by @dchigarev, thanks!

modin/core/storage_formats/pandas/query_compiler.py
```diff
@@ -541,6 +541,38 @@ def reindex(self, axis, labels, **kwargs):
         return self.__constructor__(new_modin_frame)
 
     def reset_index(self, **kwargs):
+        if self.lazy_execution:
```
YarShev, line 544: This is false for distributed engines. What are we trying to speed up?

YarShev (commented):
Ping me when PR is ready for review.

modin/core/storage_formats/pandas/query_compiler.py
dchigarev: no, it is conditionally True or False depending on whether the frame has delayed computations:
https://github.com/modin-project/modin/blob/4f78260d91042497b5d40305c3647bb00c482b14/modin/core/storage_formats/pandas/query_compiler.py#L254-L266

Here we want to delegate the logic of error-checking to the kernel function in case there's not enough metadata to do so in the main process

dchigarev:
> Ping me when PR is ready for review.

@YarShev it's ready for review

modin/core/storage_formats/pandas/query_compiler.py
YarShev: We know in advance that the keyword arg can't be None for the kernel and can check it if the higher layers pass a wrong value.

```python
def test(*args, idx=None):
    if idx is None:
        raise Exception
    print(args, "\n", idx)
```

modin/core/storage_formats/pandas/query_compiler.py
YarShev: What logic of error-checking do you mean? I don't see anything related in the kernel.

modin/core/storage_formats/pandas/query_compiler.py
```diff
@@ -541,6 +541,38 @@ def reindex(self, axis, labels, **kwargs):
...
+                    df.index = pandas.RangeIndex(start, stop)
+                return df
+
+            if self._modin_frame._columns_cache is not None and kwargs["drop"]:
+                new_columns = self._modin_frame._columns_cache
+            else:
+                new_columns = None
+
+            return self.__constructor__(
+                self._modin_frame.broadcast_apply_full_axis(
```
YarShev, line 565:
```suggestion
                self._modin_frame.apply_full_axis(
```
Change this and propagate the new added parameters to both `apply_full_axis` and `broadcast_apply_full_axis`. I think all the parameters should be in sync for the two methods. Or we could leave only one of the methods.

modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -2626,6 +2626,7 @@ def broadcast_apply_full_axis(
         dtypes=None,
         keep_partitioning=True,
         sync_labels=True,
+        pass_axis_lengths_to_partitions=False,
```
YarShev, line 2629: Why did we refuse from pass_cols_to_partitions? What if we will want to pass rows to partitions in future regardless of `axis` value?

modin/core/execution/dask/implementations/pandas_on_dask/partitioning/partition.py
```diff
@@ -228,33 +228,47 @@ def preprocess_func(cls, func):
...
-    def width(self):
+    def width(self, materialize=True):
         """
         Get the width of the object wrapped by the partition.
 
+        Parameters
+        ----------
+        materialize : bool, default: True
+            Whether to forcibly materialize the result into an integer. If ``False``
+            was specified may return a future of the result if it hasn't been
```
YarShev, line 261:
```suggestion
            was specified, may return a future of the result if it hasn't been
```

modin/core/execution/ray/implementations/cudf_on_ray/partitioning/partition.py
```diff
@@ -177,31 +178,51 @@ def preprocess_func(cls, func):
...
-    def length(self):
+    def length(self, materialize=True):
         """
         Get the length of the object wrapped by this partition.
 
+        Parameters
+        ----------
+        materialize : bool, default: True
+            Whether to forcibly materialize the result into an integer. If ``False``
+            was specified may return a future of the result if it hasn't been
```
YarShev, line 189:
```suggestion
            was specified, may return a future of the result if it hasn't been
```

modin/core/execution/ray/implementations/cudf_on_ray/partitioning/partition.py
```diff
@@ -177,31 +178,51 @@ def preprocess_func(cls, func):
...
-    def width(self):
+    def width(self, materialize=True):
         """
         Get the width of the object wrapped by this partition.
 
+        Parameters
+        ----------
+        materialize : bool, default: True
+            Whether to forcibly materialize the result into an integer. If ``False``
+            was specified may return a future of the result if it hasn't been
```
YarShev, line 212:
```suggestion
            was specified, may return a future of the result if it hasn't been
```

modin/core/execution/ray/implementations/pandas_on_ray/partitioning/partition.py
```diff
@@ -262,13 +262,20 @@ def preprocess_func(cls, func):
...
-    def length(self):
+    def length(self, materialize=True):
         """
         Get the length of the object wrapped by this partition.
 
+        Parameters
+        ----------
+        materialize : bool, default: True
+            Whether to forcibly materialize the result into an integer. If ``False``
+            was specified may return a future of the result if it hasn't been
```
YarShev, line 273:
```suggestion
            was specified, may return a future of the result if it hasn't been
```

modin/core/execution/ray/implementations/pandas_on_ray/partitioning/partition.py
```diff
@@ -278,17 +285,24 @@ def length(self):
...
-    def width(self):
+    def width(self, materialize=True):
         """
         Get the width of the object wrapped by the partition.
 
+        Parameters
+        ----------
+        materialize : bool, default: True
+            Whether to forcibly materialize the result into an integer. If ``False``
+            was specified may return a future of the result if it hasn't been
```
YarShev, line 300:
```suggestion
            was specified, may return a future of the result if it hasn't been
```

modin/core/execution/unidist/implementations/pandas_on_unidist/partitioning/partition.py
```diff
@@ -237,13 +237,20 @@ def preprocess_func(cls, func):
...
-    def length(self):
+    def length(self, materialize=True):
         """
         Get the length of the object wrapped by this partition.
 
+        Parameters
+        ----------
+        materialize : bool, default: True
+            Whether to forcibly materialize the result into an integer. If ``False``
+            was specified may return a future of the result if it hasn't been
```
YarShev, line 248:
```suggestion
            was specified, may return a future of the result if it hasn't been
```

modin/core/execution/unidist/implementations/pandas_on_unidist/partitioning/partition.py
```diff
@@ -254,17 +261,24 @@ def length(self):
...
-    def width(self):
+    def width(self, materialize=True):
         """
         Get the width of the object wrapped by the partition.
 
+        Parameters
+        ----------
+        materialize : bool, default: True
+            Whether to forcibly materialize the result into an integer. If ``False``
+            was specified may return a future of the result if it hasn't been
```
YarShev, line 276:
```suggestion
            was specified, may return a future of the result if it hasn't been
```

modin/core/execution/dask/implementations/pandas_on_dask/partitioning/partition.py
```diff
@@ -228,33 +228,47 @@ def preprocess_func(cls, func):
...
+            materialized yet.
+
         Returns
         -------
-        int
+        int or its Future
```
YarShev, line 244:
```suggestion
        int or distributed.Future
```

modin/core/execution/dask/implementations/pandas_on_dask/partitioning/partition.py
```diff
@@ -228,33 +228,47 @@ def preprocess_func(cls, func):
...
```
YarShev, line 266:
```suggestion
        int or distributed.Future
```

modin/core/execution/unidist/implementations/pandas_on_unidist/partitioning/partition.py
```diff
@@ -237,13 +237,20 @@ def preprocess_func(cls, func):
...
+            materialized yet.
+
         Returns
         -------
-        int
+        int or its Future
```
YarShev, line 253:
```suggestion
        int or unidist.ObjectRef
```

modin/core/execution/unidist/implementations/pandas_on_unidist/partitioning/partition.py
```diff
@@ -254,17 +261,24 @@ def length(self):
...
```
YarShev, line 281:
```suggestion
        int or unidist.ObjectRef
```

modin/core/storage_formats/pandas/query_compiler.py
YarShev: ping

modin/core/storage_formats/pandas/query_compiler.py
dchigarev: the kernel does this natively since we're using pandas itself (whose `.reset_index` implementation contains the logic), the difference here that in case of lazy execution we'll get an error at the time we'll materialize the results (which seems logical for lazy execution)

modin/core/dataframe/pandas/dataframe/dataframe.py
dchigarev: > Why did we refuse from pass_cols_to_partitions?

We don't need it anymore since the function is now row-wise

> What if we will want to pass rows to partitions in future regardless of axis value?

It's unlikely we would want that. We don't need column widths/columns if the `axis=1` since the whole column axis is available in the kernel and the same for `axis=0`. What we potentially want to have when applying a full-axis function is some knowledge about an opposite axis. In this PR this "knowledge" was chosen to be partition shapes.

modin/core/dataframe/pandas/partitioning/partition_manager.py
```diff
@@ -423,6 +424,8 @@ def broadcast_axis_partitions(
             Note that `apply_func` must be able to accept `partition_idx` kwarg.
         lengths : list of ints, default: None
             The list of lengths to shuffle the object.
+        apply_func_args : list-like, optional
```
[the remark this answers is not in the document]
dchigarev: idk, the `*args` approach would look confusing in my opinion:
```python
part_mgr.broadcast_apply(refs_list, 42, None, axis=0, func=func)
part_mgr.broadcast_apply(axis=0, func=func, apply_func_args=[refs_list, 42, None])
```

anmyachev:
@dchigarev your implementation of `reset_index` works well for my workload.

modin/core/storage_formats/pandas/query_compiler.py
anmyachev: Why write our own check if python can check it itself, at the time of the function call? (python will raise `TypeError: test() missing 1 required keyword-only argument: 'idx'` if we start to pass `idx` into remote functions as part of `*args`)

modin/core/dataframe/pandas/dataframe/dataframe.py
anmyachev: Any suggestion on how to make this option more general? Maybe in the form of a callback, to which it will be possible to pass the `PandasDataframe` and return an arbitrary list of futures that remote function will have to accept?

YarShev: Let's leave as is for now.

modin/core/dataframe/pandas/partitioning/partition_manager.py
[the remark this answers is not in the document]
YarShev: We should  then change `**kwargs` to `apply_func kwargs` to have the logic aligned. Create an issue for this please.

YarShev:
Ping me when all the unresolved comments are addressed.

modin/core/storage_formats/pandas/query_compiler.py
anmyachev: In the case of `def test(*args, idx):` `idx` - Keyword-only argument. 

"Keyword-only arguments are not required to have a default value." from https://peps.python.org/pep-3102/#specification

YarShev: I don't like such optional assumptions in any specification at all. That relaxes users and gives them freedom to do what they want, which may lead to misunderstanding the code by other developers/users.

anmyachev: What optional assumptions are you talking about? And how does it relax users that this parameter does not have a default value?

YarShev: > What optional assumptions are you talking about?

Exactly this -> "Keyword-only arguments are not required to have a default value.", i.e., what not required is optional.

> And how does it relax users that this parameter does not have a default value?

In this concrete case it requires THE developer to know the function declaration to work around the problem and pass the argument with no issue.

I wonder if everyone is okay with the current code? Maybe I am too meticulous.

anmyachev: > Exactly this -> "Keyword-only arguments are not required to have a default value.", i.e., what not required is optional.

It says here that this kind of parameter does not need a default value in the function signature at all, however, when calling, the user **must** specify some value and only in the following format: `idx=some_value`.

dchigarev: Wow, such a big discussion you've had here :D

Personally, I would like to leave it as it is. Given [this comment](https://github.com/modin-project/modin/pull/5547#discussion_r1087860867), it really seems more reasonable to have this parameter as a non-optional keyword argument.

modin/pandas/test/dataframe/test_indexing.py
```diff
@@ -1251,17 +1251,25 @@ def test_reindex_multiindex():
     df_equals(modin_result, pandas_result)
 
 
+@pytest.mark.parametrize("use_get_indices", [False, True])
```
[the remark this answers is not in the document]
anmyachev: removed

modin/core/storage_formats/pandas/query_compiler.py
YarShev: cc @mvashishtha, @vnlitvinov, I would like to listen your thoughts as well.

modin/core/storage_formats/pandas/query_compiler.py
anmyachev: done

modin/core/dataframe/pandas/dataframe/dataframe.py
anmyachev: ok

modin/core/dataframe/pandas/partitioning/partition_manager.py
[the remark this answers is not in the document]
anmyachev: @YarShev some of the parameters from `kwargs` dictionary are auxiliary and are used at higher levels without being passed to the custom function. Are you sure we need to rename it?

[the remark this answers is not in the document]
YarShev: I would prefer to rename it.

modin/core/storage_formats/pandas/query_compiler.py
vnlitvinov: I'm siding with **Anatoly** here, Python `TypeError` message on function misuse would be clear enough to tell the developer what to do with an error.

Come on, this feature is from Python 3.0 ("point zero"!), accepted in 2006. It's time to break mental compatibility with 2.7 😉

modin/core/dataframe/pandas/partitioning/partition_manager.py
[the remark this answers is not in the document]
anmyachev: https://github.com/modin-project/modin/issues/5615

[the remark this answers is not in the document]
YarShev: thanks!

modin/core/execution/ray/implementations/pandas_on_ray/partitioning/partition.py
```diff
@@ -262,13 +262,20 @@ def preprocess_func(cls, func):
         """
         return ray.put(func)
 
-    def length(self):
+    def length(self, materialize=True):
```
YarShev, line 265: Let's make this change for `ip` method as part of a separate issue.

modin/core/storage_formats/pandas/query_compiler.py
YarShev: Create an issue to sync params for both methods please. I think it is bit confusing to call broadcast_apply_full_axis and pass None for other.

YarShev (approved):

YarShev:
Is there any comparison of the changes against master?

modin/core/storage_formats/pandas/query_compiler.py
anmyachev: https://github.com/modin-project/modin/issues/5620

modin/core/execution/ray/implementations/pandas_on_ray/partitioning/partition.py
anmyachev: In order to make this change it would be nice to have code that uses this.

YarShev: https://github.com/modin-project/modin/blob/0df5fbf7a66663110878633467e4488ad37b0337/modin/distributed/dataframe/pandas/partitions.py#LL96C34-L96C39

anmyachev:
> Is there any comparison of the changes against master?

Version | Sync mode | Async mode
-|-|-
Master 1rst |  4.09 |  1.91
Master 2nd |  0.98 |  0.23
Master 3rd |  0.81 | 0.13
PR 1rst | 4.01 | 0.05
PR 2nd | 1.94 | 0.06
PR 3rd | 1.83 | 0.06

<details>
<summary> Benchmark: Ray, 32 cores </summary>

```bash
import modin.pandas as pd
import numpy as np
from time import time

data = np.random.randint(10000, size=(1000000, 1000))
df = pd.DataFrame(data)

for _ in range(3):
    df._query_compiler._modin_frame._index_cache = None
    start = time()
    res = df.reset_index()
    execute(res)
    print(f"reset_index time: {time()-start}")
```

</details>

modin/core/execution/ray/implementations/pandas_on_ray/partitioning/partition.py
anmyachev: It looks like we're working through cached values everywhere right now, you want to change that and use `ip(materialize=False)`?

By the way, can we return an empty dataframe here to speed up?
`self.apply(lambda df: pd.DataFrame([]))._ip_cache` instead of `self.apply(lambda df: df)._ip_cache`

dchigarev:
it's strange that the PR is slower in sync mode...

anmyachev:
> it's strange that the PR is slower in sync mode...
> 
> Am I understanding correctly that 'sync' mode is `execute(res)` and 'async' is without calling `execute()`?

Yes.
Why strange? We are doing a full axis operation.

dchigarev:
> Why strange? We are doing a full axis operation.

Right, the more column partitions we have the slower the new implementation will be. Here are some more measurements in sync mode with the same script but with different shapes:
NPartitions = 112, the table shows time of the 3rd iteration:
| | Master | PR |
|-|-|-|
|(1_000_000, 10) | 0.13s | 0.06s |
|(1_000_000, 32) | 0.16s | 0.07s|
|(1_000_000, 256) | 0.51s | 0.45s |
|(1_000_000, 512) | 1.01s | 0.99s |
|(1_000_000, 1024) | 1.41s | 1.84s |
|(10_000_000, 32) | 1.38s | 0.55s |
|(10_000_000, 64) | 1.37s | 1.11s |

dchigarev:
Here are also measurements in sync mode, but here we're triggering an '.index' materialization at the end of '.reset_index':
| | Master | PR |
|-|-|-|
|(1_000_000, 10) | 0.11s | 0.1s |
|(1_000_000, 32) | 0.13s | 0.09s|
|(1_000_000, 256) | 0.14s | 0.45s |
|(1_000_000, 512) | 0.16s | 0.98s |
|(1_000_000, 1024) | 0.18s | 1.93s |
|(10_000_000, 32) | 1.36s | 0.49s |
|(10_000_000, 64) | 1.53s | 0.96s |

<details><summary>script</summary>

```python
import modin.pandas as pd
import numpy as np
from time import time

sizes = [
    (1_000_000, 10),
    (1_000_000, 32),
    (1_000_000, 256),
    (1_000_000, 512),
    (1_000_000, 1024),
    (10_000_000, 32),
    (10_000_000, 64),
]

for size in sizes:
    data = np.random.randint(10000, size=size)
    df = pd.DataFrame(data)

    for _ in range(3):
        df._query_compiler._modin_frame._index_cache = None
        start = time()
        res = df.reset_index()
        len(res.index) # trigger index materialization
        t_Res = time()-start
    print(size, t_Res)
```

</details>

modin/core/execution/ray/implementations/pandas_on_ray/partitioning/partition.py
YarShev: > It looks like we're working through cached values everywhere right now, you want to change that and use `ip(materialize=False)`?

Yes, I think we should use public api where possible instead of private.

> By the way, can we return an empty dataframe here to speed up?
self.apply(lambda df: pd.DataFrame([]))._ip_cache instead of self.apply(lambda df: df)._ip_cache

No, we have to know where the original partition lays on.

anmyachev: > No, we have to know where the original partition lays on.

Why do you think that if we return the same object, then this will show what is needed, but if we return another object, then no?

YarShev: Ah, I see, we pass a partition to the remote function anyway. Then, it should work. I thought you just returned an empty df from the remote function without any argument.

YarShev:
Should we have some dispatching mechanism depending on the number of columns?

YarShev:
shape | Master | PR
-- | -- | --
(1_000_000, 256) | 0.14s | 0.45s
(1_000_000, 512) | 0.16s | 0.98s
(1_000_000, 1024) | 0.18s | 1.93s

Based on the last comment, it doesn't look very good.

modin/core/execution/ray/implementations/pandas_on_ray/partitioning/partition.py
anmyachev: > 

I will create issue after the PR is merged.

anmyachev:
> Should we have some dispatching mechanism depending on the number of columns?

For the workload, which I accelerated, an asynchronous implementation of this operation is needed for a dataframe with ~250 columns.

YarShev:
Even though wide frames are not a common use case.

anmyachev:
> I think we should detect the limits of `row_lengths` and `column_widths` using which we get slower. Depending on the limits we would use either previous or the new implemtation. Can you create an issue related to that?

https://github.com/modin-project/modin/issues/5633

modin/core/execution/ray/implementations/pandas_on_ray/partitioning/partition.py
anmyachev: https://github.com/modin-project/modin/issues/5634

modin/core/storage_formats/pandas/query_compiler.py
anmyachev: https://github.com/modin-project/modin/issues/5635

PR #5567, opened 2023-01-20
FIX-#5566: Enable test_indexing test on the HDK engine and add to ci

### LINKED ISSUE #5566: Enable test_indexing test on the HDK engine and add to ci
Currently, there are 337 failures, when the test is running on the HDK engine. All these failures should be fixed and the test added to CI.

AndreyPavlenko:
## What do these changes do?

Reworked column names encoding.
Fixes.

Depends on #5156

- [x] Resolves #5566 
- [x] tests added and passing

### DIFF
--- a/modin/experimental/core/execution/native/implementations/hdk_on_native/dataframe/dataframe.py
+++ b/modin/experimental/core/execution/native/implementations/hdk_on_native/dataframe/dataframe.py
@@ -44,6 +44,7 @@ from pandas.core.dtypes.common import (
 from modin.error_message import ErrorMessage
 from modin.pandas.indexing import is_range_like
 from modin.utils import MODIN_UNNAMED_SERIES_LABEL
+from modin.core.dataframe.pandas.utils import concatenate
 from modin.core.dataframe.base.dataframe.utils import join_columns
 import pandas as pd
 import typing
@@ -2522,7 +2523,7 @@ class HdkOnNativeDataframe(PandasDataframe):
                 for idx, new_type in cast.items():
                     schema = schema.set(idx, pyarrow.field(new_type, pyarrow.string()))
                 at = at.cast(schema)
-            df = arrow_to_pandas(at)
+            df = concatenate([arrow_to_pandas(at)])
         else:
             df = self._partition_mgr_cls.to_pandas(self._partitions)
 
--- a/modin/experimental/core/execution/native/implementations/hdk_on_native/dataframe/dataframe.py
+++ b/modin/experimental/core/execution/native/implementations/hdk_on_native/dataframe/dataframe.py
@@ -2649,7 +2649,7 @@ class HdkOnNativeDataframe(PandasDataframe):
         # the number of rows and, thus, unable to restore the index.
         # That's what we usually have for arrow tables and execution
         # result. Unnamed index is renamed to {IDX_COL_PREF}. Also all
-        # columns get 'F_' prefix to handle names unsupported in HDK.
+        # columns get encoded to handle names unsupported in HDK.
         elif (
             len(new_index) == 0
             and not isinstance(new_index, MultiIndex)
@@ -2667,7 +2667,7 @@ class HdkOnNativeDataframe(PandasDataframe):
         new_dtypes = df.dtypes
 
         def encoder(n):
-            return n if n == MODIN_UNNAMED_SERIES_LABEL else encode_col_name(n, False)
+            return n if n == MODIN_UNNAMED_SERIES_LABEL else encode_col_name(n, ignore_reserved=False)
 
         if index_cols is not None:
             cols = index_cols.copy()
@@ -2681,7 +2681,7 @@ class HdkOnNativeDataframe(PandasDataframe):
             new_lengths,
             new_widths,
             unsupported_cols,
-        ) = cls._partition_mgr_cls.from_pandas(df, True, False)
+        ) = cls._partition_mgr_cls.from_pandas(df, return_dims=True, encode_col_names=False)
 
         if len(unsupported_cols) > 0:
             ErrorMessage.single_warning(
--- a/modin/experimental/core/execution/native/implementations/hdk_on_native/dataframe/dataframe.py
+++ b/modin/experimental/core/execution/native/implementations/hdk_on_native/dataframe/dataframe.py
@@ -24,6 +24,10 @@ from modin.experimental.core.storage_formats.hdk.query_compiler import (
 from .utils import (
     LazyProxyCategoricalDtype,
     arrow_to_pandas,
+    encode_col_name,
+    decode_col_name,
+    IDX_COL_NAME,
+    ROWID_COL_NAME,
     check_join_supported,
     check_cols_to_join,
     get_data_for_join_by_index,
@@ -224,25 +228,12 @@ class HdkOnNativeDataframe(PandasDataframe):
         if partitions is not None:
             self._filter_empties()
 
-        # This frame uses encoding for column names to support exotic
-        # (e.g. non-string and reserved words) column names. Encoded
-        # names are used in HDK tables and corresponding Arrow tables.
-        # If we import Arrow table, we have to rename its columns for
-        # proper processing.
         if self._has_arrow_table() and self._partitions.size > 0:
             assert self._partitions.size == 1
             table = self._partitions[0][0].get()
-            column_names = table.column_names
-            if len(table) > 0 and column_names[0] != f"F_{self._table_cols[0]}":
-                column_names = [f"F_{col}" for col in column_names]
-                table = table.rename_columns(column_names)
-                self._partitions[0][
-                    0
-                ] = self._partition_mgr_cls._partition_class.put_arrow(table)
-
             for i, t in enumerate(dtypes):
                 if isinstance(t, LazyProxyCategoricalDtype):
-                    dtypes[i] = t._new(table, column_names[i])
+                    dtypes[i] = t._new(table, table.column_names[i])
 
         self._uses_rowid = uses_rowid
         self._force_execution_mode = force_execution_mode
@@ -291,7 +282,7 @@ class HdkOnNativeDataframe(PandasDataframe):
         -------
         InputRefExpr
         """
-        if col == "__rowid__":
+        if col == ROWID_COL_NAME:
             return InputRefExpr(self, col, get_dtype(int))
         return InputRefExpr(self, col, self.get_dtype(col))
 
@@ -492,7 +483,7 @@ class HdkOnNativeDataframe(PandasDataframe):
         def generate_by_name(by):
             """Generate unuqie name for `by` column in the resulted frame."""
             if as_index:
-                return f"__index__0_{by}"
+                return f"{IDX_COL_NAME}0_{by}"
             elif by in agg_cols:
                 # Aggregation columns are more prioritized than the 'by' cols,
                 # so in case of naming conflicts, we drop 'by' cols.
@@ -667,6 +658,7 @@ class HdkOnNativeDataframe(PandasDataframe):
             columns=self.columns,
             dtypes=dtypes,
             op=new_op,
+            index=self._index_cache,
             index_cols=self._index_cols,
             force_execution_mode=self._force_execution_mode,
         )
@@ -737,6 +729,7 @@ class HdkOnNativeDataframe(PandasDataframe):
             columns=self.columns,
             dtypes=dtypes,
             op=new_op,
+            index=self._index_cache,
             index_cols=self._index_cols,
             force_execution_mode=self._force_execution_mode,
         )
@@ -793,6 +786,7 @@ class HdkOnNativeDataframe(PandasDataframe):
             columns=self.columns,
             dtypes=new_dtypes,
             op=new_op,
+            index=self._index_cache,
             index_cols=self._index_cols,
             force_execution_mode=self._force_execution_mode,
         )
@@ -1033,7 +1027,7 @@ class HdkOnNativeDataframe(PandasDataframe):
                 else:
                     assert index_width == 1, "unexpected index width"
                     aligned_index = self._mangle_index_names([None])
-                    exprs[aligned_index[0]] = frame.ref("__rowid__")
+                    exprs[aligned_index[0]] = frame.ref(ROWID_COL_NAME)
                     aligned_index_dtypes = [get_dtype(int)]
                     uses_rowid = True
                 aligned_dtypes = aligned_index_dtypes + new_dtypes
@@ -1103,16 +1097,6 @@ class HdkOnNativeDataframe(PandasDataframe):
                 self.columns = frame.columns
                 self.index_cols = frame._index_cols
                 self.table: pyarrow.Table = frame._partitions[0][0].arrow_table
-                # To be removed after #5567 merge
-                if self.index_cols is not None:
-                    self.index_cols = [f"F_{n}" for n in self.index_cols]
-
-        # This function to be removed after #5567 merge
-        def mangle_index_names(names):
-            return [
-                f"F___index__{i}_{'__None__' if n is None else n}"
-                for i, n in enumerate(names)
-            ]
 
         frames: List[FrameData] = [FrameData(f) for f in [self] + other_modin_frames]
         col_fields: typing.OrderedDict[Tuple[str, str], pyarrow.Field] = OrderedDict()
@@ -1166,7 +1150,7 @@ class HdkOnNativeDataframe(PandasDataframe):
                 idx = RangeIndex(0, sum(len(f.table) for f in frames))
             else:
                 idx = frames[0].index.append([f.index for f in frames[1:]])
-            idx_cols = mangle_index_names(idx.names)
+            idx_cols = self._mangle_index_names(idx.names)
             idx_df = pd.DataFrame(index=idx).reset_index()
             union = pyarrow.Table.from_pandas(idx_df).rename_columns(idx_cols)
         else:
@@ -1176,7 +1160,7 @@ class HdkOnNativeDataframe(PandasDataframe):
                     if ignore_index:
                         frame.index = pd.RangeIndex(0, len(frame.index))
                     idx = frame.index
-                    frame.index_cols = mangle_index_names(idx.names)
+                    frame.index_cols = self._mangle_index_names(idx.names)
                     idx_df = pd.DataFrame(index=idx).reset_index()
                     frame.table = pyarrow.Table.from_pandas(idx_df)
                     frame.table = frame.table.rename_columns(frame.index_cols)
@@ -1209,11 +1193,11 @@ class HdkOnNativeDataframe(PandasDataframe):
                 if idx_equal:
                     idx = frames[0].index
                     if isinstance(idx, MultiIndex):
-                        idx_cols = mangle_index_names(idx.names)
+                        idx_cols = self._mangle_index_names(idx.names)
                     else:
                         idx_names = {f.index.name for f in frames}
                         idx_names = [None] if len(idx_names) > 1 else [idx.name]
-                        idx_cols = mangle_index_names(idx_names)
+                        idx_cols = self._mangle_index_names(idx_names)
 
                     # Rename index columns
                     for frame in frames:
@@ -1226,7 +1210,7 @@ class HdkOnNativeDataframe(PandasDataframe):
                 else:
                     # Align index columns
                     idx = frames[0].index.append([f.index for f in frames[1:]])
-                    idx_cols = mangle_index_names(idx.names)
+                    idx_cols = self._mangle_index_names(idx.names)
                     idx_df = pd.DataFrame(index=idx).reset_index()
                     idx_table = pyarrow.Table.from_pandas(idx_df)
                     idx_table = idx_table.rename_columns(idx_cols)
@@ -1256,14 +1240,11 @@ class HdkOnNativeDataframe(PandasDataframe):
                 for i, name in enumerate(idx_table.column_names):
                     union = union.add_column(i, idx_table.field(i), idx_table.column(i))
 
-        # To be removed after #5567 merge
-        if idx_cols is not None:
-            idx_cols = [n[2:] for n in idx_cols]
-
         return self.from_arrow(
             union,
             index_cols=idx_cols,
             columns=[k[0] for k in col_fields.keys()],
+            encode_col_names=False,
         )
 
     def _join_by_index(self, other_modin_frames, how, sort, ignore_index):
@@ -1289,12 +1270,15 @@ class HdkOnNativeDataframe(PandasDataframe):
         check_join_supported(how)
         lhs = self._maybe_materialize_rowid()
         reset_index_names = False
+        new_columns_dtype = self.columns.dtype
         for rhs in other_modin_frames:
             rhs = rhs._maybe_materialize_rowid()
             if len(lhs._index_cols) != len(rhs._index_cols):
                 raise NotImplementedError(
                     "join by indexes with different sizes is not supported"
                 )
+            if new_columns_dtype != rhs.columns.dtype:
+                new_columns_dtype = None
 
             reset_index_names = reset_index_names or lhs._index_cols != rhs._index_cols
 
@@ -1326,7 +1310,7 @@ class HdkOnNativeDataframe(PandasDataframe):
             )
 
             new_columns = Index.__new__(
-                Index, data=new_columns, dtype=self.columns.dtype
+                Index, data=new_columns, dtype=new_columns_dtype
             )
             lhs = lhs.__constructor__(
                 dtypes=lhs._dtypes_for_exprs(exprs),
@@ -1454,6 +1438,7 @@ class HdkOnNativeDataframe(PandasDataframe):
                 columns=self.columns,
                 dtypes=self._dtypes_for_exprs(exprs),
                 op=TransformNode(self, exprs),
+                index=self._index_cache,
                 index_cols=self._index_cols,
                 force_execution_mode=self._force_execution_mode,
             )
@@ -1471,6 +1456,7 @@ class HdkOnNativeDataframe(PandasDataframe):
                 columns=self.columns,
                 dtypes=self._dtypes_for_exprs(exprs),
                 op=TransformNode(self, exprs),
+                index=self._index_cache,
                 index_cols=self._index_cols,
                 force_execution_mode=self._force_execution_mode,
             )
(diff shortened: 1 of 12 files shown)


modin/pandas/test/utils.py
```diff
@@ -634,6 +634,7 @@ def df_equals(df1, df2):
             check_index_type=False,
             check_column_type=False,
             check_categorical=False,
+            check_freq=False,
```
anmyachev, line 637: Why?

modin/pandas/test/dataframe/test_indexing.py
```diff
@@ -2017,11 +2060,15 @@ def test___setitem__partitions_aligning():
...
 
+@pytest.mark.xfail(
+    StorageFormat.get() == "Hdk", reason="https://github.com/intel-ai/hdk/issues/165"
+)
 def test___setitem__with_mismatched_partitions():
     with ensure_clean(".csv") as fname:
         np.savetxt(fname, np.random.randint(0, 100, size=(200_000, 99)), delimiter=",")
-        modin_df = pd.read_csv(fname)
-        pandas_df = pandas.read_csv(fname)
+        # Use header=None to avoid duplicate column names
```
anmyachev, line 2069: This should be fixed already?

modin/pandas/test/dataframe/test_indexing.py
```diff
@@ -1615,7 +1642,19 @@ def test_reset_index_with_named_index(
...
-@pytest.mark.parametrize("test_async_reset_index", [False, True])
+@pytest.mark.parametrize(
+    "test_async_reset_index",
+    [
+        pytest.param(False),
+        pytest.param(
+            True,
+            marks=pytest.mark.skipif(
+                StorageFormat.get() == "Hdk",
+                reason="_propagate_index_objs is not supported by HDK",
```
anmyachev, line 1653: I believe that direct `set_index` usage should help to avoid `_propagate_index_objs` usage.

modin/pandas/test/dataframe/test_indexing.py
```diff
@@ -57,6 +57,9 @@
 # of defaulting to pandas.
 pytestmark = pytest.mark.filterwarnings(default_to_pandas_ignore_string)
 
+# Initialize env for storage format detection in @pytest.mark.xfail
+pd.DataFrame()
```
anmyachev, line 61: If it is really necessary, then it will probably need to be moved to `conftest.py` file.

modin/pandas/dataframe.py
```diff
@@ -1212,6 +1212,7 @@ def insert(
             if (
                 is_list_like(value)
                 and not isinstance(value, (pandas.Series, Series))
+                and len(self.index) != 0
```
anmyachev, line 1215: Why?

modin/experimental/core/storage_formats/hdk/query_compiler.py
```diff
@@ -557,18 +553,27 @@ def concat(self, axis, other, **kwargs):
 
     def drop(self, index=None, columns=None, errors: str = "raise"):
         if index is not None:
-            raise NotImplementedError("Row drop")
+            # Only column drop is supported by the HDK engine
+            raise NotImplementedError("Index drop")
```
anmyachev, line 557: Maybe it just defaults to pandas?

modin/experimental/core/storage_formats/hdk/query_compiler.py
```diff
@@ -557,18 +553,27 @@ def concat(self, axis, other, **kwargs):
...
+        columns = self.columns.drop(columns)
+        new_frame = self._modin_frame.take_2d_labels_or_positional(
+            row_labels=index, col_labels=columns
         )
 
+        # If all columns are dropped and the index is trivial, we are
+        # not able to restore it, since we don't know the number of rows.
+        # In this case, we copy the index from the current frame.
+        if len(columns) == 0 and new_frame._index_cols is None:
+            new_frame._index_cache = self._modin_frame.index.copy()
```
anmyachev, line 573: I suppose this is the behavior of pandas to save the index if there are no more columns? I wonder how this moment is handled by other backends.

modin/experimental/core/storage_formats/hdk/query_compiler.py
AndreyPavlenko: What do you mean?

anmyachev: Will the user see this exception or not? If he sees it, then why not execute this code on pandas, and then convert it back to hdk?

dchigarev: HDK automatically defaults to pandas on any `NotImplementedError` being raised: https://github.com/modin-project/modin/blob/7c1cb3c96aa1a9123fa9edb544a1cdcbe34a41f2/modin/experimental/core/storage_formats/hdk/query_compiler.py#L63

anmyachev: got it

modin/experimental/core/execution/native/implementations/hdk_on_native/dataframe/utils.py
```diff
@@ -13,11 +13,202 @@
...
+_BASE_DICT = dict((c, i) for i, c in enumerate(_BASE_LIST))
+_NON_ALPHANUM_PATTERN = re.compile("[^a-zA-Z0-9]+")
+# Number of bytes in the tailing chunk
+_TAIL_LEN = {"_0": 0, "_1": 1, "_2": 2, "_3": 3}
+_RESERVED_NAMES = (MODIN_UNNAMED_SERIES_LABEL, ROWID_COL_NAME)
+_COL_TYPES = Union[str, int, float, Timestamp, None]
+_COL_NAME_TYPE = Union[_COL_TYPES, Tuple[_COL_TYPES, ...]]
+
+
+def encode_col_name(
```
dchigarev, line 41: I am a bit concerned about the performance of this. `encode/decode_col_name` functions are called pretty often as I have seen and ideally we would want this code to be written in C, maybe there's already a library that does efficient encoding/decoding to alpha-num strings? But maybe we don't need to care about this much if the python's perf is okay, so, can you please measure how these functions perform and post the result?

modin/experimental/core/execution/native/implementations/hdk_on_native/dataframe/utils.py
```diff
@@ -13,11 +13,215 @@
...
+            count -= 1
+            if count != 0:
+                dst.append("_S")  # Separator
+        return "".join(dst)
+    if len(name) == 0:
+        return "_E"
+    if ignore_reserved and (name.startswith(IDX_COL_NAME) or name in _RESERVED_NAMES):
+        return name
+
+    non_alpha = _NON_ALPHANUM_PATTERN.search(name)
```
dchigarev, line 91: why having non alpha-num characters is a problem for a column name? why we're getting rid of them?

modin/pandas/test/dataframe/test_indexing.py
```diff
@@ -1376,7 +1376,19 @@ def test_reset_index(data, test_async_reset_index):
     df_equals(modin_df_cp, pd_df_cp)
 
 
-@pytest.mark.parametrize("data", test_data_values, ids=test_data_keys)
+@pytest.mark.parametrize(
+    "data",
+    [
+        pytest.param(test_data["int_data"]),
```
anmyachev, line 1382:
```suggestion
        test_data["int_data"],
```

modin/pandas/test/dataframe/test_indexing.py
```diff
@@ -1642,7 +1654,19 @@ def test_reset_index_with_multi_index_drop(
...
-@pytest.mark.parametrize("test_async_reset_index", [False, True])
+@pytest.mark.parametrize(
+    "test_async_reset_index",
+    [
+        pytest.param(False),
+        pytest.param(
+            True,
+            marks=pytest.mark.skipif(
+                StorageFormat.get() == "Hdk",
+                reason="_propagate_index_objs is not supported by HDK",
```
anmyachev, line 1665: @AndreyPavlenko please rebase on master. Tests no longer have `_propagate_index_objs ` functions.

modin/experimental/core/execution/native/implementations/hdk_on_native/test/test_utils.py
anmyachev, line 1: Are we testing this file in CI or not?

modin/experimental/core/execution/native/implementations/hdk_on_native/partitioning/partition_manager.py
```diff
@@ -122,16 +127,29 @@ def from_arrow(cls, at, return_dims=False, unsupported_cols=None):
...
-        put_func = cls._partition_class.put_arrow
+        if encode_col_names:
+            encode_col_names = False
+            encoded_names = []
+            for n in at.column_names:
+                encoded_name = encode_col_name(n)
+                encoded_names.append(encoded_name)
+                if n != encoded_name:
+                    encode_col_names = True
+            encoded_at = at.rename_columns(encoded_names) if encode_col_names else at
```
anmyachev, line 147:
```suggestion
            encoded_names = [encode_col_name(n) for n in at.column_names]
            encoded_at = at
            if encoded_names != at.column_names:
                encoded_at = at.rename_columns(encoded_names)
```

modin/experimental/core/execution/native/implementations/hdk_on_native/dataframe/utils.py
anmyachev: @AndreyPavlenko this is quite low-level and complex code. Is it possible to make the necessary utility in the HDK itself?

anmyachev (dismissed):
LGTM!

modin/experimental/core/execution/native/implementations/hdk_on_native/dataframe/utils.py
```diff
@@ -13,11 +13,215 @@
...
-
+import re
 import pandas
 import pyarrow
 
+from pandas import Timestamp
 from modin.error_message import ErrorMessage
+from modin.utils import MODIN_UNNAMED_SERIES_LABEL
+from string import ascii_uppercase, ascii_lowercase, digits
+from typing import List, Tuple, Union
```
anmyachev, line 24:
```suggestion
import re
from string import ascii_uppercase, ascii_lowercase, digits
from typing import List, Tuple, Union

import pandas
from pandas import Timestamp
import pyarrow

from modin.error_message import ErrorMessage
from modin.utils import MODIN_UNNAMED_SERIES_LABEL
```

anmyachev (dismissed):

anmyachev:
@AndreyPavlenko do you still have anything to do here?

dchigarev:
I would like to have this comment answered before we merge https://github.com/modin-project/modin/pull/5567#discussion_r1143136268

modin/experimental/core/execution/native/implementations/hdk_on_native/dataframe/utils.py
AndreyPavlenko: > @AndreyPavlenko this is quite low-level and complex code. Is it possible to make the necessary utility in the HDK itself?

I'll try to implement it in C and, if the performance is much better, I'll submit a PR to HDK.

modin/experimental/core/execution/native/implementations/hdk_on_native/dataframe/utils.py
```diff
@@ -13,11 +13,220 @@
...
+    end = len(name)
+    while idx != -1:
+        dst.append(name[off:idx])
+        off = _unquote(name, dst, idx, end)
+        idx = name.find("_Q", off)
+    dst.append(name[off:])
+    return "".join(dst)
+
+
+def _quote(src: str, dst: List[str]):  # noqa: GL08
```
dchigarev, line 162: I'm more concerned about the need to maintain this in our code base rather than the performance of this function. 

As we figured out from offline discussion, the custom implementation of the base64 algorithm was introduced in order to ensure that the encoded string will only contain alpha-num characters (excluding '+', '/' and '=' symbols that may appear in the strings encoded by 'base64' python's library), which may be useful if a user wants to interact with HDK via its SQL interface (that only supports alpha-num col names).

Well, I have a few things to disagree with here:
1. If we really want this baseX encoding then we may use 'base32' version from python's standard library. The output alphabet for the 32 version is exactly alpha-num characters, its disadvantage though is that [the result is approximately ~20% longer than with base64](https://en.wikipedia.org/wiki/Base32#Disadvantages), but I think we could put up with it.
2. As far as I understand the only reason for this special strings encoding is that if a column name originally contained an '_S' substring then it breaks the decoding as we reserved the '_S' string as a separator of the tuple values:
    ```python
    >>> multiindex_col_name = ("level1", "level2")
    >>> encode_without_base64(multiindex_col_name)
    "_T_Slevel1_Slevel2"
    >>> decode_col_name("_T_Slevel1_Slevel2")
    ("level1", "level2")
    >>> decode_col_name(encode_without_base64(("lev_Sel1", "level2")))
    ("lev", "el1", "level2")
    ```
    So for me, it feels like an overkill to use baseX encoding algorithm for every string because of that. One of the reasons I'm against it is that such kind of mangling may make debugging really hard as you just wouldn't understand what the column name is right after you imported pandas df to modin:
    ```python
    >>> base64.b32encode(bytes("simple column name", "utf-8"))
    b'ONUW24DMMUQGG33MOVWW4IDOMFWWK==='
    ```
    Maybe we can use some really special separator to split the tuple values in the encoded column names so there would be less chance we met it in a user's column name and then just raise an error if this special character is actually met (just reserve certain symbols and forbid them in column names).

dchigarev (changes requested):

modin/experimental/core/execution/native/implementations/hdk_on_native/dataframe/utils.py
AndreyPavlenko: > 2\. I'm against it is that such kind of mangling may make debugging really hard as you just wouldn't understand what the column name is right after you imported pandas df to modin

Not, on the modin level the column names are not changed. The encoded names are used in the arrow tables. Another point - alpha-num parts of the string are not encoded. In your example, only the space characters will be encoded.

I agree with you, that maintaining this code could be a pain, despite that the algorithm is quite simple. I've simplified the algorithm and removed the non-alphanum characters encoding.

modin/experimental/core/execution/native/implementations/hdk_on_native/dataframe/dataframe.py
```diff
@@ -2651,30 +2654,40 @@ def from_pandas(cls, df):
...
             df.index.names = index_cols
             df = df.reset_index()
-
             orig_df.index.names = orig_index_names
+
         new_dtypes = df.dtypes
-        df = df.add_prefix("F_")
+
+        def encoder(n):
+            return n if n == MODIN_UNNAMED_SERIES_LABEL else encode_col_name(n, False)
```
dchigarev, line 2676: could you please pass this as a kwarg so it would read better
```suggestion
            return n if n == MODIN_UNNAMED_SERIES_LABEL else encode_col_name(n, ignore_reserved=False)
```

modin/experimental/core/execution/native/implementations/hdk_on_native/dataframe/dataframe.py
```diff
@@ -2651,30 +2654,40 @@ def from_pandas(cls, df):
...
+        else:
+            df = df.rename(columns=encoder)
 
         (
             new_parts,
             new_lengths,
             new_widths,
             unsupported_cols,
-        ) = cls._partition_mgr_cls.from_pandas(df, True)
+        ) = cls._partition_mgr_cls.from_pandas(df, True, False)
```
dchigarev, line 2690:
```suggestion
        ) = cls._partition_mgr_cls.from_pandas(df, return_dims=True, encode_col_names=False)
```

modin/experimental/core/execution/native/implementations/hdk_on_native/dataframe/dataframe.py
```diff
@@ -2651,30 +2654,40 @@ def from_pandas(cls, df):
         # it into columns either because, otherwise, we don't know
         # the number of rows and, thus, unable to restore the index.
         # That's what we usually have for arrow tables and execution
-        # result. Unnamed index is renamed to __index__. Also all
+        # result. Unnamed index is renamed to {IDX_COL_PREF}. Also all
         # columns get 'F_' prefix to handle names unsupported in HDK.
```
dchigarev, line 2658: ?
```suggestion
        # columns get encoded to handle names unsupported in HDK.
```

modin/experimental/core/execution/native/implementations/hdk_on_native/dataframe/dataframe.py
```diff
@@ -2094,15 +2092,20 @@ def _build_index_cache(self):
...
-                    index_df.index.rename(
-                        self._index_names(self._index_cols), inplace=True
-                    )
-                    self._index_cache = index_df.index
+                    index_df.set_index(self._index_cols, inplace=True)
+                    idx = index_df.index
+                    idx.rename(self._index_names(self._index_cols), inplace=True)
+                    if (
+                        isinstance(idx, (pd.DatetimeIndex, pd.TimedeltaIndex))
+                        and len(idx) > 3
```
dchigarev, line 2105: why is `> 3` needed? let's add an in-code comment about this

modin/experimental/core/execution/native/implementations/hdk_on_native/dataframe/dataframe.py
```diff
@@ -2525,25 +2529,26 @@ def to_pandas(self):
                 for idx, new_type in cast.items():
                     schema = schema.set(idx, pyarrow.field(new_type, pyarrow.string()))
                 at = at.cast(schema)
-            df = at.to_pandas()
+            df = concatenate([at.to_pandas()])
```
dchigarev, line 2532: please add an in-code comment on why we call `concatenate` here

dchigarev (approved):
lgtm!

left few suggestions

modin/experimental/core/execution/native/implementations/hdk_on_native/dataframe/dataframe.py
AndreyPavlenko: Done

modin/experimental/core/execution/native/implementations/hdk_on_native/dataframe/dataframe.py
AndreyPavlenko: Done

modin/experimental/core/execution/native/implementations/hdk_on_native/dataframe/utils.py
```diff
@@ -12,12 +12,132 @@
...
 
 import pandas
+from pandas import Timestamp
+
 import pyarrow
 
 from modin.error_message import ErrorMessage
+from modin.utils import MODIN_UNNAMED_SERIES_LABEL
+
+IDX_COL_NAME = "__index__"
```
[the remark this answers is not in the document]
AndreyPavlenko: Done

modin/experimental/core/execution/native/implementations/hdk_on_native/dataframe/utils.py
```diff
@@ -12,13 +12,139 @@
...
+
+class ColNameCodec:
+    IDX_COL_NAME = "__index__"
+    ROWID_COL_NAME = "__rowid__"
+
+    _RESERVED_NAMES = (MODIN_UNNAMED_SERIES_LABEL, ROWID_COL_NAME)
+    _COL_TYPES = Union[str, int, float, Timestamp, None]
+    _COL_NAME_TYPE = Union[_COL_TYPES, Tuple[_COL_TYPES, ...]]
+
+    def _encode_tuple(values: Tuple[_COL_TYPES, ...]) -> str:  # noqa: GL08
```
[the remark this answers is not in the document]
anmyachev: @AndreyPavlenko pay attention

modin/experimental/core/execution/native/implementations/hdk_on_native/dataframe/utils.py
```diff
@@ -12,13 +12,139 @@
...
+            if isinstance(value, str):
+                dst.append(value.replace("_", "_Q"))
+            else:
+                dst.append(ColNameCodec._ENCODERS[type(value)](value))
+            count -= 1
+            if count != 0:
+                dst.append("_T")
+        return "".join(dst)
+
+    def _decode_tuple(encoded: str) -> Tuple[_COL_TYPES, ...]:  # noqa: GL08
```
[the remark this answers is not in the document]
anmyachev: @AndreyPavlenko pay attention

modin/experimental/core/execution/native/implementations/hdk_on_native/dataframe/utils.py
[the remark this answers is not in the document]
AndreyPavlenko: This method is only called from the static method encode(). The instance of this class is never created.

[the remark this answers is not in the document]
anmyachev: Let's make it static as well.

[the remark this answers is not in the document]
AndreyPavlenko: If make these 2 methods static, on Python 3.8 it fails to call these methods with `TypeError: 'staticmethod' object is not callable`. To make these methods static, additional workarounds are required. I'm not sure if we need to overcomplicate it here.

modin/experimental/core/execution/native/implementations/hdk_on_native/dataframe/utils.py
```diff
@@ -13,15 +13,141 @@
...
+            dec = (
+                None
+                if len(item) < 2 or item[0] != "_"
+                else ColNameCodec._DECODERS.get(item[1], None)
+            )
+            items.append(item.replace("_Q", "_") if dec is None else dec(item))
+        return tuple(items)
+
+    _ENCODERS = {
+        tuple: _encode_tuple,
```
anmyachev, line 64: Isn't that enough to make it work?
```suggestion
        tuple: ColNameCodec._encode_tuple,
```

dchigarev (approved):

anmyachev (approved):

modin/experimental/core/execution/native/implementations/hdk_on_native/dataframe/utils.py
```diff
@@ -12,13 +12,141 @@
...
+
+    _DECODERS = {
+        "T": _decode_tuple,
+        "N": lambda v: None,
+        "E": lambda v: "",
+        "S": lambda v: "_" + v[2:],
+        "I": lambda v: int(v[2:]),
+        "F": lambda v: float(v[2:]),
+        "D": lambda v: Timestamp.fromtimestamp(
+            float(v[2 : (idx := v.index("_", 2))]), tz=v[idx + 1 :]
```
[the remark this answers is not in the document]
anmyachev: I guess we can rely on the order in which the arguments are evaluated. https://docs.python.org/3/reference/expressions.html#evaluation-order

modin/experimental/core/execution/native/implementations/hdk_on_native/test/test_utils.py
```diff
@@ -0,0 +1,94 @@
...
+# the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific language
+# governing permissions and limitations under the License.
+
+import sys
+import pandas
+import pytz
+import timeit
+
+from random import randint, uniform, choice
```
YarShev, line 19:
```suggestion
from random import randint, uniform, choice

```

modin/experimental/core/execution/native/implementations/hdk_on_native/dataframe/utils.py
```diff
@@ -13,15 +13,141 @@
...
+
+class ColNameCodec:
+    IDX_COL_NAME = "__index__"
+    ROWID_COL_NAME = "__rowid__"
+
+    _RESERVED_NAMES = (MODIN_UNNAMED_SERIES_LABEL, ROWID_COL_NAME)
+    _COL_TYPES = Union[str, int, float, Timestamp, None]
+    _COL_NAME_TYPE = Union[_COL_TYPES, Tuple[_COL_TYPES, ...]]
+
+    def _encode_tuple(values: Tuple[_COL_TYPES, ...]) -> str:  # noqa: GL08
```
YarShev, line 39: I would suggest some reshuffling for methods and attrs. First go attrs, second go methods.

modin/experimental/core/execution/native/implementations/hdk_on_native/dataframe/utils.py
```diff
@@ -13,15 +13,141 @@
...
 
 from modin.error_message import ErrorMessage
+from modin.utils import MODIN_UNNAMED_SERIES_LABEL
+
+
+class ColNameCodec:
+    IDX_COL_NAME = "__index__"
+    ROWID_COL_NAME = "__rowid__"
+
+    _RESERVED_NAMES = (MODIN_UNNAMED_SERIES_LABEL, ROWID_COL_NAME)
```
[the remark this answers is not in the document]
AndreyPavlenko: IDX_COL_NAME is a prefix and it's checked in the encode/decode methods:
```python
            and (
                name.startswith(ColNameCodec.IDX_COL_NAME)
                or name in ColNameCodec._RESERVED_NAMES
            )
```

modin/experimental/core/execution/native/implementations/hdk_on_native/dataframe/utils.py
```diff
@@ -13,15 +13,141 @@
...
+        "S": lambda v: "_" + v[2:],
+        "I": lambda v: int(v[2:]),
+        "F": lambda v: float(v[2:]),
+        "D": lambda v: Timestamp.fromtimestamp(
+            float(v[2 : (idx := v.index("_", 2))]), tz=v[idx + 1 :]
+        ),
+    }
+
+    @staticmethod
+    @lru_cache(1024)
```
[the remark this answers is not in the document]
AndreyPavlenko: I think, 1024 should be enough for most workflows.

dchigarev (approved):

Garra1980:
Let's merge then?

anmyachev:
> Let's merge then?

@Garra1980 If we are not in a hurry, then it is better to wait for the end of testing.

Garra1980:
> > Let's merge then?
> 
> @Garra1980 If we are not in a hurry, then it is better to wait for the end of testing.

Absolutely - just thought testing is ok already

PR #5570, opened 2023-01-21
REFACTOR-#5571: Compute length/width cache in apply/drain_call_queue methods of Dask partitions

### LINKED ISSUE #5571: Compute length/width cache in apply/drain_call_queue methods of Dask partitions
This has already been done for partitions on Ray (and Unidist). I see no reason not to do the same for Dask.

anmyachev:

## What do these changes do?

- [x] Resolves #5571 
- [x] tests added and passing

### DIFF
--- a/modin/core/execution/dask/implementations/pandas_on_dask/partitioning/partition.py
+++ b/modin/core/execution/dask/implementations/pandas_on_dask/partitioning/partition.py
@@ -91,26 +91,26 @@ class PandasOnDaskDataframePartition(PandasDataframePartition):
             self._is_debug(log) and log.debug(
                 f"SUBMIT::_apply_list_of_funcs::{self._identity}"
             )
-            futures = DaskWrapper.deploy(
+            result, length, width, ip = DaskWrapper.deploy(
                 func=apply_list_of_funcs,
                 f_args=(call_queue, self._data),
-                num_returns=2,
+                num_returns=4,
                 pure=False,
             )
         else:
             # We handle `len(call_queue) == 1` in a different way because
             # this improves performance a bit.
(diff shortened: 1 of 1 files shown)


anmyachev:

dchigarev (commented):
I wonder why it wasn't done before...

Weren't there performance drawbacks when returning multiple values from a dask remote? @Yarshev don't you remember anything related?

anmyachev:
> I wonder why it wasn't done before...
> 
> Weren't there performance drawbacks when returning multiple values from a dask remote? @YarShev don't you remember anything related?

Length caching was added in https://github.com/modin-project/modin/pull/613. At that time, there was already Dask, but another api was used (delayed api). I guess it didn't matter at the time. That's all what I know.

YarShev:
I remember there was performance slowdown when testing such a case (return multiple values from a Dask remote function). That's why I didn't introduce the changes for Dask engine. Most likely, it is because Dask doesn't support multiple values return from its remote functions so we have to use additional remote calls to get the required number of the objects. I highly recommend measuring the performance before we decide if the changes are valuable.

anmyachev:
> I remember there was performance slowdown when testing such a case (return multiple values from a Dask remote function). That's why I didn't introduce the changes for Dask engine. Most likely, it is because Dask doesn't support multiple values return from its remote functions so we have to use additional remote calls to get the required number of the objects. I highly recommend measuring the performance before we decide if the changes are valuable.

Got it. Do you have any benchmarks left?

YarShev:
Unfortunately, no. That was approximately a year ago or even later but, probably, I was just checking the operations that involve block partitions mapping because the changes affect it (map, tree_reduce, broadcast_apply, etc.).

PR #5667, opened 2023-02-16
FEAT-#5666: Support logic operations on modin numpy arrays

### LINKED ISSUE #5666: FEAT: Support logic operations on modin numpy arrays
We would like to support the majority of the numpy logic operations, as listed here: https://numpy.org/doc/stable/reference/routines.logic.html

noloerino:
## What do these changes do?
This implements the following numpy [array logic operations](https://numpy.org/doc/stable/reference/routines.logic.html) in modin:
<details>
<summary>List of functions</summary>
<pre>
- all
- any
- isfinite
- isinf
- isnan
- isnat
- isneginf
- isposinf
- iscomplex
- isreal
- isscalar
- logical_not
- logical_and
- logical_or
- logical_xor
- greater
- greater_equal
- less
- less_equal
- equal
- not_equal
</pre>
</details>

`logical_and`, `or`, and `xor` currently do not support 2D to 1D broadcasts; the logic for these seems to be slightly different from the existing arithmetic binary operators, as attempting to use `Binary.register(lambda df, r, *args, **kwargs: pd.DataFrame(np.logical_and(df, r)))` yields the following error: `NotImplementedError: Cannot apply ufunc <ufunc 'logical_and'> to mixed DataFrame and Series inputs.` This can be addressed either in forthcoming refactors of Modin's numpy internals, or I can look further into it and try to introduce a fix in this PR as well.

- [x] Resolves #5666
- [x] tests added and passing


modin/core/storage_formats/pandas/query_compiler.py
```diff
@@ -418,6 +418,20 @@ def to_numpy(self, **kwargs):
         join_type="left",
     )
 
+    # Needed for numpy API
+    _logical_and = Binary.register(
+        lambda df, r, *args, **kwargs: pandas.DataFrame(np.logical_and(df, r)),
```
RehanSD, line 423: Why r? Perhaps a more descriptive name like `other` would be better?

RehanSD, line 423: Shouldn't we be passing args and kwargs to NumPy?

modin/core/storage_formats/pandas/query_compiler.py
```diff
@@ -1439,6 +1453,25 @@ def isin_func(df, values):
...
+    _isneginf = Map.register(  # Needed for numpy API
+        lambda df, *args, **kwargs: pandas.DataFrame(np.isneginf(df))
+    )
+    _isposinf = Map.register(  # Needed for numpy API
+        lambda df, *args, **kwargs: pandas.DataFrame(np.isposinf(df))
+    )
+    _iscomplex = Map.register(  # Needed for numpy API
+        lambda df, *args, **kwargs: pandas.DataFrame(np.iscomplex(df))
+    )
+    _isreal = Map.register(  # Needed for numpy API
```
RehanSD, line 1471: Shouldn't we hint the dtypes here as well?

modin/numpy/arr.py
```diff
@@ -1469,6 +1559,283 @@ def sum(
                 + initial
             )
 
+    def all(self, axis=None, out=None, keepdims=None, *, where=True):
```
RehanSD, line 1562: Is `where` a mask on the output (like with ge, or is it supposed to be a mask on the input like with mean)?

modin/numpy/arr.py
```diff
@@ -1469,6 +1559,283 @@ def sum(
...
+            if keepdims:
+                if out is not None and out.shape != (1,):
+                    raise ValueError(
+                        f"operand was set up as a reduction along axis 0, but the length of the axis is {out.shape[0]} (it has to be 1)"
+                    )
+                return fix_dtypes_and_determine_return(result, 1, bool, out, where)
+            result = result.to_numpy()[0, 0]
+            return result if not where else result and where
+        if axis is None:
+            result = self.flatten().all(
```
RehanSD, line 1577: We shouldn't be using flatten - it's pretty inefficient. We can set this up as two reductions - first along axis 1 then axis 0

modin/numpy/arr.py
```diff
@@ -1469,6 +1559,283 @@ def sum(
...
+                getattr(self, qc_method_name)(x2),
+                self._ndim,
+                dtype,
+                out,
+                where,
+            )
+        caller, callee, new_ndim, kwargs = self._binary_op(x2)
+        # breakpoint()
+        if caller._query_compiler != self._query_compiler:
+            # In this case, we are doing an operation that looks like this 1D_object,op(2D_object).
```
RehanSD, line 1750: Are we allowed to swap like this? I think so, but am curious about short circuiting - if we did or with two numerical arrays, would the result be the same if we swapped?

modin/numpy/arr.py
```diff
@@ -1469,6 +1559,283 @@ def sum(
...
+                dtype,
+                out,
+                where,
+            )
+        caller, callee, new_ndim, _ = self._binary_op(x2)
+        if caller._query_compiler != self._query_compiler:
+            # In this case, we are doing an operation that looks like this 1D_object | 2D_object.
+            # For Modin to broadcast directly, we have to swap it so that the operation is actually
+            # 2D_object.or(1D_object).
+            caller, callee = callee, caller
```
RehanSD, line 1805: Why don't we use logical bin op here?

RehanSD (changes requested):
Left some questions - great work!

modin/core/storage_formats/pandas/query_compiler.py
noloerino: I followed the same pattern as the `Binary.register` call for `__getitem_bool`. I'll change it to `other` as suggested.

modin/numpy/arr.py
noloerino: Would short-circuiting ever matter here? Also, is there a different way we can do this broadcasted operation without rewriting a bunch of code from the `_binary_op` method?

modin/numpy/arr.py
noloerino: Seems like I forgot to change it, will fix.

modin/numpy/arr.py
noloerino: It seems to be a mask on the input:
```
>>> numpy.array([[0,1],[1,0]]).all(where=[[0,1],[1,0]])
True
```

I'm not sure how to efficiently implement `where` masking properly in the `axis=None` case if we're doing a double reduction with `self.any(axis=0).any(axis=1)`. Any suggestions?

modin/core/storage_formats/pandas/query_compiler.py
noloerino: If a `where` array is specified
> This condition is broadcast over the input. At locations where the condition is True, the out array will be set to the ufunc result. Elsewhere, the out array will retain its original value. This means there might be non-boolean values in the result. I'll add a comment to that effect. 

An example:
```
>>> out = numpy.array([2.0,2.0])
>>> numpy.isfinite(numpy.array([1.0,1.0]), where=[False, True], out=out)
array([2., 1.])
```

modin/numpy/arr.py
RehanSD: Sounds good - we can just do something like `where.where(self, True)` for `all` and `where.where(self, False)` for `any`.

modin/numpy/arr.py
RehanSD: Sorry - by short circuiting I mean logic short circuiting:
```

In [1]: import numpy as np

In [2]: arr = np.array([[1, 2], [3, 4], [5, 6]])

In [3]: arr1 = arr + 1

In [4]: np.logical_or(arr, arr1)
Out[4]:
array([[ True,  True],
       [ True,  True],
       [ True,  True]])
```
seems this isn't a problem with logical or.
(For context, 
```
In [1]: 5 or 7
Out[1]: 5
```
is the short circuiting I was concerned about).

noloerino: Right, I figured short-circuiting wouldn't matter since the result is always coerced into a boolean here, so we shouldn't have to worry about keeping the exact numeric values around.

RehanSD: Sounds good!

modin/core/storage_formats/pandas/query_compiler.py
```diff
@@ -1436,9 +1456,31 @@ def isin_func(df, values):
     convert_dtypes = Map.register(pandas.DataFrame.convert_dtypes)
     invert = Map.register(pandas.DataFrame.__invert__, dtypes="copy")
     isna = Map.register(pandas.DataFrame.isna, dtypes=np.bool_)
+    # `dtypes` cannot be naively hinted for these functions when called by the numpy API:
+    # if `where` and `out` are specified, then some values in the resulting dataframe
+    # may retain their original values, which would not have the bool datatype
```
RehanSD, line 1461: We can still naively hint dtypes here - we don't pass `out` to the algebra layer, and handle that at the API Layer instead (at the `fix_dtypes_and_determine_return` function), so the output of this will always be all bools - we only do the masking at the API Layer, and there's where we don't know what the output dtype will be.

modin/numpy/arr.py
```diff
@@ -1469,6 +1559,249 @@ def sum(
                 + initial
             )
 
+    def all(self, axis=None, out=None, keepdims=None, *, where=True):
+        check_kwargs(keepdims=keepdims, where=where)
+        if self._ndim == 1:
+            if axis == 1:
+                raise numpy.AxisError(1, 1)
+            result = self._query_compiler.all(axis=0)
```
RehanSD, line 1567: We would need to apply the mask here as well.

modin/numpy/arr.py
```diff
@@ -1469,6 +1559,249 @@ def sum(
...
+            if keepdims:
+                if out is not None and out.shape != (1, 1):
+                    raise ValueError(
+                        f"operand was set up as a reduction along axis 0, but the length of the axis is {out.shape[0]} (it has to be 1)"
+                    )
+                return fix_dtypes_and_determine_return(
+                    array(numpy.array([[result]]))._query_compiler, 2, bool, out, where
+                )
+            return result
+        result = self._query_compiler.all(axis=axis)
```
RehanSD, line 1590: We would need to mask here as well.

modin/numpy/arr.py
```diff
@@ -1469,6 +1559,249 @@ def sum(
...
+                if out is not None and out.shape != (1,):
+                    raise ValueError(
+                        f"operand was set up as a reduction along axis 0, but the length of the axis is {out.shape[0]} (it has to be 1)"
+                    )
+                return fix_dtypes_and_determine_return(result, 1, bool, out, where)
+            result = result.to_numpy()[0, 0]
+            return result if not where else result and where
+        if axis is None:
+            target = where.where(self, True) if isinstance(where, array) else self
+            result = target.all(axis=1, keepdims=None).all(
```
RehanSD, line 1578: We can just use `logical_and` on the query compiler twice (once along each axis).

noloerino: I'm not quite sure what you mean here, since `logical_and` doesn't do any reductions.

modin/numpy/arr.py
```diff
@@ -1469,6 +1559,247 @@ def sum(
...
+        if self._ndim == 1:
+            if axis == 1:
+                raise numpy.AxisError(1, 1)
+            result = target._query_compiler.all(axis=0)
+            if keepdims:
+                if out is not None and out.shape != (1,):
+                    raise ValueError(
+                        f"operand was set up as a reduction along axis 0, but the length of the axis is {out.shape[0]} (it has to be 1)"
+                    )
+                return fix_dtypes_and_determine_return(result, 1, bool, out, where)
```
RehanSD, line 1574: Here we need to pass `where is not False`, since we don't want to mask what we pass to the output array (since the mask is only for the input). You can take a look at `sum` or `prod` to see how to implement this part.

modin/numpy/arr.py
```diff
@@ -1469,6 +1559,247 @@ def sum(
...
+            result = result.to_numpy()[0, 0]
+            return result if not where else result and where
+        if axis is None:
+            result = target.all(axis=1, keepdims=None).all(axis=0, keepdims=None)
```
RehanSD, line 1578: Could we do this at the QC layer rather than at the API Layer? so `target._query_compiler.all(axis=1).all(axis=0)`? Should avoid duplicate checks etc.

modin/numpy/arr.py
```diff
@@ -1469,6 +1559,247 @@ def sum(
...
+                out,
+                where,
+            )
+        caller, callee, new_ndim, kwargs = self._binary_op(x2)
+        # breakpoint()
+        if caller._query_compiler != self._query_compiler:
+            # In this case, we are doing an operation that looks like this 1D_object.op(2D_object).
+            # For Modin to broadcast directly, we have to swap it so that the operation is actually
+            # 2D_object.op(1D_object).
+            caller, callee = callee, caller
```
RehanSD, line 1745: Why are we re-swapping the results of `_binary_op` here?

modin/numpy/logic.py
```diff
@@ -0,0 +1,60 @@
...
+logical_and = _dispatch_logic("logical_and")
+logical_or = _dispatch_logic("logical_or")
+logical_xor = _dispatch_logic("logical_xor")
+greater = _dispatch_logic("greater")
+greater_equal = _dispatch_logic("greater_equal")
+less = _dispatch_logic("less")
+less_equal = _dispatch_logic("less_equal")
+equal = _dispatch_logic("equal")
+not_equal = _dispatch_logic("not_equal")
+array_equal = _dispatch_logic("array_equal")
```
RehanSD, line 60: We don't define `array_equal` anywhere. Should this be `equal`?

modin/numpy/test/test_array_logic.py
```diff
@@ -0,0 +1,118 @@
...
+    modin_result = getattr(np, operator)(x1, axis=axis)
+    # Result may be a scalar
+    if isinstance(modin_result, np.array):
+        modin_result = modin_result._to_numpy()
+    numpy.testing.assert_array_equal(
+        modin_result, numpy_result, err_msg=f"Unary operator {operator} failed."
+    )
+
+
+def test_all_any_where():
```
RehanSD, line 52: Could we add a test for `all` + `any` along an axis with a `where` argument?

RehanSD (changes requested):
Left some more comments!

modin/numpy/arr.py
noloerino: Changed. FYI, it seems like the pandas `all()` method does not support the `keepdims` argument.

modin/numpy/arr.py
noloerino: This is a mistake, I've removed it.

modin/numpy/logic.py
noloerino: `equal` is already dispatched above, I left this in by mistake. [`array_equal`](https://numpy.org/doc/stable/reference/generated/numpy.array_equal.html) should be pretty trivial to implement if we need it, but I wasn't sure whether we wanted to implement some matter of short-circuiting for efficiency, or if having users do a to_numpy conversion if they need this method is acceptable.

modin/numpy/test/test_array_logic.py
noloerino: added!

modin/numpy/logic.py
RehanSD: Honestly I think its fine to not have `array_equal`. We can always implement it later if necessary!

RehanSD: Honestly I think its fine to not have `array_equal`, since users can implement it via .equal and .all right?

noloerino: Yeah, they should be equivalent (if a little inefficient), except for cases where broadcasting applies (array_equal checks dimensions, but equal does not).

modin/numpy/arr.py
```diff
@@ -1469,6 +1559,253 @@ def sum(
...
+            result = result.to_numpy()[0, 0]
+            return result if not where else result and where
+        if axis is None:
+            result = target._query_compiler.all(axis=1).all(axis=0)
+            if keepdims:
+                if out is not None and out.shape != (1, 1):
+                    raise ValueError(
+                        f"operand was set up as a reduction along axis 0, but the length of the axis is {out.shape[0]} (it has to be 1)"
+                    )
+                return fix_dtypes_and_determine_return(
```
RehanSD, line 1586: Same here with returning the correct value when keep dims is True, but out is None.

modin/numpy/arr.py
```diff
@@ -1469,6 +1559,253 @@ def sum(
...
+                    bool,
+                    out,
+                    where is not False,
+                )
+            return result.to_numpy()[0, 0]
+        result = target._query_compiler.all(axis=axis)
+        new_ndim = self._ndim - 1 if not keepdims else self._ndim
+        if new_ndim == 0:
+            result = result.to_numpy()[0, 0]
+            return result if not where else result & where
```
RehanSD, line 1598: Again, I don't think this is correct.

modin/numpy/arr.py
```diff
@@ -1469,6 +1559,253 @@ def sum(
...
+        if not keepdims and axis != 1:
+            result = result.transpose()
+        return fix_dtypes_and_determine_return(
+            result, new_ndim, bool, out, where is not False
+        )
+
+    _all = all
+
+    def any(self, axis=None, out=None, keepdims=None, *, where=True):
```
RehanSD, line 1607: Same comments as `all`.

RehanSD: Except it seems that `any` defaults `False`:
```python
In [6]: arr.any(where=False)
Out[6]: False
```

RehanSD (changes requested):
1 small bug left, but we should be good to merge after that!

modin/core/storage_formats/pandas/query_compiler.py
```diff
@@ -1437,8 +1457,27 @@ def isin_func(df, values):
     invert = Map.register(pandas.DataFrame.__invert__, dtypes="copy")
     isna = Map.register(pandas.DataFrame.isna, dtypes=np.bool_)
     _isfinite = Map.register(
```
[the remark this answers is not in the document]
noloerino: These methods were already implemented for pandas; their kwargs might not be compatible with numpy's kwargs. Should we add separate methods to invoke the numpy versions w/ kwargs instead?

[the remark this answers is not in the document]
RehanSD: Wouldn't the args and kwargs being passed to this method be from when we call query_compiler.isinf etc. from the NumPy API Layer?

modin/numpy/arr.py
```diff
@@ -1434,41 +1529,308 @@ def sum(
...
+                    raise ValueError(
+                        f"operand was set up as a reduction along axis 0, but the length of the axis is {out.shape[0]} (it has to be 1)"
+                    )
+                if truthy_where or out is not None:
+                    return fix_dtypes_and_determine_return(
+                        result, 1, bool, out, truthy_where
+                    )
+                else:
+                    return array([True], dtype=bool)
+            result = result.to_numpy()[0, 0] if truthy_where else True
```
RehanSD, line 1586: Quick style comment - can't we just combine these two lines? return result.to_numpy()[0, 0] if truthy_where else True?

modin/numpy/arr.py
```diff
@@ -1434,41 +1529,308 @@ def sum(
...
+            result = result.to_numpy()[0, 0]
+            return result if truthy_where else True
+        if not keepdims and axis != 1:
+            result = result.transpose()
+        if truthy_where or out is not None:
+            return fix_dtypes_and_determine_return(
+                result, new_ndim, bool, out, truthy_where
+            )
+        else:
+            return numpy.ones_like(array(_query_compiler=result, _ndim=new_ndim))
```
RehanSD, line 1618: Shouldn't we throw a dtype conversion on this, so that its an array full of Trues (rather than of 1's, which are semantically equivalent)?

modin/numpy/arr.py
```diff
@@ -1434,41 +1529,308 @@ def sum(
...
+            result = result.to_numpy()[0, 0]
+            return result if truthy_where else False
+        if not keepdims and axis != 1:
+            result = result.transpose()
+        if truthy_where or out is not None:
+            return fix_dtypes_and_determine_return(
+                result, new_ndim, bool, out, truthy_where
+            )
+        else:
+            return numpy.zeros_like(array(_query_compiler=result, _ndim=new_ndim))
```
RehanSD, line 1673: same here wrt dtype conversion!

modin/numpy/test/test_array_logic.py
```diff
@@ -0,0 +1,127 @@
...
+    numpy.testing.assert_array_equal(result, numpy.array([False, True]))
+
+    where = np.array([[True, False], [False, True]])
+    result = arr.any(where=where)
+    # Result should be np.bool_ False, since mask isolates only zero elements
+    assert not result
+
+    where = np.array([[False, True], [False, False]])
+    result = arr.any(where=where, axis=1)
+    numpy.testing.assert_array_equal(result, numpy.array([True, False]))
```
RehanSD, line 70: Mind adding a test where `where=False`, to make sure we get the right result in those cases as well?

RehanSD (changes requested):
Looks great! Just a few minor style comments + one place where I'd like a little more testing - should be ready to merge soon!

modin/core/storage_formats/pandas/query_compiler.py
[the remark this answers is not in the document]
noloerino: I misread your comment and thought you were referring to isna, I'll add kwargs passing for all the below functions.

modin/numpy/arr.py
noloerino: `ones_like` inherits the type of its argument array, and `result` here is already bools after the call to the query compiler method.

https://numpy.org/doc/stable/reference/generated/numpy.ones_like.html
> Return an array of ones with the same shape and type as a given array.

modin/numpy/arr.py
noloerino: as above

modin/numpy/test/test_array_logic.py
noloerino: done

modin/core/storage_formats/pandas/query_compiler.py
```diff
@@ -418,6 +418,26 @@ def to_numpy(self, **kwargs):
         join_type="left",
     )
 
+    # Needed for numpy API
+    _logical_and = Binary.register(
+        lambda df, other, *args, **kwargs: pandas.DataFrame(
```
RehanSD, line 423: I don't think we need to wrap this in a pandas constructor, since `numpy` preserves types:
```python
In [1]: import pandas; import numpy

In [2]: arr_values = [list(range(7)) + ["a", "b"] + list(range(7))] * 7

In [3]: df = pandas.DataFrame(arr_values)

In [4]: numpy.logical_or
Out[4]: <ufunc 'logical_or'>

In [5]: numpy.logical_or(df, df)
Out[5]:
  0  1  2  3  4  5  6  7  8  9  10 11 12 13 14 15
0  0  1  2  3  4  5  6  a  b  0  1  2  3  4  5  6
1  0  1  2  3  4  5  6  a  b  0  1  2  3  4  5  6
2  0  1  2  3  4  5  6  a  b  0  1  2  3  4  5  6
3  0  1  2  3  4  5  6  a  b  0  1  2  3  4  5  6
4  0  1  2  3  4  5  6  a  b  0  1  2  3  4  5  6
5  0  1  2  3  4  5  6  a  b  0  1  2  3  4  5  6
6  0  1  2  3  4  5  6  a  b  0  1  2  3  4  5  6

In [6]: type(numpy.logical_or(df, df))
Out[6]: pandas.core.frame.DataFrame
```

modin/numpy/arr.py
```diff
@@ -1467,41 +1562,308 @@ def sum(
...
+                        f"operand was set up as a reduction along axis 0, but the length of the axis is {out.shape[0]} (it has to be 1)"
+                    )
+                if truthy_where or out is not None:
+                    return fix_dtypes_and_determine_return(
+                        result, 1, bool, out, truthy_where
+                    )
+                else:
+                    return array([True], dtype=bool)
+            result = result.to_numpy()[0, 0] if truthy_where else True
+            return result if where else True
```
RehanSD, line 1620: This should be `truthy_where`.

RehanSD: but couldn't we also just return result in this line? Since we've already made sure it has the correct value in line 1619?

modin/numpy/arr.py
```diff
@@ -1467,41 +1562,308 @@ def sum(
...
+                        f"operand was set up as a reduction along axis 0, but the length of the axis is {out.shape[0]} (it has to be 1)"
+                    )
+                if truthy_where or out is not None:
+                    return fix_dtypes_and_determine_return(
+                        result, 1, bool, out, truthy_where
+                    )
+                else:
+                    return array([False], dtype=bool)
+            result = result.to_numpy()[0, 0] if truthy_where else False
+            return result if where else False
```
RehanSD, line 1675: Same here.

modin/numpy/test/test_array_logic.py
```diff
@@ -0,0 +1,141 @@
...
+    x1 = np.array(x1)
+    modin_result = np.logical_not(x1)
+    numpy.testing.assert_array_equal(modin_result._to_numpy(), numpy_result)
+
+
+@pytest.mark.parametrize("operand1_shape", [100, (3, 100)])
+@pytest.mark.parametrize("operand2_shape", [100, (3, 100)])
+@pytest.mark.parametrize("operator", ["logical_and", "logical_or", "logical_xor"])
+def test_logical_binops(operand1_shape, operand2_shape, operator):
+    if operand1_shape != operand2_shape:
```
RehanSD, line 130: I think this has to be removed right?

RehanSD: Does not work - needs the series to be converted to a df and aligned.

RehanSD (approved):

RehanSD (changes requested):
Need to add test to ubuntu + windows on ray, dask, and python as well. This PR only adds the test to mpi

RehanSD (approved):
LGTM! Thank you for all of the awesome work on this @noloerino!

PR #5668, opened 2023-02-16
FIX-#5665: Fix operations that flatten an array, as well as handling of where argument in such operations

### LINKED ISSUE #5665: BUG: NumPy ops with axis=None call flatten, which is inefficient.
### Modin version checks

### Reproducible Example

```python
import modin.numpy as np
arr = np.array([[1, 2], [3, 4]])
arr.sum()
```

### Issue Description

Axis=None operations call flatten internally, which is inefficient.

### Expected Behavior

The operations should not be calling flatten.

### Error Logs

<details>

```python-traceback

```

</details>

### Installed Versions

<details>

</details>

RehanSD:
## What do these changes do?

- [x] Resolves #5665 ? 
- [x] tests added and passing

### DIFF
--- a/modin/numpy/arr.py
+++ b/modin/numpy/arr.py
@@ -732,7 +732,7 @@ class array(object):
                 raise numpy.AxisError(1, 1)
             target = where.where(self, numpy.nan) if isinstance(where, array) else self
             result = target._query_compiler.astype(
-                {col_name: out_dtype for col_name in result.columns}
+                {col_name: out_dtype for col_name in target._query_compiler.columns}
             ).mean(axis=0)
             if keepdims:
                 if out is not None and out.shape != (1,):
@@ -1089,8 +1089,8 @@ class array(object):
                 raise numpy.AxisError(1, 1)
             target = where.where(self, 1) if isinstance(where, array) else self
             result = target._query_compiler.astype(
-                {col_name: out_dtype for col_name in result.columns}
-            ).prod(axis=0)
+                {col_name: out_dtype for col_name in target._query_compiler.columns}
+            ).prod(axis=0, skipna=False)
             result = result.mul(initial)
             if keepdims:
                 if out is not None:
@@ -1116,8 +1116,8 @@ class array(object):
                 result = where.where(self, 1)
             result = (
                 result.astype(out_dtype)
-                ._query_compiler.prod(axis=1)
-                .prod(axis=0)
+                ._query_compiler.prod(axis=1, skipna=False)
+                .prod(axis=0, skipna=False)
                 .to_numpy()[0, 0]
             )
             result *= initial
@@ -1148,7 +1148,7 @@ class array(object):
         target = where.where(self, 1) if isinstance(where, array) else self
         result = target._query_compiler.astype(
             {col_name: out_dtype for col_name in self._query_compiler.columns}
-        ).prod(axis=axis)
+        ).prod(axis=axis, skipna=False)
         result = result.mul(initial)
         new_ndim = self._ndim - 1 if not keepdims else self._ndim
         if new_ndim == 0:
@@ -1391,8 +1391,8 @@ class array(object):
                 raise numpy.AxisError(1, 1)
             target = where.where(self, 0) if isinstance(where, array) else self
             result = target._query_compiler.astype(
-                {col_name: out_dtype for col_name in result.columns}
-            ).sum(axis=0)
+                {col_name: out_dtype for col_name in target._query_compiler.columns}
+            ).sum(axis=0, skipna=False)
             result = result.add(initial)
             if keepdims:
                 if out is not None:
@@ -1418,8 +1418,8 @@ class array(object):
                 result = where.where(self, 0)
             result = (
                 result.astype(out_dtype)
-                ._query_compiler.sum(axis=1)
-                .sum(axis=0)
+                ._query_compiler.sum(axis=1, skipna=False)
+                .sum(axis=0, skipna=False)
                 .to_numpy()[0, 0]
             )
             result += initial
@@ -1448,7 +1448,7 @@ class array(object):
         target = where.where(self, 0) if isinstance(where, array) else self
         result = target._query_compiler.astype(
             {col_name: out_dtype for col_name in self._query_compiler.columns}
-        ).sum(axis=axis)
+        ).sum(axis=axis, skipna=False)
         result = result.add(initial)
         new_ndim = self._ndim - 1 if not keepdims else self._ndim
         if new_ndim == 0:
--- a/modin/numpy/arr.py
+++ b/modin/numpy/arr.py
@@ -411,10 +411,10 @@ class array(object):
         if self._ndim == 1:
             if axis == 1:
                 raise numpy.AxisError(1, 1)
-            if not isinstance(where, array):
-                result = self._query_compiler.max(axis=0)
-            else:
-                result = where.where(self, initial)._query_compiler.max(axis=0)
+            result = self._query_compiler
+            if isinstance(where, array):
+                result = where.where(self, initial)._query_compiler
+            result = result.max(axis=0)
             if keepdims:
                 if initial is not None and result.lt(initial).any():
                     result = pd.Series([initial])._query_compiler
@@ -505,10 +505,10 @@ class array(object):
         if self._ndim == 1:
             if axis == 1:
                 raise numpy.AxisError(1, 1)
-            if not isinstance(where, array):
-                result = self._query_compiler.min(axis=0)
-            else:
-                result = where.where(self, initial)._query_compiler.min(axis=0)
+            result = self._query_compiler
+            if isinstance(where, array):
+                result = where.where(self, initial)._query_compiler
+            result = result.min(axis=0)
             if keepdims:
                 if initial is not None and result.lt(initial).any():
                     result = pd.Series([initial])._query_compiler
@@ -760,8 +760,8 @@ class array(object):
                         f"operand was set up as a reduction along axis 0, but the length of the axis is {out.shape[0]} (it has to be 1)"
                     )
                 if out is not None:
-                    out._query_compiler = array(
-                        numpy.ones(out.shape) * numpy.nan
+                    out._query_compiler = (
+                        numpy.ones_like(out) * numpy.nan
                     )._query_compiler
                 return (
                     fix_dtypes_and_determine_return(
@@ -794,8 +794,8 @@ class array(object):
                         f"operand was set up as a reduction along axis 1, but the length of the axis is {out.shape[0]} (it has to be 1)"
                     )
                 if out is not None:
-                    out._query_compiler = array(
-                        numpy.ones(out.shape) * numpy.nan
+                    out._query_compiler = (
+                        numpy.ones_like(out) * numpy.nan
                     )._query_compiler
                 return (
                     fix_dtypes_and_determine_return(
@@ -823,9 +823,7 @@ class array(object):
         if not keepdims and axis != 1:
             result = result.transpose()
         if out is not None:
-            out._query_compiler = array(
-                numpy.ones(out.shape) * numpy.nan
-            )._query_compiler
+            out._query_compiler = (numpy.ones_like(out) * numpy.nan)._query_compiler
         return (
             fix_dtypes_and_determine_return(
                 result, new_ndim, dtype, out, where is not False
--- a/modin/numpy/arr.py
+++ b/modin/numpy/arr.py
@@ -750,7 +750,7 @@ class array(object):
             if isinstance(where, array):
                 result = where.where(self, numpy.nan)._query_compiler
             result = result.astype(
-                {col_name: out_dtype for col_name in self._query_compiler.columns}
+                {col_name: out_dtype for col_name in result.columns}
             ).mean(axis=0)
             if keepdims:
                 if out is not None and out.shape != (1,):
@@ -1110,7 +1110,7 @@ class array(object):
             if isinstance(where, array):
                 result = where.where(self, 1)._query_compiler
             result = result.astype(
-                {col_name: out_dtype for col_name in self._query_compiler.columns}
+                {col_name: out_dtype for col_name in result.columns}
             ).prod(axis=0)
             result = result.mul(initial)
             if keepdims:
@@ -1420,7 +1420,7 @@ class array(object):
             if isinstance(where, array):
                 result = where.where(self, 0)._query_compiler
             result = result.astype(
-                {col_name: out_dtype for col_name in self._query_compiler.columns}
+                {col_name: out_dtype for col_name in result.columns}
             ).sum(axis=0)
             result = result.add(initial)
             if keepdims:
--- a/modin/numpy/arr.py
+++ b/modin/numpy/arr.py
@@ -29,6 +29,7 @@ from modin.core.dataframe.algebra import (
 
 from .utils import try_convert_from_interoperable_type
 
+
 def check_kwargs(order="C", subok=True, keepdims=None, casting="same_kind", where=True):
     if order not in ["K", "C"]:
         ErrorMessage.single_warning(
@@ -87,7 +88,7 @@ def fix_dtypes_and_determine_return(
             out._query_compiler = result._query_compiler
         return out
     if isinstance(where, array) and out is None:
-        from array_creation import zeros_like
+        from .array_creation import zeros_like
 
         out = zeros_like(result).astype(dtype if dtype is not None else result.dtype)
         out._query_compiler = where.where(result, out)._query_compiler
@@ -281,7 +282,20 @@ class array(object):
             args += [
                 input._query_compiler if hasattr(input, "_query_compiler") else input
             ]
-        return array(_query_compiler=new_ufunc(*args, **kwargs), _ndim=out_ndim)
+        out_kwarg = kwargs.get("out", None)
+        if out_kwarg is not None:
+            out_kwarg = out_kwarg[0]
+        where_kwarg = kwargs.get("where", True)
+        kwargs["out"] = None
+        kwargs["where"] = True
+        result = new_ufunc(*args, **kwargs)
+        return fix_dtypes_and_determine_return(
+            result,
+            out_ndim,
+            dtype=kwargs.get("dtype", None),
+            out=out_kwarg,
+            where=where_kwarg,
+        )
 
     def __array_function__(self, func, types, args, kwargs):
         from . import array_creation as creation, array_shaping as shaping, math
@@ -744,11 +758,9 @@ class array(object):
                         f"operand was set up as a reduction along axis 0, but the length of the axis is {out.shape[0]} (it has to be 1)"
                     )
                 if out is not None:
-                    out._query_compiler = (
-                        array(numpy.ones(out.shape) * numpy.nan)
-                        .astype(out_dtype)
-                        ._query_compiler
-                    )
+                    out._query_compiler = array(
+                        numpy.ones(out.shape) * numpy.nan
+                    )._query_compiler
                 return (
                     fix_dtypes_and_determine_return(
                         result, 1, dtype, out, where is not False
@@ -763,14 +775,14 @@ class array(object):
                 result = where.where(self, numpy.nan)
             result = (
                 numpy.array(
-                    [result._query_compiler.sum(axis=0).sum(axis=1).to_numpy()[0, 0]],
+                    [result._query_compiler.sum(axis=1).sum(axis=0).to_numpy()[0, 0]],
                     dtype=out_dtype,
                 )
                 / (
                     prod(self.shape)
                     - result._query_compiler.isna()
-                    .sum(axis=0)
                     .sum(axis=1)
+                    .sum(axis=0)
                     .to_numpy()[0, 0]
                 )
             ).flatten()[0]
@@ -780,11 +792,9 @@ class array(object):
                         f"operand was set up as a reduction along axis 1, but the length of the axis is {out.shape[0]} (it has to be 1)"
                     )
                 if out is not None:
-                    out._query_compiler = (
-                        array(numpy.ones(out.shape) * numpy.nan)
-                        .astype(out_dtype)
-                        ._query_compiler
-                    )
+                    out._query_compiler = array(
+                        numpy.ones(out.shape) * numpy.nan
+                    )._query_compiler
                 return (
                     fix_dtypes_and_determine_return(
                         array(numpy.array([[result]]))
@@ -811,11 +821,9 @@ class array(object):
         if not keepdims and axis != 1:
             result = result.transpose()
         if out is not None:
-            out._query_compiler = (
-                array(numpy.ones(out.shape) * numpy.nan)
-                .astype(out_dtype)
-                ._query_compiler
-            )
+            out._query_compiler = array(
+                numpy.ones(out.shape) * numpy.nan
+            )._query_compiler
         return (
             fix_dtypes_and_determine_return(
                 result, new_ndim, dtype, out, where is not False
@@ -1104,10 +1112,9 @@ class array(object):
             result = result.astype(
                 {col_name: out_dtype for col_name in self._query_compiler.columns}
             ).prod(axis=0)
-            if initial is not None:
-                result = result.mul(initial)
+            result = result.mul(initial)
             if keepdims:
-                if initial is not None and out is not None:
+                if out is not None:
                     out._query_compiler = (
                         (numpy.ones_like(out) * initial)
                         .astype(out_dtype)
@@ -1131,19 +1138,18 @@ class array(object):
                 result = where.where(self, 1)
             result = (
                 result.astype(out_dtype)
-                ._query_compiler.prod(axis=0)
-                .prod(axis=1)
+                ._query_compiler.prod(axis=1)
+                .prod(axis=0)
                 .to_numpy()
                 .flatten()[0]
             )
-            if initial is not None:
-                result *= initial
+            result *= initial
             if keepdims:
                 if out is not None and out.shape != (1, 1):
                     raise ValueError(
                         f"operand was set up as a reduction along axis 1, but the length of the axis is {out.shape[0]} (it has to be 1)"
                     )
-                if initial is not None and out is not None:
+                if out is not None:
                     out._query_compiler = (
                         (numpy.ones_like(out) * initial)
                         .astype(out_dtype)
@@ -1169,8 +1175,7 @@ class array(object):
         result = result.astype(
             {col_name: out_dtype for col_name in self._query_compiler.columns}
         ).prod(axis=axis)
-        if initial is not None:
-            result = result.mul(initial)
+        result = result.mul(initial)
         new_ndim = self._ndim - 1 if not keepdims else self._ndim
         if new_ndim == 0:
             return result.to_numpy()[0, 0] if where is not False else initial
@@ -1417,10 +1422,9 @@ class array(object):
             result = result.astype(
                 {col_name: out_dtype for col_name in self._query_compiler.columns}
             ).sum(axis=0)
-            if initial is not None:
-                result = result.add(initial)
+            result = result.add(initial)
             if keepdims:
-                if initial is not None and out is not None:
+                if out is not None:
                     out._query_compiler = (
                         (numpy.ones_like(out, dtype=out_dtype) * initial)
                         .astype(out_dtype)
@@ -1444,19 +1448,18 @@ class array(object):
                 result = where.where(self, 0)
             result = (
                 result.astype(out_dtype)
-                ._query_compiler.sum(axis=0)
-                .sum(axis=1)
+                ._query_compiler.sum(axis=1)
+                .sum(axis=0)
                 .to_numpy()
                 .flatten()[0]
             )
-            if initial is not None:
-                result += initial
+            result += initial
             if keepdims:
                 if out is not None and out.shape != (1, 1):
                     raise ValueError(
                         f"operand was set up as a reduction along axis 1, but the length of the axis is {out.shape[0]} (it has to be 1)"
                     )
-                if initial is not None and out is not None:
+                if out is not None:
                     out._query_compiler = (
                         (numpy.ones_like(out) * initial)
                         .astype(out_dtype)
@@ -1480,14 +1483,13 @@ class array(object):
         result = result.astype(
             {col_name: out_dtype for col_name in self._query_compiler.columns}
         ).sum(axis=axis)
-        if initial is not None:
-            result = result.add(initial)
+        result = result.add(initial)
         new_ndim = self._ndim - 1 if not keepdims else self._ndim
         if new_ndim == 0:
             return result.to_numpy()[0, 0] if where is not False else initial
         if not keepdims and axis != 1:
             result = result.transpose()
-        if initial is not None and out is not None:
+        if out is not None:
             out._query_compiler = (
                 (numpy.ones_like(out) * initial).astype(out_dtype)._query_compiler
             )
--- a/modin/numpy/arr.py
+++ b/modin/numpy/arr.py
@@ -284,6 +284,8 @@ class array(object):
             ]
         out_kwarg = kwargs.get("out", None)
         if out_kwarg is not None:
+            # If `out` is a modin.numpy.array, `kwargs.get("out")` returns a 1-tuple
+            # whose only element is that array, so we need to unwrap it from the tuple.
             out_kwarg = out_kwarg[0]
         where_kwarg = kwargs.get("where", True)
         kwargs["out"] = None
--- a/modin/numpy/arr.py
+++ b/modin/numpy/arr.py
@@ -411,10 +411,8 @@ class array(object):
         if self._ndim == 1:
             if axis == 1:
                 raise numpy.AxisError(1, 1)
-            result = self._query_compiler
-            if isinstance(where, array):
-                result = where.where(self, initial)._query_compiler
-            result = result.max(axis=0)
+            target = where.where(self, initial) if isinstance(where, array) else self
+            result = target._query_compiler.max(axis=0)
             if keepdims:
                 if initial is not None and result.lt(initial).any():
                     result = pd.Series([initial])._query_compiler
@@ -426,23 +424,20 @@ class array(object):
                     raise ValueError(
                         f"operand was set up as a reduction along axis 0, but the length of the axis is {out.shape[0]} (it has to be 1)"
                     )
-                return (
-                    fix_dtypes_and_determine_return(
-                        result, 1, dtype, out, (where is not False)
+                if where is not False or out is not None:
+                    return fix_dtypes_and_determine_return(
+                        result, 1, dtype, out, where is not False
                     )
-                    if where is not False or out is not None
-                    else array([initial])
-                )
+                else:
+                    return array([initial])
             if initial is not None:
                 result = max(result.to_numpy()[0, 0], initial)
             else:
                 result = result.to_numpy()[0, 0]
             return result if where is not False else initial
         if axis is None:
-            result = self._query_compiler
-            if isinstance(where, array):
-                result = where.where(self, initial)._query_compiler
-            result = result.max(axis=0).max(axis=1).to_numpy().flatten()[0]
+            target = where.where(self, initial) if isinstance(where, array) else self
+            result = target._query_compiler.max(axis=0).max(axis=1).to_numpy()[0, 0]
             if initial is not None:
                 result = max(result, initial)
             if keepdims:
@@ -454,22 +449,19 @@ class array(object):
                     out._query_compiler = (
                         numpy.ones_like(out) * initial
                     )._query_compiler
-                return (
-                    fix_dtypes_and_determine_return(
+                if where is not False or out is not None:
+                    return fix_dtypes_and_determine_return(
                         array(numpy.array([[result]]))._query_compiler,
                         2,
                         dtype,
                         out,
                         where is not False,
                     )
-                    if where is not False or out is not None
-                    else array([[initial]])
-                )
+                else:
+                    return array([[initial]])
             return result if where is not False else initial
-        result = self._query_compiler
-        if isinstance(where, array):
-            result = where.where(self, initial)._query_compiler
-        result = result.max(axis=axis)
+        target = where.where(self, initial) if isinstance(where, array) else self
+        result = target._query_compiler.max(axis=axis)
         new_ndim = self._ndim - 1 if not keepdims else self._ndim
         if new_ndim == 0:
             if initial is not None:
@@ -482,17 +474,16 @@ class array(object):
         if initial is not None and out is not None:
             out._query_compiler = (numpy.ones_like(out) * initial)._query_compiler
         intermediate = fix_dtypes_and_determine_return(
-            result, new_ndim, dtype, out, (where is not False)
+            result, new_ndim, dtype, out, where is not False
         )
         if initial is not None:
             intermediate._query_compiler = (
                 (intermediate > initial).where(intermediate, initial)._query_compiler
             )
-        return (
-            intermediate
-            if where is not False or out is not None
-            else numpy.ones_like(intermediate) * initial
-        )
+        if where is not False or out is not None:
+            return intermediate
+        else:
+            return numpy.ones_like(intermediate) * initial
 
     def min(
         self, axis=None, dtype=None, out=None, keepdims=None, initial=None, where=True
@@ -505,12 +496,10 @@ class array(object):
         if self._ndim == 1:
             if axis == 1:
                 raise numpy.AxisError(1, 1)
-            result = self._query_compiler
-            if isinstance(where, array):
-                result = where.where(self, initial)._query_compiler
-            result = result.min(axis=0)
+            target = where.where(self, initial) if isinstance(where, array) else self
+            result = target._query_compiler.min(axis=0)
             if keepdims:
-                if initial is not None and result.lt(initial).any():
+                if initial is not None and result.gt(initial).any():
                     result = pd.Series([initial])._query_compiler
                 if initial is not None and out is not None:
                     out._query_compiler = (
@@ -520,23 +509,20 @@ class array(object):
                     raise ValueError(
                         f"operand was set up as a reduction along axis 0, but the length of the axis is {out.shape[0]} (it has to be 1)"
                     )
-                return (
-                    fix_dtypes_and_determine_return(
+                if where is not False or out is not None:
+                    return fix_dtypes_and_determine_return(
                         result, 1, dtype, out, where is not False
                     )
-                    if where is not False or out is not None
-                    else array([initial])
-                )
+                else:
+                    return array([initial])
             if initial is not None:
                 result = min(result.to_numpy()[0, 0], initial)
             else:
                 result = result.to_numpy()[0, 0]
             return result if where is not False else initial
         if axis is None:
-            result = self._query_compiler
-            if isinstance(where, array):
-                result = where.where(self, initial)._query_compiler
-            result = result.min(axis=0).min(axis=1).to_numpy().flatten()[0]
+            target = where.where(self, initial) if isinstance(where, array) else self
+            result = target._query_compiler.min(axis=0).min(axis=1).to_numpy()[0, 0]
             if initial is not None:
                 result = min(result, initial)
             if keepdims:
@@ -548,22 +534,19 @@ class array(object):
                     out._query_compiler = (
                         numpy.ones_like(out) * initial
                     )._query_compiler
-                return (
-                    fix_dtypes_and_determine_return(
+                if where is not False or out is not None:
+                    return fix_dtypes_and_determine_return(
                         array(numpy.array([[result]]))._query_compiler,
                         2,
                         dtype,
                         out,
                         where is not False,
                     )
-                    if where is not False or out is not None
-                    else array([[initial]])
-                )
+                else:
+                    return array([[initial]])
             return result if where is not False else initial
-        result = self._query_compiler
-        if isinstance(where, array):
-            result = where.where(self, initial)._query_compiler
-        result = result.min(axis=axis)
+        target = where.where(self, initial) if isinstance(where, array) else self
+        result = target._query_compiler.min(axis=axis)
         new_ndim = self._ndim - 1 if not keepdims else self._ndim
         if new_ndim == 0:
             if initial is not None:
@@ -582,11 +565,10 @@ class array(object):
             intermediate._query_compiler = (
                 (intermediate < initial).where(intermediate, initial)._query_compiler
             )
-        return (
-            intermediate
-            if where is not False or out is not None
-            else numpy.ones_like(intermediate) * initial
-        )
+        if where is not False or out is not None:
+            return intermediate
+        else:
+            return numpy.ones_like(intermediate) * initial
 
     def __abs__(
         self,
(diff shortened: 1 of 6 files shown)


anmyachev:
@RehanSD this is a fairly large pull request, could you summarize the main changes?

RehanSD:
> @RehanSD this is a fairly large pull request, could you summarize the main changes?

Sure! The main changes are as follows:
1. Ensure that all top level Modin.NumPy math functions (eg sum, max, min) can accept Modin.pandas objects (and internally converts them to arrays for operations)
2. Make `__array_ufunc__` handle the `out` and `where` kwargs correctly (it previously ignored them). 
3. Prevent operations on `axis=None` from calling `.flatten()` internally, which creates a bunch of query compilers and is inefficient. 
4. Ensure that operations that take the axis argument handle the `where` kwarg correctly. (For context: on operations like `add` where is a mask that masks the output, so entries where `where` is `False` keep the default value, while for operations like `sum`, `where` is a mask on the input, so entries where `where` is `False` are discarded before summation.)

modin/numpy/math.py
```diff
@@ -44,6 +46,7 @@ def absolute(
 def add(
     x1, x2, out=None, where=True, casting="same_kind", order="K", dtype=None, subok=True
 ):
+    x1 = try_convert_from_interoperable_type(x1)
```
anmyachev, line 49: There are many functions that are not tested: `add`, `divide`, `float_power`, `floor_divide`, `power`, `prod`, `multiply`, `remainder`, `subtract`, `true_divide`, `mean`, `maximum`, `minimum`, `amax`, `amin`. What is the plan for adding tests?

modin/numpy/utils.py
```diff
@@ -0,0 +1,30 @@
...
+
+"""Collection of array utility functions for internal use."""
+
+import modin.pandas as pd
+import modin.numpy as np
+
+_INTEROPERABLE_TYPES = (pd.DataFrame, pd.Series)
+
+
+def try_convert_from_interoperable_type(obj):
```
anmyachev, line 22: It looks like it's not being tested.

modin/numpy/utils.py
```diff
@@ -0,0 +1,30 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software distributed under
+# the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific language
+# governing permissions and limitations under the License.
+
+"""Collection of array utility functions for internal use."""
+
+import modin.pandas as pd
+import modin.numpy as np
```
anmyachev, line 17: You are importing a whole `np` module, which itself imports `utils.py` this can very easily spawn cyclic imports.
```suggestion
from modin.numpy.arr import array
```

modin/numpy/utils.py
```diff
@@ -0,0 +1,30 @@
...
+
+_INTEROPERABLE_TYPES = (pd.DataFrame, pd.Series)
+
+
+def try_convert_from_interoperable_type(obj):
+    if isinstance(obj, _INTEROPERABLE_TYPES):
+        new_qc = obj._query_compiler.reset_index(drop=True)
+        new_qc.columns = range(len(new_qc.columns))
+        obj = np.array(
```
anmyachev, line 26:
```suggestion
        obj = array(
```

modin/numpy/math.py
```diff
@@ -348,12 +362,48 @@ def mean(x1, axis=None, dtype=None, out=None, keepdims=None, *, where=True):
...
 # implementation will automatically handle this, so we can just use NumPy's maximum/minimum
 # since that will route to our array's ufunc.
-maximum = numpy.maximum
+def maximum(
+    x1, x2, out=None, where=True, casting="same_kind", order="K", dtype=None, subok=True
+):
+    x1 = try_convert_from_interoperable_type(x1)
+    if not isinstance(x1, array):
+        ErrorMessage.single_warning(
+            f"Modin NumPy only supports objects of modin.numpy.array types for maximum, not {type(x1)}. Defaulting to NumPy."
```
anmyachev, line 371: This error message is very easily parametrized by a function name.

modin/numpy/math.py
```diff
@@ -348,12 +362,48 @@ def mean(x1, axis=None, dtype=None, out=None, keepdims=None, *, where=True):
...
+        dtype=dtype,
+        subok=subok,
+    )
+
 
-minimum = numpy.minimum
+def minimum(
+    x1, x2, out=None, where=True, casting="same_kind", order="K", dtype=None, subok=True
+):
+    x1 = try_convert_from_interoperable_type(x1)
```
anmyachev, line 388:
```suggestion
    x1 = try_convert_from_interoperable_type(x1)
    x2 = try_convert_from_interoperable_type(x2)
```

modin/numpy/math.py
```diff
@@ -348,12 +362,48 @@ def mean(x1, axis=None, dtype=None, out=None, keepdims=None, *, where=True):
 # Maximum and minimum are ufunc's in NumPy, which means that our array's __array_ufunc__
 # implementation will automatically handle this, so we can just use NumPy's maximum/minimum
 # since that will route to our array's ufunc.
-maximum = numpy.maximum
+def maximum(
+    x1, x2, out=None, where=True, casting="same_kind", order="K", dtype=None, subok=True
+):
+    x1 = try_convert_from_interoperable_type(x1)
```
anmyachev, line 368:
```suggestion
    x1 = try_convert_from_interoperable_type(x1)
    x2 = try_convert_from_interoperable_type(x2)
```

modin/numpy/math.py
```diff
@@ -309,6 +321,7 @@ def sum(arr, axis=None, dtype=None, out=None, keepdims=None, initial=None, where
 def true_divide(
     x1, x2, out=None, where=True, casting="same_kind", order="K", dtype=None, subok=True
 ):
+    x1 = try_convert_from_interoperable_type(x1)
```
anmyachev, line 324:
```suggestion
    x1 = try_convert_from_interoperable_type(x1)
    x2 = try_convert_from_interoperable_type(x2)
```

modin/numpy/math.py
```diff
@@ -262,6 +272,7 @@ def remainder(
 def subtract(
     x1, x2, out=None, where=True, casting="same_kind", order="K", dtype=None, subok=True
 ):
+    x1 = try_convert_from_interoperable_type(x1)
```
anmyachev, line 275:
```suggestion
    x1 = try_convert_from_interoperable_type(x1)
    x2 = try_convert_from_interoperable_type(x2)
```

modin/numpy/math.py
```diff
@@ -231,6 +240,7 @@ def multiply(
 def remainder(
     x1, x2, out=None, where=True, casting="same_kind", order="K", dtype=None, subok=True
 ):
+    x1 = try_convert_from_interoperable_type(x1)
```
anmyachev, line 243:
```suggestion
    x1 = try_convert_from_interoperable_type(x1)
    x2 = try_convert_from_interoperable_type(x2)
```

modin/numpy/math.py
```diff
@@ -203,6 +211,7 @@ def prod(a, axis=None, out=None, keepdims=None, where=True, dtype=None, initial=
 def multiply(
     x1, x2, out=None, where=True, casting="same_kind", order="K", dtype=None, subok=True
 ):
+    x1 = try_convert_from_interoperable_type(x1)
```
anmyachev, line 214:
```suggestion
    x1 = try_convert_from_interoperable_type(x1)
    x2 = try_convert_from_interoperable_type(x2)
```

modin/numpy/math.py
```diff
@@ -156,6 +162,7 @@ def floor_divide(
 def power(
     x1, x2, out=None, where=True, casting="same_kind", order="K", dtype=None, subok=True
 ):
+    x1 = try_convert_from_interoperable_type(x1)
```
anmyachev, line 165:
```suggestion
    x1 = try_convert_from_interoperable_type(x1)
    x2 = try_convert_from_interoperable_type(x2)
```

modin/numpy/math.py
```diff
@@ -128,6 +133,7 @@ def float_power(
 def floor_divide(
     x1, x2, out=None, where=True, casting="same_kind", order="K", dtype=None, subok=True
 ):
+    x1 = try_convert_from_interoperable_type(x1)
```
anmyachev, line 136:
```suggestion
    x1 = try_convert_from_interoperable_type(x1)
    x2 = try_convert_from_interoperable_type(x2)
```

modin/numpy/math.py
```diff
@@ -100,6 +104,7 @@ def divide(
 def float_power(
     x1, x2, out=None, where=True, casting="same_kind", order="K", dtype=None, subok=True
 ):
+    x1 = try_convert_from_interoperable_type(x1)
```
anmyachev, line 107:
```suggestion
    x1 = try_convert_from_interoperable_type(x1)
    x2 = try_convert_from_interoperable_type(x2)
```

modin/numpy/math.py
```diff
@@ -72,6 +75,7 @@ def add(
 def divide(
     x1, x2, out=None, where=True, casting="same_kind", order="K", dtype=None, subok=True
 ):
+    x1 = try_convert_from_interoperable_type(x1)
```
anmyachev, line 78:
```suggestion
    x1 = try_convert_from_interoperable_type(x1)
    x2 = try_convert_from_interoperable_type(x2)
```

modin/numpy/math.py
anmyachev, line 49:
```suggestion
    x1 = try_convert_from_interoperable_type(x1)
    x2 = try_convert_from_interoperable_type(x2)
```

modin/numpy/arr.py
```diff
@@ -294,7 +282,20 @@ def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
             args += [
                 input._query_compiler if hasattr(input, "_query_compiler") else input
             ]
-        return array(_query_compiler=new_ufunc(*args, **kwargs), _ndim=out_ndim)
+        out_kwarg = kwargs.get("out", None)
+        if out_kwarg is not None:
+            out_kwarg = out_kwarg[0]
```
anmyachev, line 287: Why?

modin/numpy/arr.py
```diff
@@ -408,71 +409,88 @@ def max(
         if self._ndim == 1:
             if axis == 1:
                 raise numpy.AxisError(1, 1)
-            result = self._query_compiler.max(axis=0)
+            if not isinstance(where, array):
+                result = self._query_compiler.max(axis=0)
+            else:
+                result = where.where(self, initial)._query_compiler.max(axis=0)
```
anmyachev, line 415: I think this is an important case, we need to test it.

modin/numpy/math.py
RehanSD: We test the `array.add`, etc, which these call, so I think that should be ok?

modin/numpy/utils.py
RehanSD: I'm not sure we need to test `try_convert` do we? I can add tests if you think its necessary!

modin/numpy/utils.py
RehanSD: Makes sense to me!

modin/numpy/math.py
RehanSD: Sounds good to me!

modin/numpy/math.py
RehanSD: We don't need to convert the second operand, since we do so inside the array class!

modin/numpy/arr.py
RehanSD: If the `out_kwarg` is a modin.numpy.array, getting it from kwargs returns a 1-length tuple containing the array. I'll add a comment clarifying this as well!

modin/numpy/arr.py
RehanSD: Makes sense to me!

modin/numpy/utils.py
RehanSD: Tried this and got a circular import error. Reverted!

modin/numpy/utils.py
anmyachev: At least for one function but `_INTEROPERABLE_TYPES`, yes.

modin/numpy/utils.py
anmyachev: oops :)

modin/numpy/math.py
anmyachev: Apparently I do not fully understand how the code works.

Could you explain where we convert the second operand?
And why can't we convert the first operand there?

At least the next check can be rewritten as follows:
```python
if not isinstance(x1, (array, pd.DataFrame, pd.Series)):
```

modin/numpy/math.py
anmyachev: So is this a mistake in `codecov`?

modin/numpy/arr.py
```diff
@@ -704,50 +741,101 @@ def mean(self, axis=None, dtype=None, out=None, keepdims=None, *, where=True):
...
-            return result.to_numpy()[0, 0] if where else numpy.nan
+            return result.to_numpy()[0, 0] if where is not False else numpy.nan
         if not keepdims and axis != 1:
             result = result.transpose()
         if out is not None:
-            out._query_compiler = (out * numpy.nan).astype(out_dtype)._query_compiler
-        return fix_dtypes_and_determine_return(result, new_ndim, dtype, out, where)
+            out._query_compiler = array(
+                numpy.ones(out.shape) * numpy.nan
+            )._query_compiler
```
anmyachev, line 828: Why is the `array` call here?
What is the difference with `out._query_compiler = (numpy.ones_like(out) * initial)._query_compiler`?

anmyachev: Also why is `.astype(out_dtype)` not called now?

RehanSD: The main difference is in the dtype of the created array. I changed it to this because I thought that the old dtype was causing the ubuntu test to fail, but it was actually the as type part - which was causing it to fail since we had an array of NaN's and tried to convert that to an array of dtype int, which was failing.

modin/numpy/math.py
RehanSD: Yeah - I think codecov is complaining that `modin.numpy.add` isn't being tested (which is technically true), but `modin.numpy.add` is literally just a wrapper for `modin.numpy.array.add`, which is tested, so I think its fine not to test the wrapper?

modin/numpy/utils.py
RehanSD: To clarify - you think we should add a test for the `try_convert_from_interoperable_type` function to ensure that it correctly converts convertible objects, and flags for objects that can't be converted? I will add it!

modin/numpy/math.py
RehanSD: So all of the `modin.numpy` methods are simply wrappers that internally call `modin.numpy.array.<method>`. E.g.:
```python
def add(...):
x1 = try_convert_from_interoperable_type(x1)
if isinstance(x1, array):
x1.add ...
```
So we need to make sure that the first operand is converted from a modin pandas object to a modin numpy object so that it can call the appropriate method, but inside the `modin.numpy.array.add` method, the first line attempts to convert the other operand to an array.

Take `__add__` as an example. On [this](https://github.com/RehanSD/modin/blob/a94b84cd6ebc63a0d951476ee1fd34629ae7c6de/modin/numpy/arr.py#L868) line, we call `_binary_op` - `_binary_op`, makes sure that the second operand is of type array, and errors if it isn't on these lines: https://github.com/RehanSD/modin/blob/a94b84cd6ebc63a0d951476ee1fd34629ae7c6de/modin/numpy/arr.py#L634-L638 - so we don't need to convert the second operand at the numpy layer.

RehanSD:
Thank you for all of the reviews @anmyachev!

modin/numpy/utils.py
anmyachev: Exactly! Thanks

modin/numpy/math.py
anmyachev: Ok! Thank you for the clarification.

modin/numpy/math.py
anmyachev: Got it! Thank you for the clarification.

modin/numpy/arr.py
anmyachev, line 828: Will it work? More consistent with other places, of course, if it does not bring any undesirable effects in this case or slow it down.
```suggestion
            out._query_compiler = (numpy.ones_like(out) * numpy.nan)._query_compiler
```

modin/numpy/arr.py
```diff
@@ -704,50 +741,101 @@ def mean(self, axis=None, dtype=None, out=None, keepdims=None, *, where=True):
...
             if keepdims:
                 if out is not None and out.shape != (1,):
                     raise ValueError(
                         f"operand was set up as a reduction along axis 0, but the length of the axis is {out.shape[0]} (it has to be 1)"
                     )
-                return fix_dtypes_and_determine_return(result, 1, dtype, out, where)
+                if out is not None:
+                    out._query_compiler = array(
+                        numpy.ones(out.shape) * numpy.nan
+                    )._query_compiler
```
anmyachev, line 765: Will it work?
```suggestion
                    out._query_compiler = (numpy.ones_like(out) * numpy.nan)._query_compiler
```

modin/numpy/arr.py
```diff
@@ -704,50 +741,101 @@ def mean(self, axis=None, dtype=None, out=None, keepdims=None, *, where=True):
...
             if keepdims:
                 if out is not None and out.shape != (1, 1):
                     raise ValueError(
-                        f"operand was set up as a reduction along axis 0, but the length of the axis is {out.shape[0]} (it has to be 1)"
+                        f"operand was set up as a reduction along axis 1, but the length of the axis is {out.shape[0]} (it has to be 1)"
+                    )
+                if out is not None:
+                    out._query_compiler = array(
+                        numpy.ones(out.shape) * numpy.nan
+                    )._query_compiler
```
anmyachev, line 799: Will it work?
```suggestion
                    out._query_compiler = (numpy.ones_like(out) * numpy.nan)._query_compiler
```

modin/numpy/arr.py
```diff
@@ -408,71 +411,88 @@ def max(
...
-                dtype=dtype,
-                out=out,
-                keepdims=None,
-                initial=initial,
-                where=where,
-            )
+            result = self._query_compiler
+            if isinstance(where, array):
+                result = where.where(self, initial)._query_compiler
+            result = result.max(axis=0).max(axis=1).to_numpy().flatten()[0]
```
anmyachev, line 445: Will it work?
```suggestion
            result = result.max().to_numpy().flatten()[0]
```

modin/numpy/arr.py
```diff
@@ -485,71 +505,88 @@ def min(
...
-                dtype=dtype,
-                out=out,
-                keepdims=None,
-                initial=initial,
-                where=where,
-            )
+            result = self._query_compiler
+            if isinstance(where, array):
+                result = where.where(self, initial)._query_compiler
+            result = result.min(axis=0).min(axis=1).to_numpy().flatten()[0]
```
anmyachev, line 539: Will it work?
```suggestion
            result = result.min().to_numpy().flatten()[0]
```

modin/numpy/arr.py
```diff
@@ -408,71 +411,88 @@ def max(
         if self._ndim == 1:
             if axis == 1:
                 raise numpy.AxisError(1, 1)
-            result = self._query_compiler.max(axis=0)
+            if not isinstance(where, array):
+                result = self._query_compiler.max(axis=0)
+            else:
+                result = where.where(self, initial)._query_compiler.max(axis=0)
```
anmyachev, line 417:
```suggestion
            if isinstance(where, array):
                result = where.where(self, initial)._query_compiler
            result = result.max(axis=0)
```

modin/numpy/arr.py
```diff
@@ -485,71 +505,88 @@ def min(
         if self._ndim == 1:
             if axis == 1:
                 raise numpy.AxisError(1, 1)
-            result = self._query_compiler.min(axis=0)
+            if not isinstance(where, array):
+                result = self._query_compiler.min(axis=0)
+            else:
+                result = where.where(self, initial)._query_compiler.min(axis=0)
```
anmyachev, line 511:
```suggestion
            if isinstance(where, array):
                result = where.where(self, initial)._query_compiler
            result = result.min(axis=0)
```

anmyachev:
@RehanSD these are some of my last comments, please answer them and if no one minds, I'll merge it on Monday evening.

anmyachev (approved):

modin/numpy/math.py
RehanSD: Added a method to `ErrorMessage` class to refactor this!

modin/numpy/arr.py
RehanSD: I don't think it will slow anything down - I'll go ahead and change it - good catch, thank you!

modin/numpy/arr.py
RehanSD: This won't work - the query compiler expects an axis argument to functions like `max` and `min`. If none is provided, it errors out. (this is because `df.max(axis=None)` is the same as `df.max(axis=0)`, unlike with NumPy, where `axis=None` means a reduction to a single value).

RehanSD:
Thank you @anmyachev - really appreciate the help getting this over the finish line! I've gone ahead and addressed your comments, and am pushing a revised version of the code!

vnlitvinov:
> and if no one minds, I'll merge it on Monday evening.

Do note that as of right now, CI is minding the merge 😉

modin/numpy/arr.py
```diff
@@ -294,7 +282,22 @@ def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
             args += [
                 input._query_compiler if hasattr(input, "_query_compiler") else input
             ]
-        return array(_query_compiler=new_ufunc(*args, **kwargs), _ndim=out_ndim)
+        out_kwarg = kwargs.get("out", None)
+        if out_kwarg is not None:
+            # If `out` is a modin.numpy.array, `kwargs.get("out")` returns a 1-tuple
```
vnlitvinov, line 287: but what if `out` is _not_ a `modin.numpy.array`?.. shouldn't we `isinstance()` check it instead of `is not None`?

modin/numpy/arr.py
```diff
@@ -408,71 +411,88 @@ def max(
         if self._ndim == 1:
             if axis == 1:
                 raise numpy.AxisError(1, 1)
-            result = self._query_compiler.max(axis=0)
+            result = self._query_compiler
+            if isinstance(where, array):
+                result = where.where(self, initial)._query_compiler
+            result = result.max(axis=0)
```
vnlitvinov, line 417:
```suggestion
            target = where.where(self, initial) if isinstance(where, array) else self
            result = target._query_compiler.max(axis=0)
```
?

modin/numpy/arr.py
```diff
@@ -408,71 +411,88 @@ def max(
...
                         f"operand was set up as a reduction along axis 0, but the length of the axis is {out.shape[0]} (it has to be 1)"
                     )
-                return fix_dtypes_and_determine_return(result, 1, dtype, out, where)
+                return (
+                    fix_dtypes_and_determine_return(
+                        result, 1, dtype, out, (where is not False)
+                    )
+                    if where is not False or out is not None
+                    else array([initial])
+                )
```
vnlitvinov, line 435:
```suggestion
                if where is not False or out is not None:
                    return fix_dtypes_and_determine_return(result, 1, dtype, out, where is not False)
                return array([initial])
```
I wonder if this reads better...

modin/numpy/arr.py
```diff
@@ -408,71 +411,88 @@ def max(
...
-                axis=axis,
-                dtype=dtype,
-                out=out,
-                keepdims=None,
-                initial=initial,
-                where=where,
-            )
+            result = self._query_compiler
+            if isinstance(where, array):
+                result = where.where(self, initial)._query_compiler
```
vnlitvinov, line 444:
```suggestion
            result = (where.where(self, initial) if isinstance(where, array) else self)._query_compiler
```
?
or like in my suggestion for `max()` with creation of intermediate `target`

modin/numpy/arr.py
```diff
@@ -408,71 +411,88 @@ def max(
...
+            result = result.max(axis=0).max(axis=1).to_numpy().flatten()[0]
+            if initial is not None:
+                result = max(result, initial)
             if keepdims:
                 if out is not None and out.shape != (1, 1):
                     raise ValueError(
-                        f"operand was set up as a reduction along axis 0, but the length of the axis is {out.shape[0]} (it has to be 1)"
+                        f"operand was set up as a reduction along axis 1, but the length of the axis is {out.shape[0]} (it has to be 1)"
```
vnlitvinov, line 451: this error looks misleading, it talks about axis 1, reports axis 0, and _really checks_ both axes 😆

modin/numpy/arr.py
```diff
@@ -408,71 +411,88 @@ def max(
...
+                    fix_dtypes_and_determine_return(
+                        array(numpy.array([[result]]))._query_compiler,
+                        2,
+                        dtype,
+                        out,
+                        where is not False,
                     )
-                return fix_dtypes_and_determine_return(
-                    array(numpy.array([[result]]))._query_compiler, 2, dtype, out, where
+                    if where is not False or out is not None
```
vnlitvinov, line 465: I'm not sure I like this `a if foo else b` inside the `return` statement when it's so lengthy... I personally like it done like 
```python
if foo:
    return a
return b
```

modin/numpy/arr.py
```diff
@@ -408,71 +411,88 @@ def max(
...
+                    else array([[initial]])
                 )
-            return result
-        result = self._query_compiler.max(axis=axis)
+            return result if where is not False else initial
+        result = self._query_compiler
+        if isinstance(where, array):
+            result = where.where(self, initial)._query_compiler
+        result = result.max(axis=axis)
```
vnlitvinov, line 472:
```suggestion
        target = where.where(self, initial) if isinstance(where, array) else self
        result = target._query_compiler.max(axis=axis)
```

modin/numpy/arr.py
```diff
@@ -408,71 +411,88 @@ def max(
...
-            else:
-                out = (
-                    numpy.ones_like(array(_query_compiler=result, _ndim=new_ndim))
-                    * initial
-                ).astype(self.dtype)
+        if initial is not None and out is not None:
+            out._query_compiler = (numpy.ones_like(out) * initial)._query_compiler
         intermediate = fix_dtypes_and_determine_return(
-            result, new_ndim, dtype, out, where
+            result, new_ndim, dtype, out, (where is not False)
```
vnlitvinov, line 485: why do we need braces around the `where is not False`?

modin/numpy/arr.py
```diff
@@ -485,71 +505,88 @@ def min(
         if self._ndim == 1:
             if axis == 1:
                 raise numpy.AxisError(1, 1)
-            result = self._query_compiler.min(axis=0)
```
vnlitvinov, line 488: starting from here, it looks like a copy of `.max()` but with `min()` instead... why did we copy-paste so much?

I won't repeat my suggestions on `target = where.where(...) if foo else self` here, but they do apply

modin/numpy/arr.py
```diff
@@ -485,71 +505,88 @@ def min(
...
+            result = self._query_compiler
+            if isinstance(where, array):
+                result = where.where(self, initial)._query_compiler
+            result = result.min(axis=0)
             if keepdims:
-                if initial is not None and result.lt(initial):
+                if initial is not None and result.lt(initial).any():
```
vnlitvinov, line 513:
```suggestion
                if initial is not None and result.gt(initial).any():
```
I wonder why we `.lt()` both in `.max()` and in `.min()`, feels like it should be `.gt()` here?..

modin/numpy/arr.py
```diff
@@ -704,50 +741,99 @@ def mean(self, axis=None, dtype=None, out=None, keepdims=None, *, where=True):
             if dtype is not None
             else (out.dtype if out is not None else self.dtype)
         )
+        out_type = out_dtype if not hasattr(out_dtype, "type") else out_dtype.type
```
vnlitvinov, line 744:
```suggestion
        out_type = getattr(out_dtype, "type", out_dtype)
```
😉 
does the same, works slightly faster (because `hasattr()` internally does the `getattr()` thing and returns True if no error happened)

modin/numpy/arr.py
```diff
@@ -704,50 +741,99 @@ def mean(self, axis=None, dtype=None, out=None, keepdims=None, *, where=True):
...
+        if isinstance(where, array) and issubclass(out_type, numpy.integer):
+            out_dtype = numpy.float64
```
vnlitvinov, line 746: why?..

modin/numpy/arr.py
```diff
@@ -704,50 +741,99 @@ def mean(self, axis=None, dtype=None, out=None, keepdims=None, *, where=True):
...
         check_kwargs(keepdims=keepdims, where=where)
         if self._ndim == 1:
             if axis == 1:
                 raise numpy.AxisError(1, 1)
-            result = self._query_compiler.astype(
-                {col_name: out_dtype for col_name in self._query_compiler.columns}
+            result = self._query_compiler
+            if isinstance(where, array):
+                result = where.where(self, numpy.nan)._query_compiler
+            result = result.astype(
```
vnlitvinov, line 754:
```suggestion
            target = where.where(self, numpy.nan) if isinstance(where, array) else self
            result = target._query_compiler.astype(
```

modin/numpy/arr.py
```diff
@@ -704,50 +741,99 @@ def mean(self, axis=None, dtype=None, out=None, keepdims=None, *, where=True):
...
+                        numpy.ones_like(out) * numpy.nan
+                    )._query_compiler
+                return (
+                    fix_dtypes_and_determine_return(
+                        result, 1, dtype, out, where is not False
+                    )
+                    if where is not False or out is not None
+                    else array([numpy.nan], dtype=out_dtype)
+                )
             return result.to_numpy()[0, 0] if where else numpy.nan
```
vnlitvinov, line 773: shouldn't we change this to `if where is not False` as well?

modin/numpy/arr.py
```diff
@@ -704,50 +741,99 @@ def mean(self, axis=None, dtype=None, out=None, keepdims=None, *, where=True):
...
+                    [result._query_compiler.sum(axis=1).sum(axis=0).to_numpy()[0, 0]],
+                    dtype=out_dtype,
+                )
+                / (
+                    prod(self.shape)
+                    - result._query_compiler.isna()
+                    .sum(axis=1)
+                    .sum(axis=0)
+                    .to_numpy()[0, 0]
+                )
```
vnlitvinov, line 789: I wonder if extracting the calculation of "now many NAs is there" to a separate variable would improve the readability... my first impression was that this piece was accidentally calling `.sum()` on only part of the equation, until I realized what was going on

modin/numpy/arr.py
```diff
@@ -704,50 +741,99 @@ def mean(self, axis=None, dtype=None, out=None, keepdims=None, *, where=True):
...
+                    if where is not False or out is not None
+                    else array([[numpy.nan]], dtype=out_dtype)
                 )
-            return result
-        result = self._query_compiler.astype(
+            return result if where is not False else numpy.nan
+        result = self._query_compiler
+        if isinstance(where, array):
+            result = where.where(self, numpy.nan)._query_compiler
+        result = result.astype(
```
vnlitvinov, line 817:
```suggestion
        target = where.where(self, numpy.nan) if isinstance(where, array) else self
        result = target._query_compiler.astype(
```

modin/numpy/arr.py
```diff
@@ -1020,74 +1106,94 @@ def prod(
         if self._ndim == 1:
             if axis == 1:
                 raise numpy.AxisError(1, 1)
-            result = self._query_compiler.astype(
-                {col_name: out_dtype for col_name in self._query_compiler.columns}
+            result = self._query_compiler
+            if isinstance(where, array):
+                result = where.where(self, 1)._query_compiler
+            result = result.astype(
```
vnlitvinov, line 1112: I will stop my suggestion of `target = where.where(...) if foo else self` from now on, but it does apply here and in other places

modin/numpy/arr.py
```diff
@@ -1020,74 +1106,94 @@ def prod(
...
+                {col_name: out_dtype for col_name in result.columns}
             ).prod(axis=0)
-            if initial is not None:
-                result = result.mul(initial)
+            result = result.mul(initial)
```
vnlitvinov, line 1115: what would happen if `initial` is None?

modin/numpy/arr.py
```diff
@@ -1310,72 +1416,93 @@ def sum(
...
-                {col_name: out_dtype for col_name in self._query_compiler.columns}
+            result = self._query_compiler
+            if isinstance(where, array):
+                result = where.where(self, 0)._query_compiler
+            result = result.astype(
+                {col_name: out_dtype for col_name in result.columns}
             ).sum(axis=0)
-            if initial is not None:
-                result = result.add(initial)
+            result = result.add(initial)
```
vnlitvinov, line 1425: what would happen if `initial` is None?

modin/numpy/arr.py
```diff
@@ -1310,72 +1416,93 @@ def sum(
...
-                    out=out,
-                    keepdims=None,
-                    initial=initial,
-                    where=where,
-                )
+                result.astype(out_dtype)
+                ._query_compiler.sum(axis=1)
+                .sum(axis=0)
+                .to_numpy()
+                .flatten()[0]
```
vnlitvinov, line 1454: why do we need to `.flatten()`, cannot we just `[0, 0]` here? it should be reduced to a single cell anyway

modin/numpy/math.py
```diff
@@ -44,10 +44,9 @@ def absolute(
 def add(
     x1, x2, out=None, where=True, casting="same_kind", order="K", dtype=None, subok=True
 ):
+    x1 = try_convert_from_interoperable_type(x1)
     if not isinstance(x1, array):
-        ErrorMessage.single_warning(
-            f"Modin NumPy only supports objects of modin.numpy.array types for add, not {type(x1)}. Defaulting to NumPy."
-        )
+        ErrorMessage.bad_type_for_numpy_op("add", type(x1))
```
vnlitvinov, line 49: why don't we convert `x2` the same way? here and for other binary functions

modin/numpy/math.py
```diff
@@ -14,16 +14,16 @@
...
 
 def absolute(
     x, out=None, where=True, casting="same_kind", order="K", dtype=None, subok=True
 ):
+    x = try_convert_from_interoperable_type(x)
     if not isinstance(x, array):
-        ErrorMessage.single_warning(
-            f"Modin NumPy only supports objects of modin.numpy.array types for absolute, not {type(x)}. Defaulting to NumPy."
-        )
+        ErrorMessage.bad_type_for_numpy_op("absolute", type(x))
```
vnlitvinov, line 26: I wonder if this should be hidden in a decorator, like this (untested):
```python
from functools import wraps

def _default_to_numpy(func):
    @wraps(func):
    def wrapped(arr, *args, **kw):
        arr = try_convert_from_interoperable_type(arr)
        if not isinstance(arr, array):
            ErrorMessage.bad_type_for_numpy_op(func.__name__, type(arr))
            return getattr(numpy, func.__name__)(arr, *args, **kw)
        return func(arr, *args, **kw)
    return wrapped


@_default_to_numpy
def absolute(
    x, out=None, where=True, casting="same_kind", order="K", dtype=None, subok=True
):
    return x.absolute(
        out=out, where=where, casting=casting, order=order, dtype=dtype, subok=subok
    )

@_default_to_numpy
def add(
    x1, x2, out=None, where=True, casting="same_kind", order="K", dtype=None, subok=True
):
    return x1.__add__(
        x2,
        out=out,
        where=where,
        casting=casting,
        order=order,
        dtype=dtype,
        subok=subok,
    )

# etc.
```
?

modin/numpy/math.py
```diff
@@ -348,16 +336,46 @@ def mean(x1, axis=None, dtype=None, out=None, keepdims=None, *, where=True):
 # Maximum and minimum are ufunc's in NumPy, which means that our array's __array_ufunc__
 # implementation will automatically handle this, so we can just use NumPy's maximum/minimum
 # since that will route to our array's ufunc.
```
vnlitvinov, line 338: this comment seems outdated now

modin/numpy/test/test_array_axis_functions.py
```diff
@@ -122,6 +137,13 @@ def test_min():
     numpy_result = numpy_arr.min(keepdims=True)
     assert modin_result.shape == numpy_result.shape
     numpy.testing.assert_array_equal(modin_result._to_numpy(), numpy_result)
+    numpy_arr = numpy.array([1, -10000, 2, 3, 4, 5])
+    modin_arr = np.array(numpy_arr)
+    numpy_mask = numpy.array([True, False, True, True, True, True])
+    modin_mask = np.array(numpy_mask)
+    assert numpy_arr.max(where=numpy_mask, initial=5) == modin_arr.max(
```
vnlitvinov, line 144:
```suggestion
    assert numpy_arr.min(where=numpy_mask, initial=5) == modin_arr.min(
```
I don't see a point of testing .`max()` within `test_min()` function 😉

modin/numpy/test/test_array_axis_functions.py
```diff
@@ -189,6 +211,14 @@ def test_min():
     numpy_result = numpy_arr.min(axis=0, initial=4, out=numpy_out, where=numpy_where)
     numpy.testing.assert_array_equal(modin_result._to_numpy(), numpy_result)
     numpy.testing.assert_array_equal(modin_out._to_numpy(), numpy_out)
+    numpy_arr = numpy.array([[1, -10000, 2], [3, 4, 5]])
+    modin_arr = np.array(numpy_arr)
+    numpy_mask = numpy.array([[True, False, True], [True, True, True]])
+    modin_mask = np.array(numpy_mask)
+    numpy.testing.assert_equal(
+        numpy_arr.max(where=numpy_mask, initial=5),
+        modin_arr.max(where=modin_mask, initial=5),
```
vnlitvinov, line 220:
```suggestion
        numpy_arr.min(where=numpy_mask, initial=5),
        modin_arr.min(where=modin_mask, initial=5),
```

vnlitvinov (changes requested):

vnlitvinov:
I should've posted a summary... most of my suggestions are a matter of taste, but I believe that `.min()` implementation is bugged when `initial` is not None (https://github.com/modin-project/modin/pull/5668#discussion_r1118326375), and I'm sure tests for `.min()` have a copy-paste error as they're testing for `.max()` instead

modin/numpy/arr.py
RehanSD: That's handled by the function `fix_dtypes_and_determine_return`. It calls `check_can_broadcast_to_output`, which first checks if the out variable is a modin.numpy.array, and errors if it is not!

modin/numpy/arr.py
RehanSD: I added them for clarity - it's not really necessary, just a personal style thing.

modin/numpy/arr.py
RehanSD: True, but in this case I was just copying the error NumPy gives.

modin/numpy/arr.py
RehanSD: A lot of the code is very similar with some small changes - I think @noloerino is planning a refactor for the future!

modin/numpy/arr.py
RehanSD: Good catch, thank you!

modin/numpy/arr.py
RehanSD: Fixed!

modin/numpy/arr.py
RehanSD: if we try and convert out to the out dtype on ubuntu, it errors, since out will contain NaNs.

modin/numpy/arr.py
RehanSD: I think it would fail, but we set `initial` to 1 if it is `None` at the top, so it doesn't matter.

modin/numpy/math.py
RehanSD: We don't need to - since the `binary_op` code takes care of converting x2!

modin/numpy/math.py
RehanSD: Perhaps - I think its little enough code that its fine for now IMO!

modin/numpy/math.py
RehanSD: Fixed!

modin/numpy/test/test_array_axis_functions.py
RehanSD: Good catch, thank you!

modin/numpy/test/test_array_axis_functions.py
RehanSD: Good catch, thank you!

modin/numpy/arr.py
RehanSD: No - in this case, where should just be evaluated for truthiness, since we're returning a single value.

modin/numpy/arr.py
RehanSD: I can break this up and add some comments to help explain what's going on!

RehanSD:
Thank you for the review @vnlitvinov! I've gone ahead and addressed your comments!

modin/numpy/arr.py
vnlitvinov: TBH I still don't understand why do we pass a variable using a 1-element tuple... feels super-weird. Could you pls explain?

modin/numpy/arr.py
vnlitvinov: As far as I can see, this handles the case when after `where.where()` we have some `NaN`-s inside... but it's not always going to happen? What does numpy do in this case, does it convert the output to `float` regardless of whether `NaN`-s are present?

P.S. I don't understand what "ubuntu" means here 🙃 so I "played by ear" and guessed what you meant, hopefully correctly...

modin/numpy/arr.py
vnlitvinov: I wonder if this should be a comment in the code, as this is the only place where the pattern is different...

modin/numpy/math.py
vnlitvinov: I guess this can be handled during the refactor you and @noloerino were planning ahead...

modin/numpy/arr.py
```diff
@@ -426,23 +424,20 @@ def max(
...
-                        result, 1, dtype, out, (where is not False)
+                if where is not False or out is not None:
+                    return fix_dtypes_and_determine_return(
+                        result, 1, dtype, out, where is not False
                     )
-                    if where is not False or out is not None
-                    else array([initial])
-                )
+                else:
+                    return array([initial])
```
vnlitvinov, line 432: technically you can do this:
```python
if where is not False or out is not None:
    return fix_dtypes_and_determine_return(...)
return array([initial])
```
i.e. get rid of `else:` line completely... though I think that it may actually read clearer your way now

modin/numpy/arr.py
```diff
@@ -763,31 +744,33 @@ def mean(self, axis=None, dtype=None, out=None, keepdims=None, *, where=True):
...
-                / (
-                    prod(self.shape)
-                    - result._query_compiler.isna()
-                    .sum(axis=1)
-                    .sum(axis=0)
-                    .to_numpy()[0, 0]
-                )
-            ).flatten()[0]
+                / num_total_elements
+            )[0]
```
vnlitvinov, line 773: now I wonder what is the reason of making a single-value numpy array to only take its sole element straight away... cannot we do it like
```python
result = result._query_compiler.sum(axis=1).sum(axis=0).to_numpy()[0, 0] / num_total_elements
```
?

modin/numpy/arr.py
```diff
@@ -1106,9 +1085,8 @@ def prod(
         if self._ndim == 1:
             if axis == 1:
                 raise numpy.AxisError(1, 1)
-            result = self._query_compiler
-            if isinstance(where, array):
-                result = where.where(self, 1)._query_compiler
+            target = where.where(self, 1) if isinstance(where, array) else self
+            result = target._query_compiler
             result = result.astype(
```
vnlitvinov, line 1090:
```python
result = target._query_compiler.astype(
```
?

modin/numpy/arr.py
```diff
@@ -1165,13 +1141,11 @@ def prod(
...
-                )
+                else:
+                    return array([[initial]], dtype=out_dtype)
             return result if where is not False else initial
-        result = self._query_compiler
-        if isinstance(where, array):
-            result = where.where(self, 1)._query_compiler
+        target = where.where(self, 1) if isinstance(where, array) else self
+        result = target._query_compiler
         result = result.astype(
```
vnlitvinov, line 1149:
```python
result = target._query_compiler.astype(
```
?

modin/numpy/arr.py
```diff
@@ -1416,9 +1389,8 @@ def sum(
         if self._ndim == 1:
             if axis == 1:
                 raise numpy.AxisError(1, 1)
-            result = self._query_compiler
-            if isinstance(where, array):
-                result = where.where(self, 0)._query_compiler
+            target = where.where(self, 0) if isinstance(where, array) else self
+            result = target._query_compiler
             result = result.astype(
```
vnlitvinov, line 1394:
```python
result = target._query_compiler.astype(
```
?

modin/numpy/arr.py
```diff
@@ -1465,21 +1435,19 @@ def sum(
...
-                )
+                else:
+                    return array([[initial]], dtype=out_dtype)
             return result if where is not False else initial
-        result = self._query_compiler
-        if isinstance(where, array):
-            result = where.where(self, 0)._query_compiler
+        target = where.where(self, 0) if isinstance(where, array) else self
+        result = target._query_compiler
         result = result.astype(
```
vnlitvinov, line 1451:
```python
result = target._query_compiler.astype(
```
?

vnlitvinov (approved):
Taste changes and questions left... bugs have been squashed, so I'm stamping a green mark. 
🚢!

modin/numpy/arr.py
RehanSD: Sorry - to clarify, we don't do that - that's just how NumPy handles the arguments. Say someone calls `numpy.tanh(modin.numpy.array, *args, **kwargs)` - the following happens:
1. NumPy recognizes that our `modin.numpy.array` object has an `__array_ufunc__` method.
2. NumPy translates `numpy.tanh` to `modin.numpy.array.__array_ufunc__(numpy.tanh, "__call__", *args, **kwargs)`
Our `out` argument is inside the `**kwargs`, but the packaging (it being inside a 1-tuple) is handled by NumPy, so I just unpack it if its 1-tuple.

modin/numpy/arr.py
RehanSD: I wanted to keep it a NumPy array so that we can get NumPy division rules - say we have 0 non-na elements in the list - rather than having an error (because we're dividing by 0), NumPy will handle the division correctly, and give us the appropriate result (i.e. NaN vs. INF depending on what our output dtype should be)

modin/numpy/arr.py
RehanSD: Fixed!

modin/numpy/arr.py
RehanSD: Sorry - to clarify, I meant that if we don't do this, our tests fail on the Ubuntu OS (but not on macOS or windows).

Interesting - I just checked `NumPy`, and it looks like if `NaN` is present in the array, it returns `NaN` as the mean (or a junk value if the output dtype is int). We ignore the `NaN` and compute the mean or sum as if it wasn't there - because that's how pandas does it, but NumPy propagates the NaN.

Do you think we should try and resolve this in this PR? I'm concerned its already kind of big, and if we do decide to fix this, this will also mean we need to handle our masks differently, since we currently take advantage of pandas ignoring `NaN` behavior to replace values we're making with `NaN` without affecting the sum, so we'd probably need to come up with a new method for masking. (Although we could technically convert this to a Binary algebra operator maybe? Where the second operand is the where array for masking?)

modin/numpy/arr.py
RehanSD: Sure - I can add a comment!

modin/numpy/arr.py
vnlitvinov: I would say, let's make a separate issue, as this PR is really already big enough 🙃

vnlitvinov (approved):
Awesome, LGTM!

modin/numpy/arr.py
RehanSD: Issue created: https://github.com/modin-project/modin/issues/5720

anmyachev (approved):

PR #5673, opened 2023-02-19
FIX-#5628: BUG: HDK: Unable to concatenate tables with different number of non-numeric columns

### LINKED ISSUE #5628: BUG: HDK: Unable to concatenate tables with different number of non-numeric columns
### Modin version checks

### Reproducible Example

```python
import modin.pandas as pd

df1 = pd.DataFrame({"A": [1, 2], "B": ["1", "2"]})
df2 = pd.DataFrame({"A": [1, 2]})
df = pd.concat([df1, df2])
print(df)
```

### Issue Description

Unable to concatenate tables with different number of non-numeric columns. If the columns are numeric, concatenation works, but it's ~100 times slower, than pandas (see https://github.com/intel-ai/hdk/issues/182)

### Expected Behavior

```
   A     B
0  1     1
1  2     2
0  1  None
1  2  None
```

### Error Logs

<details>

```python-traceback

modin/experimental/core/execution/native/implementations/hdk_on_native/hdk_worker.py", line 90, in executeRA
    ra = cls._calcite.process(query, db_name="hdk")
  File "_sql.pyx", line 41, in pyhdk._sql.Calcite.process
ValueError: Cannot compute compatible row type for arguments to set op: RecordType(BIGINT __index__, BIGINT A, VARCHAR(50) B), RecordType(BIGINT __index__, BIGINT A, BIGINT B)

```

</details>

### Installed Versions

<details>

</details>

AndreyPavlenko:
## What do these changes do?

- [x] Resolves #5628 
- [x] tests added and passing

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -475,6 +475,7 @@ jobs:
       - run: python -m pytest modin/test/interchange/dataframe_protocol/test_general.py
       - run: python -m pytest modin/test/interchange/dataframe_protocol/hdk
       - run: python -m pytest modin/experimental/sql/test/test_sql.py
+      - run: python -m pytest -n 2 modin/pandas/test/test_concat.py
       - run: python examples/docker/modin-hdk/census-hdk.py examples/data/census_1k.csv -no-ml
       - run: python examples/docker/modin-hdk/nyc-taxi-hdk.py examples/data/nyc-taxi_1k.csv
       - run: |
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -475,7 +475,7 @@ jobs:
       - run: python -m pytest modin/test/interchange/dataframe_protocol/test_general.py
       - run: python -m pytest modin/test/interchange/dataframe_protocol/hdk
       - run: python -m pytest modin/experimental/sql/test/test_sql.py
-      - run: python -m pytest -n 2 modin/pandas/test/test_concat.py
+      - run: python -m pytest modin/pandas/test/test_concat.py
       - run: python examples/docker/modin-hdk/census-hdk.py examples/data/census_1k.csv -no-ml
       - run: python examples/docker/modin-hdk/nyc-taxi-hdk.py examples/data/nyc-taxi_1k.csv
       - run: |
--- a/modin/experimental/core/execution/native/implementations/hdk_on_native/dataframe/dataframe.py
+++ b/modin/experimental/core/execution/native/implementations/hdk_on_native/dataframe/dataframe.py
@@ -30,6 +30,7 @@ from modin.error_message import ErrorMessage
 from modin.pandas.indexing import is_range_like
 from modin.utils import MODIN_UNNAMED_SERIES_LABEL
 import pandas as pd
+import typing
 from typing import List, Hashable, Optional, Tuple, Union
 
 from ..df_algebra import (
@@ -969,6 +970,19 @@ class HdkOnNativeDataframe(PandasDataframe):
         HdkOnNativeDataframe
             The new frame.
         """
+        # In case of different number of columns, HDK performs
+        # slowly and supports only numeric column types.
+        # See https://github.com/intel-ai/hdk/issues/182
+        # To work around this issue, perform concatenation
+        # with arrow.
+        if (
+            len(other_modin_frames) == 0
+            or len(self.columns) == 0
+            or any((len(f.columns) != len(self.columns) for f in other_modin_frames))
+            or any((set(f._dtypes) != set(self._dtypes) for f in other_modin_frames))
+        ):
+            return self._union_all_arrow(other_modin_frames, join, sort, ignore_index)
+
         # determine output columns
         new_cols_map = OrderedDict()
         for col in self.columns:
@@ -1016,8 +1030,8 @@ class HdkOnNativeDataframe(PandasDataframe):
                         exprs[col] = frame.ref(col)
                 else:
                     assert index_width == 1, "unexpected index width"
-                    aligned_index = ["__index__"]
-                    exprs["__index__"] = frame.ref("__rowid__")
+                    aligned_index = self._mangle_index_names([None])
+                    exprs[aligned_index[0]] = frame.ref("__rowid__")
                     aligned_index_dtypes = [get_dtype(int)]
                     uses_rowid = True
                 aligned_dtypes = aligned_index_dtypes + new_dtypes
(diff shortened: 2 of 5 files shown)


modin/pandas/general.py
```diff
@@ -404,26 +404,34 @@ def concat(
                 sort=sort,
             )
         )
-    if join not in ["inner", "outer"]:
+    if join == "outer":
+        # Filter out empties
```
anmyachev, line 408: Why does this filtering need to be done at the modin level of the dataframe, and not at the hdk backend level?

modin/experimental/core/execution/native/implementations/hdk_on_native/dataframe/dataframe.py
```diff
@@ -1054,6 +1067,206 @@ def _union_all(
...
+            idx_cols = None
+            idx_table = None
+            idx_ranges = None
+            idx_fields: typing.OrderedDict[str, pyarrow.Field] = OrderedDict()
+
+            if not ignore_index:
+                if all(f.index_cols is None for f in frames):
+                    # All frames have a trivial index
+                    idx_cols = mangle_index_names(self.index.names)
+                    idx_ranges = list(range(len(frames[0].table)))
```
ienkovich, line 1184: When you create `FrameData` objects, you materialize all indexes, even default ones. There is no need to create new ranges here then. Actually, it looks like you don't need any special handling of the case when all frames use default indexes. The `else` code below should handle it correctly by appending all indexes.

modin/experimental/core/execution/native/implementations/hdk_on_native/dataframe/dataframe.py
```diff
@@ -2505,7 +2718,12 @@ def from_arrow(cls, at, index_cols=None, index=None, columns=None):
 
         if columns is not None:
             new_columns = columns
-            new_index = pd.RangeIndex(at.num_rows) if index is None else index
+            if index is not None:
+                new_index = index
+            elif index_cols is not None:
+                new_index = None
+            else:
+                new_index = pd.RangeIndex(at.num_rows)
```
ienkovich, line 2726: Why not materialize the index on access as usual? This materialization might be useless if the index is not used later.

ienkovich (changes requested):
The concatenation code using `pyarrow` looks good!

My main concern here is that we kill laziness here with no good reason. Most probably, we will want to modify the current behavior someday, maybe choose more wisely between pyarrow and HDK or make some additional optimizations (e.g. why to concat columns if we later drop them?). To make such decisions we will need a bigger picture and therefore should allow building execution trees with concatenations.

I think we should make the decision about the execution engine and trigger the actual execution as usual. We already have code that allows us to execute sub-trees and run arrow-based execution. I don't see any reason to not utilize it and introduce additional execution entry points instead.

modin/pandas/general.py
AndreyPavlenko: The filtering was already implemented here - https://github.com/modin-project/modin/pull/5673/files#diff-270ead8b42be5edcb8f1979a86ef8cbc663d116fb7f0fa14e51999ded69e6c08L422 . I've changed the logic a bit to be compatible with Pandas.

modin/experimental/core/execution/native/implementations/hdk_on_native/dataframe/dataframe.py
```diff
@@ -1054,6 +1067,206 @@ def _union_all(
...
+
+        if join == "outer":
+            for frame in frames:
+                table = frame.table
+                idx_width = 0 if frame.index_cols is None else len(frame.index_cols)
+                table_cols = table.column_names[idx_width:]
+                for col_name, table_col_name in zip(frame.columns, table_cols):
+                    add_col_field(table, col_name, table_col_name)
+        else:
+            col_names = {c for c in frames[0].columns} if len(frames) > 1 else []
```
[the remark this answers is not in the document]
AndreyPavlenko: Empty frames are filtered out at modin/pandas/general.py. In case of inner join, we have to remove all columns, but keep the index.

modin/experimental/core/execution/native/implementations/hdk_on_native/dataframe/dataframe.py
AndreyPavlenko: I added the ranges in my initial implementation for performance optimization and it worked much faster. Now I see, it does not make any sense. Removed this code. Thanks.

modin/pandas/general.py
anmyachev: It is at least strange that filtering is now only for `outer` type of join. What about `inner` join?

modin/pandas/general.py
```diff
@@ -454,6 +462,18 @@ def concat(
...
+    if len(list_of_objs) == 0:
+        idx = input_list_of_objs[0].index.append(
+            [f.index for f in input_list_of_objs[1:]]
+        )
+        if axis == 1 or any(
+            isinstance(obj, (DataFrame, pandas.DataFrame)) for obj in input_list_of_objs
+        ):
+            return DataFrame(index=idx)
+        else:
+            return Series(index=idx)
```
anmyachev, line 475: It looks like a correction of the consequences of the fact that those objects were filtered out, information about which we need in the future. Let's try to avoid it.

modin/pandas/general.py
```diff
@@ -404,26 +404,34 @@ def concat(
...
             )
         )
-    if join not in ["inner", "outer"]:
+    if join == "outer":
+        # Filter out empties
+        list_of_objs = [
+            obj
+            for obj in list_of_objs
+            if (
+                isinstance(obj, (Series, pandas.Series))
```
anmyachev, line 413: Why is this part of the condition not checked for `lazy_execution` for Modin Series?

AndreyPavlenko:
> e.g. why to concat columns if we later drop them?

In fact, we don not concatenate columns. In case of an empty frame in the list, a new table is created, that contains only the index columns - https://github.com/modin-project/modin/pull/5673/files#diff-d1a0db76b871ca475d604ff109cc3ebf13d7ead8ce07cca7caa54ace5f736cc0R1160

modin/pandas/general.py
AndreyPavlenko: > It is at least strange that filtering is now only for `outer` type of join. What about `inner` join?

It's incorrect to filter out empty frames in case of inner join. In this case, we have to return an empty frame, but preserve the index.

modin/pandas/general.py
AndreyPavlenko: If all objects are filtered out, we need concatenate indexes and return either an empty DataFrame or Series. According to the documentation:


When concatenating all Series along the index (axis=0), a Series is returned. When objs contains at least one DataFrame, a DataFrame is returned. When concatenating along the columns (axis=1), a DataFrame is returned.

modin/pandas/general.py
AndreyPavlenko: We should not filter out Series - https://github.com/pandas-dev/pandas/blob/main/pandas/core/reshape/concat.py#L476

modin/pandas/general.py
anmyachev: So you have found another bug that is present for all backends. We need to at least create a separate issue and add tests not only for hdk.

AndreyPavlenko: > So you have found another bug that is present for all backends. We need to at least create a separate issue and add tests not only for hdk.

What about moving the test here https://github.com/modin-project/modin/blob/master/modin/pandas/test/test_concat.py ?

anmyachev: If these tests reproduce the problem with `inner` concat (for example, for Ray engine and pandas backend) and it is possible to reduce the amount of hdk-specific code, then why not?

modin/pandas/general.py
anmyachev: For Series we have a different code branch https://github.com/modin-project/modin/blob/b113f47a1787a68b3846d861dd2526e38b1bdc72/modin/pandas/general.py#L391 where we don't filter them.

AndreyPavlenko: This branch checks if all objs are modin Series, but here we filter out empty DataFrames, while preserving empty Series.

modin/experimental/core/execution/native/implementations/hdk_on_native/dataframe/dataframe.py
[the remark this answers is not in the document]
ienkovich: I see, thanks

modin/pandas/general.py
anmyachev: > We should not filter out Series - https://github.com/pandas-dev/pandas/blob/main/pandas/core/reshape/concat.py#L476

Judging by your link, empty Series are not filtered only when all objects are Series types. Isn't it?

modin/pandas/test/test_concat.py
```diff
@@ -236,3 +242,58 @@ def test_sort_order(sort, join, axis):
         modin_concat,
     )
     assert list(pandas_concat.columns) == list(modin_concat.columns)
+
+
+@pytest.mark.skipif(
+    StorageFormat.get() not in ("Hdk", "Base"),
```
anmyachev, line 248: Looks like, your changes in `general.py` do not fix the problem on the pandas backend?

modin/pandas/general.py
anmyachev, line 475: This case should not be here. Do you have a specific example when we need to explicitly create a Series here?

modin/pandas/general.py
AndreyPavlenko: According to the following condition, any Series are not filtered out:

```python
            # filter out the empties if we have not multi-index possibilities
            # note to keep empty Series as it affect to result columns / name
            non_empties = [
                obj for obj in objs if sum(obj.shape) > 0 or isinstance(obj, ABCSeries)
            ]
```

modin/pandas/test/test_concat.py
AndreyPavlenko: Right. On the master branch, the test fails on all backends. On this branch, the test passes on HDK and Base.

modin/pandas/general.py
AndreyPavlenko: As I mentioned here https://github.com/modin-project/modin/pull/5673#discussion_r1115074467, this case is described in the documentation.

modin/pandas/general.py
anmyachev:
```python
        # figure out what our result ndim is going to be
        ndims = set()
        for obj in objs:
            if not isinstance(obj, (ABCSeries, ABCDataFrame)):
                msg = (
                    f"cannot concatenate object of type '{type(obj)}'; "
                    "only Series and DataFrame objs are valid"
                )
                raise TypeError(msg)

            ndims.add(obj.ndim)

        # get the sample
        # want the highest ndim that we have, and must be non-empty
        # unless all objs are empty
        sample: NDFrame | None = None
        if len(ndims) > 1:
            max_ndim = max(ndims)
            for obj in objs:
                if obj.ndim == max_ndim and np.sum(obj.shape):
                    sample = obj
                    break

        else:
            # filter out the empties if we have not multi-index possibilities
            # note to keep empty Series as it affect to result columns / name
            non_empties = [
```
We get into `else` branch only if the input contains concatenation objects of the same types. These are either all dataframes, but this has nothing to do with series, or all series and this is exactly the case you are talking about. We also have a separate branch for series, in which objects are not filtered in the same way as in pandas. I want to say that from this point of view, the already existing code should correspond to it. Am I missed something?

modin/pandas/general.py
anmyachev: in the only case when we need to return a series, we have a separate branch with return statement. See https://github.com/modin-project/modin/pull/5673#discussion_r1115126900 and https://github.com/modin-project/modin/pull/5673#discussion_r1116179037

AndreyPavlenko: You are right. This branch is dead. I've removed it. Thanks.

modin/pandas/general.py
```diff
@@ -404,27 +404,35 @@ def concat(
...
         if isinstance(obj, (pandas.Series, Series)) and axis == 0
-        else DataFrame(obj)
+        else DataFrame(obj)._query_compiler
         for obj in list_of_objs
     ]
-    list_of_objs = [
-        obj._query_compiler
-        for obj in list_of_objs
-        if (not obj._query_compiler.lazy_execution and len(obj.index))
-        or len(obj.columns)
```
anmyachev, line 426: Wouldn't it be enough to change this condition?
```python
or len(obj.columns) if join == "outer" else True
```
In order not to filter dataframes for the inner join in which there is not an empty index, but empty columns?

modin/pandas/test/test_concat.py
```diff
@@ -23,11 +23,17 @@
...
 )
-from modin.config import NPartitions
+from modin.config import NPartitions, StorageFormat
 
 NPartitions.put(4)
 
+pytestmark = pytest.mark.filterwarnings(default_to_pandas_ignore_string)
+
+# Initialize env for storage format detection in @pytest.mark.*
+pd.DataFrame()
```
anmyachev, line 35: Are you sure this is necessary? `StorageFormat` should get the right value for `@pytest.mark.skipif(` without explicit initialization.

AndreyPavlenko: If I comment out this line, I'm getting 1152 tests skipped in the HDK environment. StorageFormat.get() returns "Pandas".

modin/pandas/general.py
AndreyPavlenko: Probably, it's enough, but, it this case, we just create a copy of the list. I think, it's better to avoid filtering and creating a new list in case of the inner join.

modin/pandas/test/test_concat.py
anmyachev: I can't reproduce this problem. Moreover, we already have a place in the tests that checks the backend, but at the same time, the dataframe is not explicitly initialized in that file: https://github.com/modin-project/modin/blob/8df0030e1c1cdde73591fefe1d2fd7a164312b3a/modin/pandas/test/test_io.py#L70

Let's delete it and see how the CI goes.

anmyachev, line 35:
```suggestion
```

modin/pandas/general.py
anmyachev: Isn't it safe to filter out objects that have an empty index **and** empty columns for an inner join?

modin/pandas/test/test_concat.py
AndreyPavlenko: > I can't reproduce this problem.

Have you checked this env - requirements/env_hdk.yml ?

modin/pandas/test/test_concat.py
anmyachev: Do you have a separate reproducer for the pandas backend where the previous filtering implementation was not working properly? It is highly desirable to add a test for this in order to have protection in the future.

AndreyPavlenko: I've filed a bug with a reproducer - https://github.com/modin-project/modin/issues/5696

modin/pandas/test/test_concat.py
anmyachev: No, I use the following code:
```python
import os
os.environ["MODIN_EXPERIMENTAL"] = "True"
os.environ["MODIN_ENGINE"] = "native"
os.environ["MODIN_STORAGE_FORMAT"] = "hdk"

from modin.config import StorageFormat
print(StorageFormat.get())
```

modin/pandas/test/test_concat.py
anmyachev: You describe another problem in the title. Isn't it?

modin/pandas/test/test_concat.py
AndreyPavlenko: It works if you initialise the environment explicitly.
Try to remove this line - pd.DataFrame() and run the test in this - requirements/env_hdk.yml environment, without setting the environment variables:

```
mamba create -f requirements/env_hdk.yml
mamba activate modin_on_hdk
python -m pytest -n 2 modin/pandas/test/test_concat.py
```

anmyachev: Now it is clear. However, although Modin can determine the engine and backend based on installed dependencies, the design of the tests does not imply that the environment is not configured.

modin/pandas/general.py
AndreyPavlenko: It's not. For example, we filter out an empty frame and pass down, to the backend, non-empty frames. The backend concatenates the non-empty frames, but, in case of the inner join, the result should be an empty frame with merged index.

modin/pandas/test/test_concat.py
anmyachev: @AndreyPavlenko for this behavior, we need a centralized solution (for all files with tests), I will be glad if you create an issue for this. Good suggestion, thanks!

AndreyPavlenko: I absolutely agree we need a solution. The solution, implemented in this fix, is just an ugly workaround.

modin/pandas/general.py
anmyachev: Got it, thanks for clarification!

anmyachev: Pretty simple reproducer:
```python
import modin.pandas as pd
df1 = pd.DataFrame(index=[1,2,3])
df2 = pd.DataFrame(index=[])
pd.concat((df1, df2), axis=1, join="inner")
```

modin/pandas/test/test_concat.py
AndreyPavlenko: The test, in the issues description, reproduces multiple problems and this is just one of them.
Changed the issue title.

anmyachev (changes requested):
@AndreyPavlenko this pull request solves at least 2 problems, and they are related to different backends.

Inner join bug for pandas backend should be fixed in a separate pull request. Reproducer can be found here: https://github.com/modin-project/modin/pull/5673#discussion_r1118173416. You can add it as a new test.

Let me know if you need help with this.

AndreyPavlenko:
> @AndreyPavlenko this pull request solves at least 2 problems, and they are related to different backends.
> 
> Inner join bug for pandas backend should be fixed in a separate pull request. Reproducer can be found here: [#5673 (comment)](https://github.com/modin-project/modin/pull/5673#discussion_r1118173416). You can add it as a new test.
> 
> Let me know if you need help with this.

What about creating a separate PR, that fixes https://github.com/modin-project/modin/issues/5696 issue for the Pandas backend only? I'll move the test from this PR to the new one. This test (pytest modin/pandas/test/test_concat.py --execution=BaseOnPython) fails with 280 failures, if it's running on the master branch.

modin/pandas/test/test_concat.py
AndreyPavlenko: > I will be glad if you create an issue for this.

I've created the issue - https://github.com/modin-project/modin/issues/5712

anmyachev:
> > @AndreyPavlenko this pull request solves at least 2 problems, and they are related to different backends.
> > Inner join bug for pandas backend should be fixed in a separate pull request. Reproducer can be found here: [#5673 (comment)](https://github.com/modin-project/modin/pull/5673#discussion_r1118173416). You can add it as a new test.
> > Let me know if you need help with this.
> 
> What about creating a separate PR, that fixes #5696 issue for the Pandas backend only? I'll move the test from this PR to the new one. This test (pytest modin/pandas/test/test_concat.py --execution=BaseOnPython) fails with 280 failures, if it's running on the master branch.

This looks like a third separate problem. I'm more concerned (for now) with the changes in `general.py` file, which should be done separately and with the addition of a test (not your complex test, but a simple one that checks for one specific problem).

.github/workflows/ci.yml
```diff
@@ -475,6 +475,7 @@ jobs:
       - run: python -m pytest modin/test/interchange/dataframe_protocol/test_general.py
       - run: python -m pytest modin/test/interchange/dataframe_protocol/hdk
       - run: python -m pytest modin/experimental/sql/test/test_sql.py
+      - run: python -m pytest -n 2 modin/pandas/test/test_concat.py
```
anmyachev, line 478: Not sure if this is needed for hdk
```suggestion
      - run: python -m pytest modin/pandas/test/test_concat.py
```

anmyachev (approved):
LGTM!

Garra1980:
can we merge this already?

ienkovich:
My concern about non-lazy concat execution was not addressed. We should either fix it or agree on a technical dept this PR introduces.

AndreyPavlenko:
@ienkovich What about implementing laziness in a separate PR? I think, we need a general approach to the arrow-based execution, that covers not only this case.

ienkovich:
> @ienkovich What about implementing laziness in a separate PR? I think, we need a general approach to the arrow-based execution, that covers not only this case.

What's wrong with the current general approach to arrow-based execution?

ienkovich (approved):
We've discussed the lazy execution issue and decided to make further improvements in a follow-up patch. The solution will be to generate proper TransformNodes to prepare frames for concat and support arrow execution for those nodes.

PR #5678, opened 2023-02-20
PERF-#5674: reduce data transferring in `_launch_tasks` function

### LINKED ISSUE #5674: PERF: internal `_launch_tasks` function takes too long while `read_csv` is executing
We can reduce the amount of data transferred to workers.

Profile (on Ray engine):
```bash
   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
      400    0.050    0.000   18.654    0.047 .../modin/modin/core/io/text/text_file_dispatcher.py:599(_launch_tasks)
```

anmyachev:
## What do these changes do?

The main idea behind this pull request is to reduce the amount of data that needs to be serialized. For this, the following is done:
* `cls.parse` function is cached (only serialized on first use).
* Most of the dictionary data that we pass to remote functions (for `read_csv` function) is the same across partitions, allowing us to serialize it once.
* There is a problem in the serialization of dictionaries by Dask, whose behavior is different from other engines. If a dictionary is passed, then the dask serializes only the values, and the keys are used for service purposes. As far as I understood, there is a problem very similar to a key collision: https://github.com/dask/distributed/issues/3965#issuecomment-659422586. However, a simple workaround - wrapping the dictionary in a new type (UserDict) - circumvents this problem, and, as shown [here](https://github.com/modin-project/modin/pull/5678/files#r1126411771), does not slow down the execution.

Some thoughts:
* When passing references to remote functions, we need to pass them as **positional** parameters so that the engines can automatically deserialize them (at least this is true for Ray, but for unification, I want to do the same for all engines). This requires a change in the signatures of methods `_launch_tasks`, `parse` and `single_worker_read`.
* I guess one of the debatable parts here is that not all `parse` functions have the same signature (for other read functions). If these changes are made now, they will be unused, which I do not really like.

Perf results:
* On one of the workloads, where there is an intensive reading of many files, quite small ones. I'm seeing a speedup of `_launch_tasks` function by about 15 seconds (from ~18 sec down to ~3 sec).

- [x] Resolves #5674 
- [x] tests added and passing


anmyachev:
@YarShev please take a look at unidist error:
```bash
data_id = DataID(rank_0_id_201891)

    def get_impl(data_id):
        if object_store.contains(data_id):
            value = object_store.get(data_id)
        else:
            value = request_worker_data(data_id)
    
        if isinstance(value, Exception):
>           raise value
E           TypeError: unsupported operand type(s) for //: 'int' and 'NoneType'

/usr/share/miniconda3/envs/modin_on_unidist/lib/python3.8/site-packages/unidist/core/backends/mpi/core/controller/api.py:252: TypeError

---------- coverage: platform linux, python 3.8.16-final-0 -----------
Coverage XML written to file coverage.xml

=========================== short test summary info ============================
FAILED modin/pandas/test/dataframe/test_indexing.py::test___setitem__with_mismatched_partitions - TypeError: unsupported operand type(s) for //: 'int' and 'NoneType'
```
Link: https://github.com/modin-project/modin/actions/runs/4262914506/jobs/7419069177

modin/core/execution/dask/common/engine_wrapper.py
```diff
@@ -121,5 +123,13 @@ def put(cls, data, **kwargs):
...
         List, dict, iterator, or queue of futures matching the type of input.
         """
+        if isinstance(data, dict):
+            # there is a bug that looks similar to https://github.com/dask/distributed/issues/3965;
+            # to avoid this we could change behaviour for serialization:
+            # <Future: finished, type: collections.UserDict, key: UserDict-b8a15c164319c1d32fd28481125de455>
+            # vs
+            # {'sep': <Future: finished, type: pandas._libs.lib._NoDefault, key: sep>, \
+            #  'delimiter': <Future: finished, type: NoneType, key: delimiter> ...
+            data = UserDict(data)
```
anmyachev, line 133: Fairly fast operation, total conversion time: `~0.03 sec` for the following command:
`pytest modin\pandas\test\test_io.py::TestCsv -s`

anmyachev: Total time for scatter(dict) op: `35.20 sec` vs  total time for scatter(user_dict) op: `35.21 sec`.

anmyachev: How was it measured? (changes for `DaskWrapper.put`)
```python
            start = time()
            _ = client.scatter(data, **kwargs)
            cls._dict_scatter += time()-start
            print(f"DICT SCATTER: {cls._dict_scatter}")
            start = time()
            data = UserDict(data)
            cls._user_dict_conversion += time()-start
            print(f"USER DICT CONVERSION: {cls._user_dict_conversion}")
            start = time()
            _ = client.scatter(data, **kwargs)
            cls._use_dict_scatter += time()-start
            print(f"USER DICT SCATTER: {cls._use_dict_scatter}\n")
```

YarShev: What happens if we don't wrap data into UserDict?

modin/core/io/text/text_file_dispatcher.py
```diff
@@ -591,13 +591,26 @@ def _define_metadata(
 
         return column_widths, num_splits
 
+    _parse_func = None
+
+    @classmethod
+    def get_parse_func(cls):  # noqa: RT01
```
YarShev, line 597: I would suggest to rename this to `preprocess_func` like we have in partitions.

modin/core/io/text/text_file_dispatcher.py
```diff
@@ -591,13 +591,26 @@ def _define_metadata(
...
+        """Prepare a function for transmission to remote workers."""
+        if cls._parse_func is None:
+            cls._parse_func = cls.put(cls.parse)
+        return cls._parse_func
+
     @classmethod
-    def _launch_tasks(cls, splits: list, **partition_kwargs) -> Tuple[list, list, list]:
+    def _launch_tasks(
+        cls, *args, splits: list, **partition_kwargs
```
YarShev, line 605:
```suggestion
        cls, splits: list, *partition_args, **partition_kwargs
```
I don't really like this order. Can we use as suggested?

modin/core/io/text/text_file_dispatcher.py
```diff
@@ -619,7 +632,8 @@ def _launch_tasks(cls, splits: list, **partition_kwargs) -> Tuple[list, list, li
         for idx, (start, end) in enumerate(splits):
             partition_kwargs.update({"start": start, "end": end})
             *partition_ids[idx], index_ids[idx], dtypes_ids[idx] = cls.deploy(
-                func=cls.parse,
+                func=cls.get_parse_func(),
```
YarShev, line 635: We could preprocess the func outside the loop like we do in partitions.

modin/core/io/text/text_file_dispatcher.py
```diff
@@ -1101,7 +1115,6 @@ def _read(cls, filepath_or_buffer, **kwargs):
         partition_kwargs = dict(
             kwargs,
             fname=filepath_or_buffer_md,
-            num_splits=num_splits,
```
YarShev, line 1104: Why removed?

modin/core/io/text/text_file_dispatcher.py
```diff
@@ -1110,8 +1123,13 @@ def _read(cls, filepath_or_buffer, **kwargs):
             nrows=None,
             compression=compression_infered,
         )
+        fname = partition_kwargs.pop("fname")
```
YarShev, line 1126: Why do we not want to serialize the func along with partition_kwargs?

modin/core/execution/dask/common/engine_wrapper.py
anmyachev: Something like that:
```python
                            st = self.futures[key]
                            exception = st.exception
                            traceback = st.traceback
                        except (KeyError, AttributeError):
                            exc = CancelledError(key)
                        else:
                            raise exception.with_traceback(traceback)
>                       raise exc
E                       concurrent.futures._base.CancelledError: lambda-07539c5507424ccd64320fffc3e31fdb

/usr/share/miniconda3/envs/modin/lib/python3.8/site-packages/distributed/client.py:2170: CancelledError
------------------------------ Captured log call -------------------------------
INFO     distributed.scheduler:scheduler.py:4355 User asked for computation on lost data, _deploy_dask_func-f7749d8a41c1b73530edaaf9ba6bff81
INFO     distributed.scheduler:scheduler.py:4355 User asked for computation on lost data, _deploy_dask_func-0e25adbc074f4d4de7bb67abc25840e8
INFO     distributed.scheduler:scheduler.py:4355 User asked for computation on lost data, _deploy_dask_func-c21c49c627cbfb1d2fa268b0eae99420
```
https://github.com/modin-project/modin/actions/runs/4266413378/jobs/7426978720

modin/core/io/text/text_file_dispatcher.py
anmyachev: By doing this separately, we are serializing the function once for the entire workflow, and not every time `read_csv` is called.

modin/core/io/text/text_file_dispatcher.py
anmyachev: It doesn't need to be done since we are serializing and caching the result. Caching for the [entire ](https://github.com/modin-project/modin/pull/5678#discussion_r1131145293) workflow.

modin/core/io/text/text_file_dispatcher.py
anmyachev: This parameter is now explicitly passed to `_launch_tasks` function. This had to be done because `partition_kwargs` is now passed to `_launch_tasks` function as a reference/future.

modin/core/io/text/text_file_dispatcher.py
anmyachev: The current solution implies that this parameter must always be passed by name. For me, this is more explicit behavior. Why don't you like it?

modin/core/io/text/text_file_dispatcher.py
YarShev: We could call cls.put(fname) right here rather than calling preprocess_func multiple times in the loop.

YarShev: I just don't like that we call preprocess_func multiple times in the loop.

modin/core/io/text/text_file_dispatcher.py
```diff
@@ -1110,8 +1123,13 @@ def _read(cls, filepath_or_buffer, **kwargs):
...
+        kwargs_ref = cls.put(partition_kwargs)
         partition_ids, index_ids, dtypes_ids = cls._launch_tasks(
-            splits, **partition_kwargs
+            fname,
+            kwargs_ref,
+            splits=splits,
+            num_splits=num_splits,
```
YarShev, line 1132: splits and num_splits look confusing. we could probably serialize kwargs inside of launch_tasks.

modin/core/io/text/text_file_dispatcher.py
anmyachev: > in the loop.

moved out of the loop

modin/core/io/text/text_file_dispatcher.py
anmyachev: `launch_tasks` is a common function for several dispatchers, so in order not to make changes to, for example, `read_custom_text` function, I do not serialize inside it.

anmyachev:
@YarShev ready for review

modin/core/io/text/text_file_dispatcher.py
```diff
@@ -1110,8 +1124,13 @@ def _read(cls, filepath_or_buffer, **kwargs):
             nrows=None,
             compression=compression_infered,
         )
+        fname = partition_kwargs.pop("fname")
+        kwargs_ref = cls.put(partition_kwargs)
```
YarShev, line 1128: maybe it is worth adding comments regarding why we preprocess `partition_kwargs` here and `func` inside of `_launch_tasks`?

modin/core/storage_formats/pandas/parsers.py
```diff
@@ -311,10 +320,14 @@ def single_worker_read(cls, fname, *, reason: str, **kwargs):
 @doc(_doc_pandas_parser_class, data_type="CSV files")
 class PandasCSVParser(PandasParser):
     @staticmethod
-    @doc(_doc_parse_func, parameters=_doc_parse_parameters_common)
-    def parse(fname, **kwargs):
-        kwargs["callback"] = PandasCSVParser.read_callback
-        return PandasParser.generic_parse(fname, **kwargs)
+    @doc(_doc_parse_func, parameters=_doc_parse_parameters_common2)
+    def parse(fname, common_read_kwargs, **kwargs):
```
YarShev, line 324: is `common_read_kwargs` a dict here? Before adding this parameter we used `**kwargs`, which can be smth like `d = {"A": object_ref, "B": 1}`. Ray itself would deserialize `object_ref` in the remote task. If `common_read_kwargs` is a dict, Ray won't be able to deserialze `object_ref` containing in `common_read_kwargs`, which may slow down the flow.

modin/core/io/text/text_file_dispatcher.py
anmyachev: added

YarShev (approved):

YarShev (approved):

PR #5731, opened 2023-03-02
FIX-#5730: Add __repr__, __len__, size, and make dtype changing lazy.

### LINKED ISSUE #5730: BUG: Modin.NumPy __repr__ materializes full data onto head node
### Modin version checks

### Reproducible Example

```python
import modin.numpy as np
arr = np.array([[1, 2], [3, 4], [5, 6]])
arr
```

### Issue Description

__repr__ materializes all of the data onto the head node and relies on NumPy's repr, when we shouldn't be doing that for big arrays. Additionally, we reset index quite frequently (which isn't necessary for non-Pandas storage formats), and try to change types very frequently.

### Expected Behavior

Only materialize necessary data to head node.

### Error Logs

<details>

```python-traceback

```

</details>

### Installed Versions

<details>

</details>

RehanSD:
## What do these changes do?

- [x] Resolves #5730 ? 
- [x] tests added and passing

### DIFF
--- a/modin/numpy/arr.py
+++ b/modin/numpy/arr.py
@@ -15,6 +15,7 @@
 
 from math import prod
 import numpy
+import pandas
 from pandas.core.dtypes.common import is_list_like, is_numeric_dtype, is_bool_dtype
 from pandas.api.types import is_scalar
 from inspect import signature
@@ -101,17 +102,6 @@ def fix_dtypes_and_determine_return(
     return result
 
 
-def find_common_dtype(dtypes):
-    if len(dtypes) == 1:
-        return dtypes[0]
-    elif len(dtypes) == 2:
-        return numpy.promote_types(*dtypes)
-    midpoint = len(dtypes) // 2
-    return numpy.promote_types(
-        find_common_dtype(dtypes[:midpoint]), find_common_dtype(dtypes[midpoint:])
-    )
-
-
 class array(object):
     """
     Modin distributed representation of ``numpy.array``.
@@ -144,7 +134,9 @@ class array(object):
         if _query_compiler is not None:
             self._query_compiler = _query_compiler
             self._ndim = _ndim
-            new_dtype = find_common_dtype(self._query_compiler.dtypes.values)
+            new_dtype = pandas.core.dtypes.cast.find_common_type(
+                self._query_compiler.dtypes.values
+            )
         elif is_list_like(object) and not is_list_like(object[0]):
             series = pd.Series(object)
             self._query_compiler = series._query_compiler
@@ -834,7 +826,7 @@ class array(object):
         operand_dtype = (
             self.dtype
             if not isinstance(x2, array)
-            else find_common_dtype([self.dtype, x2.dtype])
+            else pandas.core.dtypes.cast.find_common_type([self.dtype, x2.dtype])
         )
         out_dtype = (
             dtype
@@ -884,7 +876,7 @@ class array(object):
         operand_dtype = (
             self.dtype
             if not isinstance(x2, array)
-            else find_common_dtype([self.dtype, x2.dtype])
+            else pandas.core.dtypes.cast.find_common_type([self.dtype, x2.dtype])
         )
         out_dtype = (
             dtype
@@ -933,7 +925,7 @@ class array(object):
         operand_dtype = (
             self.dtype
             if not isinstance(x2, array)
-            else find_common_dtype([self.dtype, x2.dtype])
+            else pandas.core.dtypes.cast.find_common_type([self.dtype, x2.dtype])
         )
         out_dtype = (
             dtype
@@ -977,7 +969,7 @@ class array(object):
         operand_dtype = (
             self.dtype
             if not isinstance(x2, array)
-            else find_common_dtype([self.dtype, x2.dtype])
+            else pandas.core.dtypes.cast.find_common_type([self.dtype, x2.dtype])
         )
         out_dtype = (
             dtype
@@ -1042,7 +1034,7 @@ class array(object):
         operand_dtype = (
             self.dtype
             if not isinstance(x2, array)
-            else find_common_dtype([self.dtype, x2.dtype])
+            else pandas.core.dtypes.cast.find_common_type([self.dtype, x2.dtype])
         )
         out_dtype = (
             dtype
@@ -1186,7 +1178,7 @@ class array(object):
         operand_dtype = (
             self.dtype
             if not isinstance(x2, array)
-            else find_common_dtype([self.dtype, x2.dtype])
+            else pandas.core.dtypes.cast.find_common_type([self.dtype, x2.dtype])
         )
         out_dtype = (
             dtype
@@ -1241,7 +1233,7 @@ class array(object):
         operand_dtype = (
             self.dtype
             if not isinstance(x2, array)
-            else find_common_dtype([self.dtype, x2.dtype])
+            else pandas.core.dtypes.cast.find_common_type([self.dtype, x2.dtype])
         )
         out_dtype = (
             dtype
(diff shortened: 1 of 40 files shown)


modin/numpy/arr.py
```diff
@@ -1546,7 +1565,140 @@ def astype(self, dtype, order="K", casting="unsafe", subok=True, copy=True):
...
+                        ).to_numpy()
+                    ),
+                ),
+            ).rstrip(")]")
+            repr_str += ", ...,"
+            right_str = re.sub(
+                ", dtype=.*",
+                "",
+                re.sub(
+                    "\[",  # noqa: W605
```
vnlitvinov, line 1647:
```suggestion
                    r"\[",
```

modin/numpy/arr.py
```diff
@@ -1546,7 +1565,140 @@ def astype(self, dtype, order="K", casting="unsafe", subok=True, copy=True):
         return array(_query_compiler=result, _ndim=self._ndim)
 
     def __repr__(self):
```
vnlitvinov, line 1567: this method's body is super-lengthy, can we split it into logical pieces, like `__repr_1dim()`, `__repr_1xN()`, `__repr_Nx1()`, etc., so that all this `if shape == foo` is not getting in the way of understanding what's happening?

modin/numpy/arr.py
```diff
@@ -140,12 +142,16 @@ def __init__(
         ErrorMessage.single_warning(
             "Using Modin's new NumPy API. To convert from a Modin object to a NumPy array, either turn off the ExperimentalNumPyAPI flag, or use `modin.utils.to_numpy`."
         )
+        if isinstance(object, pd.DataFrame):
+            _query_compiler = object._query_compiler
+            _ndim = 2
+        elif isinstance(object, pd.Series):
+            _query_compiler = object._query_compiler
+            _ndim = 1
```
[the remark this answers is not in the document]
RehanSD: This is actually not the best way to handle these objects in the constructor - I have another PR #5736 that handles this, so I'll just remove this for now!

modin/numpy/arr.py
RehanSD: Sure!

modin/numpy/arr.py
```diff
@@ -1545,8 +1558,155 @@ def astype(self, dtype, order="K", casting="unsafe", subok=True, copy=True):
...
+                    ).to_numpy()
+                )
+                .replace("array([[", "")
+                .rstrip("])"),
+            ),
+        )
+        right_str = right_str.rstrip("]")
+        top_str = []
+        for l_str, r_str in zip(repr_str.split("\n"), right_str.split("\n")):
+            top_str.append(l_str + " " + r_str.lstrip())
```
vnlitvinov, line 1656:
```suggestion
            top_str.append(f"{l_str} {r_str.lstrip()}")
```
?

RehanSD:
@vnlitvinov I ended up rewriting the code so its much shorter and cleaner (we go with an approach similar to the modin DF repr, where we gather just enough data to the head node to build the representative array)!

modin/numpy/arr.py
```diff
@@ -1545,8 +1549,68 @@ def astype(self, dtype, order="K", casting="unsafe", subok=True, copy=True):
...
 
+    def _build_repr_array(self):
+        # We want to rely on NumPy for creating a string representation of this array; however
+        # we also don't want to materialize all of the data to the head node. Instead, we will
+        # materialize enough data that NumPy can build the summarized representation of the array
+        # (while changing with the NumPy print options so it will format this smaller array as
+        # abridged) and return this smaller array. In the worst case, this array will have
+        # (2*numpy.get_printoptions()["edgeitems"] + 1)^2 items, so 49 items max for the default
+        # value of 3.
+        num_elements = numpy.get_printoptions()["edgeitems"]
```
vnlitvinov, line 1560: I wonder if we should do `+1` here to account for middle element being removed instead of reducing the threshold down below, so that _shown_ elements' count is correct.
As right now, if I understand correctly, we'd gather N*2 elements (where N is the threshold), and then show N*2-1 elements, while numpy would show N*2?..

modin/numpy/test/test_array.py
```diff
@@ -18,11 +18,28 @@
...
+        (100, 100),
+        (6, 100),
+        (100, 6),
+        (100, 7),
+        (7, 100),
+    ],
+)
 def test_repr(size):
+    prev_threshold = numpy.get_printoptions()["threshold"]
+    numpy.set_printoptions(threshold=50)
```
vnlitvinov, line 38: this should actually be configured within pytest fixture... or at least wrap it in `try..finally`, as otherwise when e.g. `assert` would fire the threshold would not be restored back

modin/numpy/arr.py
```diff
@@ -1545,8 +1549,68 @@ def astype(self, dtype, order="K", casting="unsafe", subok=True, copy=True):
...
+        if self._ndim == 1 or self.shape[1] == 0:
+            if len(self) > num_elements * 2:
+                idxs = list(range(num_elements + 1)) + list(
+                    range(len(self) - num_elements, len(self))
+                )
+            elif len(self) == num_elements * 2:
+                idxs = list(range(num_elements * 2))
```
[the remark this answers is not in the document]
RehanSD: Makes sense, thank you!

modin/numpy/arr.py
```diff
@@ -1545,8 +1549,68 @@ def astype(self, dtype, order="K", casting="unsafe", subok=True, copy=True):
...
+        # If we are dealing with a small array, we can just collate all the data on the
+        # head node and let numpy handle the logic to get a string representation.
+        if self.size <= numpy.get_printoptions()["threshold"]:
+            return repr(self._to_numpy())
+        arr = self._build_repr_array()
+        prev_threshold = numpy.get_printoptions()["threshold"]
+        numpy.set_printoptions(threshold=arr.size - 1)
+        repr_str = repr(arr)
+        numpy.set_printoptions(threshold=prev_threshold)
+        return repr_str
```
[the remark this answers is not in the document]
RehanSD: Makes sense, thank you!

modin/numpy/arr.py
```diff
@@ -1545,8 +1549,68 @@ def astype(self, dtype, order="K", casting="unsafe", subok=True, copy=True):
...
+        if self._ndim == 1 or self.shape[1] == 0:
+            if len(self) > num_elements * 2:
+                idxs = list(range(num_elements + 1)) + list(
+                    range(len(self) - num_elements, len(self))
+                )
+            elif len(self) == num_elements * 2:
+                idxs = list(range(num_elements * 2))
+            else:
+                idxs = list(range(len(self)))
+            arr = self._query_compiler.getitem_row_array(idxs).to_numpy()
```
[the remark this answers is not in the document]
RehanSD: Makes sense!

modin/numpy/arr.py
RehanSD: So we do add +1. The reason we have to modify threshold is because if the size (so total number of elements in the array) of the array is less than the threshold, it will just print the entire array. We need to reduce the threshold to get the `...` numbers if the array is bigger than the threshold.

E.g. the default threshold is 1000. If we have a 2000 long array, and want 3 edge items, we'd get 49 items, and try to repr that. But when we try and get the repr of the 49 items, NumPy thinks that we're trying to repr an array of size 49, so it prints the whole thing out. We need to modify the repr so NumPy still prints the `...` etc. even though it thinks its actually dealing with a relatively smaller array.

To get the `...`, we do the logic above, where if the axis is long enough, we get 0...num_elements and len(self)-num_elements...num_elements elements.

modin/numpy/test/test_array.py
RehanSD: Sounds good to me!

vnlitvinov (approved):
LGTM!

devin-petersohn (approved):
LGTM, thanks @RehanSD !

PR #5735, opened 2023-03-02
FIX-#5720: Ensure that modin.numpy.array's propagate NaN values when computing mean

### LINKED ISSUE #5720: BUG: modin.numpy.mean does not propagate NaN values correctly
### Modin version checks

### Reproducible Example

```python
In [1]: import modin.numpy as np

In [2]: arr = np.array([[1, 2], [3, 4], [5, np.nan]])

In [3]: arr.mean()
Out[3]: 3.0

In [4]: arr._to_numpy().mean()
Out[4]: nan
```

### Issue Description

modin.numpy.mean does not propagate NA values present in the input. We could do so by using `skipna=False` in the `mean` argument, but we use the `skipna=True` behavior to support the `where` argument, by replacing masked values with `nan`. We'd need to replace this behavior and swap `skipna` to `False`.

### Expected Behavior

propagate nans

### Error Logs

<details>

```python-traceback

```

</details>

### Installed Versions

<details>

</details>

RehanSD:
## What do these changes do?

- [x] Resolves #5720 ? 
- [x] tests added and passing

### DIFF
--- a/modin/numpy/arr.py
+++ b/modin/numpy/arr.py
@@ -717,6 +717,24 @@ class array(object):
         result = caller._query_compiler.ne(callee._query_compiler, **kwargs)
         return array(_query_compiler=result, _ndim=new_ndim)
 
+    def _compute_masked_mean(self, mask, output_dtype, axis):
+        # By default, pandas ignores NaN values when doing computations.
+        # NumPy; however, propagates the value by default. We use pandas
+        # default behaviour in order to mask values (by replacing them)
+        # with NaN when initially computing the mean, but we need to propagate
+        # NaN values that were not masked to the final output, so we do a
+        # sum along the same axis (where masked values are 0) to see where
+        # NumPy would propagate NaN, and swap out those values in our result
+        # with NaN.
+        target = mask.where(self, numpy.nan)._query_compiler
+        target = target.astype(
+            {col_name: output_dtype for col_name in target.columns}
+        ).mean(axis=axis)
+        na_propagation_mask = mask.where(self, 0)._query_compiler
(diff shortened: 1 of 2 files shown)


modin/numpy/arr.py
```diff
@@ -717,6 +717,24 @@ def __ne__(self, x2):
         result = caller._query_compiler.ne(callee._query_compiler, **kwargs)
         return array(_query_compiler=result, _ndim=new_ndim)
 
+    def _compute_masked_mean(self, mask, output_dtype, axis):
+        # By default, pandas ignores NaN values when doing computations.
+        # NumPy; however, propagates the value by default. We use pandas
```
vnlitvinov, line 722:
```suggestion
        # NumPy, however, propagates the value by default. We use pandas
```

modin/numpy/arr.py
```diff
@@ -1147,7 +1180,7 @@ def prod(
             return result if where is not False else initial
         target = where.where(self, 1) if isinstance(where, array) else self
         result = target._query_compiler.astype(
-            {col_name: out_dtype for col_name in self._query_compiler.columns}
+            {col_name: out_dtype for col_name in target._query_compiler.columns}
```
vnlitvinov, line 1183: why? shouldn't these be the same?

vnlitvinov (approved):

modin/numpy/arr.py
RehanSD: Yes, but I think it's better from a style perspective to have `target`. It's also less error prone in case we change how target is derived later on!

PR #5736, opened 2023-03-02
FIX-#5733: Allow all Modin objects in all Modin object constructors, and make sure copy=False works

### LINKED ISSUE #5733: BUG: Interoperability between `modin.pandas` and `modin.numpy`
### Modin version checks

### Reproducible Example

```python
In [1]: import modin.numpy as np
import modin.
In [2]: import modin.pandas as pd

In [3]: arr = np.array([[1, 2], [3, 4], [5, 6]])
UserWarning: Using Modin's new NumPy API. To convert from a Modin object to a NumPy array, either turn off the ExperimentalNumPyAPI flag, or use `modin.utils.to_numpy`.
df UserWarning: Ray execution environment not yet initialized. Initializing...

    import ray
    ray.init(runtime_env={'env_vars': {'__MODIN_AUTOIMPORT_PANDAS__': '1'}})

= pd.DataFrUserWarning: Distributing <class 'numpy.ndarray'> object. This may take some time.

In [4]: df = pd.DataFrame(arr)
UserWarning: Distributing <class 'modin.numpy.arr.array'> object. This may take some time.
Input In [4], in <cell line: 1>()
----> 1 df = pd.DataFrame(arr)

File ~/Documents/modin/modin/logging/logger_decorator.py:128, in enable_logging.<locals>.decorator.<locals>.run_and_log(*args, **kwargs)
    113 """
    114 Compute function with logging if Modin logging is enabled.
    115
   (...)
    125 Any
    126 """
    127 if LogMode.get() == "disable":
--> 128     return obj(*args, **kwargs)
    130 logger = get_logger()
    131 logger_level = getattr(logger, log_level)

File ~/Documents/modin/modin/pandas/dataframe.py:216, in DataFrame.__init__(self, data, index, columns, dtype, copy, query_compiler)
    210             return
    212         data = {
    213             k: v._to_pandas() if isinstance(v, Series) else v
    214             for k, v in data.items()
    215         }
--> 216     pandas_df = pandas.DataFrame(
    217         data=data, index=index, columns=columns, dtype=dtype, copy=copy
    218     )
    219     self._query_compiler = from_pandas(pandas_df)._query_compiler
    220 else:

File ~/.miniconda3/envs/modin/lib/python3.8/site-packages/pandas/core/frame.py:781, in DataFrame.__init__(self, data, index, columns, dtype, copy)
    778 # For data is scalar
    779 else:
    780     if index is None or columns is None:
--> 781         raise ValueError("DataFrame constructor not properly called!")
    783     index = ensure_index(index)
    784     columns = ensure_index(columns)

ValueError: DataFrame constructor not properly called!
```

### Issue Description

Passing a `modin.numpy` array to the constructors for `modin.pandas.DataFrame` and `modin.pandas.Series` errors out. We should make it so that they can be passed in to the constructors.

### Expected Behavior

We should get a dataframe that is equivalent to the array.

### Error Logs

<details>

```python-traceback

```

</details>

### Installed Versions

<details>

</details>

RehanSD:
## What do these changes do?

- [x] Resolves #5733 ? 
- [x] tests added and passing


modin/numpy/utils.py
```diff
@@ -19,12 +19,7 @@
 _INTEROPERABLE_TYPES = (pd.DataFrame, pd.Series)
 
 
-def try_convert_from_interoperable_type(obj):
+def try_convert_from_interoperable_type(obj, copy=True):
```
RehanSD, line 22: Given that the default behavior of converting to NumPy from pandas is copy=False, should we make copy False here? This is used in the conversion for when we call NumPy methods on Modin objects that aren't arrays.

modin/numpy/arr.py
```diff
@@ -137,9 +137,20 @@ def __init__(
         _query_compiler=None,
         _ndim=None,
     ):
+        self._siblings = []
         ErrorMessage.single_warning(
             "Using Modin's new NumPy API. To convert from a Modin object to a NumPy array, either turn off the ExperimentalNumPyAPI flag, or use `modin.utils.to_numpy`."
         )
+        if isinstance(object, array):
+            _query_compiler = object._query_compiler.copy()
```
RehanSD, line 145: I followed the convention of the Dataframe class here, and copy the QC - but this creates a deep copy of the query compiler - should we really do this even when copy is False?

RehanSD (commented):
Left a couple of questions that I was hoping for folks insight on

modin/numpy/utils.py
vnlitvinov: I would say yes, as defaulting to `True` here would break the scenario of modifying pandas objects from which numpy arrays were made...

modin/numpy/arr.py
```diff
@@ -137,9 +137,20 @@ def __init__(
...
+            if not copy:
+                object._add_sibling(self)
+            _ndim = object._ndim
+        if isinstance(object, (pd.DataFrame, pd.Series)):
```
vnlitvinov, line 149:
```suggestion
        elif isinstance(object, (pd.DataFrame, pd.Series)):
```
super-small perf speedup 😆

modin/numpy/arr.py
```diff
@@ -165,7 +176,11 @@ def __init__(
                     target_kwargs.pop(key)
                 else:
                     target_kwargs[key] = locals()[key]
-            arr = numpy.array(object, **target_kwargs)
+            if not isinstance(object, numpy.ndarray):
+                arr = numpy.array(object, **target_kwargs)
+            else:
+                # Avoid creating a second NumPy array in memory before distribution.
+                arr = object
```
vnlitvinov, line 183:
```suggestion
            arr = numpy.asarray(object, **target_kwargs)
```
should do what you want - avoid copying if `object` is already an array of desired type etc., see [numpy docs](https://numpy.org/doc/stable/reference/generated/numpy.asarray.html)

modin/numpy/arr.py
```diff
@@ -190,6 +205,38 @@ def __init__(
             {col_name: new_dtype for col_name in self._query_compiler.columns}
         )
 
+    def _add_sibling(self, sibling):
```
vnlitvinov, line 208: can we re-use that function from `BasePandasDataset` instead of copy-pasting it here somehow?

modin/numpy/arr.py
```diff
@@ -190,6 +205,38 @@ def __init__(
...
+        ----------
+        sibling : BasePandasDataset
+            Dataset to add to siblings list.
+        """
+        sibling._siblings = self._siblings + [self]
+        self._siblings += [sibling]
+        for sib in self._siblings:
+            sib._siblings += [sibling]
+
+    def _update_inplace(self, new_query_compiler):
```
vnlitvinov, line 225: same goes for this function, it's exact copy of the one in `BasePandasDataset` - I would propose to extract the `._siblings` and these methods into a [common mixin](https://www.pythontutorial.net/python-oop/python-mixin/)

modin/numpy/test/test_array.py
```diff
@@ -63,6 +63,78 @@ def test_conversion():
...
+        index=pd.Index([4, 6]),
+        dtype=str,
+    )
+    pandas_df = pandas.DataFrame(
+        numpy.array([[1, 2, 3], [4, 5, 6]]),
+        columns=["col 0", "col 1", "col 2"],
+        index=pandas.Index([4, 6]),
+        dtype=str,
+    )
+    df_equals(pandas_df, modin_df)
```
vnlitvinov, line 101:
```suggestion
    for kw in [{}, {"dtype": str}]:
        modin_df, pandas_df = [
            lib.DataFrame(
                np.array([[1, 2, 3], [4, 5, 6]]),
                columns=["col 0", "col 1", "col 2"],
                index=pd.Index([4, 6]),
                **kw
            )
            for lib in (pd, pandas)
        ]
        df_equals(pandas_df, modin_df)
    df_equals(pandas_df, modin_df)
```

modin/numpy/utils.py
```diff
@@ -19,12 +19,7 @@
...
     if isinstance(obj, _INTEROPERABLE_TYPES):
-        new_qc = obj._query_compiler.reset_index(drop=True)
-        new_qc.columns = range(len(new_qc.columns))
-        obj = np.array(
-            _query_compiler=new_qc,
-            _ndim=2 if isinstance(obj, pd.DataFrame) else 1,
-        )
```
vnlitvinov, line 29: why is this no longer needed?

modin/numpy/utils.py
```diff
@@ -19,12 +19,7 @@
...
+        obj = np.array(obj, copy=copy)
```
vnlitvinov, line 24: if this is all what this function does, can we replace its calls with `numpy.asarray(obj, copy=copy)` everywhere instead?

modin/pandas/dataframe.py
```diff
@@ -160,7 +162,18 @@ def __init__(
...
+            if copy is not None and not copy:
+                data._add_sibling(self)
+            if columns is not None and not isinstance(columns, pandas.Index):
+                columns = pandas.Index(columns)
+            if columns is not None:
+                self.set_axis(columns, axis=1, inplace=True)
+            if index is not None:
+                self.set_axis(index, axis=0, inplace=True)
+            if dtype is not None:
+                self.astype(dtype, copy=False)
```
vnlitvinov, line 176: shouldn't we either pass `inplace=True` or assign the result somewhere?

modin/numpy/arr.py
RehanSD: Digging into the code, it seems that `QC.copy()` creates a new query compiler object that points to the exact same partitions, so I think its fine to keep the copy here?

modin/numpy/utils.py
RehanSD: Since we now natively accept modin objects in the modin array constructor, we can just call the constructor. We don't need the reset index and columns part since thats handled in the constructor as well.

modin/numpy/utils.py
RehanSD: Not quite, since this is checking if object is an interoperable type, and making it a `modin.numpy.array` (not a `numpy.array`), and returning the original object if it isn't.

modin/pandas/dataframe.py
RehanSD: Setting `copy=False` makes the change happen inplace! (We do test that this works, see the test where we pass dtype=str).

modin/numpy/arr.py
RehanSD: Spoke to @devin-petersohn and he said he didn't like mixins + it would be more code to add it to a mixin or inheritance class, so he recommends just leaving it!

modin/numpy/utils.py
RehanSD: Sounds good to me, although we set `copy=False` explicitly anywhere where the array (and hence the df) should be modified, and in the other places, the array doesn't get modified (so neither should the df).

modin/numpy/arr.py
RehanSD: Fixed!

modin/numpy/utils.py
vnlitvinov: with that said, maybe we should not provide _any_ default value, hence requiring the user of this function to decide if they need a copy or not every time?

modin/numpy/arr.py
vnlitvinov: in that case, probably yeah, should be safe (and not too much memory, so better eat a little more mem than fail sometime down the drain)

modin/numpy/arr.py
vnlitvinov: I can see a point of "don't like", but it would actually be less code - you'd cut it from `BasePandasDataset` and paste somewhere else, and inherit in two places.

Technically it can even be not a mixin, but a common ancestor class for `BasePandasDataset` and this `array`, would it be better?

vnlitvinov (dismissed):
Reducing the copy-pasteness on siblings would be good IMHO, but I can live with some duplication 🙃 

LGTM!

modin/numpy/arr.py
RehanSD: That makes sense in the long run, but I think for now, we want to avoid that architectural complexity.

modin/numpy/arr.py
```diff
@@ -490,10 +490,10 @@ def max(
                         f"operand was set up as a reduction along axis 1, but the length of the axis is {out.shape[0]} (it has to be 1)"
                     )
                 if initial is not None and out is not None:
-                    out._update_inplace(
-                        (numpy.ones_like(out) * initial)._query_compiler
-                    )
-                if where is not False or out is not None:
+                    out._update_inplace = (
+                        numpy.ones_like(out) * initial
+                    )._query_compiler
```
vnlitvinov, line 495: uhh, `out._update_inplace =` feels awfully wrong

modin/numpy/arr.py
```diff
@@ -494,10 +490,10 @@ def max(
...
+                if truthy_where or out is not None:
```
vnlitvinov, line 496: where `truthy_where` is defined?

modin/numpy/arr.py
RehanSD: Yep - that's definitely wrong!

vnlitvinov (dismissed):
LGTM!

devin-petersohn (approved):
LGTM, thanks @RehanSD !

PR #5755, opened 2023-03-07
FEAT-#5754: Add np.linalg operations

### LINKED ISSUE #5754: FEAT: Add np.linalg operations
We should make basic linear algebra operations (mainly just `dot` and `linalg.norm`) available in `modin.numpy`.

noloerino:
## What do these changes do?
This PR adds support for `np.dot` and `np.linalg.norm`.

- [x] Resolves #5754 
- [x] tests added and passing

### DIFF
--- a/modin/numpy/arr.py
+++ b/modin/numpy/arr.py
@@ -1432,7 +1432,7 @@ class array(object):
         check_kwargs(keepdims=keepdims)
         if ord is not None and ord not in ("fro",):  # , numpy.inf, -numpy.inf, 0):
             raise NotImplementedError("unsupported ord argument for norm:", ord)
-        if axis < 0:
+        if isinstance(axis, int) and axis < 0:
             axis = self._ndim + axis
         apply_axis = axis or 0
         if apply_axis >= self._ndim or apply_axis < 0:
--- a/modin/numpy/arr.py
+++ b/modin/numpy/arr.py
@@ -1417,7 +1417,7 @@ class array(object):
             result = self._query_compiler.dot(other._query_compiler)
             result_ndim = 1
         if out is not None:
-            out._query_compiler = result
+            out._update_inplace(result)
         return array(_query_compiler=result, _ndim=result_ndim)
 
     def __matmul__(self, other):
--- a/modin/numpy/arr.py
+++ b/modin/numpy/arr.py
@@ -1390,6 +1390,60 @@ class array(object):
     ):
         return self.multiply(x2, out, where, casting, order, dtype, subok)
 
+    def dot(self, other, out=None):
+        other = try_convert_from_interoperable_type(other)
+        if numpy.isscalar(other):
+            # other is scalar -- result is an array
+            result = self._query_compiler.mul(other)
+        elif not isinstance(other, array):
+            raise TypeError(
+                f"Unsupported operand type(s): '{type(self)}' and '{type(other)}'"
+            )
+        elif self._ndim == 1 and other._ndim == 1:
+            # both 1D arrays -- result is a scalar
+            result = self._query_compiler.dot(
+                other._query_compiler, squeeze_self=True, squeeze_other=True
+            )
+            return result.to_numpy()[0, 0]
+        elif self._ndim == 2 and other._ndim == 2:
+            # both 2D arrays -- result is a 2D array
+            result = self._query_compiler.dot(other._query_compiler)
+        elif self._ndim == 1 and other._ndim == 2:
+            result = self._query_compiler.dot(other._query_compiler)
+        elif self._ndim == 2 and other._ndim == 1:
+            if other.shape[0] == 1:
+                result = self._query_compiler.dot(other._query_compiler.transpose())
+            else:
+                result = self._query_compiler.dot(other._query_compiler)
+        if out is not None:
+            out._query_compiler = result
+        return array(_query_compiler=result)
+
+    def _norm(self, ord=None, axis=None, keepdims=False):
+        check_kwargs(keepdims=keepdims)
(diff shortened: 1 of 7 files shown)


modin/numpy/arr.py
```diff
@@ -1347,6 +1347,60 @@ def __rmul__(
...
+            result = self._query_compiler.dot(other._query_compiler)
+        elif self._ndim == 1 and other._ndim == 2:
+            result = self._query_compiler.dot(other._query_compiler)
+        elif self._ndim == 2 and other._ndim == 1:
+            if other.shape[0] == 1:
+                result = self._query_compiler.dot(other._query_compiler.transpose())
+            else:
+                result = self._query_compiler.dot(other._query_compiler)
+        if out is not None:
+            out._query_compiler = result
```
RehanSD, line 1376: This should call `update_inplace`.

modin/numpy/arr.py
```diff
@@ -1347,6 +1347,60 @@ def __rmul__(
...
+        return array(_query_compiler=result)
+
+    def _norm(self, ord=None, axis=None, keepdims=False):
+        check_kwargs(keepdims=keepdims)
+        if ord is not None and ord not in ("fro",):  # , numpy.inf, -numpy.inf, 0):
+            raise NotImplementedError("unsupported ord argument for norm:", ord)
+        apply_axis = axis or 0
```
RehanSD, line 1383: We need to account for negative axes as well.

modin/numpy/arr.py
```diff
@@ -1347,6 +1347,60 @@ def __rmul__(
     ):
         return self.multiply(x2, out, where, casting, order, dtype, subok)
 
+    def dot(self, other, out=None):
```
RehanSD, line 1350: We need to alias `__matmul__` to `dot`.

RehanSD (changes requested):
Left a few minor comments

modin/numpy/arr.py
noloerino: The behavior of `matmul` and `dot` are not quite the same: https://numpy.org/doc/stable/reference/generated/numpy.matmul.html

That said, it looks like the "stacks of matrices" case only applies for higher dimensions, and the scalar multiplication case is easy to special case out. I'll add this extra logic.

modin/numpy/arr.py
noloerino: It seems like negative axes are supported (`axis=-1` is equivalent to `axis=1` for a 2D array, and `axis=-2` to `axis=0`). I'll add in a special case for this, though it seems like most of our other functions don't fully support this behavior.

modin/numpy/arr.py
noloerino: What object is `update_inplace` defined on? I also don't see it being used anywhere else in the numpy API for functions with `out` arguments.

RehanSD:
Mind using all close for testing norm, with error around 1e-12? That should clear the CI issue up!

modin/numpy/arr.py
```diff
@@ -1390,6 +1390,71 @@ def __rmul__(
...
+            result = self._query_compiler.dot(other._query_compiler)
+            result_ndim = 2
+        elif self._ndim == 1 and other._ndim == 2:
+            result = self._query_compiler.dot(other._query_compiler, squeeze_self=True)
+            result_ndim = 1
+        elif self._ndim == 2 and other._ndim == 1:
+            result = self._query_compiler.dot(other._query_compiler)
+            result_ndim = 1
+        if out is not None:
+            out._update_inplace(result)
```
RehanSD, line 1420: We should check if out has the correct dimensions, and error if it doesn't.

modin/numpy/arr.py
```diff
@@ -1390,6 +1390,71 @@ def __rmul__(
...
+                "cannot call matmul with a scalar argument (use np.dot instead)"
+            )
+        return self.dot(other)
+
+    def _norm(self, ord=None, axis=None, keepdims=False):
+        check_kwargs(keepdims=keepdims)
+        if ord is not None and ord not in ("fro",):  # , numpy.inf, -numpy.inf, 0):
+            raise NotImplementedError("unsupported ord argument for norm:", ord)
+        if isinstance(axis, int) and axis < 0:
+            axis = self._ndim + axis
```
RehanSD, line 1436: We can't do this, since we need the original axis in the error (I.e. if users passed in -3 for axis, it should error and say axis=-3 is invalid, not that axis =-1 is invalid)

modin/numpy/arr.py
```diff
@@ -1390,6 +1390,71 @@ def __rmul__(
...
+        apply_axis = axis or 0
+        if apply_axis >= self._ndim or apply_axis < 0:
+            raise numpy.AxisError(apply_axis, self._ndim)
+        result = self._query_compiler.applymap(lambda x: x**2)
```
RehanSD, line 1440: pow may be faster here and below.

RehanSD (changes requested):
LGTM, few minor comments

modin/numpy/arr.py
noloerino: I've replaced this with `fix_dtypes_and_determine_return`.

modin/numpy/arr.py
noloerino: Fixed.

modin/numpy/arr.py
noloerino: Fixed, I thought your earlier comment was specifically in reference to replacing np.sqrt.

RehanSD (approved):
LGTM! Great work Jonathan!

Issue #5122, opened 2022-10-12
TEST: epic: Make codecov more useful and less annoying


mvashishtha:
This is an epic covering a few smaller goals for codecov.

# Current state:

## Pull requests

1. Many jobs in Modin's CI that successfully finish tests will upload all coverage data (an xml from pytest) to the codecov app. The jobs upload data using `codecov/codecov-action@v2`, e.g. [here](https://github.com/modin-project/modin/blob/88f7b27d32b099c508e93251cafe503f2908d426/.github/workflows/ci.yml#L328)
2. (does not necessarily happen before 3, but in practice always does) we get an initial email and PR comment when codecov gets coverage data for the first test (see step 3 for what these results mean). codecov will keep updating this comment as more CI jobs finish. Example comment [here](https://github.com/modin-project/modin/pull/5121#issuecomment-1276241373).
3. Aggregating all the "hits" / "misses" (i.e. lines covered / not covered by tests) from all tests that ran from all workflows for the PR, codecov compares coverage on the CI jobs to coverage of the last green commit on master
    - For a PR with no changes and assuming all tests pass for both PR and master, codecov will typically give about 4% extra coverage. This is mainly because the set of CI runs for master vs a PR are different. the pull request jobs actually do have different coverage than the push-to-master jobs. e.g. here we get +100% coverage in compat because we test compat on pull request but not on merge: https://app.codecov.io/gh/modin-project/modin/pull/5048

## Commits on master
On merging commits to master, codecov calculates coverage for the project and calculates how much of the diff from the commit was covered by tests. e.g. for [this commit](https://github.com/modin-project/modin/commit/4bcce6ef12c0c80d6df15ef4a4b2b3a85713c106) we have both results [here](https://app.codecov.io/gh/modin-project/modin/commit/4bcce6ef12c0c80d6df15ef4a4b2b3a85713c106). These give confusing results, e.g. I don't understand the losses [here](https://app.codecov.io/gh/modin-project/modin/commit/d0054290ec8305d01b040e740e3f4524815ea77e). Also they don't always compare to the last green commit, so e.g. [here](https://app.codecov.io/gh/modin-project/modin/commit/0215a133ac32969ca0a4126f6aef82be099ae8ef) we seem to get a bunch of gains but that's just because the omnisci tests failed on the previous commit.

[...]

PR #5782, opened 2023-03-14
TEST-#2544: delay `codecov` notifications until all reports have been sent

### LINKED ISSUE #2544: Use codecov' delayed notifications in our CI

It would be great if we did not see premature numbers from codecov.

For details see:
https://docs.codecov.io/docs/merging-reports#delayed-notifications

anmyachev:
## What do these changes do?

* The main change is that all coverage reports are first sent to a temporary storage location using `actions/download-artifact@v3.0.2`. After all the reports have been received, we combine them, convert them to `xml` format and send them in the same way as before using `codecov/codecov-action@v3`.
* Also, `fail_ci_if_error: true` new option  is now used, very often reports were not loaded for some reason, now, if such a situation occurs, it will be easy to restart `upload-coverage` job. It's much faster than rerunning tests.
* This is a step forward to https://github.com/modin-project/modin/issues/5122.

- [x] Resolves #2544 
- [x] tests added and passing


setup.cfg
```diff
@@ -12,7 +12,7 @@ tag_prefix =
 parentdir_prefix = modin-
 
 [tool:pytest]
-addopts = --disable-pytest-warnings --cov-config=setup.cfg --cov=modin --cov-append --cov-report=xml
+addopts = --disable-pytest-warnings --cov-config=setup.cfg --cov=modin --cov-append
```
anmyachev, line 15: Now coverage reports are merged manually, this requires using the raw format (simply `.coverage`).

.github/workflows/ci.yml
```diff
@@ -1002,3 +1002,30 @@ jobs:
...
+        with:
+          fetch-depth: 1
+      - name: Download coverage data
+        uses: actions/download-artifact@v3.0.2
+        with:
+          name: coverage-data
+      - uses: actions/setup-python@v4
+        with:
+          python-version: "3.8.x"
+          architecture: "x64"
```
vnlitvinov, line 1023: I wonder if this is necessary... I think it might be faster to install if we let github use its defaults (as they're usually pre-cached on the workers already)

.github/workflows/upload-coverage/action.yml
```diff
@@ -0,0 +1,18 @@
+name: Upload Coverage
+description: Upload coverage files
+
+runs:
+  using: "composite"
+
+  steps:
+    - run: |
+        COVERAGE_UUID=$(python3 -c "import uuid; print(uuid.uuid4())")
```
vnlitvinov, line 9: shouldn't we use `uuid.uuid4().hex` here? or does `print()` already provide us with that result?

setup.cfg
```diff
@@ -62,6 +62,7 @@ omit =
     # Skip CLI part
     modin/__main__.py
 parallel = True
+relative_files = true
```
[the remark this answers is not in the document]
vnlitvinov: I would rather you put this as an in-file comment 😉

.github/workflows/upload-coverage/action.yml
anmyachev: So it seems to work well too (without `.hex`).
```bash
(modin) C:\projects\modin>python -c "import uuid; print(uuid.uuid4())     
9990da74-437a-4a2a-94c6-bbd718e5c9b5
```

```bash
(modin) C:\projects\modin>python -c "import uuid; print(uuid.uuid4().hex) 
ef5c0022e478403cae1d9df6dc89719e
```

anmyachev:
@vnlitvinov thanks for review!

anmyachev:
@vnlitvinov answered all your comments

.github/workflows/ci.yml
anmyachev: It looks like the code works with the default python.

vnlitvinov (approved):
LGTM!

PR #5814, opened 2023-03-17
TEST-#0000: use custom Distributed

anmyachev:
## What do these changes do?

In the custom version of this `Distributed`, we remove the call of `self._get_computation_code()`, which is called in `submit->_graph_to_futures`. `_get_computation_code` takes about 75% of the total call time of `submit` function and it looks like it's just collecting additional information that might not be collected in a non-debug mode. With this change, Dask tests are faster than Ray tests. It is interesting to try to propose to implement a mechanism that disables the call to this function in `Distributed` library itself.

- [ ] Resolves #? 
- [ ] tests added and passing

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -692,6 +692,8 @@ jobs:
           # we set use-only-tar-bz2 to false in order for conda to properly find new packages to be installed
           # for more info see https://github.com/conda-incubator/setup-miniconda/issues/264
           use-only-tar-bz2: false
+      - name: Install custom Distributed
+        run: pip install git+https://github.com/anmyachev/distributed.git@get-computation-code
       - name: Conda environment
         run: |
           conda info


mvashishtha:
@anmyachev this looks very promising. Should we try contributing this to dask?

anmyachev:
> @anmyachev this looks very promising. Should we try contributing this to dask?

@mvashishtha I think we should at least suggest introducing an environment variable that would regulate the collection of information using `_get_computation_code` function. For Modin, this could only be enabled in debug mode.

I don't have time yet, but if someone has the opportunity, then he can start doing right now :)

mvashishtha:
Okay, I filed https://github.com/dask/distributed/issues/7740 with dask/distributed

mvashishtha:
@anmyachev how did you measure how long `_get_computation_code` took? Some dask tests, like ` MODIN_ENGINE=DASK python -m pytest -n 2  modin/pandas/test/dataframe/test_indexing.py`, are excruciatingly slow in CI as well as on my laptop.

anmyachev:
> @anmyachev how did you measure how long `_get_computation_code` took? Some dask tests, like ` MODIN_ENGINE=DASK python -m pytest -n 2 modin/pandas/test/dataframe/test_indexing.py`, are excruciatingly slow in CI as well as on my laptop.

IIRC ran a full set of tests on Dask with https://pypi.org/project/pytest-profiling/.

PR #5847, opened 2023-03-23
FIX-#5829: fix ndarray assignment via loc

### LINKED ISSUE #5829: BUG: Syntax Error in modin > pandas > indexing.py:829
### Modin version checks

### Reproducible Example

```python
import modin.pandas as mp
import numpy as np

temp = mp.DataFrame({'a': [1,2,3,4,5], 'b': [11,12,13,14,15]}, dtype=object)

temp.loc[[3, 4], ['c', 'd', 'e']] = np.array([[24, 34, 44], [25, 35, 45]])
```

### Issue Description

I think there is a syntax error in the line `829 of modin/pandas/indexing.py`

### Expected Behavior

The expected code is : 

```python

... if item.shape != (len(self.qc.index), len(col_loc)):
...

```

Therefore, the error of the mismatch of the shapes should be as following :

```python-traceback

...\modin\pandas\indexing.py:830, in _LocIndexer._setitem_with_new_columns(self, row_loc, col_loc, item)
    828     else:
    829         if item.shape != (len(self.qc.index), len(col_loc)):
--> 830             raise ValueError(
    831                 "Must have equal len keys and value when setting with an iterable"
    832             )
    833     exist_items = (
    834         item[:, common_label_loc]
    835         if len(item.shape) > 1
    836         else item[common_label_loc]
    837     )
    838 if not all(common_label_loc):
    839     # In this case we have some new cols and some old ones

```

### Error Logs

<details>

```python-traceback

..\lib\site-packages\modin\pandas\indexing.py:829, in _LocIndexer._setitem_with_new_columns(self, row_loc, col_loc, item)
    825         raise ValueError(
    826             "Must have equal len keys and value when setting with an iterable"
    827         )
    828 else:
--> 829     if item.shape != (len(self.qc.index, len(col_loc))):
    830         raise ValueError(
    831             "Must have equal len keys and value when setting with an iterable"
    832         )
    833 exist_items = (
    834     item[:, common_label_loc]
    835     if len(item.shape) > 1
    836     else item[common_label_loc]
    837 )

TypeError: len() takes exactly one argument (2 given)

```

</details>

### Installed Versions

<details>

Modin dependencies
------------------
modin            : 0.19.0
ray              : 2.3.0
...

pandas dependencies
-------------------
pandas           : 1.5.3
...

</details>

anmyachev:
## What do these changes do?

- [x] Resolves #5829 
- [x] tests added and passing

### DIFF
--- a/modin/pandas/indexing.py
+++ b/modin/pandas/indexing.py
@@ -838,22 +838,23 @@ class _LocIndexer(_LocationIndexerBase):
                         "Must have equal len keys and value when setting with an iterable"
                     )
             else:
-                if item.shape != (len(self.qc.index, len(col_loc))):
+                if item.shape != (len(row_loc), len(col_loc)):
                     raise ValueError(
                         "Must have equal len keys and value when setting with an iterable"
                     )
-            exist_items = (
-                item[:, common_label_loc]
-                if len(item.shape) > 1
-                else item[common_label_loc]
-            )
+            if not isinstance(item, np.ndarray):
+                exist_items = (
+                    item[:, common_label_loc]
+                    if len(item.shape) > 1
+                    else item[common_label_loc]
+                )
         if not all(common_label_loc):
             # In this case we have some new cols and some old ones
             columns = self.qc.columns
             for i in range(len(common_label_loc)):
                 if not common_label_loc[i]:
                     columns = columns.insert(len(columns), col_loc[i])
-            self.qc = self.qc.reindex(labels=columns, axis=1, fill_value=0)
+            self.qc = self.qc.reindex(labels=columns, axis=1, fill_value=np.NaN)
             self.df._update_inplace(new_query_compiler=self.qc)
         self._set_item_existing_loc(row_loc, np.array(col_loc), exist_items)
 
--- a/modin/pandas/indexing.py
+++ b/modin/pandas/indexing.py
@@ -829,7 +829,6 @@ class _LocIndexer(_LocationIndexerBase):
             Value that should be assigned to located dataset.
         """
         exist_items = item
-        common_label_loc = np.isin(col_loc, self.qc.columns.values)
         if is_list_like(item) and not isinstance(item, (DataFrame, Series)):
             item = np.array(item)
             if len(item.shape) == 1:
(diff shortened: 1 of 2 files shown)


modin/pandas/indexing.py
```diff
@@ -821,22 +821,23 @@ def _setitem_with_new_columns(self, row_loc, col_loc, item):
                         "Must have equal len keys and value when setting with an iterable"
                     )
             else:
-                if item.shape != (len(self.qc.index, len(col_loc))):
+                if item.shape != (len(row_loc), len(col_loc)):
```
anmyachev, line 824: I believe that we need to compare the lengths of the keys with the object that we insert.

modin/pandas/indexing.py
```diff
@@ -821,22 +821,23 @@ def _setitem_with_new_columns(self, row_loc, col_loc, item):
...
+                    else item[common_label_loc]
+                )
         if not all(common_label_loc):
             # In this case we have some new cols and some old ones
             columns = self.qc.columns
             for i in range(len(common_label_loc)):
                 if not common_label_loc[i]:
                     columns = columns.insert(len(columns), col_loc[i])
-            self.qc = self.qc.reindex(labels=columns, axis=1, fill_value=0)
+            self.qc = self.qc.reindex(labels=columns, axis=1, fill_value=np.NaN)
```
anmyachev, line 840: Apparently the behavior in pandas has been changed.

modin/pandas/indexing.py
```diff
@@ -821,22 +821,23 @@ def _setitem_with_new_columns(self, row_loc, col_loc, item):
...
+                if item.shape != (len(row_loc), len(col_loc)):
                     raise ValueError(
                         "Must have equal len keys and value when setting with an iterable"
                     )
-            exist_items = (
-                item[:, common_label_loc]
-                if len(item.shape) > 1
-                else item[common_label_loc]
-            )
+            if not isinstance(item, np.ndarray):
```
anmyachev, line 828: At least in the case when the keys are strings, we will get an empty array here, which is not true.

anmyachev: The previous comment is not entirely correct.

This part of the code was responsible for not taking those columns from the assigned value that were not in the query compiler, however this behavior is not true for `loc` operation, since missed columns will be added in the next few lines via `reindex` op.

anmyachev: Considering also that in this code branch there can only be a numpy array, I completely deleted this code.

dchigarev (approved):

PR #5882, opened 2023-03-28
REFACTOR-#2863: remove `other_name` from broadcast_apply

### LINKED ISSUE #2863: BaseFrameManager.broadcast_apply has argument "other_name" but it is impossible to pass from query_compiler level
### System information

Ubuntu 20.04.2 LTS

0.9.0+15.g371b1abe.dirty

- **Python version**:

Python 3.8.6

### Describe the problem

There is a function `BaseFrameManager.broadcast_apply` https://github.com/modin-project/modin/blob/master/modin/engines/base/frame/partition_manager.py#L251 which has argument `other_name` which means name of second frame argument in the applied function. But when calling this function via `BasePandasFrame.broadcast_apply` it is impossible to specify this name because `BasePandasFrame.broadcast_apply` has no such argument and always calls `BaseFrameManager.broadcast_apply` implicitly without specifying `other_name` parameter https://github.com/modin-project/modin/blob/master/modin/engines/base/frame/data.py#L1509.
Because of this developer is forced to write apply function with implicit 2nd parameter name `r` which isn't very meaningful and isn't documented anywhere.

anmyachev:
## What do these changes do?

Using `other_name` option looks like a complication. In all cases where `broadcast_apply` function is used, it is assumed that the user-defined function has the following signature: `func(df, other)` (only the name of `other` parameter changes on different calls). Nothing prevents us from passing this parameter as a positional one.

- [x] Resolves #2863 
- [x] tests added and passing

### DIFF
--- a/modin/core/dataframe/pandas/partitioning/partition_manager.py
+++ b/modin/core/dataframe/pandas/partitioning/partition_manager.py
@@ -242,7 +242,7 @@ class PandasDataframePartitionManager(ClassLogger, ABC):
 
         if by is not None:
             mapped_partitions = cls.broadcast_apply(
-                axis, map_func, left=partitions, right=by, other_name="other"
+                axis, map_func, left=partitions, right=by
             )
         else:
             mapped_partitions = cls.map_partitions(partitions, map_func)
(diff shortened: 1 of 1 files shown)


anmyachev:
@dchigarev let's merge

PR #5027, opened 2022-09-23
REFACTOR-#5026: Change exception names to simplify grepping

anmyachev:
Signed-off-by: Myachev <anatoly.myachev@intel.com>

## What do these changes do?

- [x] commit message follows format outlined [here](https://modin.readthedocs.io/en/latest/development/contributing.html#commit-message-formatting)
- [x] passes `flake8 modin/ asv_bench/benchmarks scripts/doc_checker.py`
- [x] passes `black --check modin/ asv_bench/benchmarks scripts/doc_checker.py`
- [x] signed commit with `git commit -s` 
- [x] Resolves #5026 
- [x] tests added and passing
- [x] module layout described at `docs/development/architecture.rst` is up-to-date 
- [x] added (Issue Number: PR title (PR Number)) and github username to release notes for next major release

PR #5946, opened 2023-04-05
FIX-#4954: defaults to pandas in `read_json` in case of rows having different columns

### LINKED ISSUE #4954: Reading large json lines dataset fails because of rows having different columns
### System information
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: Ubuntu 20.04.4 LTS
- **Modin version** (`modin.__version__`): 0.15.3
- **Python version**: 3.10.6
```python
import modin.pandas as pd
pd.read_json("Automotive_5.json", lines=True)
```
where `Automotive_5.json` is the "small" (1711519 lines) automotive dataset from https://nijianmo.github.io/amazon/index.html
I tried to reproduce it by generating synthetic json lines file where different json objects have different fields with no avail. Have not yet tried to reduce the dataset to smaller to see when it stops failing.

### Describe the problem

I am trying to load large json lines file (which is actually the whole dataset, but it reproduces with the smaller one) and it errors out with `ValueError('Columns must be the same across all rows.')`. I can work around this by loading the dataset with `pandas` and then transforming that to `moding` dataframe, but that is unnecessarily slow.

### Source code / logs

[output.txt](https://github.com/modin-project/modin/files/9540968/output.txt)

anmyachev:
## What do these changes do?

- [x] Resolves #4954 
- [x] tests added and passing


mvashishtha (approved):

mvashishtha:
Thank you @anmyachev !

modin/core/io/file_dispatcher.py
```diff
@@ -154,7 +155,12 @@ def read(cls, *args, **kwargs):
         dispatcher class `_read` function with passed parameters and performs some
         postprocessing work on the resulting query_compiler object.
         """
-        query_compiler = cls._read(*args, **kwargs)
+        try:
+            query_compiler = cls._read(*args, **kwargs)
+        except ModinAssumptionError as exc:
```
YarShev, line 160: exc -> err

https://github.com/modin-project/modin/pull/5027/files

anmyachev: @YarShev https://github.com/modin-project/modin/pull/6252

PR #5981, opened 2023-04-11
FIX-#0000: use `wrapper.materialize` instead of `wait_partitions`

anmyachev:
## What do these changes do?

- [ ] Resolves #? 
- [ ] tests added and passing


pyrito (approved):
I see - so we were writing it out async? That could explain why the parquet files weren't found.

pyrito:
@anmyachev could you confirm that this fixes the issue by running the push-to-main to trigger on changes to your branch? I missed that `test-ray-client` was the test that was failing when I made my "fix".

anmyachev:
@pyrito I see `botocore.exceptions.NoCredentialsError: Unable to locate ` in https://github.com/modin-project/modin/actions/runs/4671549630/jobs/8272679284?pr=5982

pyrito:
@anmyachev I think I figured it out. If you take a look at how we initialize Ray normally here: https://github.com/modin-project/modin/blob/master/modin/core/execution/ray/common/utils.py#L81-L88, you can see that we actually do pass in the mock keys to Ray init. We initialize the Ray client differently on CI: https://github.com/modin-project/modin/blob/master/modin/conftest.py#L587-L591. My guess is that we need to pass in these env variables through here as well.

anmyachev:
> @anmyachev I think I figured it out. If you take a look at how we initialize Ray normally here: https://github.com/modin-project/modin/blob/master/modin/core/execution/ray/common/utils.py#L81-L88, you can see that we actually do pass in the mock keys to Ray init. We initialize the Ray client differently on CI: https://github.com/modin-project/modin/blob/master/modin/conftest.py#L587-L591. My guess is that we need to pass in these env variables through here as well.

@pyrito good catch! Let's try.

pyrito:
@anmyachev please see here: https://github.com/modin-project/modin/pull/5984/commits/464185f91f50267a3336685cfd20581be2cdb30b

pyrito (approved):

YarShev (approved):

PR #5983, opened 2023-04-11
TEST-#5991: add simple tests for `read_orc`, `read_spss`, `json_normalize`, `read_xml`, `read_gbq`

### LINKED ISSUE #5991: add tests for `read_orc`, `read_spss`, `json_normalize`, `read_xml`, `read_gbq`

anmyachev:
## What do these changes do?

- [x] Resolves #5991 
- [x] tests added and passing

### DIFF
--- a/modin/core/io/io.py
+++ b/modin/core/io/io.py
@@ -174,42 +174,9 @@ class BaseIO:
         summary="Load data from Google BigQuery into query compiler",
         returns=_doc_returns_qc,
     )
-    def read_gbq(
-        cls,
-        query: str,
-        project_id=None,
-        index_col=None,
-        col_order=None,
-        reauth=False,
-        auth_local_webserver=False,
-        dialect=None,
-        location=None,
-        configuration=None,
-        credentials=None,
-        use_bqstorage_api=None,
-        private_key=None,
-        verbose=None,
-        progress_bar_type=None,
-        max_results=None,
-    ):  # noqa: PR01
+    def read_gbq(cls, query: str, **kwargs):  # noqa: PR01
         ErrorMessage.default_to_pandas("`read_gbq`")
-        return cls.from_pandas(
-            pandas.read_gbq(
-                query,
-                project_id=project_id,
-                index_col=index_col,
-                col_order=col_order,
-                reauth=reauth,
-                auth_local_webserver=auth_local_webserver,
-                dialect=dialect,
-                location=location,
-                configuration=configuration,
-                credentials=credentials,
-                use_bqstorage_api=use_bqstorage_api,
-                progress_bar_type=progress_bar_type,
-                max_results=max_results,
-            )
-        )
+        return cls.from_pandas(pandas.read_gbq(query, **kwargs))
 
     @classmethod
     @_inherit_docstrings(pandas.read_html, apilink="pandas.read_html")
--- a/modin/core/io/io.py
+++ b/modin/core/io/io.py
@@ -174,9 +174,42 @@ class BaseIO:
         summary="Load data from Google BigQuery into query compiler",
         returns=_doc_returns_qc,
     )
-    def read_gbq(cls, query: str, **kwargs):  # noqa: PR01
+    def read_gbq(
+        cls,
+        query: str,
+        project_id=None,
+        index_col=None,
+        col_order=None,
+        reauth=False,
+        auth_local_webserver=False,
+        dialect=None,
+        location=None,
+        configuration=None,
+        credentials=None,
+        use_bqstorage_api=None,
+        private_key=None,
+        verbose=None,
+        progress_bar_type=None,
+        max_results=None,
+    ):  # noqa: PR01
         ErrorMessage.default_to_pandas("`read_gbq`")
-        return cls.from_pandas(pandas.read_gbq(query, **kwargs))
+        return cls.from_pandas(
+            pandas.read_gbq(
+                query,
+                project_id=project_id,
+                index_col=index_col,
+                col_order=col_order,
+                reauth=reauth,
+                auth_local_webserver=auth_local_webserver,
+                dialect=dialect,
+                location=location,
+                configuration=configuration,
+                credentials=credentials,
+                use_bqstorage_api=use_bqstorage_api,
+                progress_bar_type=progress_bar_type,
+                max_results=max_results,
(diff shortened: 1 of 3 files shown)


modin/pandas/test/test_io.py
```diff
@@ -2613,6 +2629,56 @@ def test_to_pickle(self):
...
+ <row>
+   <shape>circle</shape>
+   <degrees>360</degrees>
+   <sides/>
+ </row>
+"""
+        eval_io("read_xml", path_or_buffer=data)
+
+
+class TestOrc:
```
anmyachev, line 2651: For some formats (`orc`, `spss`), it is not easy to add infrastructure. In cases where we default to pandas, I think it's enough to check that the parameters are passed to pandas.

modin/pandas/io.py
```diff
@@ -761,7 +761,7 @@ def read_orc(
     Load an ORC object from the file path, returning a DataFrame.
     """
     ErrorMessage.default_to_pandas("read_orc")
-    return DataFrame(pandas.read_orc(path, columns, **kwargs))
+    return DataFrame(pandas.read_orc(path, columns=columns, **kwargs))
```
anmyachev, line 764: This is not a mistake, it's just more convenient to test, from the point of view of logic, nothing changes.

modin/core/io/io.py
```diff
@@ -174,42 +174,9 @@ def read_json(
         summary="Load data from Google BigQuery into query compiler",
         returns=_doc_returns_qc,
     )
-    def read_gbq(
```
anmyachev, line 177: Some parameters are outdated (for example `private_key`). I guess we can just use `**kwargs`.

modin/pandas/io.py
```diff
@@ -716,17 +716,17 @@ def to_pickle(
...
 ):  # noqa: PR01, RT01, D200
     """
     Load an SPSS file from the file path, returning a DataFrame.
     """
+    _, _, _, kwargs = inspect.getargvalues(inspect.currentframe())
+
     from modin.core.execution.dispatching.factories.dispatcher import FactoryDispatcher
 
-    return DataFrame(
-        query_compiler=FactoryDispatcher.read_spss(path, usecols, convert_categoricals)
```
anmyachev, line 728: It didn't work at all before. Parameters must be passed as keywords.

modin/core/io/io.py
```diff
@@ -174,42 +174,9 @@ def read_json(
...
-        location=None,
-        configuration=None,
-        credentials=None,
-        use_bqstorage_api=None,
-        private_key=None,
-        verbose=None,
-        progress_bar_type=None,
-        max_results=None,
-    ):  # noqa: PR01
+    def read_gbq(cls, query: str, **kwargs):  # noqa: PR01
```
vnlitvinov, line 177: I don't like that you're changing API in a PR marked as `TEST-`.

Plus this change hurts IDE hinting and interactive help...

modin/pandas/io.py
```diff
@@ -716,17 +716,17 @@ def to_pickle(
 @enable_logging
 def read_spss(
     path: Union[str, pathlib.Path],
-    usecols: Union[Sequence[str], type(None)] = None,
+    usecols: Sequence[str] | None = None,
```
vnlitvinov, line 719: This is Python 3.10 syntax, and our minimum version is 3.8. I suggest using `Optional`:

```suggestion
    usecols: Optional[Sequence[str]] = None,
```

modin/pandas/io.py
```diff
@@ -716,17 +716,17 @@ def to_pickle(
...
     convert_categoricals: bool = True,
 ):  # noqa: PR01, RT01, D200
     """
     Load an SPSS file from the file path, returning a DataFrame.
     """
+    _, _, _, kwargs = inspect.getargvalues(inspect.currentframe())
+
```
vnlitvinov, line 726: let's not introduce this line... it's kinda slow and not that much useful when we have only 3 arguments

modin/pandas/io.py
```diff
@@ -716,17 +716,17 @@ def to_pickle(
...
     from modin.core.execution.dispatching.factories.dispatcher import FactoryDispatcher
 
-    return DataFrame(
-        query_compiler=FactoryDispatcher.read_spss(path, usecols, convert_categoricals)
-    )
+    return DataFrame(query_compiler=FactoryDispatcher.read_spss(**kwargs))
```
vnlitvinov, line 729: I would rather we pass explicitly, it's only three arguments to spell, not ~19 like in `.read_csv`

modin/pandas/test/test_io.py
```diff
@@ -2491,6 +2493,20 @@ def test_to_gbq(self):
...
             modin_df.to_gbq("modin.table")
 
+    def test_read_gbq_mock(self):
+        test_args = ("fake_query",)
+        test_kwargs = {"project_id": "test_id", "dialect": "standart"}
+        all_params = inspect.signature(pd.read_gbq).parameters
+        for param in all_params:
+            # 'query' - positional arg
+            if param not in test_kwargs and param != "query":
+                test_kwargs[param] = all_params[param].default
```
vnlitvinov, line 2503: maybe something like (untested)
```suggestion
        test_kwargs = inspect.signature(pd.read_gbq).parameters.copy()
        test_kwargs.update(project_id="test_id", dialect="standart")
        test_kwargs.pop("query", None)
```
?

modin/pandas/test/test_io.py
vnlitvinov: this should be in-code comment 😉

modin/pandas/io.py
anmyachev: You are right, but there are many such places in our code. Something will need to be done about this as well.

modin/pandas/test/test_io.py
anmyachev: added

modin/core/io/io.py
anmyachev: reverted

anmyachev:
@vnlitvinov thanks for review! Ready for another one :)

anmyachev:
@vnlitvinov ping

vnlitvinov (approved):
LGTM!

PR #5995, opened 2023-04-12
FEAT-#5936: support pandas 2.0.2

### LINKED ISSUE #5936: Support pandas 2.x
Pandas 2.0.0 is officially [out](https://pandas.pydata.org/pandas-docs/version/2.0/whatsnew/v2.0.0.html). Probably the most notable improvement is added support for pyarrow backend. Would be great to assess the compatibility with the 2.0.0 API & add support

anmyachev:
## What do these changes do?

- [x] Resolves #5936 
- [ ] tests added and passing


anmyachev:
Do not forget to add a check for the equivalence of parameters for groupby objects.

anmyachev:
Do not forget to add a check for the equivalence of parameters for methods of `str`, and `dt` accessor.

anmyachev:
It will be possible to close https://github.com/modin-project/modin/issues/3142.

anmyachev:
I guess we can go straight to support 2.0.1

anmyachev:
Current status:
* test_api.py - works
* test_concat.py - works
* test_expanding.py - doesn't work
* test_general.py - doesn't work
* test_groupby.py - doesn't work
* test_io.py - works
* test_reshape.py - works
*

anmyachev:
One of the remaining problems is converting a pyarrow table to a pandas dataframe while saving timestamp resolution.

@AndreyPavlenko can you suggest a solution?

```python
Desired behaviour:
timestamp[s] -> datetime64[s]
timestamp[ms] -> datetime64[ms]
timestamp[us] -> datetime64[us]
timestamp[ns] -> datetime64[ns]

Current behaviour:
timestamp[s] -> datetime64[ns]
timestamp[ms] -> datetime64[ns]
timestamp[us] -> datetime64[ns]
timestamp[ns] -> datetime64[ns]
```

anmyachev:
I guess we can start the review, as the pull request is large and it is very easy to miss something.

AndreyPavlenko:
> can you suggest a solution?

I don't know an efficient solution. Moreover, pandas forcibly converts all datetime64 values to datetime64[ns]:
```python
array = pa.array([1678966200, 1678966260, 1678966320], type=pa.timestamp('s'))
at = pa.Table.from_arrays([array], names=['A'])
data = {f._name: f.to_numpy() for f in at.columns}
print(data["A"].dtype)
print(pandas.DataFrame(data).dtypes["A"])
```
```
datetime64[s]
datetime64[ns]
```

anmyachev:
> Moreover, pandas forcibly converts all datetime64 values to datetime64[ns]:

@AndreyPavlenko since pandas 2.0 the behavior is changing.
```bash
datetime64[s]
datetime64[s]
```

modin/core/io/io.py
```diff
@@ -436,6 +438,8 @@ def read_sql(
         parse_dates=None,
         columns=None,
         chunksize=None,
+        dtype_backend=no_default,
+        dtype=None,
```
YarShev, line 442: Why are the arguments not propagated into the original method?

modin/core/io/text/text_file_dispatcher.py
```diff
@@ -868,10 +868,16 @@ def _define_index(
             Partitions rows lengths.
         """
         index_objs = cls.materialize(index_ids)
-        if len(index_objs) == 0 or isinstance(index_objs[0], int):
+
+        # fix for 'test_read_csv_error_handling'
```
YarShev, line 872: The comment looks confusing and redundant. I think it can be removed.

modin/core/storage_formats/pandas/query_compiler.py
```diff
@@ -907,16 +907,16 @@ def mean(self, axis, **kwargs):
         # TODO-FIX: this function may work incorrectly with user-defined "numeric" values.
         # Since `count(numeric_only=True)` discards all unknown "numeric" types, we can get incorrect
         # divisor inside the reduce function.
-        def map_fn(df, **kwargs):
+        def map_fn(df, numeric_only=False, **kwargs):
```
YarShev, line 910: Why do we need this change?

modin/core/storage_formats/pandas/query_compiler.py
```diff
@@ -1965,7 +1956,7 @@ def astype(self, col_dtypes, errors: str = "raise"):
         # invalid type keys.
         return self.__constructor__(self._modin_frame.astype(col_dtypes, errors=errors))
 
-    def infer_objects(self):
+    def infer_objects(self, copy):
```
YarShev, line 1959: Why is `copy` not propagated into the method of the `modin_frame`?

modin/pandas/__init__.py
```diff
@@ -14,7 +14,7 @@
 import pandas
 import warnings
 
-__pandas_version__ = "1.5.3"
+__pandas_version__ = "2.0.1"
```
YarShev, line 17: Don't we want to do the upgrade to 2.0.1 in a separate release?

modin/pandas/base.py
```diff
@@ -569,6 +566,10 @@ def _get_axis_number(cls, axis):
 
         return cls._pandas_class._get_axis_number(axis) if axis is not None else 0
 
+    def _get_axis_name(cls, axis):
```
YarShev, line 569:
```suggestion
    def _get_axis_name(self, axis):
```

modin/pandas/base.py
```diff
@@ -569,6 +566,10 @@ def _get_axis_number(cls, axis):
...
+        axis_number = cls._get_axis_number(axis)
+        return cls._AXIS_ORDERS[axis_number]
```
YarShev, line 571:
```suggestion
        axis_number = self._get_axis_number(axis)
        return self._AXIS_ORDERS[axis_number]
```

modin/pandas/base.py
```diff
@@ -1133,7 +1067,7 @@ def clip(
             axis = self._get_axis_number(axis)
         self._validate_dtypes(numeric_only=True)
         inplace = validate_bool_kwarg(inplace, "inplace")
-        axis = numpy_compat.function.validate_clip_with_axis(axis, args, kwargs)
+        axis = numpy_compat.function.validate_clip_with_axis(axis, (), kwargs)
```
YarShev, line 1070: Why do we pass just `()` now?

modin/pandas/base.py
```diff
@@ -2021,37 +1903,22 @@ def _stat_operation(
...
+
         if not numeric_only:
-            try:
-                self._validate_dtypes(numeric_only=True)
-            except TypeError:
-                if numeric_only is not None:
-                    raise
-            else:
-                numeric_only = False
+            # fix for 'test_reduce_specific'
```
YarShev, line 1915: redundant, should be removed before merging. look for other similar places

modin/pandas/base.py
```diff
@@ -3602,30 +3415,6 @@ def __array__(self, dtype=None):
         arr = self.to_numpy(dtype)
         return arr
 
-    def __array_wrap__(self, result, context=None):
```
YarShev, line 3605: why removed?

modin/pandas/dataframe.py
```diff
@@ -112,6 +112,8 @@ class DataFrame(BasePandasDataset):
     ``pd.read_csv``).
     """
 
+    _AXIS_ORDERS = ["index", "columns"]
+    _AXIS_LEN = len(_AXIS_ORDERS)
```
YarShev, line 116:
```suggestion
    _AXES = ["index", "columns"]
    _NDIM = len(_AXIS_ORDERS)
```

Why do we need these new attributes?

modin/pandas/dataframe.py
```diff
@@ -1609,21 +1499,11 @@ def prod(
...
-            ):
-                raise ValueError("level > 0 or level < -1 only valid with MultiIndex")
-            return self.groupby(level=level, axis=axis, sort=False).prod(
-                numeric_only=numeric_only, min_count=min_count
-            )
 
         axis_to_apply = self.columns if axis else self.index
         if (
             skipna is not False
+            # potential place to remove
```
YarShev, line 1506: Elaborate please or remove

modin/pandas/test/dataframe/test_default.py
```diff
@@ -69,10 +69,9 @@
         ("from_records", lambda df: {"data": to_pandas(df)}),
         ("hist", lambda df: {"column": "int_col"}),
         ("interpolate", None),
-        ("lookup", lambda df: {"row_labels": [0], "col_labels": ["int_col"]}),
         ("mask", lambda df: {"cond": df != 0}),
         ("pct_change", None),
-        ("to_xarray", None),
+        # ("to_xarray", None),
```
YarShev, line 74: Why commented?

asv_bench/benchmarks/benchmarks.py
```diff
@@ -488,7 +471,7 @@ def setup(self, shape, axis, drop_ncols):
         self.labels = self.df.axes[axis][:drop_count]
 
     def time_drop(self, shape, axis, drop_ncols):
-        execute(self.df.drop(self.labels, axis))
+        execute(self.df.drop(self.labels, axis=axis))
```
anmyachev, line 474: `drop` operation accepts `axis` parameter only as keyword

modin/pandas/base.py
anmyachev: This operation has been removed in https://pandas.pydata.org/docs/whatsnew/v2.0.0.html.

modin/pandas/test/dataframe/test_default.py
anmyachev: https://github.com/pydata/xarray/issues/7873

environment-dev.yml
```diff
@@ -12,7 +12,7 @@ dependencies:
   - dask>=2.22.0
   - distributed>=2.22.0
   - fsspec
-  - xarray
+  # - xarray
```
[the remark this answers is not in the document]
YarShev: Should we remove this at all then if CI passes?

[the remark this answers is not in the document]
anmyachev: I wouldn't want to do this, as soon we will have to switch to python 3.9 anyway. It looks like pandas 2.0.1 is the last one to support python 3.8.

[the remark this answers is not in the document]
YarShev: I don't get why we should keep `xarray` commented out. If Modin or its dependencies doesn't required `xarray` to be installed, why we should keep it in env files?

Also, we should probably move to testing on python 3.9 as part of a separate issue.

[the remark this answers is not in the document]
anmyachev: > I don't get why we should keep xarray commented out. If Modin or its dependencies doesn't required xarray to be installed, why we should keep it in env files?

We still have tests (for `to_xarray`) that are temporarily skipped if those tests are run under python 3.8.

modin/pandas/series.py
```diff
@@ -2077,8 +1921,8 @@ def value_counts(
             ascending=ascending,
             dropna=dropna,
         )
-        # pandas sets output index names to None because the Series name already contains it
-        counted_values._query_compiler.set_index_name(None)
+        # https://pandas.pydata.org/pandas-docs/version/2.0/whatsnew/v2.0.0.html#value-counts-sets-the-resulting-name-to-count
+        counted_values.name = "proportion" if normalize else "count"
```
dchigarev, line 1925: we already did this at the base implementation
```suggestion
```

modin/core/io/column_stores/parquet_dispatcher.py
```diff
@@ -614,7 +614,10 @@ def _read(cls, path, engine, columns, **kwargs):
         ParquetFile API is used. Please refer to the documentation here
         https://arrow.apache.org/docs/python/parquet.html
         """
-        if any(arg not in ("storage_options", "use_nullable_dtypes") for arg in kwargs):
+        if any(
+            arg not in ("storage_options", "use_nullable_dtypes", "dtype_backend")
```
dchigarev, line 618: don't we want to default to pandas for every `.read_*` method if `dtype_backend` was specified?

dchigarev (commented):
great work! left some minor comments

modin/core/io/io.py
anmyachev: Fixed

modin/core/io/text/text_file_dispatcher.py
anmyachev: Removed

modin/core/storage_formats/pandas/query_compiler.py
anmyachev: https://github.com/modin-project/modin/issues/6216

modin/pandas/base.py
anmyachev: removed

modin/pandas/dataframe.py
```diff
@@ -353,20 +355,20 @@ def shape(self):  # noqa: RT01, D200
...
         return len(self.index), len(self.columns)
 
-    def add_prefix(self, prefix):  # noqa: PR01, RT01, D200
+    def add_prefix(self, prefix, axis=None):  # noqa: PR01, RT01, D200
         """
         Prefix labels with string `prefix`.
         """
         return self.__constructor__(
-            query_compiler=self._query_compiler.add_prefix(prefix)
+            query_compiler=self._query_compiler.add_prefix(prefix, axis or 1)
```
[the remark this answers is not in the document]
anmyachev: fixed

modin/pandas/dataframe.py
```diff
@@ -353,20 +355,20 @@ def shape(self):  # noqa: RT01, D200
...
         )
 
-    def add_suffix(self, suffix):  # noqa: PR01, RT01, D200
+    def add_suffix(self, suffix, axis=None):  # noqa: PR01, RT01, D200
         """
         Suffix labels with string `suffix`.
         """
         return self.__constructor__(
-            query_compiler=self._query_compiler.add_suffix(suffix)
+            query_compiler=self._query_compiler.add_suffix(suffix, axis or 1)
```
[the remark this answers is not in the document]
anmyachev: fixed

modin/pandas/groupby.py
```diff
@@ -196,39 +197,27 @@ def default_handler(*args, **kwargs):
...
+        # default behaviour for aggregations; for the reference see
+        # `_op_via_apply` func in pandas==2.0.1
+        if axis is None or axis is no_default:
+            axis = self._axis
+
+        # `groupby_skew` can't handle `axis`, `skipna` parameters
+        # that should be added into `agg_kwargs`;
+        # looks like an implicit supported combination of parameters in the
+        # previous implementation: axis == 1, skipna==True
+        if axis != 1 or not skipna:
```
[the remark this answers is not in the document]
anmyachev: It was a bug, actually `axis=0` was supported.

modin/core/io/column_stores/parquet_dispatcher.py
anmyachev: Not every function needs it. I'll go through all the functions and check https://github.com/modin-project/modin/pull/5995#issuecomment-1573727109.

anmyachev:
@YarShev @dchigarev ready for review

anmyachev:
How to enable this option on workers https://pandas.pydata.org/docs/whatsnew/v2.0.0.html#copy-on-write-improvements?

Maybe to use `ray.worker.global_worker.run_function_on_all_workers`? https://github.com/modin-project/modin/pull/4603/files

modin/core/storage_formats/pandas/query_compiler.py
```diff
@@ -2008,33 +1999,15 @@ def last_valid_index_builder(df):
...
-                percentiles, datetime_is_numeric=datetime_is_numeric, include="all"
-            )
+            .describe(percentiles, include="all")
         )
         new_index = empty_df.index
 
-        # Note: `describe` convert timestamp type to object type
-        # which results in the loss of two values in index: `first` and `last`
-        # for empty DataFrame.
-        if not any(map(is_numeric_dtype, empty_df.dtypes)) and not datetime_is_numeric:
```
anmyachev, line 2029: "Removed argument datetime_is_numeric from [DataFrame.describe()](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.describe.html#pandas.DataFrame.describe) and [Series.describe()](https://pandas.pydata.org/docs/reference/api/pandas.Series.describe.html#pandas.Series.describe) as datetime data will always be summarized as numeric data ([GH34798](https://github.com/pandas-dev/pandas/issues/34798))"

In pandas 2.0 this code branch is not reachable.

anmyachev:
Disallow passing non-keyword arguments:
* reindex +
* sort_values +
* replace +
* bfill +
* ffill +
* clip +
* rename_axis +
* set_axis +
* where +
* to_stata +
* mask +
* read_* +
* pivot +
* concat +
* any +
* interpolate +
* sort_index +
* drop_duplicates +
* dropna +
* reset_index +
* set_index +
* split/rsplit +
* fillna +
* drop +
*

modin/pandas/test/test_groupby.py
```diff
@@ -2507,10 +2516,10 @@ def run_test(eval_function, *args, **kwargs):
     run_test(eval_any)
     run_test(eval_apply, func=lambda df: df.mean())
     run_test(eval_count)
-    run_test(eval_cummax)
-    run_test(eval_cummin)
-    run_test(eval_cumprod)
-    run_test(eval_cumsum)
+    run_test(eval_cummax, numeric_only=True)
```
anmyachev, line 2519: Behavior in pandas 1.5.3:
```bash
FutureWarning: Dropping invalid columns in DataFrameGroupBy.cummin is deprecated. In a future version, a TypeError will be raised. Before calling .cummin, select only columns which should be valid for the function.
```

modin/pandas/test/test_groupby.py
```diff
@@ -2355,7 +2384,7 @@ def test_groupby_sort(sort, is_categorical_by):
     pd_grp = pd_df.groupby("key_col", sort=sort)
 
     modin_groupby_equals_pandas(md_grp, pd_grp)
-    eval_general(md_grp, pd_grp, lambda grp: grp.sum())
+    eval_general(md_grp, pd_grp, lambda grp: grp.sum(numeric_only=True))
```
anmyachev, line 2387: Behavior in pandas 1.5.3:
```bash
FutureWarning: The default value of numeric_only in DataFrameGroupBy.sum is deprecated. In a future version, numeric_only will default to False. Either specify numeric_only or select only columns which should be valid for the function.
```

vnlitvinov:
> Maybe to use `ray.worker.global_worker.run_function_on_all_workers`

This doesn't exist 🙃 but you could modify the `modin-autoimport-pandas.pth` so that it would do what you want, like
```python
import os; os.environ.get("__MODIN_AUTOIMPORT_PANDAS__", None) and __import__("pandas").set_option("mode.copy_on_write", True)
```

modin/pandas/test/test_groupby.py
```diff
@@ -633,7 +664,8 @@ def test_single_group_row_groupby():
     eval_general(
         modin_groupby,
         pandas_groupby,
-        lambda df: df.pct_change(),
+        # AttributeError: 'DataFrameGroupBy' object has no attribute 'pad'
```
vnlitvinov, line 667: what does this comment mean?

modin/pandas/test/test_groupby.py
```diff
@@ -1145,12 +1152,19 @@ def eval_ngroups(modin_groupby, pandas_groupby):
...
 
 
-def eval_skew(modin_groupby, pandas_groupby):
-    modin_df_almost_equals_pandas(modin_groupby.skew(), pandas_groupby.skew())
+def eval_skew(modin_groupby, pandas_groupby, numeric_only=False):
+    kwargs = dict(numeric_only=numeric_only)
+    modin_df_almost_equals_pandas(
+        modin_groupby.skew(**kwargs),
+        pandas_groupby.skew(**kwargs),
+    )
```
vnlitvinov, line 1160:
```suggestion
    modin_df_almost_equals_pandas(
        modin_groupby.skew(numeric_only=numeric_only),
        pandas_groupby.skew(numeric_only=numeric_only),
    )
```
?

modin/pandas/test/test_groupby.py
```diff
@@ -1340,7 +1371,8 @@ def test(grp):
 
         return test
 
-    # issue-#3252
+    # issue-#3252, https://github.com/pandas-dev/pandas/issues/52760
+    """
```
vnlitvinov, line 1375: please do not disable code using block-comment, it isn't possible to see in diffs or when `git blame`-ing; comment each line if you don't want to remove the code

modin/pandas/test/test_io.py
```diff
@@ -835,20 +808,12 @@ def test_read_csv_quoting(
             )
 
     # Error Handling parameters tests
-    @pytest.mark.parametrize("warn_bad_lines", [True, False, None])
-    @pytest.mark.parametrize("error_bad_lines", [True, False, None])
+    @pytest.mark.skip
```
vnlitvinov, line 811: why do we skip this test?

modin/pandas/test/test_io.py
```diff
@@ -2694,12 +2783,6 @@ def test_to_pickle(self, tmp_path):
             fn="to_pickle",
             extension="pkl",
         )
-        unique_filename_modin = get_unique_filename(extension="pkl", data_dir=tmp_path)
-        unique_filename_pandas = get_unique_filename(extension="pkl", data_dir=tmp_path)
-        pd.to_pickle(modin_df, unique_filename_modin)
-        pandas.to_pickle(pandas_df, unique_filename_pandas)
-
-        assert assert_files_eq(unique_filename_modin, unique_filename_pandas)
```
vnlitvinov, line 2702: why did you remove this?

modin/pandas/test/test_series.py
```diff
@@ -2602,7 +2551,7 @@ def h(x):
 
     def g(x, arg1=0):
         for _ in range(arg1):
-            x = x.append(x)
+            x = (pd if isinstance(x, pd.Series) else pandas).concat((x, x))
```
vnlitvinov, line 2554: same suggestion on concat-ing a tuple of `arg1 + 1` elements

examples/tutorial/jupyter/execution/pandas_on_dask/test/test_notebooks.py
```diff
@@ -95,7 +95,8 @@ def sq_mad_func(self, axis=None, skipna=True, level=None, **kwargs):
     _replace_str(nb, "modin_mad_custom = ...", user_mad_implementation)
 
     nbformat.write(nb, modified_notebook_path)
-    _execute_notebook(modified_notebook_path)
+    # need to update example, `.mad` doesn't exist
```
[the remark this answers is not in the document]
anmyachev: https://github.com/modin-project/modin/issues/6235

modin/core/dataframe/base/dataframe/dataframe.py
```diff
@@ -252,7 +252,7 @@ def groupby(
         passed to the groupby may be at most the number of rows in the group, and
         may be as small as a single row.
 
-        Unlike the pandas API, an intermediate “GROUP BY” object is not present in this
+        Unlike the pandas API, an intermediate `GROUP BY` object is not present in this
```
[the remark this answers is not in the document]
anmyachev: On windows I see:
```bash
python scripts\doc_checker.py modin\core\dataframe\pandas\dataframe\dataframe.py
Output:
....
UnicodeDecodeError: 'charmap' codec can't decode byte 0x9d in position 137655: character maps to <undefined>
```

I suppose this may refer to my default encoding, but how important is it for us to have unicode characters here?

modin/core/storage_formats/base/query_compiler.py
```diff
@@ -5029,14 +5035,6 @@ def resample_asfreq(self, resample_kwargs, fill_value):
             self, resample_kwargs, fill_value
         )
 
-    # FIXME: `resample_backfill` is an alias for `resample_bfill`, one of these method
-    # should be removed (Modin issue #3107).
```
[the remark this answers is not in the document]
anmyachev: no, this issue has a wider range of tasks

modin/core/storage_formats/pandas/parsers.py
```diff
@@ -779,8 +781,10 @@ def parse(files_for_parser, engine, **kwargs):
         columns = kwargs.get("columns", None)
         storage_options = kwargs.get("storage_options", {})
         chunks = []
-        # `single_worker_read` just passes in a string path
-        if isinstance(files_for_parser, str):
+        # `single_worker_read` just passes in a string path or path-like object
+        if not isinstance(files_for_parser, collections.abc.Iterable) or isinstance(
```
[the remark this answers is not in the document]
anmyachev: now check for path-like type

modin/pandas/test/dataframe/test_map_metadata.py
```diff
@@ -830,6 +767,28 @@ def test_convert_dtypes_single_partition(
...
+        "b": pd.Series(["x", "y", "z"], dtype=np.dtype("O")),
+        "c": pd.Series([True, False, np.nan], dtype=np.dtype("O")),
+        "d": pd.Series(["h", "i", np.nan], dtype=np.dtype("O")),
+        "e": pd.Series([10, np.nan, 20], dtype=np.dtype("float")),
+        "f": pd.Series([np.nan, 100.5, 200], dtype=np.dtype("float")),
+    }
+
+    def comparator(df1, df2):
+        df_equals(df1, df2)
+        df_equals(df1.dtypes, df2.dtypes)
```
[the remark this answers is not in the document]
anmyachev: checks but does it with some assumptions, and here an exact match is needed, otherwise the difference in types is not caught

modin/pandas/test/dataframe/test_udf.py
```diff
@@ -390,7 +391,7 @@ def h(x):
 
     def g(x, arg1=0):
         for _ in range(arg1):
-            x = x.append(x)
+            x = (pd if isinstance(x, pd.DataFrame) else pandas).concat((x, x))
         return x
```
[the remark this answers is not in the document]
anmyachev: I'm not sure why, but your replacement is not equivalent to what it was before.

modin/pandas/test/test_api.py
```diff
@@ -237,7 +237,7 @@ def test_groupby_api_equality(obj):
     modin_dir = [x for x in dir(getattr(pd.groupby, obj)) if x[0] != "_"]
     pandas_dir = [x for x in dir(getattr(pandas.core.groupby, obj)) if x[0] != "_"]
     # This attribute is hidden from the DataFrameGroupBy object
-    ignore = ["keys"]
+    ignore = ["keys", "level"]
```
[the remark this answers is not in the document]
anmyachev: Pandas has logic that hides the public attributes of a class in its instance. Here we ignore them, so as not to implement the same logic in Modin.

.github/workflows/ci.yml
```diff
@@ -477,7 +477,8 @@ jobs:
       - run: python -m pytest modin/experimental/core/execution/native/implementations/hdk_on_native/test/test_utils.py
       - run: python -m pytest modin/pandas/test/test_io.py --verbose
       - run: python -m pytest modin/test/interchange/dataframe_protocol/test_general.py
-      - run: python -m pytest modin/test/interchange/dataframe_protocol/hdk
+      # TODO: uncomment after fix
+      # - run: python -m pytest modin/test/interchange/dataframe_protocol/hdk
```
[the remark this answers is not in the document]
anmyachev: done

modin/core/io/text/text_file_dispatcher.py
```diff
@@ -868,10 +868,14 @@ def _define_index(
             Partitions rows lengths.
         """
         index_objs = cls.materialize(index_ids)
-        if len(index_objs) == 0 or isinstance(index_objs[0], int):
+        if len(index_objs) == 0 or all((isinstance(obj, int) for obj in index_objs)):
```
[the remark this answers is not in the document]
anmyachev: reverted

modin/pandas/base.py
```diff
@@ -3139,7 +2949,11 @@ def to_excel(
             storage_options=storage_options,
         )
 
-    def to_dict(self, orient="dict", into=dict):  # pragma: no cover
+    def to_dict(self, orient="dict", into=dict, index=True):
+        if not index:
+            return self._default_to_pandas(
+                "to_dict", orient=orient, into=into, index=index
+            )
```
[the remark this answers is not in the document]
anmyachev: removed

modin/pandas/dataframe.py
```diff
@@ -1919,8 +1799,32 @@ def is_dtype_instance_mapper(column, dtype):
         ]
         return self.drop(columns=self.columns[indicate], inplace=False)
 
+    def __array_wrap__(self, result, context=None):
```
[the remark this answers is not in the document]
anmyachev: removed

modin/pandas/dataframe.py
```diff
@@ -167,9 +167,9 @@ def __init__(
             if columns is not None and not isinstance(columns, pandas.Index):
                 columns = pandas.Index(columns)
             if columns is not None:
-                self.set_axis(columns, axis=1, inplace=True)
+                self = self.set_axis(columns, axis=1, copy=False)
             if index is not None:
-                self.set_axis(index, axis=0, inplace=True)
+                self = self.set_axis(index, axis=0, copy=False)
```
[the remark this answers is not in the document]
anmyachev: changed

modin/pandas/test/test_series.py
anmyachev: I'm not sure, your replacement is not equivalent to what it was before.

modin/pandas/test/test_io.py
anmyachev: The reason is unknown, I can not reproduce locally. Let's fix it separately.

modin/pandas/test/test_groupby.py
anmyachev: removed

modin/pandas/test/test_groupby.py
```diff
@@ -418,7 +407,14 @@ def maybe_get_columns(df, by):
         lambda df: df.sem(),
         modin_df_almost_equals_pandas,
     )
-    eval_mean(modin_groupby, pandas_groupby)
+    # TypeError: 'Categorical' with dtype category does not support reduction 'mean'
```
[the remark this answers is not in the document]
anmyachev: removed

modin/pandas/test/test_groupby.py
```diff
@@ -167,19 +167,19 @@ def test_mixed_dtypes_groupby(as_index):
         )
         eval_general(modin_groupby, pandas_groupby, lambda df: df.ewm(com=0.5).std())
         eval_shift(modin_groupby, pandas_groupby)
-        eval_mean(modin_groupby, pandas_groupby)
+        eval_mean(modin_groupby, pandas_groupby, numeric_only=True)
```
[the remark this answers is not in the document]
anmyachev: No. I use this option explicitly to test the same as before. In the previous release of pandas, this parameter was determined automatically.

modin/core/storage_formats/base/query_compiler.py
```diff
@@ -3124,6 +3119,17 @@ def groupby_skew(
...
+            return GroupByDefault.register(pandas.core.groupby.DataFrameGroupBy.skew)(
+                self,
+                by=by,
+                axis=axis,
+                groupby_kwargs=groupby_kwargs,
+                agg_args=agg_args,
+                agg_kwargs=agg_kwargs,
+                drop=drop,
+            )
+        # ValueError: Operation skew does not support axis=1
```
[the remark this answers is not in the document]
anmyachev: made the comment more clear.

modin/core/storage_formats/pandas/query_compiler.py
```diff
@@ -3747,7 +3714,7 @@ def compute_groupby(df, drop=False, partition_idx=0):
 
         # that means that exception in `compute_groupby` was raised
         # in every partition, so we also should raise it
-        # TODO: we should be able to drop this logic with pandas 2.0 as it removes `numeric_only=None`
+        # TODO: we should be able to drop this logic with pandas 2.0.0 as it removes `numeric_only=None`
```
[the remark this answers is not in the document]
anmyachev: let's try

modin/core/storage_formats/pandas/query_compiler.py
anmyachev: > Which part of the code passes this argument to `map_fn` then?

When calling `mean` function through `**kwargs`.

anmyachev:
@vnlitvinov ready for review

modin/pandas/dataframe.py
```diff
@@ -167,9 +167,11 @@ def __init__(
             if columns is not None and not isinstance(columns, pandas.Index):
                 columns = pandas.Index(columns)
             if columns is not None:
-                self = self.set_axis(columns, axis=1, copy=False)
+                obj_with_new_columns = self.set_axis(columns, axis=1, copy=False)
+                self._query_compiler = obj_with_new_columns._query_compiler
```
vnlitvinov, line 171: We have a method `._update_inplace()` which is aimed at updating in-place, I would think it makes sense to use that (as it's doing a little more than just updating a QC)

modin/core/dataframe/base/dataframe/dataframe.py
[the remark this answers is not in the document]
vnlitvinov: In that case, I would suggest using double quotes as a symbol (`"`), instead of unicode-double-quotes.
Apostrophe is reserved for referencing arguments of a class or function in our doc style AFAIR.

BTW, thanks for digging that failure, it bothered me a while back (though stopped in my new OS install, no idea why).

modin/pandas/test/dataframe/test_udf.py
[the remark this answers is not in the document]
vnlitvinov: Ah, I see, it was squaring `x` in each loop, so my multiplication might be wrong... whatever, we can just keep this code, it's for tests anyway

modin/pandas/test/test_api.py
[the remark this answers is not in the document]
vnlitvinov: So Modin is going to have some public attributes on instance where pandas would not? This feels like worth mentioning somewhere, maybe in the docs...

modin/pandas/test/test_groupby.py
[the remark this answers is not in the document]
vnlitvinov: Makes sense. However I think we should also test for `numeric_only=False` (most likely in a separate issue, as I would presume it would incur some bugfixing - I would be very surprised if we don't have any bugs on an area we don't test!)

modin/pandas/test/test_io.py
vnlitvinov: At least make an umbrella issue and reference it in all places where tests started to fail after migration to newer pandas so you had to disable them. Then we can split the issue by root causes and create new ones when something is known better, but I believe there should exist at least one tracker.

modin/pandas/dataframe.py
```diff
@@ -716,7 +648,7 @@ def compare(
         )
 
     def corr(
-        self, method="pearson", min_periods=1, numeric_only=no_default
+        self, method="pearson", min_periods=1, numeric_only=False
```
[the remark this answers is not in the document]
vnlitvinov: cc @dchigarev

vnlitvinov:
@anmyachev there are a few open questions left from previous review, but most of them are minor.
I think the only major blocker for me is my last comment, about updating QC in-place.

modin/pandas/dataframe.py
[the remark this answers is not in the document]
dchigarev: > I would rather we add a `FIXME` comment in here which you could drop in your upcoming PR...

agree

modin/pandas/test/test_groupby.py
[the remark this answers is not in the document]
anmyachev: https://github.com/modin-project/modin/issues/6241

modin/test/interchange/dataframe_protocol/hdk/test_protocol.py
```diff
@@ -30,6 +30,9 @@
 from .utils import get_data_of_all_types, split_df_into_chunks, export_frame
 
 
+@pytest.mark.xfail(
+    reason="conversion from 'pyarrow' ends up with the wrong datetime64 resolution"
+)
```
dchigarev, line 35: there's an `exclude_dtypes` parameter in `get_data_of_all_types()`, can we simply exclude datetimes instead of disabling all these tests?

modin/pandas/test/dataframe/test_udf.py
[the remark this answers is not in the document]
anmyachev: ok, let's keep

dchigarev (dismissed):
looks good from my side

thank you for this massive work!

modin/pandas/test/test_general.py
```diff
@@ -543,6 +543,18 @@ def test_pivot():
     with pytest.raises(ValueError):
         pd.pivot(test_df["bar"], index="foo", columns="bar", values="baz")
 
+    if get_current_execution() != "BaseOnPython" and StorageFormat.get() != "Hdk":
+        # Failed for some reason on 'BaseOnPython' and 'HDK'
```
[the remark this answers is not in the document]
anmyachev: created

modin/pandas/test/test_io.py
anmyachev: made an umbrella issue

anmyachev:
@dchigarev ready for review

dchigarev (dismissed):

environment-dev.yml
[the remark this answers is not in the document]
YarShev: At some point we were testing two python versions, 3.6 and 3.7, as I remember. Please open an issue where we could discuss our further testing.

YarShev (approved):

YarShev:
@vnlitvinov, any comments?

dchigarev (approved):

mvashishtha:
> @vnlitvinov, any comments?

@YarShev @dchigarev I see that @vnlitvinov comment about `_update_inplace` is still open

https://github.com/modin-project/modin/pull/5995/files/7edd9bdd8bcaa16466f666949d5314d460f610ea..6d7e8d260bf08a7bb304c258e902c237736aa043#r1219040039

anmyachev:
> > @vnlitvinov, any comments?
> 
> @YarShev @dchigarev I see that @vnlitvinov comment about `_update_inplace` is still open
> 
> https://github.com/modin-project/modin/pull/5995/files/7edd9bdd8bcaa16466f666949d5314d460f610ea..6d7e8d260bf08a7bb304c258e902c237736aa043#r1219040039

@mvashishtha I fixed it but forgot to reply to the comment

modin/pandas/test/test_io.py
```diff
@@ -2152,16 +2229,45 @@ def test_read_sql(self, tmp_path, make_sql_connection, read_sql_engine):
...
 
+        old_sql_engine = ReadSqlEngine.get()
         ReadSqlEngine.put(read_sql_engine)
         if ReadSqlEngine.get() == "Connectorx":
             modin_df = pd.read_sql(sql=query, con=conn)
         else:
             modin_df = pd.read_sql(
                 sql=query, con=ModinDatabaseConnection("sqlalchemy", conn)
             )
+        ReadSqlEngine.put(old_sql_engine)
```
vnlitvinov, line 2240: this should be done in `try..finally` to be absolutely sure we return the value back... not critical in tests I suppose, so mostly a note for the future patterns

modin/pandas/test/test_io.py
```diff
@@ -2152,16 +2229,45 @@ def test_read_sql(self, tmp_path, make_sql_connection, read_sql_engine):
...
+    def test_read_sql_dtype_backend(self, tmp_path, make_sql_connection, dtype_backend):
+        filename = get_unique_filename(extension="db")
+
+        table = "test_read_sql_dtype_backend"
+        conn = make_sql_connection(tmp_path / filename, table)
+        query = f"select * from {table}"
+
+        def comparator(df1, df2):
+            df_equals(df1, df2)
+            df_equals(df1.dtypes, df2.dtypes)
```
vnlitvinov, line 2260: feels like it should be extracted to a common place, this comparator is written in a few places already; could go in some tests refactoring PR I suppose

vnlitvinov:
@anmyachev thanks for the tremendous amount of effort into making this PR by the way!

PR #6045, opened 2023-04-25
REFACTOR-#6044: remove code duplication for `get_objects_from_partitions`

### LINKED ISSUE #6044: CLN: remove code duplication for `get_objects_from_partitions`

anmyachev:
## What do these changes do?

- [x] Resolves #6044 
- [x] tests added and passing

### DIFF
--- a/modin/core/dataframe/pandas/partitioning/partition_manager.py
+++ b/modin/core/dataframe/pandas/partitioning/partition_manager.py
@@ -844,7 +844,7 @@ class PandasDataframePartitionManager(ClassLogger, ABC):
     @classmethod
     def get_objects_from_partitions(cls, partitions):
         """
-        Get the objects wrapped by `partitions`.
+        Get the objects wrapped by `partitions` (in parallel if supported).
 
         Parameters
         ----------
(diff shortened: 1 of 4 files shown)


modin/core/dataframe/pandas/partitioning/partition_manager.py
```diff
@@ -855,12 +855,18 @@ def get_objects_from_partitions(cls, partitions):
         -------
         list
             The objects wrapped by `partitions`.
-
-        Notes
-        -----
-        This method should be implemented in a more efficient way for engines that support
-        getting objects in parallel.
         """
+        if hasattr(cls, "_execution_wrapper"):
```
YarShev, line 859:
```suggestion
        if hasattr(cls._partition_class, "execution_wrapper"):
```
Would this work? I am just thinking it would good for one class to keep the execution wrapper.

anmyachev: I don't really like that kind of thing. In the code below, we will materialize multiple partitions, and it will look a little strange that to materialize multiple partitions, we use the attribute of the partition, and not the manager, which is designed to work with multiple partitions.

On the other hand, all wrappers are implemented as mixins, so we can switch to inheritance.

YarShev: > On the other hand, all wrappers are implemented as mixins, so we can switch to inheritance.

We should probably give it a try.

YarShev: Should we wait for #6052 in order to avoid double refactoring commits?

anmyachev: I'd like to merge it without waiting. Making these changes is faster than having to wait for the CI for this PR to complete again.

YarShev (approved):

PR #6092, opened 2023-05-04
REFACTOR-#6091: Push more of memory_usage down to query compiler.

### LINKED ISSUE #6091: REFACTOR: Push more of memory_usage down to query compiler.
The current implementation creates a separate modin.pandas dataframe with the index memory usage and concatenates that to the memory usage without the index. This is convenient for partitioned dataframes as in modin on ray, because computing the extra index column in a distributed way is difficult. However, there are cases (e.g. when using a database for storage and execution) where creating a pandas dataframe and concatenating it may be expensive or impossible.

mvashishtha:
## What do these changes do?

- [x] Resolves #6091 
- [x] tests added and passing

### DIFF
--- a/modin/core/storage_formats/pandas/query_compiler.py
+++ b/modin/core/storage_formats/pandas/query_compiler.py
@@ -844,11 +844,30 @@ class PandasQueryCompiler(BaseQueryCompiler):
     prod = TreeReduce.register(pandas.DataFrame.prod)
     any = TreeReduce.register(pandas.DataFrame.any, pandas.DataFrame.any)
     all = TreeReduce.register(pandas.DataFrame.all, pandas.DataFrame.all)
-    memory_usage = TreeReduce.register(
+    # memory_usage adds an extra column for index usage, but we don't want to distribute
+    # the index memory usage calculation.
+    _memory_usage_without_index = TreeReduce.register(
         pandas.DataFrame.memory_usage,
         lambda x, *args, **kwargs: pandas.DataFrame.sum(x),
         axis=0,
     )
+
+    def memory_usage(self, **kwargs):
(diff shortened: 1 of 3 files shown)


RehanSD (approved):
LGTM!

modin/core/storage_formats/pandas/query_compiler.py
```diff
@@ -844,12 +844,31 @@ def is_series_like(self):
...
         pandas.DataFrame.memory_usage,
         lambda x, *args, **kwargs: pandas.DataFrame.sum(x),
         axis=0,
     )
 
+    def memory_usage(self, **kwargs):
+        index = kwargs.get("index", True)
+        deep = kwargs.get("deep", False)
+        usage_without_index = self._memory_usage_without_index(index=False, deep=deep)
+        return (
```
RehanSD, line 859: won't this error out if we do `df.memory_usage(index=True)`? since this result will be passed to base's memory usage, which will call `transpose`, and `reduce_dimension` on this, which will try to `reduce_dimension` on a dataframe with two rows - which is invalid?

RehanSD (commented):
Have a quick question. Besides that, lgtm!

modin/core/storage_formats/pandas/query_compiler.py
mvashishtha: There's a passing test case with `index=True` [here](https://github.com/modin-project/modin/blob/cf9e7fcc917c7acc156c9e18747fecfc2da57cdd/modin/pandas/test/dataframe/test_reduce.py#L264-L267).  Dataframe reduce dimension will convert the result the a series, which is correct whether or not `index` is `True`.

PR #6137, opened 2023-05-15
FIX-#6107: Allow pass through of `tz_convert` and `tz_localize` to QC if possible

### LINKED ISSUE #6107: REFACTOR: pass tz_convert and tz_localize directly to query compiler
this is porting commit https://github.com/ponder-org/modin-public/commit/844d6a146fbe8d952f20bbea6725005f4f00b791

modin API layer shouldn’t assume that it’s cheap to create pandas.Series or to do operations on the index. that should be a choice of the particular engine.

RehanSD:
## What do these changes do?

- [ ] Resolves #6107 ? 
- [ ] tests added and passing

### DIFF
--- a/modin/pandas/base.py
+++ b/modin/pandas/base.py
@@ -3455,15 +3455,20 @@ class BasePandasDataset(ClassLogger):
         """
         Convert tz-aware axis to target time zone.
         """
-        axis = self._get_axis_number(axis)
-        if level is not None:
-            new_labels = (
-                pandas.Series(index=self.axes[axis]).tz_convert(tz, level=level).index
+        if hasattr(self._query_compiler, "tz_convert"):
+            return self.__constructor__(
+                query_compiler=self._query_compiler.tz_convert(tz, axis, level, copy)
             )
         else:
-            new_labels = self.axes[axis].tz_convert(tz)
-        obj = self.copy() if copy else self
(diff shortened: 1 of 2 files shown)


modin/pandas/base.py
```diff
@@ -3455,36 +3455,50 @@ def tz_convert(self, tz, axis=0, level=None, copy=True):  # noqa: PR01, RT01, D2
         """
         Convert tz-aware axis to target time zone.
         """
-        axis = self._get_axis_number(axis)
-        if level is not None:
-            new_labels = (
-                pandas.Series(index=self.axes[axis]).tz_convert(tz, level=level).index
+        if hasattr(self._query_compiler, "tz_convert"):
```
vnlitvinov, line 3458: this doesn't feel right... I would propose to require this method but default-to-pandas in normal implementation.
This way - allowing different QCs to have or not to have a method - opens a can of worms IMHO I would rather not let be opened...

vnlitvinov (commented):
Do we have tests to `tz_convert` and `tz_localize`?

modin/pandas/base.py
RehanSD: Makes sense - let me wait on #6145 for this PR then, since we'll need the base's `_create_or_update_inplace` for these

vnlitvinov (approved):
LGTM!

PR #6140, opened 2023-05-16
FEAT-#6139: 🔥 DataLoader interplay.

### LINKED ISSUE #6139: FEAT: add interplay with PyTorch DataLoader
Hey devs, thanks for the great work!

As an avid PyTorch user myself, is it possible you could make this work with pytorch's dataloaders? I tried this myself but do not know how to do this efficiently.

rentruewang:
Added wrappers for modin dataset s.t. it can be used with torch's DataLoader. Closes: #6139 

### Problem

Using a plain `torch`'s `Dataset` to wrap `modin`'s dataframe is..., well, slow. The reason for that is because of `modin`'s relative high latency, coupled with `torch`'s `DataLoader` always extracting element one by one, making the latency very noticeable compared to say, batch access.

### Solution

Custom `ModinDataLoader` that circumvents the `DataFrame -> Dataset -> DataLoader` pattern, s.t. `DataLoader` knows to extract one batch at a time, significantly speeding up the dataloader.
It also works with `torch`'s `Sampler`s, and aims to be a drop in replacement for `DataLoader` that directly wraps a `modin` dataframe.

- [x] Resolves #6139? 
- [x] tests added and passing


modin/experimental/torch/datasets.py
```diff
@@ -0,0 +1,108 @@
...
+
+from modin.pandas import DataFrame as ModinDataFrame
+from modin.pandas import Series as ModinSeries
+
+
+class ModinIterableDataset(IterableDataset):
+    """
+    PandasIterableDataset is responsible for converting a pandas/modin dataframe into a torch-compatible dataset.
+
+    The optional with_index option determines if index would be part of the output. Default is no.
```
noloerino, line 18: No need to include this line if you're mentioning it in the __init__ docstring as well.

```suggestion
```

modin/experimental/torch/datasets.py
```diff
@@ -0,0 +1,108 @@
...
+    ) -> None:
+        """
+        Parameters
+        ----------
+
+        df: DataFrame
+            Anything that looks like a pandas dataframe. The data inside would be served row by row.
+
+        with_index: bool = False
+            Whether to include the index object with the result (like iterrows). Default is no.
```
noloerino, line 34: Formatting for consistency with the rest of code; check also what `scripts/doc_checker.py` reports for your changes.
```suggestion
        df : DataFrame | ModinDataFrame
            DataFrame-like object from which the iterable will draw data.

        with_index : bool, default: False
            If true, include the index object with the resulting iterable, similar to `DataFrame.iterrows`.
```

modin/experimental/torch/datasets.py
```diff
@@ -0,0 +1,108 @@
...
+    batch_size: int = 1
+        Batch size that dataloader uses.
+
+    with_index: bool = False
+        Whether to include the index object with the result (like iterrows). Default is no.
+
+    Returns
+    -------
+
+    The desired dataloader.
```
noloerino, line 104:
```suggestion
    batch_size : int, default: 1
        Batch size that dataloader uses.

    with_index : bool, default: False
	If true, include the index object with the resulting iterable, similar to `DataFrame.iterrows`.

    Returns
    -------
    DataLoader
        DataLoader object backed by desired data.
```

noloerino (changes requested):
Nice work. A couple of style comments; make sure you run `python3 scripts/doc_checker.py` on your changes. I'll let someone more knowledgeable with pytorch comment on the substance.

modin/experimental/torch/datasets.py
```diff
@@ -0,0 +1,104 @@
...
+            yield from self._multithread_iter()
+
+    def _iter(self):
+        """
+        Single threaded iterator.
+        """
+        for idx, row in self._df.iterrows():
+            yield self._proc_row(idx, row)
+
+    def _proc_row(self, index: Hashable, row: Series | ModinSeries):
```
vnlitvinov, line 60: I wonder why do you want a separate function

modin/experimental/torch/datasets.py
```diff
@@ -0,0 +1,104 @@
...
+        """
+        Convert row to ndarray.
+        """
+
+        np_row = np.array(row)
+
+        if self._with_index:
+            return index, np_row
+        else:
+            return np_row
```
vnlitvinov, line 70:
```suggestion
        return (index, np_row) if self._with_index else np_row
```
? (doing the same thing, but a little denser)

vnlitvinov (commented):
Please fix CI failures (for now they seem to be linting-related).

modin/experimental/torch/datasets.py
rentruewang: It's because I personally dislike using multiple yield statements when there are loops. NBD, changed it.

anmyachev:
@vnlitvinov looks like CI is fixed

modin/experimental/torch/datasets.py
```diff
@@ -0,0 +1,94 @@
...
+from pandas import DataFrame
+from torch.utils.data import SequentialSampler
+from modin.pandas import DataFrame as ModinDataFrame
+from abc import abstractmethod
+
+
+class Sampler(Protocol):
+    @abstractmethod
+    def __iter__(self) -> Iterable[int]:
+        ...
```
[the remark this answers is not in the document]
rentruewang: I know.

modin/experimental/torch/datasets.py
```diff
@@ -0,0 +1,94 @@
...
+            idx_buffer.append(idx)
+
+            if self._end_of_batch(cnt):
+                yield self._df.iloc[idx_buffer].to_numpy()
+                idx_buffer = []
+
+    def _end_of_batch(self, counter: int):
+        return (
+            False
+            or counter % self._batch_size == self._batch_size - 1
```
vnlitvinov, line 89:
```suggestion
            counter % self._batch_size == self._batch_size - 1
```

modin/experimental/torch/datasets.py
```diff
@@ -0,0 +1,94 @@
...
+            or counter == len(self._sampler) - 1
+        )
+
+
+PandasDataLoader = ModinDataLoader
```
vnlitvinov, line 94: uh? pls add a code comment if this is really necessary...

modin/experimental/torch/tests/test_dataloader.py
```diff
@@ -0,0 +1,74 @@
...
+# ANY KIND, either express or implied. See the License for the specific language
+# governing permissions and limitations under the License.
+from __future__ import annotations
+
+from types import ModuleType
+
+import pandas as pd
+from torch.utils.data import RandomSampler
+
+from modin import pandas as mpd
```
vnlitvinov, line 20: we write the tests using `import modin.pandas as pd` form, like we say in README for the users; could you please also use that form of import so everything is uniform in the codebase?

modin/experimental/torch/tests/test_dataloader.py
```diff
@@ -0,0 +1,74 @@
...
+from modin.experimental.torch.datasets import ModinDataLoader
+
+
+def _load_test_dataframe(pandas: ModuleType):
```
vnlitvinov, line 24: I would suggest naming the argument as `lib` or `library`, as we do `import pandas` to use it in our codebase all around, and it looks confusing here

vnlitvinov (changes requested):

vnlitvinov:
@rentruewang please respond to unresolved comments

rentruewang:
@vnlitvinov Sorry I think I missed the notif. Should be good now.

modin/experimental/torch/tests/test_dataloader.py
```diff
@@ -41,34 +55,24 @@ def _test_torch_dataloader(pandas: ModuleType):
...
-            "POPULATION",
-            "PRICE",
-        ],
-        sampler=RandomSampler,
-    )
-    for batch in loader:
-        assert batch.shape[0] <= 16, batch.shape
-        assert batch.shape[1] == 6, batch.shape
+@pytest.mark.parametrize("sampler_cls", [RandomSampler, SequentialSampler])
+def compare_dataloaders(sampler_cls: Type[Sampler]):
```
vnlitvinov, line 72: Is this even being called? I think pytest won't consider this as a test because of its name not starting with `test_`...

vnlitvinov (commented):
Mostly good, but had a question on testing.

modin/experimental/torch/tests/test_dataloader.py
rentruewang: Oops, my bad.

vnlitvinov (approved):
LGTM, thanks @rentruewang!

modin/experimental/torch/tests/test_dataloader.py
```diff
@@ -0,0 +1,84 @@
...
+import ray
+import torch
+from torch.utils.data import RandomSampler, Sampler, SequentialSampler
+
+import modin.pandas as pd
+from modin.experimental.torch.datasets import ModinDataLoader
+
+
+@pytest.fixture(scope="module", autouse=True)
+def ray_fix():
```
YarShev, line 30: What is this needed for? I think we should add a comment for sure.

PR #6142, opened 2023-05-16
FEAT-#6141: ray.data.from_modin should work with non-partitioned storage format

### LINKED ISSUE #6141: FEAT: ray.data.from_modin should work with non-partitioned modin storage format
Right now it doesn't work because `unwrap_partitions` here:

https://github.com/ray-project/ray/blob/5a5155c7f49f720a2a548812057724a16d5d564f/python/ray/data/read_api.py#L1523-L1541

doesn't work when modin is not backed by partitions. Maybe this could be changed?

rentruewang:
I added a hack that would make `from_modin` work. I think a long term solution would be to change the unwrap_partition function completely / provide abstractions instead of relying on implementation details. Fixes #6141.

- [ ] Resolves #? 
- [ ] tests added and passing

### DIFF
--- a/modin/distributed/dataframe/pandas/partitions.py
+++ b/modin/distributed/dataframe/pandas/partitions.py
@@ -86,7 +86,7 @@ def _ray_from_modin_non_partition(api_layer_object: DataFrame | Series):
         """
         start_idx = part_idx * part_size
         end_idx = (part_idx + 1) * part_size
-        return pd.DataFrame(api_layer_object.iloc[start_idx:end_idx])
+        return pandas.DataFrame(api_layer_object.iloc[start_idx:end_idx])
 
     # Store things into ray.
     return [store_range.remote(api_layer_object, i) for i in range(num_parts)]
--- a/modin/distributed/dataframe/pandas/partitions.py
+++ b/modin/distributed/dataframe/pandas/partitions.py
@@ -24,11 +24,11 @@ from modin.core.storage_formats.pandas.query_compiler import PandasQueryCompiler
 from modin.pandas.dataframe import DataFrame, Series
 
 if TYPE_CHECKING:
-    from modin.core.execution.dask.implementations.pandas_on_dask.partitioning import (
-        PandasOnDaskDataframePartition,
-    )
     from modin.core.execution.ray.implementations.pandas_on_ray.partitioning import (
         PandasOnRayDataframePartition,
+    )
+    from modin.core.execution.dask.implementations.pandas_on_dask.partitioning import (
+        PandasOnDaskDataframePartition,
     )
     from modin.core.execution.unidist.implementations.pandas_on_unidist.partitioning.partition import (
         PandasOnUnidistDataframePartition,
@@ -80,6 +80,10 @@ def _ray_from_modin_non_partition(api_layer_object: DataFrame | Series):
 
     @remote
     def store_range(api_layer_object: DataFrame | Series, part_idx: int):
+        """
+        This function stores chunks of the `api_layer_object` into Ray's distributed object store.
+        Since ray automatically puts objects into shared memory, the "upload" is done implicitly.
+        """
         start_idx = part_idx * part_size
         end_idx = (part_idx + 1) * part_size
         return pd.DataFrame(api_layer_object.iloc[start_idx:end_idx])
--- a/modin/distributed/dataframe/pandas/partitions.py
+++ b/modin/distributed/dataframe/pandas/partitions.py
@@ -45,10 +45,12 @@ else:
     PartitionUnionType = Any
 
 
-def _ray_from_modin_hack(api_layer_object: DataFrame | Series):
+def _ray_from_modin_non_partition(api_layer_object: DataFrame | Series):
     """
     HACK:
-    This is a hack that enables support of soda within the ray.data.from_modin function here: https://docs.ray.io/en/latest/data/api/doc/ray.data.from_modin.html#ray-data-from-modin.
+    This is a hack that makes ray.data.from_modin work with modin dataframe not backed by partitions.
+
+    See also: https://docs.ray.io/en/latest/data/api/doc/ray.data.from_modin.html#ray-data-from-modin for ray.data.from_modin's documentation.
 
     Parameters
     ----------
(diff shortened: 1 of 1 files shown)


modin/distributed/dataframe/pandas/partitions.py
```diff
@@ -13,20 +13,23 @@
...
 if TYPE_CHECKING:
-    from modin.core.execution.ray.implementations.pandas_on_ray.partitioning import (
-        PandasOnRayDataframePartition,
-    )
     from modin.core.execution.dask.implementations.pandas_on_dask.partitioning import (
         PandasOnDaskDataframePartition,
     )
+    from modin.core.execution.ray.implementations.pandas_on_ray.partitioning import (
+        PandasOnRayDataframePartition,
+    )
```
vnlitvinov, line 32: why reorder?

modin/distributed/dataframe/pandas/partitions.py
```diff
@@ -42,6 +45,49 @@
...
+
+    # Calculate the actual number of partitions.
+    # This might be fewer than length / CPUs because there's a minimum chunk size.
+    num_parts = length // part_size
+
+    @remote
+    def store_range(api_layer_object: DataFrame | Series, part_idx: int):
+        start_idx = part_idx * part_size
+        end_idx = (part_idx + 1) * part_size
+        return pd.DataFrame(api_layer_object.iloc[start_idx:end_idx])
```
vnlitvinov, line 85: this feels super-weird - it's going to store the whole object in the store to only extract its piece remotely...

modin/distributed/dataframe/pandas/partitions.py
```diff
@@ -78,6 +124,13 @@ def unwrap_partitions(
         )
 
     modin_frame = api_layer_object._query_compiler._modin_frame
+
+    is_modin_distributed = hasattr(modin_frame, "_partition_mgr_cls")
+
+    if not is_modin_distributed and axis == 0:
+        # Directly uses the iloc API to extract data.
+        return _ray_from_modin_non_partition(api_layer_object=api_layer_object)
```
vnlitvinov, line 132: this feels wrong - what if user selected non-Ray engine?

rentruewang: This is to support ray.data.from_modin, like the code said, it's a hack

noloerino: Can you clarify what scenario this would happen in? Adding a short code example to your linked issue would be helpful.

Unless you're assuming that a user would want to call `ray.data.from_modin` when modin isn't using the Ray backend, you should add an extra if statement to check the [`modin.config.envvars.Engine`](https://github.com/modin-project/modin/blob/01278aeacd9e8943f00a59b2c8715a56d7ba033b/modin/config/envvars.py#L76) environment variable.

modin/distributed/dataframe/pandas/partitions.py
vnlitvinov: BTW @rentruewang did you test that this works on non-pandas storage format (like HDK or some SQL)?

modin/distributed/dataframe/pandas/partitions.py
vnlitvinov: oh, this thread was about me being stupid; disregard the comment, I misunderstood what code was doing, sorry about that 🙃

modin/distributed/dataframe/pandas/partitions.py
vnlitvinov: Could you please revert unnecessary changes then?

modin/distributed/dataframe/pandas/partitions.py
```diff
@@ -42,6 +45,49 @@
...
+    api_layer_object : DataFrame or Series
+        The API layer object (backed by datawarehouses).
+
+    Returns
+    -------
+
+    A list of Ray.ObjectRef that can be combined into the original ``api_layer_object``
+    """
+
+    import pandas as pd
```
RehanSD, line 67:
```suggestion
    import pandas
```

modin/distributed/dataframe/pandas/partitions.py
```diff
@@ -42,6 +45,49 @@
...
+    num_cpus = ray.cluster_resources()["CPU"]
+    length = len(api_layer_object)
+    part_size: int = compute_chunksize(length, num_cpus)  # type: ignore
+
+    # Calculate the actual number of partitions.
+    # This might be fewer than length / CPUs because there's a minimum chunk size.
+    num_parts = length // part_size
+
+    @remote
+    def store_range(api_layer_object: DataFrame | Series, part_idx: int):
```
RehanSD, line 82: Can we add a comment explaining what API Layer object is, and how this function is expected to work.

RehanSD (commented):
Left some comments that should resolve @vnlitvinov

modin/distributed/dataframe/pandas/partitions.py
rentruewang: Sorry that is not what this commit is for, it's for modin instances that do not have a partition manager.

modin/distributed/dataframe/pandas/partitions.py
```diff
@@ -78,6 +128,13 @@ def unwrap_partitions(
         )
 
     modin_frame = api_layer_object._query_compiler._modin_frame
+
+    is_modin_distributed = hasattr(modin_frame, "_partition_mgr_cls")
+
+    if not is_modin_distributed and axis == 0:
+        # Directly uses the iloc API to extract data.
+        return _ray_from_modin_non_partition(api_layer_object=api_layer_object)
```
vnlitvinov, line 136: Hmm, I think I have to ask again - what if user activated Dask as current engine? This is regarding the docstring of the outer function:
```
    Returns
    -------
    list
        A list of Ray.ObjectRef/Dask.Future to partitions of the ``api_layer_object``
        if Ray/Dask is used as an engine.
```
I wonder if we should augment the check on line 134 with `Engine.get() == 'Ray'`?..

anmyachev:
@rentruewang feel free to reopen or create a new PR when you have time to continue.

PR #6145, opened 2023-05-16
FEAT-#6144: Stop defaulting at API layer for a bunch of methods

### LINKED ISSUE #6144: Stop defaulting-to-pandas at frontend for a bunch of Series/DataFrame methods
Upstreaming the following commits:

- https://github.com/ponder-org/modin-public/commit/325be0ea84bcd48cc032f36b66cfc47a5309f33e
- https://github.com/ponder-org/modin-public/commit/f319c2933567c075cf38f16d6ee568c951115db3
- https://github.com/ponder-org/modin-public/commit/2c228f357b7505501e566408dfa8c7157cd61630
- https://github.com/ponder-org/modin-public/commit/6d25e7abb913f5498273256a27f2546205d62ccf

vnlitvinov:
## What do these changes do?

- [x] Resolves #6144
- [ ] tests added and passing

### DIFF
--- a/modin/core/storage_formats/base/query_compiler.py
+++ b/modin/core/storage_formats/base/query_compiler.py
@@ -1252,6 +1252,17 @@ class BaseQueryCompiler(ClassLogger, abc.ABC):
     def sum(self, **kwargs):  # noqa: PR02
         return DataFrameDefault.register(pandas.DataFrame.sum)(self, **kwargs)
 
+    @doc_utils.add_refer_to("DataFrame.mask")
+    def mask(self, cond, **kwargs):
+        squeeze_self = kwargs.pop("squeeze_self", False)
+
+        def mask(df, cond, **kwargs):
+            if squeeze_self:
+                df = df.squeeze(axis=1)
+            return df.mask(cond, **kwargs)
+
+        return DataFrameDefault.register(mask)(self, cond, **kwargs)
+
     @doc_utils.add_refer_to("to_datetime")
     def to_datetime(self, *args, **kwargs):
         """
--- a/modin/core/storage_formats/base/query_compiler.py
+++ b/modin/core/storage_formats/base/query_compiler.py
@@ -1495,6 +1495,32 @@ class BaseQueryCompiler(ClassLogger, abc.ABC):
         """
         return DataFrameDefault.register(pandas.DataFrame.replace)(self, **kwargs)
 
+    @doc_utils.add_refer_to("Series.argsort")
+    def argsort(self, **kwargs):  # noqa: PR02
+        """
+        Return the integer indices that would sort the Series values.
+
+        Override ndarray.argsort. Argsorts the value, omitting NA/null values,
+        and places the result in the same locations as the non-NA values.
+
+        Parameters
+        ----------
+        axis : {0 or 'index'}
+            Unused. Parameter needed for compatibility with DataFrame.
+        kind : {'mergesort', 'quicksort', 'heapsort', 'stable'}, default 'quicksort'
+            Choice of sorting algorithm. See :func:`numpy.sort` for more
+            information. 'mergesort' and 'stable' are the only stable algorithms.
+        order : None
+            Has no effect but is accepted for compatibility with numpy.
+
+        Returns
+        -------
+        BaseQueryCompiler
+            One-column QueryCompiler with positions of values within the
+            sort order with -1 indicating nan values.
+        """
+        return SeriesDefault.register(pandas.Series.argsort)(self, **kwargs)
+
     @doc_utils.add_one_column_warning
     # FIXME: adding refer-to note will create two instances of the "Notes" section,
     # this breaks numpydoc style rules and also crashes the doc-style checker script.
--- a/modin/core/storage_formats/base/query_compiler.py
+++ b/modin/core/storage_formats/base/query_compiler.py
@@ -448,6 +448,12 @@ class BaseQueryCompiler(ClassLogger, abc.ABC):
     # such that columns/rows that don't have an index on the other DataFrame
     # result in NaN values.
 
+    @doc_utils.add_refer_to("DataFrame.align")
+    def align(self, other, **kwargs):
+        return DataFrameDefault.register(pandas.DataFrame.align)(
+            self, other=other, **kwargs
+        )
+
     @doc_utils.doc_binary_method(operation="addition", sign="+")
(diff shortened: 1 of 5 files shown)


modin/core/storage_formats/base/query_compiler.py
```diff
@@ -1246,6 +1284,28 @@ def prod(self, **kwargs):  # noqa: PR02
...
+    @doc_utils.add_refer_to("DataFrame.mask")
+    def mask(self, cond, **kwargs):
+        squeeze_self = kwargs.pop("squeeze_self", False)
+
+        def mask(df, cond, **kwargs):
+            if squeeze_self:
+                df = df.squeeze(axis=1)
+            return df.mask(cond, **kwargs)
+
+        return DataFrameDefault.register(mask)(self, cond, **kwargs)
```
RehanSD, line 1296: Why can't we pass in `pandas.DataFrame.mask` here?

modin/core/storage_formats/base/query_compiler.py
```diff
@@ -1246,6 +1284,28 @@ def prod(self, **kwargs):  # noqa: PR02
...
+    @doc_utils.add_refer_to("DataFrame.pct_change")
+    def pct_change(self, **kwargs):
+        squeeze_self = kwargs.pop("squeeze_self", False)
+
+        def pct_change(df, **kwargs):
+            if squeeze_self:
+                df = df.squeeze(axis=1)
+            return df.pct_change(**kwargs)
+
+        return DataFrameDefault.register(pct_change)(self, **kwargs)
```
RehanSD, line 1307: Ditto

modin/core/storage_formats/base/query_compiler.py
```diff
@@ -1449,6 +1513,32 @@ def replace(self, **kwargs):  # noqa: PR02
         """
         return DataFrameDefault.register(pandas.DataFrame.replace)(self, **kwargs)
 
+    @doc_utils.add_refer_to("Series.argsort")
+    def argsort(self, **kwargs):  # noqa: PR02
+        """
+        Return the integer indices that would sort the Series values.
```
RehanSD, line 1519: Minor nit but why have we added a docstring here?

modin/pandas/base.py
```diff
@@ -1875,15 +1896,18 @@ def mask(
         """
         Replace values where the condition is True.
         """
-        return self._default_to_pandas(
-            "mask",
-            cond,
-            other=other,
+        return self._create_or_update_from_compiler(
```
RehanSD, line 1899: Why are we using this instead of just the constructor? Or will this method just use the constructor if in place is false, and otherwise do an in place update?

modin/pandas/series.py
```diff
@@ -1376,14 +1341,28 @@ def nlargest(self, n=5, keep="first"):  # noqa: PR01, RT01, D200
         """
         Return the largest `n` elements.
         """
-        return self._default_to_pandas(pandas.Series.nlargest, n=n, keep=keep)
+        if len(self._query_compiler.columns) == 0:
+            raise NotImplementedError(
+                "Series.nlargest is not implemented for empty Series."
```
RehanSD, line 1346: Is this true for pandas as well?

RehanSD (commented):
Left a few quick questions but overall LGTM!

modin/core/storage_formats/base/query_compiler.py
vnlitvinov: Our doc checker insists on having docstrings, so we'll have to either silence it with pragmas or fill in at least some docstrings... and for methods which are a direct match for pandas I feel it useful to copy their docstrings for clarity.

modin/pandas/base.py
vnlitvinov: Indeed. It handles the `inplace` properly.

modin/pandas/series.py
vnlitvinov: Hmm, just checked, and in pandas it returns `Series([], dtype: float64)` thingy...

modin/core/storage_formats/base/query_compiler.py
vnlitvinov: We probably can, I think I've copy-pasted some of my other fix where the behaviour was different... I had already implemented the change locally before your comment, but haven't tested to it wasn't pushed. Should be fixed now.

modin/core/storage_formats/base/query_compiler.py
vnlitvinov: same

modin/pandas/base.py
```diff
@@ -1905,7 +1905,6 @@ def mask(
                 level=level,
                 errors=errors,
                 try_cast=try_cast,
-                squeeze_self=not self._is_dataframe,
```
RehanSD: Why was this argument removed?

modin/pandas/series.py
```diff
@@ -1258,8 +1258,10 @@ def mask(
         errors: IgnoreRaise | NoDefault = no_default,
         try_cast=no_default,
     ):  # noqa: PR01, RT01, D200
-        # This method exists purely because `errors` has different default value for Series :(
-        return super().mask(cond, other, inplace, axis, level, errors, try_cast)
+        # This method exists because `errors` has different default value for Series :(
+        return super(Series, self).mask(
+            cond, other, inplace, axis, level, "raise", try_cast
```
RehanSD, line 1263: Shouldn't we pass in `errors` here?

RehanSD (commented):
Left 2 quick questions - lgtm otherwise!

modin/pandas/series.py
RehanSD: Also seems this argument has been deprecated since 1.5? So we should be coercing types anyways?

modin/pandas/base.py
RehanSD: I'm assuming this is because `mask` and `pct_change` are equivalent in pandas whether applied to a Series or a 1 col DataFrame?

RehanSD (approved):
LGTM

modin/pandas/base.py
vnlitvinov: yeah, they seem to behave identically in pandas

modin/pandas/series.py
vnlitvinov: That deprecation is the reason why I'm not passing the `errors` through, as it is expected to always behave as if `raise` is given.

PR #6168, opened 2023-05-19
FIX-#6070: Simplify implementation of `shift`

### LINKED ISSUE #6070: REFACTOR: Fix `df.shift`
- https://github.com/ponder-org/modin-public/commit/73bd234fc89f89501928afc3267e6d3f83cb5977
- https://github.com/ponder-org/modin-public/commit/3d9516f804224c30cb8bb20f1559eceb051c3542

These should be upstreamed, despite the commit message.

RehanSD:
## What do these changes do?

- [ ] Resolves #6070 ? 
- [ ] tests added and passing

### DIFF
--- a/modin/pandas/base.py
+++ b/modin/pandas/base.py
@@ -2833,71 +2833,16 @@ class BasePandasDataset(ClassLogger):
         if periods == 0:
             # Check obvious case first
             return self.copy()
-
-        empty_frame = False
-        if axis == "index" or axis == 0:
-            if abs(periods) >= len(self.index):
-                fill_index = self.index
-                empty_frame = True
-            else:
-                fill_index = pandas.RangeIndex(start=0, stop=abs(periods), step=1)
-        else:
-            fill_index = self.index
-        from .dataframe import DataFrame
-
-        fill_columns = None
-        if isinstance(self, DataFrame):
-            if axis == "columns" or axis == 1:
-                if abs(periods) >= len(self.columns):
-                    fill_columns = self.columns
-                    empty_frame = True
-                else:
-                    fill_columns = pandas.RangeIndex(start=0, stop=abs(periods), step=1)
-            else:
-                fill_columns = self.columns
-
-        filled_df = (
-            self.__constructor__(index=fill_index, columns=fill_columns)
-            if isinstance(self, DataFrame)
-            else self.__constructor__(index=fill_index, name=self.name)
+        return (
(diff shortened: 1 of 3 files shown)


modin/pandas/base.py
```diff
@@ -2833,71 +2833,16 @@ def shift(
...
-            else:
-                fill_columns = self.columns
-
-        filled_df = (
-            self.__constructor__(index=fill_index, columns=fill_columns)
-            if isinstance(self, DataFrame)
-            else self.__constructor__(index=fill_index, name=self.name)
+        return (
+            self._create_or_update_from_compiler(
+                new_query_compiler=self._query_compiler.shift(
```
anmyachev, line 2838: @RehanSD looks like `_query_compiler` doesn't have `shift` method.

mvashishtha:
@RehanSD I think you should implement `shift` in the base and pandas query compilers. The base implementation should translate the existing API layer implementation.

devin-petersohn (commented):
I added `shift` to the `BaseQueryCompiler` as a `Default` method. Previously we were pretending to do distributed things here, but this line actually means that we are not really doing anything distributed: https://github.com/modin-project/modin/blob/492934c078a01ecf18e70c25fa7b1e2d8b8271a0/modin/pandas/base.py#L2860-L2864

RehanSD:
> I added `shift` to the `BaseQueryCompiler` as a `Default` method. Previously we were pretending to do distributed things here, but this line actually means that we are not really doing anything distributed:
> 
> https://github.com/modin-project/modin/blob/492934c078a01ecf18e70c25fa7b1e2d8b8271a0/modin/pandas/base.py#L2860-L2864

I'm not sure that `shift` is defaulting. Although we do call the constructor with a new index + new column, we also do operations on the new (empty) dataframe. Although the first fillna may default since the dataframe is empty, the remaining operations (e.g. the drop) will not be defaulted right?

devin-petersohn:
>I'm not sure that shift is defaulting. Although we do call the constructor with a new index + new column, we also do operations on the new (empty) dataframe. Although the first fillna may default since the dataframe is empty, the remaining operations (e.g. the drop) will not be defaulted right?

The dataframe will not be empty, it will be full of null values, so we are in effect defaulting to pandas on the constructor and then pretending to be distributed. I think with a query compiler method this will encourage a true distributed implementation.

mvashishtha (approved):

PR #6211, opened 2023-05-25
FIX-#3371: Remove pandas patch level pin

### LINKED ISSUE #3371: Do not pin pandas down to patch level
From https://github.com/conda-forge/modin-feedstock/pull/21:

> > @h-vetinari: Regarding the pandas pin, I understand that it is tempting to pin against the exact version being tested, but it is not scalable (what if another library pins to pandas 1.3.1, then skips to 1.3.3 in the next update?, etc). Pandas follows SemVer pretty strictly (so even minor-releases should stay fully backward compatible), and IMO pinning down to patch-level is an antipattern that should be removed upstream.
> 
> @devin-petersohn: Understandable, would you open an issue to discuss this on the Modin repo?

Xref #3369

mvashishtha:
## What do these changes do?

Remove pandas patch level pin  as we agreed [here](https://github.com/modin-project/modin/issues/3371#issuecomment-1317976944).

- [x] Resolves #3371 
- [x] tests added and passing

### DIFF
--- a/environment-dev.yml
+++ b/environment-dev.yml
@@ -2,7 +2,7 @@ name: modin
 channels:
   - conda-forge
 dependencies:
-  - pandas==2.0.2
+  - pandas>1,<2.1
   - numpy>=1.18.5
   - ray-default>=1.13.0
   - pyarrow
--- a/environment-dev.yml
+++ b/environment-dev.yml
@@ -2,7 +2,7 @@ name: modin
 channels:
   - conda-forge
 dependencies:
-  - pandas>1,<2.1
+  - pandas>=2,<2.1
   - numpy>=1.18.5
   - ray-default>=1.13.0
   - pyarrow
--- a/setup.py
+++ b/setup.py
@@ -47,7 +47,7 @@ setup(
     long_description=long_description,
     long_description_content_type="text/markdown",
     install_requires=[
-        "pandas==2.0.2",
+        "pandas>1,<2.1",
         "packaging",
         "numpy>=1.18.5",
         "fsspec",
--- a/setup.py
+++ b/setup.py
@@ -47,7 +47,7 @@ setup(
     long_description=long_description,
     long_description_content_type="text/markdown",
     install_requires=[
-        "pandas>1,<2.1",
+        "pandas>=2,<2.1",
         "packaging",
         "numpy>=1.18.5",
         "fsspec",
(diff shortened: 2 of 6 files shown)


environment-dev.yml
```diff
@@ -2,7 +2,7 @@ name: modin
 channels:
   - conda-forge
 dependencies:
-  - pandas==2.0.2
+  - pandas>1,<2.1
```
anmyachev, line 5:
```suggestion
  - pandas>=2,<2.1
```

anmyachev (changes requested):

setup.py
```diff
@@ -47,7 +47,7 @@ def make_distribution(self):
     long_description=long_description,
     long_description_content_type="text/markdown",
     install_requires=[
-        "pandas==2.0.2",
+        "pandas>1,<2.1",
```
YarShev, line 50: Imagine a case when the user has pandas=1.3.5 and python 3.7. Then he wants to install modin. However, Modin only supports python>=3.8. Maybe we should remove pandas patch level pin starting 2.0?

environment-dev.yml
mvashishtha: This was my mistake :) In the original issue, Devin had written ">1.4,<1.6", and that works well when there's no pandas 2.0, but we need >= to require >= pandas 2

YarShev (approved):

anmyachev (approved):

setup.py
mvashishtha: users on pandas 1.3.5 and python 3.7 can't use the latest version of modin anyway because modin now requires python >= 3.8. I think this is a separate issue. What do you suggest that we change? 

I suspect that even importing the latest modin code on python 3.7 would break because we make so many assumptions about what we can import from pandas. That was why backwards compatibility with python 3.6 was so much work. We should only consider backwards compatibility with python 3.7 if we know of a desire for it.

YarShev: The changes now are fine by me.

mvashishtha:
@YarShev @anmyachev ready for merge

vnlitvinov:
Do note that this change is _potentially_ opening a can of worms for us because we use some internal pandas things (be them functions, classes or variables), for which pandas is free to change or even remove those without breaking SemVer logic (as they're not public API).

But then, we can always pin version again if such thing happens...

PR #6228, opened 2023-05-31
FEAT-#6225: Allow set_index to take an object to be handled by the backend (a backend index, etc)

### LINKED ISSUE #6225: Allow set_index to take an object to be handled by the backend (a backend index, etc)
As discussed in #1780, the set_index API assumes that we either have a set of column names/objects and that the set of columns passed in matches the set of known columns. We then call set_index_from_columns to let the backend handle the operation. A problem occurs when we need to call set_index with a backend type which the modin layer does not know about:

```
df.set_index(MyBackendIndex)
```

`MyBackendIndex` is not a normal Pandas type, nor a type which modin knows about, nor an actual column in the dataframe. In this case we want the backend to be responsible for as much of the handling here as possible. We *also* want to retain the nice error handling already present in set_index to get for missing columns.

One solution I have is to get to see if keys is non-"primitive" and when it is considered missing from the columns we raising an error and throw it over the wall to the query_compiler. Another method might be to agree on an attribute or method set on each backend which can tell the dataframe set_index method to circumvent the column checks.

*dataframe.set_index* approach:
```
-        if missing:
+        # If the missing column is a "primitive", return the errors.
+        # Otherwise we let the query compiler figure out what to do with
+        # the keys
+        if missing and not hasattr(missing[0], "__dict__"):
+            # The keys are a primitive type
             raise KeyError(f"None of {missing} are in the columns")
```
 
https://github.com/modin-project/modin/issues/1780

jkew:
## What do these changes do?
As discussed in #1780, the set_index API assumes that we either have a set of column names/objects and that the set of columns passed in matches the set of known columns. We then call set_index_from_columns to let the backend handle the operation. A problem occurs when we need to call set_index with a backend type which the modin layer does not know about:

df.set_index(MyBackendIndex)
MyBackendIndex is not a normal Pandas type, nor a type which modin knows about, nor an actual column in the dataframe. In this case we want the backend to be responsible for as much of the handling here as possible. We also want to retain the nice error handling already present in set_index to get for missing columns.

One solution I have is to get to see if keys is non-"primitive" and when it is considered missing from the columns we raising an error and throw it over the wall to the query_compiler. Another method might be to agree on an attribute or method set on each backend which can tell the dataframe set_index method to circumvent the column checks.

- [x] Resolves #6225 
- [x] tests added and passing


modin/pandas/test/dataframe/test_indexing.py
```diff
@@ -1232,6 +1232,23 @@ def test_rename_bug():
...
+    df_years = pandas.DataFrame(data=data)
+    modin_df_years = modin_df_years.set_index("YEAR")
+    df_years = df_years.set_index("YEAR")
+    modin_datetime_index = pd.to_datetime(modin_df_years.index, format="%Y")
+    pandas_datetime_index = pandas.to_datetime(df_years.index)
+
+    modin_df_years.index = modin_datetime_index
+    df_years.index = pandas_datetime_index
+
+    modin_df_years.set_index(modin_datetime_index)
```
RehanSD, line 1247: I'm not sure how this test tests the new functionality?

RehanSD (approved):
LGTM, but I'm not sure that the test is necessary?

modin/pandas/test/dataframe/test_indexing.py
jkew: This is a similar test to the one in ponder and mostly this just makes sure basic things do not break at the modin layer. Suggestions are welcome on testing.

modin/pandas/test/dataframe/test_indexing.py
```diff
@@ -1232,6 +1232,23 @@ def test_rename_bug():
...
 
 
+def test_index_to_datetime_using_set_index():
+    data = {"YEAR": ["1992", "1993", "1994"], "ALIENS": [1, 99, 1]}
+    modin_df_years = pd.DataFrame(data=data)
+    df_years = pandas.DataFrame(data=data)
+    modin_df_years = modin_df_years.set_index("YEAR")
+    df_years = df_years.set_index("YEAR")
+    modin_datetime_index = pd.to_datetime(modin_df_years.index, format="%Y")
+    pandas_datetime_index = pandas.to_datetime(df_years.index)
```
mvashishtha, line 1242:
```suggestion
    pandas_datetime_index = pandas.to_datetime(df_years.index, format="%Y")
```

nit: being consistent with modin seems clearer

mvashishtha (commented):
@jkew LGTM except for a nit. It's good that you found #1780, which brings up the same issue [here](https://github.com/modin-project/modin/issues/1780#issuecomment-773593183).

RehanSD (approved):
LGTM!

PR #6262, opened 2023-06-13
FIX-#6232: support DataFrame.cov(numeric_only=False) without fallback to pandas

### LINKED ISSUE #6232: Do not fallback to pandas at front-end in case of `DataFrame.cov(numeric_only=False)`
Similar to https://github.com/modin-project/modin/issues/6215

anmyachev:
## What do these changes do?

- [x] Resolves #6232 
- [ ] tests added and passing

### DIFF
--- a/modin/pandas/test/dataframe/test_default.py
+++ b/modin/pandas/test/dataframe/test_default.py
@@ -358,6 +358,14 @@ def test_cov(min_periods, ddof):
     )
 
 
+@pytest.mark.parametrize("numeric_only", [True, False, None])
+def test_cov_numeric_only(numeric_only):
+    eval_general(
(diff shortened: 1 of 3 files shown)


modin/pandas/test/dataframe/test_default.py
```diff
@@ -358,6 +358,14 @@ def comparator1(df1, df2):
     )
 
 
+@pytest.mark.parametrize("numeric_only", [True, False, None])
```
YarShev, line 361: Can we just add this parameter to the test above?

anmyachev: We can, but then the parameter matrix will increase (although this is not necessary), and I would like not to do this.

YarShev (approved):

PR #6283, opened 2023-06-20
TEST-#6282: Reduce copy-pasteness in ci.yml

### LINKED ISSUE #6282: Refactor ci.yml to reduce the amount of copy-pasting

vnlitvinov:
## What do these changes do?

Reduce the amount of copy-pasteness in `ci.yml` to facilitate its further refactoring.

- [x] Resolves #6282 
- [x] tests added and passing


RehanSD (approved):
Had a couple of quick q's. but lgtm otherwise!

.github/actions/mamba-env/action.yml
```diff
@@ -0,0 +1,44 @@
...
+    description: "Conda environment yml"
+    required: true
+  activate-environment:
+    description: "Conda environment to activate"
+    default: "modin"
+
+runs:
+  using: "composite"
+  steps:
+    - name: Get current week
```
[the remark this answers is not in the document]
vnlitvinov: In a previous version, we didn't have a staling at all, and (usually) it didn't hurt. 🙃 
The probable reason is that we have so many different cached variants, that GH evicts some of our caches quite frequently.

So I would propose to keep this change as it is now and (maybe) address the eviction later when some statistics on it could be gathered.

.github/workflows/ci.yml
```diff
@@ -528,14 +307,15 @@ jobs:
           pip install git+https://github.com/airspeed-velocity/asv.git@ef016e233cb9a0b19d517135104f49e0a3c380e9
       - name: Running benchmarks
         run: |
-          # ASV correctly creates environments for testing only from the branch
```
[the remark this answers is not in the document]
vnlitvinov: We don't need to run `asv check` if there is nothing changed for us to benchmark, and skipping it saves a few minutes of total CI runtime.

See the pic from some CI run of another PR, where the section "Running benchmark" takes 2.5 minutes while here it takes 5 seconds.
image

setup.py
```diff
@@ -5,7 +5,8 @@
     long_description = fh.read()
 
 dask_deps = ["dask>=2.22.0", "distributed>=2.22.0"]
-ray_deps = ["ray[default]>=1.13.0", "pyarrow"]
+# ray==2.5.0 broken: https://github.com/conda-forge/ray-packages-feedstock/issues/100
+ray_deps = ["ray[default]>=1.13.0,!=2.5.0", "pyarrow"]
```
YarShev, line 9: We should probably update the min supported version of Ray for the config, as well as exclude the broken release there. Note that MIN_RAY_VERSION is still equal to 1.4.0.

https://github.com/modin-project/modin/blob/13fa0c47178a00f9314a62776396c84db394aae3/modin/config/envvars.py#L113

cc @vnlitvinov, @anmyachev

vnlitvinov: We shouldn't be excluding `2.5.0` in there (Ray itself is working, it's only `ray` command-line tool which is broken on Windows in that release), but should definitely bump the `MIN_RAY_VERSION` indeed.

YarShev: Why should we exclude Ray 2.5.0 here and in other env files?

vnlitvinov: For the consistency of what we test (right now we test 2.5.1 which has released the fix for the issue)

YarShev: I do not understand why we have to stop users from using Ray 2.5.0 with Modin if that Ray version has only ray command-line tool broken?

vnlitvinov: Because we do not actually know if it is working on Windows (because we cannot test this)...

YarShev: Then I do not understand why we do not want to exclude that release from the config.

YarShev: Also, are we excluding the release unconditionally here, but not depending on OSs?

anmyachev: I will update `MIN_RAY_VERSION`.

Issue #6287, opened 2023-06-23
BUG: Using vanilla NumPy methods on a Modin Dataframe fails


RehanSD:
### Modin version checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the latest released version of Modin.

- [X] I have confirmed this bug exists on the main branch of Modin. (In order to do this you can follow [this guide](https://modin.readthedocs.io/en/stable/getting_started/installation.html#installing-from-the-github-master-branch).)

### Reproducible Example

```python
# Modin
import modin.pandas as pd
import modin.numpy as np

# Vanilla 
import pandas as vpd
import numpy as vnp

cols = [f"feature_{i}" for i in range(10)]
# This is not enabled in modin.numpy hence using vanilla
npy_array = vnp.random.rand(50, 10)

vdf = vpd.DataFrame(npy_array, columns=cols)
df = pd.DataFrame(vdf)

# Works
tmp = vnp.where(
    df["feature_0"] + df["feature_1"] > df["feature_2"] + df["feature_3"], 1, 0
)

# Doesn't work
tmp = vnp.where(
    df["feature_0"] + df["feature_1"] > df["feature_2"] + df["feature_3"], 1, 0
)
# Doesn't work
tmp = np.where(
    df["feature_0"] + df["feature_1"] > df["feature_2"] + df["feature_3"], 1, 0
)
# Doesn't work
tmp = np.where(np.array(df["feature_0"]) + np.array(["feature_1"]) > np.array(df["feature_2"]) + np.array(df["feature_3"]), 1, 0)
```

### Issue Description

We need to fix some of the numpy hooks we provide.

### Expected Behavior

This should work - without ExperimentalNumPyAPI on, it should default to pandas, and with it on, it should work distributed.

### Error Logs

<details>

```python-traceback

Replace this line with the error backtrace (if applicable).

```

</details>

### Installed Versions

<details>

Replace this line with the output of pd.show_versions()

</details>

PR #6288, opened 2023-06-23
FIX-#6287: Fix handling of numpy array_function

RehanSD:
## What do these changes do?

- [ ] Resolves #? 
- [ ] tests added and passing

### DIFF
--- a/modin/numpy/arr.py
+++ b/modin/numpy/arr.py
@@ -437,11 +437,15 @@ class array(object):
             modin_func = getattr(shaping, func_name)
         elif hasattr(creation, func_name):
             modin_func = getattr(creation, func_name)
+        if func_name == "where":
+            return self.where(*args[1:])
         if modin_func is None:
             return NotImplemented
         return modin_func(*args, **kwargs)
 
     def where(self, x=None, y=None):
+        if x is None != y is None:
+            raise ValueError("either both or neither of x and y should be given")
         if not is_bool_dtype(self.dtype):
             raise NotImplementedError(
                 "Modin currently only supports where on condition arrays with boolean dtype."
--- a/modin/pandas/base.py
+++ b/modin/pandas/base.py
@@ -61,7 +61,7 @@ from modin.utils import try_cast_to_pandas, _inherit_docstrings
 from modin.error_message import ErrorMessage
 from modin import pandas as pd
 from modin.pandas.utils import is_scalar
-from modin.config import IsExperimental
+from modin.config import IsExperimental, ExperimentalNumPyAPI
 from modin.logging import disable_logging, ClassLogger
 
 # Similar to pandas, sentinel value to use as kwarg in place of None when None has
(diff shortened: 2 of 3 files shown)


anmyachev:
@RehanSD CI is broken

modin/pandas/base.py
```diff
@@ -3426,6 +3426,80 @@ def __and__(self, other):
     def __rand__(self, other):
         return self._binary_op("__rand__", other, axis=0)
 
+    def __array_function__(self, func, types, args, kwargs):
```
RehanSD, line 3429: both __array_ufunc__ and __array_function__ will not work if the user passes in a Modin NumPy Array for `out` if we end up defaulting to NumPy. In the Modin NumPy API, I've fixed it so that we do copy in the values to a Modin NumPy Array, but I'm not sure what the solution should be for this, and so have defaulted to having it be unsupported.

modin/numpy/arr.py
```diff
@@ -397,10 +397,30 @@ def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
                 if isinstance(input, pd.Series):
                     input = input._query_compiler.to_numpy().flatten()
                 args += [input]
+            out_kwarg = kwargs.get("out", None)
+            if out_kwarg is not None:
+                # If `out` is a modin.numpy.array, `kwargs.get("out")` returns a 1-tuple
+                # whose only element is that array, so we need to unwrap it from the tuple.
+                out_kwarg = out_kwarg[0]
+            kwargs.pop("out", None)
```
vnlitvinov, line 405: I think we should merge the `.get` and `.pop` like
```python
out_kwarg = kwargs.pop("out", None)
```

I see no point of keeping them separated

modin/numpy/arr.py
```diff
@@ -397,10 +397,30 @@ def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
...
+            where_kwarg = kwargs.get("where", None)
+            if where_kwarg is not None:
+                if isinstance(where_kwarg, type(self)):
+                    where_kwarg = where_kwarg._to_numpy()
+                kwargs["where"] = where_kwarg
```
vnlitvinov, line 410:
```suggestion
                if isinstance(where_kwarg, type(self)):
                    kwargs["where"] = where_kwarg._to_numpy()
```
no point of re-assigning stuff which didn't change 😺 

also, why do we want to cast `where` to plain numpy?

modin/pandas/base.py
```diff
@@ -3426,6 +3426,80 @@ def __and__(self, other):
...
+            converted_args = []
+            for input in args:
+                if hasattr(input, "_query_compiler"):
+                    input = input.__array__()
+                converted_args += [input]
+            where_kwarg = kwargs.get("where")
+            if where_kwarg is not None:
+                if hasattr(where_kwarg, "_query_compiler"):
+                    kwargs["where"] = where_kwarg.__array__()
+            return func(arr, *converted_args[1:], **kwargs)
```
vnlitvinov, line 3466: this seems to be losing `converteg_args[0]` (which was `args[0]`), feels weird

modin/pandas/base.py
vnlitvinov: I don't see where this "unsupported" is placed in code. I think it should be an exception or something similar, shouldn't it?

modin/pandas/base.py
RehanSD: I can add a comment - but basically args[0] is self. Since we’re calling the function directly instead of calling the special method again, I’m passing in the converted self as the first argument, and only need to pass in any remaining args (after making sure to convert them)

modin/numpy/arr.py
```diff
@@ -397,10 +397,29 @@ def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
                 if isinstance(input, pd.Series):
                     input = input._query_compiler.to_numpy().flatten()
                 args += [input]
+            out_kwarg = kwargs.get("out", None)
```
vnlitvinov, line 400:
```suggestion
            out_kwarg = kwargs.pop("out", None)
```

modin/numpy/arr.py
```diff
@@ -397,10 +397,29 @@ def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
...
+            if out_kwarg is not None:
+                # If `out` is a modin.numpy.array, `kwargs.get("out")` returns a 1-tuple
+                # whose only element is that array, so we need to unwrap it from the tuple.
+                out_kwarg = out_kwarg[0]
+            kwargs.pop("out", None)
```
vnlitvinov, line 405:
```suggestion
```

modin/pandas/base.py
vnlitvinov: bump

modin/pandas/base.py
vnlitvinov: a comment would be nice; also, if we're dropping `converted_args[0]`, then we should probably be doing so in the upper loop and skip the call of `args[0].__array__()` altogether

PR #6295, opened 2023-06-26
FIX-#6294: implement `__iter__` for `array` class that defaults to numpy

### LINKED ISSUE #6294: Implement `__iter__` for `array` class
Iteration over a distributed object works very slowly, it is better to default in numpy, similar to how it is done for a Series.

anmyachev:
## What do these changes do?

- [x] Resolves #6294 
- [ ] tests added and passing

### DIFF
--- a/modin/numpy/arr.py
+++ b/modin/numpy/arr.py
@@ -734,6 +734,16 @@ class array(object):
 
     absolute = __abs__
 
+    def __iter__(self):
+        """
+        Return an iterator of the values.
+
+        Returns
+        -------
+        iterable
+        """
+        return self._to_numpy().__iter__()
+
     def __invert__(self):
         """
         Apply bitwise inverse to each element of the `BasePandasDataset`.


RehanSD (changes requested):
Thank you @anmyachev - this fix looks good, but I think we want to defer the specific computation to the query compiler - that way we support other query compilers that may not want to move data out of distributed storage, or may incur costs doing so. Would you mind making the API Layer for the Array, Series, and DF route to a query_compiler method, and have the query_compiler method just return this? We could have it look something like this maybe?
```python
def __iter__(self, lib="pandas", squeeze=False):
     return self.to_numpy().__iter__() if lib != "pandas" else (self.to_pandas().squeeze() if squeeze else self.to_pandas().__iter__()
```

anmyachev:
> Thank you @anmyachev - this fix looks good, but I think we want to defer the specific computation to the query compiler - that way we support other query compilers that may not want to move data out of distributed storage, or may incur costs doing so. Would you mind making the API Layer for the Array, Series, and DF route to a query_compiler method, and have the query_compiler method just return this? We could have it look something like this maybe?
> 
> ```python
> def __iter__(self, lib="pandas", squeeze=False):
>      return self.to_numpy().__iter__() if lib != "pandas" else (self.to_pandas().squeeze() if squeeze else self.to_pandas()).__iter__()
> ```

ok, will do

anmyachev:
> ```python
> def __iter__(self, lib="pandas", squeeze=False):
>      return self.to_numpy().__iter__() if lib != "pandas" else (self.to_pandas().squeeze() if squeeze else self.to_pandas()).__iter__()
> ```

@RehanSD this will also require `Series._to_pandas` and `array._to_numpy` methods to be moved to the query compiler. `query_compiler.to_pandas` and `query_compiler.to_numpy` are not suitable. I would like to clarify whether you think it is necessary to continue moving methods in such a situation.

modin/numpy/arr.py
```diff
@@ -734,6 +734,16 @@ def __abs__(
...
 
+    def __iter__(self):
+        """
+        Return an iterator of the values.
+
+        Returns
+        -------
+        iterable
+        """
+        return self._to_numpy().__iter__()
```
anmyachev, line 745: @RehanSD I can implement in this way. This is more flexible, and there is no need to transfer the implementation of `_to_numpy` function to the query compiler. What do you think?

```suggestion
        if hasattr(self._query_compiler, "numpy__iter__"):
            return self._query_compiler.numpy__iter__()
        return self._to_numpy().__iter__()
```

vnlitvinov: I personally do not like the idea of such a dynamic interface... I would rather this `numpy__iter__()` would do this `._to_numpy().__iter__()` thing itself...

That said, why are we even wanting to default to pandas (or numpy) in such a simple case? Maybe just implement the iterator ourselves without even defaulting?

anmyachev: It probably makes sense to implement an iterator, but right now I don't have the ability to do it.

PR #6300, opened 2023-06-27
FIX-#6299: __array__ method always returns array of vanilla numpy

### LINKED ISSUE #6299: Should `__array__` method always return array of vanilla numpy? (instead of experimental Modin wrapper)
In some cases inside Modin, the series and dataframe objects are converted to a numpy array as follows: `numpy.array(value)`. However, the functions further do not expect that there can be an distributed object after this call.

@RehanSD what do you think about?

anmyachev:
## What do these changes do?

- [x] Resolves #6299 
- [x] tests added and passing

### DIFF
--- a/modin/pandas/base.py
+++ b/modin/pandas/base.py
@@ -3104,11 +3104,23 @@ class BasePandasDataset(ClassLogger):
             storage_options=storage_options,
         )
 
-    def to_numpy(
+    def _to_bare_numpy(
         self, dtype=None, copy=False, na_value=no_default
     ):  # noqa: PR01, RT01, D200
         """
         Convert the `BasePandasDataset` to a NumPy array.
+        """
+        return self._query_compiler.to_numpy(
+            dtype=dtype,
+            copy=copy,
+            na_value=na_value,
+        )
+
+    def to_numpy(
+        self, dtype=None, copy=False, na_value=no_default
+    ):  # noqa: PR01, RT01, D200
+        """
+        Convert the `BasePandasDataset` to a NumPy array or a Modin wrapper for NumPy array.
         """
         from modin.config import ExperimentalNumPyAPI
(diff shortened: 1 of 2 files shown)


modin/pandas/base.py
```diff
@@ -3440,7 +3440,16 @@ def __array__(self, dtype=None):
...
+        from modin.config import ExperimentalNumPyAPI
+
+        exp_numpy_api = ExperimentalNumPyAPI.get()
+        try:
+            if exp_numpy_api:
+                ExperimentalNumPyAPI.put(False)
+            arr = self.to_numpy(dtype)
+        finally:
+            if exp_numpy_api:
+                ExperimentalNumPyAPI.put(True)
```
vnlitvinov, line 3452: I don't like this juggling... I propose we introduce a non-standard `_to_bare_numpy()` (or something similar named) API method which would always return non-wrapped Numpy arrays instead of all this configuration trickery.

anmyachev: Some of Modin's code itself is tied to the fact that this function returns a regular numpy, and all third-party libraries that check if `__array__` method exists, without checking the type of the object.

I think that we have to enforce the contract of `__array__` function in some way.

P.S. `numpy.array(modin_array)` will failed if we don't make the changes.

vnlitvinov: How this relates to what I said? :) What I'm saying is that `__array__` implementation should be different, so that it doesn't set-unset the configuration variable.

See `modin/utils.py::to_numpy()` implementation.

anmyachev: misunderstood you

anmyachev:
@vnlitvinov ready for review

vnlitvinov (approved):
LGTM!

PR #6303, opened 2023-06-27
FEAT-#6296: Add pyhdk feature flags

### LINKED ISSUE #6296: Add support for pyhdk v0.6.1 flags
In pyhdk v0.6.1 we are introducing two new features:

* Ability to place logs in a specific directory (e.g. the temp directory, if log files are not required). This was a request from one of our benchmarking teams. 
* Ability to lazily materialize dictionary encoded string columns. This feature speeds up the import process, particularly if only a few text columns are processed later in a query. However, it makes benchmarking more difficult, and can slow down workflows with large numbers of queries on text columns, so we're making it configurable. 

We have also introduced a version API in HDK v0.5 which we will use to remove these flags from the command string for HDK versions older than 0.6.1. 

I plan to submit a PR with these new config variables + version processing shortly.

alexbaden:
## What do these changes do?

* Adds feature flag controlling lazy dictionary materialization. Lazy dict materialization delays building dictionaries for text types until query time. For datasets with large numbers of text columns that are not directly processed, this can speed up import and querying. 
* Adds feature flag to set the log file directory. Can be set to the tmp directory to suppress log artifacts. This was a request from a benchmarking team. 

These features are available in HDK main branch now and will be released this week. The version check prevents passing the flags to earlier versions of HDK which do not support them.

- [x] Resolves #6296 
- [ ] tests added and passing

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -13,6 +13,7 @@ import types
 
 import ray
 
+
 # stub ray.remote to be a no-op so it doesn't shadow docstrings
 def noop_decorator(*args, **kwargs):
     if len(args) == 1 and len(kwargs) == 0 and callable(args[0]):
@@ -43,6 +44,9 @@ if not hasattr(sys.modules["pyhdk.hdk"], "ExecutionResult"):
     sys.modules["pyhdk.hdk"].ExecutionResult = type("ExecutionResult", (object,), {})
 if not hasattr(sys.modules["pyhdk.hdk"], "RelAlgExecutor"):
     sys.modules["pyhdk.hdk"].RelAlgExecutor = type("RelAlgExecutor", (object,), {})
+if not hasattr(sys.modules["pyhdk"], "__version__"):
+    # Show all known pyhdk config options in documentation
+    sys.modules["pyhdk"].__version__ = "999"
 
 sys.path.insert(0, os.path.abspath(os.path.join(os.path.dirname(__file__), "..")))
 import modin
--- a/modin/config/envvars.py
+++ b/modin/config/envvars.py
@@ -517,15 +517,6 @@ class HdkLaunchParameters(EnvironmentVariable, type=dict):
     """
 
     varname = "MODIN_HDK_LAUNCH_PARAMETERS"
-    default = {
-        "enable_union": 1,
-        "enable_columnar_output": 1,
-        "enable_lazy_fetch": 0,
-        "null_div_by_zero": 1,
-        "enable_watchdog": 0,
-        "enable_thrift_logs": 0,
-        "cpu_only": 1,
-    }
 
     @classmethod
     def get(cls) -> dict:
@@ -558,11 +549,43 @@ class HdkLaunchParameters(EnvironmentVariable, type=dict):
             Decoded and verified config value.
         """
         custom_parameters = super().get()
-        result = cls.default.copy()
+        result = cls._get_default().copy()
         result.update(
             {key.replace("-", "_"): value for key, value in custom_parameters.items()}
         )
         return result
+
+    @classmethod
+    def _get_default(cls) -> Any:
+        """
+        Get default value of the config. Checks the pyhdk version and omits variables unsupported in prior versions.
+
+        Returns
+        -------
+        dict
+            Config keys and corresponding values.
+        """
+        if (default := getattr(cls, "default", None)) is None:
+            cls.default = default = {
+                "enable_union": 1,
+                "enable_columnar_output": 1,
+                "enable_lazy_fetch": 0,
+                "null_div_by_zero": 1,
+                "enable_watchdog": 0,
+                "enable_thrift_logs": 0,
+                "cpu_only": 1,
+            }
+
+            try:
+                import pyhdk
+
+                if version.parse(pyhdk.__version__) >= version.parse("0.6.1"):
+                    default["enable_lazy_dict_materialization"] = 0
+                    default["log_dir"] = "pyhdk_log"
+            except ImportError:
+                # if pyhdk is not available, do not show any additional options
+                pass
+        return default
 
 
 class OmnisciLaunchParameters(HdkLaunchParameters, type=dict):
--- a/modin/config/test/test_envvars.py
+++ b/modin/config/test/test_envvars.py
@@ -16,6 +16,8 @@ import pytest
 import modin.config as cfg
 from modin.config.envvars import EnvironmentVariable, _check_vars, ExactStr
 
+from packaging import version
+
 
 @pytest.fixture
 def make_unknown_env():
@@ -63,9 +65,22 @@ def test_custom_help(make_custom_envvar):
 
 
 def test_hdk_envvar():
+    try:
+        import pyhdk
+
+        defaults = cfg.HdkLaunchParameters.get()
+        assert defaults["enable_union"] == 1
+        if version.parse(pyhdk.__version__) >= version.parse("0.6.1"):
+            assert defaults["log_dir"] == "pyhdk_log"
+        del cfg.HdkLaunchParameters._value
+    except ImportError:
+        # This test is intended to check pyhdk internals. If pyhdk is not available, skip the version check test.
+        pass
+
     os.environ[
         cfg.OmnisciLaunchParameters.varname
     ] = "enable_union=2,enable_thrift_logs=3"
+    del cfg.OmnisciLaunchParameters._value
     params = cfg.OmnisciLaunchParameters.get()
     assert params["enable_union"] == 2
     assert params["enable_thrift_logs"] == 3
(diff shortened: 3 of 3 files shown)


anmyachev (approved):
LGTM! Thanks @alexbaden.

@AndreyPavlenko please also review

modin/config/envvars.py
```diff
@@ -524,6 +524,8 @@ class HdkLaunchParameters(EnvironmentVariable, type=dict):
         "null_div_by_zero": 1,
         "enable_watchdog": 0,
         "enable_thrift_logs": 0,
+        "enable_lazy_dict_materialization": 0,
```
AndreyPavlenko, line 527: What about something like this?

```python
    @classmethod
    def _get_default(cls) -> Any:
        if (default := getattr(cls, "default", None)) is None:
            import pyhdk
            from packaging import version

            cls.default = default = {
                "enable_union": 1,
                "enable_columnar_output": 1,
                "enable_lazy_fetch": 0,
                "null_div_by_zero": 1,
                "enable_watchdog": 0,
                "enable_thrift_logs": 0,
            }

            if version.parse(pyhdk.__version__) >= version.parse("0.6.1"):
                default["enable_lazy_dict_materialization"] = 0
                default["log_dir"] = "pyhdk_log"
        return default
```

alexbaden: Sure, I can make that change. Would we still need to check here, though? https://github.com/modin-project/modin/pull/6303/files#diff-229f59982e185af8b8f88540fae04f56bbb400295a6116bad2feff4ea5ba1f9cR44 
If for some reason a user set the feature flag `log_dir` but had an older version of HDK, the command line parse could fail.

AndreyPavlenko: I think we do not need to check here. If a user set the `log_dir` option, it's better to fail with a corresponding error message, rather than removing the unsupported option silently.

anmyachev:
@alexbaden doc build failed with the following error:

```bash
  File "/home/docs/checkouts/readthedocs.org/user_builds/modin/envs/6303/lib/python3.8/site-packages/sphinx/config.py", line 350, in eval_config_file
    exec(code, namespace)
  File "/home/docs/checkouts/readthedocs.org/user_builds/modin/checkouts/6303/docs/conf.py", line 50, in <module>
    export_config_help(configs_file_path)
  File "/home/docs/checkouts/readthedocs.org/user_builds/modin/checkouts/6303/modin/config/__main__.py", line 61, in export_config_help
    "Default Value": default_values.get(obj.__name__, obj._get_default()),
  File "/home/docs/checkouts/readthedocs.org/user_builds/modin/checkouts/6303/modin/config/envvars.py", line 573, in _get_default
    if version.parse(pyhdk.__version__) >= version.parse("0.6.1"):
AttributeError: module 'pyhdk' has no attribute '__version__'
```

AndreyPavlenko:
@alexbaden I think, you need to add this attribute here - https://github.com/modin-project/modin/blob/master/docs/conf.py#L38

docs/conf.py
```diff
@@ -37,6 +38,7 @@ def noop_decorator(*args, **kwargs):
     sys.modules["cupy"].ndarray = type("ndarray", (object,), {})
 if not hasattr(sys.modules["pyhdk"], "PyDbEngine"):
     sys.modules["pyhdk"].PyDbEngine = type("PyDbEngine", (object,), {})
+    sys.modules["pyhdk"].__version__ = "0.6.0"
```
anmyachev, line 41: Shouldn't this variable be defined in `pyhdk` itself?

modin/config/envvars.py
```diff
@@ -557,12 +549,32 @@ def _get(cls) -> dict:
...
         custom_parameters = super().get()
-        result = cls.default.copy()
+        result = cls._get_default().copy()
         result.update(
             {key.replace("-", "_"): value for key, value in custom_parameters.items()}
         )
         return result
 
+    @classmethod
+    def _get_default(cls) -> Any:
```
anmyachev, line 559: @alexbaden could you add a docstring?

alexbaden: Done.

docs/conf.py
alexbaden: I put it under its own attribute check. Is that what you were referring to?

anmyachev: Ah, these are fake modules, I get it.

modin/config/envvars.py
```diff
@@ -557,12 +549,40 @@ def _get(cls) -> dict:
...
+            cls.default = default = {
+                "enable_union": 1,
+                "enable_columnar_output": 1,
+                "enable_lazy_fetch": 0,
+                "null_div_by_zero": 1,
+                "enable_watchdog": 0,
+                "enable_thrift_logs": 0,
+            }
+
+            if version.parse(pyhdk.__version__) >= version.parse("0.6.1"):
```
anmyachev, line 581: Let's wait for https://github.com/modin-project/modin/pull/6328 to be merged. After that, we can remove `pyhdk.__version__` and this will fix the test errors.

alexbaden: We introduced the pyhdk version API specifically for use in modin to allow us to introduce and use new features without tightly coupling the two releases. If we can guarantee the two versions always match, then perhaps we can remove this version checking. But, I was under the impression that was not something we could guarantee (@ienkovich @Garra1980). 

To fix the test error I can simply import pyhdk under a try/except and ignore the config params if either the pyhdk version is not high enough or if the module import fails.

docs/conf.py
anmyachev: Is it supposed to keep the minimum supported version of the hdk here? Synchronizing with the version in `env_hdk.yml`?

modin/config/envvars.py
anmyachev: It turns out that depending on the version, and whether it is installed in the environment or not, different default parameters will be used?

@Garra1980 which of the options is supposed to be used to generate documentation?

modin/config/envvars.py
```diff
@@ -557,12 +549,40 @@ def _get(cls) -> dict:
...
+        Get default value of the config. Checks the pyhdk version and omits variables unsupported in prior versions.
+
+        Returns
+        -------
+        dict
+            Config keys and corresponding values.
+        """
+        if (default := getattr(cls, "default", None)) is None:
+            import pyhdk
+            from packaging import version
```
anmyachev, line 570: This import is already done at the beginning of the file.
```suggestion
```

modin/config/envvars.py
Garra1980: I think for the latest supported pyhdk… in general, @alexbaden is there a place in hdk docs where all options are  described? I think we need to post a link to that place somewhere in modin docs

docs/conf.py
anmyachev: > Is it supposed to keep the minimum supported version of the hdk here? Synchronizing with the version in `env_hdk.yml`?

@alexbaden in that case let's update to `0.7`.

AndreyPavlenko: These are fake attributes. Do we need an up to date value here?

anmyachev: Documentation is built in an environment where hdk does not installed (only fake module). If we do not update this variable, then not all hdk's parameters will be specified in https://modin.readthedocs.io/en/latest/flow/modin/config.html.

AndreyPavlenko: Got it. Then, to avoid updating this attribute each time, when a new parameter is added, I would suggest:
```python
sys.modules["pyhdk"].__version__ = "999"
```

anmyachev: It is also desirable to leave a comment why it is done this way.

modin/config/test/test_envvars.py
```diff
@@ -16,6 +16,9 @@
 import modin.config as cfg
 from modin.config.envvars import EnvironmentVariable, _check_vars, ExactStr
 
+import pyhdk
```
anmyachev, line 19: This file does not assume that there will be engines in the environment.

We need to either make mock tests, or transfer the tests to files specific to `hdk`.

alexbaden: These tests are sanity tests for HDK, not for the logic in modin - so I don't think a mock is particularly useful. I hid the import under a try/catch, it should pass if `pyhdk` isn't in the env.

modin/config/envvars.py
alexbaden: I set the version to 999 as suggested.

anmyachev:
@alexbaden please fix flake8 error

alexbaden:
Should be fixed -- but one would think catching a specific exception would be self-documenting based on context (vs just an empty `except:`).

anmyachev:
@alexbaden still failing

anmyachev (approved):
@alexbaden thanks!

modin/config/envvars.py
```diff
@@ -558,12 +549,44 @@ def _get(cls) -> dict:
...
+        if (default := getattr(cls, "default", None)) is None:
+            cls.default = default = {
+                "enable_union": 1,
+                "enable_columnar_output": 1,
+                "enable_lazy_fetch": 0,
+                "null_div_by_zero": 1,
+                "enable_watchdog": 0,
+                "enable_thrift_logs": 0,
+                "cpu_only": 1,
+            }
```
YarShev, line 577: Why do we need this double assignment?

modin/config/envvars.py
```diff
@@ -558,12 +549,44 @@ def _get(cls) -> dict:
...
+
+            try:
+                import pyhdk
+
+                if version.parse(pyhdk.__version__) >= version.parse("0.6.1"):
+                    default["enable_lazy_dict_materialization"] = 0
+                    default["log_dir"] = "pyhdk_log"
```
YarShev, line 584: Note that this change affects cls.default. Is this behavior we want to have?

modin/config/envvars.py
AndreyPavlenko: The dictionary is created only once and assigned to the class attribute `cls.default`. All subsequent calls will not get into this branch.

modin/config/envvars.py
AndreyPavlenko: Yes, this is correct. All this code is used to initialize `cls.default`.

YarShev: ok

PR #6342, opened 2023-07-01
FEAT-#6407: update minimum dependency versions

### LINKED ISSUE #6407: Update minimum dependency versions
Part of the work of keeping environments up to date.

anmyachev:
## What do these changes do?

* This pull request adds version lower bounds for packages, in line with [those in pandas](https://github.com/pandas-dev/pandas/blob/478d340667831908b5b4bf09a2787a11a14560c9/pyproject.toml) for the lowest supported version: pandas 2.0.0.
* This should simplify building the environment and save us from potential errors if the code is executed in pandas, where there will be an interaction with any third-party library whose version is not compatible with pandas itself.
* If the library is used by Modin, but not by pandas, I tried to use the version that was released more than a year ago as the minimum boundary.

- [x] Resolves #6407 
- [x] tests added and passing


Garra1980:
Is this still needed or #6350 is enough and we can close this one?

docs/supported_apis/io_supported.rst
```diff
@@ -56,8 +56,6 @@ default to pandas.
 +--------------------+---------------------------------+----------------------------------------------------+
 | `read_feather`_    | Y                               |                                                    |
 +--------------------+---------------------------------+----------------------------------------------------+
-| `read_msgpack`_    | D                               |                                                    |
```
anmyachev, line 59: Deprecated functionality. Found by looking at `msgpack-python` (which has been removed).

environment-dev.yml
```diff
@@ -6,48 +6,45 @@ dependencies:
...
   - dask>=2.22.0
   - distributed>=2.22.0
   # TODO: uncomment after Modin switch to python>=3.9
   # - xarray
-  - Jinja2
-  - scipy
+  - Jinja2>=3.0.0
+  - scipy>=1.7.1
   - s3fs>=2021.8
-  - feather-format
```
anmyachev, line 30: This is a outdated package, the required functionality is now in `pyarrow`. So I removed it.

environment-dev.yml
```diff
@@ -6,48 +6,45 @@ dependencies:
...
-  - xlrd
-  - matplotlib
+  - lxml>=4.6.3
+  - openpyxl>=3.0.7
+  - xlrd>=2.0.1
+  - matplotlib>=3.6.1
   - sqlalchemy>=1.4.0,<1.4.46
-  - pandas-gbq
-  - pytables
-  - msgpack-python
```
anmyachev, line 38: Unused package. Removed it.

environment-dev.yml
```diff
@@ -6,48 +6,45 @@ dependencies:
...
-  - scikit-learn
```
anmyachev, line 39: Looks like unused package. Need to double check it.

anmyachev: The package is used in `test_dmatrix.py`, `test_xgboost.py`.

scripts/doc_checker.py
```diff
@@ -533,6 +533,13 @@ def load_obj(name, old_load_obj=Docstring._load_obj):
     # enable docs testing on windows
     sys.getdlopenflags = Mock()
     sys.setdlopenflags = Mock()
+    xgboost_mock = Mock()
+
+    class Booster:
+        ...
+
+    xgboost_mock.Booster = Booster
```
anmyachev, line 541: Otherwise, the documentation check will break, since the object inheriting (`Booster(xgb.Booster)`) from the mock itself becomes a mock. The functionality of the inspect module does not work with mock objects:
```bash
Miniconda3\envs\test-modin\lib\inspect.py", line 676, in getfile
    raise TypeError('module, class, method, function, traceback, frame, or '
TypeError: module, class, method, function, traceback, frame, or code object was expected, got Mock
```

.github/workflows/ci.yml
```diff
@@ -142,7 +142,12 @@ jobs:
           environment-file: environment-dev.yml
       - name: Install HDF5
         run: sudo apt update && sudo apt install -y libhdf5-dev
-      - run: python -m pytest modin/experimental/xgboost/test/test_default.py --execution=${{ matrix.execution }}
+      - name: xgboost tests
+        run: |
+          # TODO(https://github.com/modin-project/modin/issues/5194): Uncap xgboost
+          # when we use collective instead of rabit.
+          mamba install "xgboost>=1.7.1,<2.0.0" scikit-learn -c conda-forge
```
anmyachev, line 149: These libraries are rarely needed in unit testing, there is no real reason to put them in the general environment (which slows down the environment creation).

environment-dev.yml
```diff
@@ -6,48 +6,42 @@ dependencies:
...
+  - packaging>=21.0
+  - psutil>=5.8.0
 
   # optional dependencies
   # ray==2.5.0 broken: https://github.com/conda-forge/ray-packages-feedstock/issues/100
   - ray-default>=1.13.0,!=2.5.0
   # https://github.com/modin-project/modin/issues/6336
   - pydantic<2
-  - pyarrow
+  - pyarrow>=7.0.0
```
vnlitvinov, line 19: this is pretty aggressive bump, isn't it? why cannot we use earlier verions?

vnlitvinov (commented):
I surely hope you did check that all the files specify the same version requirements now...

environment-dev.yml
anmyachev: Looks like `pandas 2.*` isn't compatible with `pyarrow 6.*`
https://github.com/modin-project/modin/actions/runs/5649681599/job/15304749494

anmyachev:
> I surely hope you did check that all the files specify the same version requirements now...

yes

vnlitvinov (approved):
LGTM!

mvashishtha (approved):
xgboost changes LGTM

modin/experimental/xgboost/test/test_dmatrix.py
```diff
@@ -19,9 +19,17 @@
 import xgboost as xgb
 
 import modin.pandas as pd
+from modin.config import Engine
 import modin.experimental.xgboost as mxgb
 
 
+if Engine.get() == "Python":
```
YarShev, line 26: Our xgb implementation doesn't work on dask or mpi too for now. Why do we check for python only?

anmyachev: This file is tested with Dask in the CI and the tests pass.

YarShev: As I see we tested this file only  with Ray previously but you enabled testing with Dask. Testing for this file passes with Dask because there are no tests specific to our xgb implementation but this may change. I would recommend to leave testing only with Ray.

anmyachev: I can do it right now, but why not make changes the moment the tests fail?

YarShev: Because we explicitly say that we support xgb on ray only, but at the same time we test it on other engine, which accidendtly passes. Let's test only on Ray for now.

YarShev: The reason of the test pass is that we don't check for any xgb-related things in it. Anyway, I want to be explicit and test xgb only on the engines we explicitly support.

anmyachev: Ok

PR #6357, opened 2023-07-06
Release version 0.23.0

mvashishtha:


vnlitvinov:
@mvashishtha 
I would like us to pin badges, pypi and doc links to specific release, so those would stay correct even when we make other changes, like in ee312dad11213942d0b03ee97d8c63dbb1efa07c

YarShev:
> @mvashishtha I would like us to pin badges, pypi and doc links to specific release, so those would stay correct even when we make other changes, like in [ee312da](https://github.com/modin-project/modin/commit/ee312dad11213942d0b03ee97d8c63dbb1efa07c)

Does the release procedure require an update?

YarShev:
I am actually wondering why we should start pinning badges, pypi and doc links to specific release?

anmyachev:
> I am actually wondering why we should start pinning badges, pypi and doc links to specific release?

The main branch is now showing release 0.22.3 when it should be 0.23.0rc0.

YarShev:
Why does the main branch have to show 0.23.0rc0? It is a RC, but not a stable release.

anmyachev:
> Why does the main branch have to show 0.23.0rc0? It is a RC, but not a stable release.

I believe that the release should be shown not the latest in terms of release date, but the latest in terms of versioning, where 0.23.0rc0 >= 0.22.3.

YarShev:
> I believe that the release should be shown not the latest in terms of release date, but the latest in terms of versioning, where 0.23.0rc0 >= 0.22.3.

On the one hand this makes sense, but on the other hand I am not sure we want the badges to point at non-stable releases. I would rather we keep the current concept we have in the release procedure. @modin-project/modin-core, more thoughts?

mvashishtha:
> On the one hand this makes sense, but on the other hand I am not sure we want the badges to point at non-stable releases. I would rather we keep the current concept we have in the release procedure.

I agree. @vnlitvinov what is the benefit of pinning badge versions on master? In this case the badges will be incorrect until we really publish the 0.23.0 release on pypi.

anmyachev:
If this is a question that requires discussion, then let's do it as before. And then, if we decide, we will update the badges.

mvashishtha:
@anmyachev LGTM, but I can't approve my own PR

YarShev (approved):

YarShev:
@vnlitvinov, if you still have any concern, let's discuss that offline and decide on our future release procedure.

Issue #3211, opened 2021-07-02
Fallback to Pandas and mixed types


AlexHellen:
### System information
WIndows10
0.10.0
Python 3.7
- **Code we can use to reproduce**:

### Describe the problem
Types get mixed up betweean regular pandas and modin, causing problems in the following functions

### Source code / logs
import modin.pandas as pd
df1 = pd.DataFrame(data = [[10], [20], [70]], index = ['USD', 'CHF', 'EUR'], columns=['curr'])
print (type(df1.index))
print(type(df1.curr))

Result:
<class 'pandas.core.indexes.base.Index'>
<class 'modin.pandas.series.Series'>

As a consequence for example:

import plotly.express as px
fig = px.pie(df1, names=df1.index, values = df1.curr, title='Curr Anteil an Cashflows')
fig.show()

Yields:
ValueError: The truth value of a Series is ambiguous. Use a.empty, a.bool(), a.item(), a.any() or a.all().

Workaround:
fig = px.pie(df1, names=df1.index, values = df1.curr._to_pandas(), title='Curr Anteil an Cashflows')
fig.show()

This works because of

devin-petersohn:
Hi @AlexHellen thanks for posting!

I am not sure this is something we can fix in Modin, since plotly is doing something with the Modin object.

RehanSD:
Hi @devin-petersohn @mvashishtha has this issue been resolved?

mvashishtha:
@RehanSD I get same error at Modin 88f7b27d32b099c508e93251cafe503f2908d426 using the reproducer in the first post, so this is still relevant. See #3967 for almost the same bug for Seaborn.

anmyachev:
@jbrockmendel this problem arises from the fact that `modin.Series` passes into the `pandas.Series()` constructor.

In the case of a dataframe, a protocol would help (`pandas.api.interchange.from_dataframe`). However, it will not help for series. When the protocol was planned, there was no discussion about what to do in this case?

What do you think about considering adding an extra code branch to the `pandas.Series()` constructor like the following?
```python
elif hasattr(data, "__series__"):
  pandas_series = data.__series__()
elif is_dict_like(data):
```

Your feedback will be very valuable here, thanks!

jbrockmendel:
> What do you think about considering adding an extra code branch to the pandas.Series() constructor like the following?

You can open an issue for this on the pandas tracker, but I don't think its very likely to happen.

[...]

PR #6361, opened 2023-07-06
FEAT-#0000: enable interchange protocol for Series

anmyachev:
## What do these changes do?

This pull request makes it possible to use the dataframe interchange protocol on a Series object. If on the side of the third-party library one need to get a pandas series from the Modin series, then this can be done in the following way:
`pandas_series = pandas.api.interchange.from_dataframe(modin_series).squeeze()`.

The need for such a conversion can be seen from https://github.com/modin-project/modin/issues/3211. 

@vnlitvinov @YarShev please provide your opinion.

- [ ] Resolves #? 
- [ ] tests added and passing

### DIFF
--- a/modin/pandas/series.py
+++ b/modin/pandas/series.py
@@ -210,6 +210,38 @@ class Series(BasePandasDataset):
         """
         return super(Series, self).__array__(dtype).flatten()
 
+    def __dataframe__(self, nan_as_null: bool = False, allow_copy: bool = True):
+        """
+        Get a Modin DataFrame that implements the dataframe exchange protocol.
+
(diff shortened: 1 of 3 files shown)


rgommers:
This indeed seems a bit too hacky to me. Not only from the protocol perspective, but it's kinda declaring that a `Series` instance is a kind of dataframe by adding the `__dataframe__` method. I have the uneasy feeling that that's going to lead to some problems when end users or downstream library authors may start checking objects for the presence of `__dataframe__` as perhaps the most portable way of checking if that object _is_ a dataframe (not that I'd advocate for that, but I also don't know of a better way of doing that).

anmyachev:
Thanks for the feedback @vnlitvinov @rgommers!

PR #6363, opened 2023-07-06
REFACTOR-#6362: `__setitem__` on multiple columns should be evaluated lazily

### LINKED ISSUE #6362: PERF: `__setitem__` on multiple columns should be evaluated lazily
NOTE: I haven't done any clear benchmarking for this.

In the following reproducer:
```python
import modin.pandas as pd
import pandas as vpd
import numpy as np

cols = [f"feature_{i}" for i in range(4)]
cols.append('labels')
df = pd.read_csv('data.txt', header=None)
df.columns = cols
val_df = df.sample(frac=0.2)
train_df = df.drop(val_df.index)

train_means = train_df[cols[:-1]].mean()
train_std = train_df[cols[:-1]].std()
train_df[cols[:-1]] = (train_df[cols[:-1]]- train_means)/train_std
```

data: [data.txt](https://github.com/modin-project/modin/files/11969394/data.txt)

`__setitem__` on multiple columns should not materialize partitions immediately.

pyrito:
## What do these changes do?

- [ ] Resolves #6362 
- [ ] tests added and passing

### DIFF
--- a/modin/pandas/dataframe.py
+++ b/modin/pandas/dataframe.py
@@ -56,6 +56,7 @@ from .utils import (
     from_pandas,
     from_non_pandas,
     broadcast_item,
+    _maybe_reindex_item,
     cast_function_modin2pandas,
     SET_DATAFRAME_ATTRIBUTE_WARNING,
 )
@@ -2451,13 +2452,22 @@ class DataFrame(BasePandasDataset):
                         value = np.array(value)
                     if len(key) != value.shape[-1]:
                         raise ValueError("Columns must be same length as key")
-                item = broadcast_item(
-                    self,
-                    slice(None),
-                    key,
-                    value,
-                    need_columns_reindex=False,
-                )
+
+                # If we have a modin Series or DataFrame, let's not convert it to
+                # a numpy array if we can help it.
+                if isinstance(value, (DataFrame, Series)):
+                    item = _maybe_reindex_item(
+                        self, slice(None), key, value, need_columns_reindex=False
+                    )
+                else:
+                    item = broadcast_item(
+                        self,
+                        slice(None),
+                        key,
+                        value,
+                        need_columns_reindex=False,
+                    )
+
                 new_qc = self._query_compiler.write_items(
                     slice(None), self.columns.get_indexer_for(key), item
                 )
--- a/modin/pandas/dataframe.py
+++ b/modin/pandas/dataframe.py
@@ -56,7 +56,6 @@ from .utils import (
     from_pandas,
     from_non_pandas,
     broadcast_item,
-    _maybe_reindex_item,
     cast_function_modin2pandas,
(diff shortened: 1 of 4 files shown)


modin/pandas/utils.py
```diff
@@ -345,6 +340,28 @@ def broadcast_item(
...
+            return np.broadcast_to(item, to_shape)
+    except ValueError:
+        from_shape = np.array(item).shape
+        raise ValueError(
+            f"could not broadcast input array from shape {from_shape} into shape "
+            + f"{to_shape}"
+        )
+
+
+def _maybe_reindex_item(obj, row_lookup, col_lookup, item, need_columns_reindex=True):
```
mvashishtha, line 359: Can we remove the "maybe" and make this method only valid for series and dataframes?

modin/pandas/dataframe.py
```diff
@@ -2451,13 +2452,22 @@ def __setitem__(self, key, value):
...
-                )
+
+                # If we have a modin Series or DataFrame, let's not convert it to
+                # a numpy array if we can help it.
+                if isinstance(value, (DataFrame, Series)):
+                    item = _maybe_reindex_item(
+                        self, slice(None), key, value, need_columns_reindex=False
+                    )
+                else:
+                    item = broadcast_item(
```
mvashishtha, line 2463: So `_maybe_reindex_item` gets called in either code path here (either directly here, or in `broadcast_item`)? Can we always call it directly here?

mvashishtha (commented):
@pyrito have you checked whether the PandasOnPython and BaseOnPython query compilers can handle inserting a dataframe instead of a numpy array?

I also left some comments about the implementation details.

modin/pandas/dataframe.py
pyrito: Yeah actually this was dumb thinking on my part. I could just return early in `broadcast_item`.

mvashishtha (commented):
this solution generally looks ok. I left some comments about making it neater

modin/pandas/utils.py
```diff
@@ -358,6 +348,24 @@ def broadcast_item(
         # New value for columns/index make that reindex add NaN values
         if axes_to_reindex:
             item = item.reindex(**axes_to_reindex)
+
+    # Allow the query_compiler to dictate what to do instead of materializing the data
+    if isinstance(item, (DataFrame, Series)):
```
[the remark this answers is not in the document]
pyrito: @mvashishtha technically we are still doing a broadcast in the sense that we are reshaping the dataframe, it's just that we may not always be returning a numpy array so I think only the docs really need to be updated.

anmyachev:
@dchigarev so we can close it, right?

anmyachev:
Looks like something similar already done in #6618. Feel free to reopen if needed.

PR #6374, opened 2023-07-11
FIX-#0000: don't initialize engines at import time

anmyachev:
## What do these changes do?

- [ ] Resolves #? 
- [ ] tests added and passing


anmyachev:
Note for reviewers: don't forget to change `FIX-#0000` -> `FIX-#6375` in the final commit message.

modin/core/execution/ray/implementations/pandas_on_ray/partitioning/partition.py
```diff
@@ -45,6 +45,9 @@ class PandasOnRayDataframePartition(PandasDataframePartition):
 
     execution_wrapper = RayWrapper
 
+    # these variables are intentionally initialized at runtime
```
YarShev, line 48: I would also suggest adding some words why we initialize the vars at runtime.

modin/core/execution/ray/implementations/pandas_on_ray/partitioning/partition.py
```diff
@@ -173,9 +176,11 @@ def __copy__(self):
             call_queue=self.call_queue,
         )
 
-    # If Ray has not been initialized yet by Modin,
-    # it will be initialized when calling `RayWrapper.put`.
-    _iloc = execution_wrapper.put(PandasDataframePartition._iloc)
+    @classmethod
+    def _get_iloc_func(cls):  # noqa: GL08
```
YarShev, line 180: docstring + explanation why we use it

YarShev (commented):
LGTM! Left a couple of comments.

modin/core/execution/ray/implementations/pandas_on_ray/partitioning/partition.py
```diff
@@ -173,9 +177,21 @@ def __copy__(self):
...
+    @classmethod
+    def _get_iloc_func(cls):
+        """
+        Places `_iloc` function into the storage to speed up remote function calls and caches the result.
+
+        It also postpones engine initialization, which happens implicitly when function `put` is called.
+
+        Returns
+        -------
+        ray.ObjectID
```
YarShev, line 189:
```suggestion
        ray.ObjectRef
```

YarShev (dismissed):
LGTM! Left one suggestion.

modin/core/dataframe/pandas/partitioning/partition.py
```diff
@@ -198,6 +198,18 @@ def _iloc(df, row_labels, col_labels):  # noqa: RT01, PR01
...
 
+    @classmethod
+    def _get_iloc_func(cls):
+        """
+        Allow specific implementations to have a more optimal way to get `iloc` function.
+
+        Returns
+        -------
+        callable
+            `iloc` function
```
YarShev, line 209:
```suggestion
            `iloc` function.
```

YarShev (dismissed):

vnlitvinov:
@anmyachev I've tried to improve a little bit, being as bold as pushing to your branch. Feel free to revert my commit if you disagree, but IMHO that version is both shorter and should run slightly faster in a general case (i.e. when we take masks more times than we make partitions).

anmyachev:
> @anmyachev I've tried to improve a little bit, being as bold as pushing to your branch. Feel free to revert my commit if you disagree, but IMHO that version is both shorter and should run slightly faster in a general case (i.e. when we take masks more times than we make partitions).

@vnlitvinov thanks! LGTM

modin/core/dataframe/pandas/partitioning/partition.py
```diff
@@ -38,6 +38,24 @@ class PandasDataframePartition(ABC):  # pragma: no cover
...
+    def __init__(self):
+        if type(self)._iloc_func is None:
+            # Places `_iloc` function into the storage to speed up
+            # remote function calls and caches the result.
+            # It also postpones engine initialization, which happens
+            # implicitly when `execution_wrapper.put` is called.
+            if self.execution_wrapper is not None:
+                type(self)._iloc_func = staticmethod(
+                    self.execution_wrapper.put(self._iloc)
+                )
```
YarShev, line 56:
```suggestion
                type(self)._iloc_func = self.execution_wrapper.put(self._iloc)
```
Why do we need to pass an obj ref to staticmethod?

modin/core/dataframe/pandas/partitioning/partition.py
```diff
@@ -38,6 +38,24 @@ class PandasDataframePartition(ABC):  # pragma: no cover
...
+            else:
+                type(self)._iloc_func = staticmethod(self._iloc)
```
YarShev, line 58:
```suggestion
                type(self)._iloc_func = self._iloc
```
self._iloc is already a static method. Why do we need to wrap it over again?

anmyachev: Consider the following example:
```python
>>> class A:
...   @staticmethod
...   def test_func():
...     print("call")
...
>>> A.test_func
<function A.test_func at 0x0000015615BAE280>
>>> A.test_func()
call
>>> A().test_func()
call
>>> A.test_func2 = A().test_func  # emulate behavior as in __init__
>>> A.test_func2()
call
>>> A().test_func2()
TypeError: test_func() takes 0 positional arguments but 1 was given
>>> A.test_func2 = staticmethod(A().test_func)
>>> A.test_func2()
call
>>> A().test_func2()
call
```

YarShev: I see, thanks!

YarShev (approved):

anmyachev:
@YarShev pls merge it

modin/core/dataframe/pandas/partitioning/partition.py
vnlitvinov: Class-level `__setattr__` behaves differently and does quite a few magic tricks, one of which is when you assign an attribute which is a function, it becomes an unbound method _during assignment_, etc. Hence if no `staticmethod()` call is done here, such transformation is done, and function starts to receive an extra `self` argument when called, which leads to errors ([ref](https://github.com/modin-project/modin/actions/runs/5519619408/jobs/10065219105)).

PR #6379, opened 2023-07-11
PERF-#6378: use numpy.array instead of list in internals of iloc/loc operation

### LINKED ISSUE #6378: use numpy.array instead of list in internals of iloc/loc operation

anmyachev:
## What do these changes do?

numpy arrays are much faster to serialize than lists , what will need to be done next when performing `map_axis_partitions` operation.

<details>

<summary>
Script to measure
</summary>

```python
import numpy as np
import ray
from time import time

# initialize it
ray.put([1,2,3,4])

np.random.seed(42)

len_rows = 10**7
sorted_rows_position = np.arange(len_rows)
row_positions = sorted_rows_position.copy()
np.random.shuffle(row_positions)


row_order_mapping = dict(
    zip(sorted_rows_position, range(len(row_positions)))
)

print("old way")
for _ in range(3):
    start = time()
    new_row_order = [row_order_mapping[idx] for idx in row_positions]


start = time()
ref = ray.put(new_row_order)
print(f"ray put {time()-start}")
del ref


print("from_iter way")
for _ in range(3):
    start = time()
    new_row_order = np.fromiter((row_order_mapping[idx] for idx in row_positions), dtype="int64", count=len(row_positions))

start = time()
ref = ray.put(new_row_order)
print(f"ray put {time()-start}")
```

</details>

- [x] Resolves #6378 
- [x] tests added and passing

### DIFF
--- a/modin/core/dataframe/pandas/dataframe/dataframe.py
+++ b/modin/core/dataframe/pandas/dataframe/dataframe.py
@@ -1110,14 +1110,22 @@ class PandasDataframe(ClassLogger):
             row_order_mapping = dict(
                 zip(sorted_row_positions, range(len(row_positions)))
             )
-            new_row_order = [row_order_mapping[idx] for idx in row_positions]
+            new_row_order = np.fromiter(
+                (row_order_mapping[idx] for idx in row_positions),
+                dtype="int64",
+                count=len(row_positions),
+            )
         else:
             new_row_order = None
         if col_positions is not None:
             col_order_mapping = dict(
                 zip(sorted_col_positions, range(len(col_positions)))
             )
-            new_col_order = [col_order_mapping[idx] for idx in col_positions]
+            new_col_order = np.fromiter(
+                (col_order_mapping[idx] for idx in col_positions),
+                dtype="int64",
+                count=len(col_positions),
+            )
         else:
             new_col_order = None
         return intermediate._reorder_labels(


anmyachev:
@dchigarev please review

modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -1110,14 +1110,22 @@ def _maybe_reorder_labels(
             row_order_mapping = dict(
                 zip(sorted_row_positions, range(len(row_positions)))
             )
-            new_row_order = [row_order_mapping[idx] for idx in row_positions]
+            new_row_order = np.fromiter(
+                (row_order_mapping[idx] for idx in row_positions),
+                dtype="int64",
+                count=len(row_positions),
+            )
```
vnlitvinov, line 1117: from the surrounding code, I believe we can make the following:
1. `sorted_row_positions` is a numpy array (we can make it 100% sure by calling `numpy.asarray()` which is a no-op for a thing which is already an array)
2. same for `row_positions`
3. then making the dictionary and then building an array via `np.fromiter` is just a super-slow version of `sorted_row_positions[row_positions]` 😉 

If I'm right, this piece could be sped up even further (and code greatly simplified) by rewriting as
```python
new_row_order = np.asarray(sorted_row_positions)[np.asarray(row_positions)]
```

vnlitvinov: What does it matter if it's a contiguous array (if you mean in-memory contiguous)? Numpy is smart enough:
```python
>>> np.array([1,5,7,2,9,12,19,-5,6])[np.array([2,6])]
array([ 7, 19])
>>> x1=np.array(np.arange(10))
>>> x1.flags
  C_CONTIGUOUS : True
  F_CONTIGUOUS : True
  OWNDATA : True
  WRITEABLE : True
  ALIGNED : True
  WRITEBACKIFCOPY : False

>>> x2=x1[::-1]
>>> x2.flags
  C_CONTIGUOUS : False
  F_CONTIGUOUS : False
  OWNDATA : False
  WRITEABLE : True
  ALIGNED : True
  WRITEBACKIFCOPY : False

>>> x2[np.array([2,3])]
array([7, 6])
>>> x1[x2]
array([9, 8, 7, 6, 5, 4, 3, 2, 1, 0])
```

anmyachev: I have the following example in my mind:
```python
>>> import numpy as np
>>> row_positions = np.array([7,1,3])
>>> sorted_row_positions = np.array([1,3,7])
>>> sorted_row_positions[row_positions]
IndexError: index 7 is out of bounds for axis 0 with size 3
>>>
```

vnlitvinov: I have built an alternative approach, need to benchmark it before making a PR though...

PR #6380, opened 2023-07-12
FEAT-#6377: HDK: Do not keep reference to arrow table imported to HDK

### LINKED ISSUE #6377: HDK: Do not keep reference to arrow table imported to HDK
When the table is imported, it's not required any more and could be garbage collected.

AndreyPavlenko:
## What do these changes do?

- [x] Resolves #6377 
- [ ] tests added and passing

### DIFF
--- a/modin/experimental/core/execution/native/implementations/hdk_on_native/hdk_worker.py
+++ b/modin/experimental/core/execution/native/implementations/hdk_on_native/hdk_worker.py
@@ -12,12 +12,12 @@
 # governing permissions and limitations under the License.
 
 """Module provides ``HdkWorker`` class."""
-from typing import Optional, Tuple, List
+from typing import Optional, Tuple, List, Union
 
 import pyarrow as pa
 import os
 
-from pyhdk.hdk import HDK, ExecutionResult, RelAlgExecutor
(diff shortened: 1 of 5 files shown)


modin/experimental/core/execution/native/implementations/hdk_on_native/hdk_worker.py
```diff
@@ -56,13 +62,11 @@ def column_names(self) -> List[str]:
...
-        if at is None:
-            at = self._result.to_arrow()
-            if (names := getattr(self, "_column_names", None)) is not None:
-                at = at.rename_columns(names)
-            self._at = at
-        return at
+        return (
+            self._table.to_arrow()
+            if isinstance(self._table, ExecutionResult)
+            else self._table.run().to_arrow()
```
ienkovich, line 68: This can cause performance degradation. Previously, you had zero cost for calling this method for imported arrow tables. Now, it would run actual query execution + arrow conversion for its result.

AndreyPavlenko: Right, but it allows the arrow table to be garbage collected.
In case of a normal execution (no fallback to pandas), this method will not be called.

Issue #6389, opened 2023-07-14
Ignore `.modin` creation


rentruewang:
**Is your feature request related to a problem? Please describe.**
Modin creates its logs in `.modin`, which becomes a headache when I run testing in a git repository, because it gets created automatically and pollutes the repo if not ignored.

PR #6390, opened 2023-07-14
FEAT-#6389: Ignore logs.

rentruewang:
This commit creates a gitignore file in the 🪵 directory. Closes: #6389.

## What do these changes do?

Creates a gitignore file with `*` as content, to prevent addition of logs to git repos when using modin in git repositories. Inspired by `pytype`.

- [x] Resolves #? 
- [x] tests added and passing


RehanSD (dismissed):
Shouldn’t we just modify the .gitignore file to ignore everything inside .modin instead of doing it programmatically?

rentruewang:
It's for the people using modin. E.g. when you have tests using modin, the .modin would get added to the git repository, which is usually never the intention.

anmyachev:
Wouldn't it be better to move `.modin` folder to `tmp` folder?

rentruewang:
@anmyachev Not my decision to make unfortunately. However, I don't think `/tmp` is the best way as there "might be" persistent cache etc in the `.modin` folder.

modin/logging/config.py
```diff
@@ -117,8 +117,19 @@ def _create_logger(
     Logger
         Logger object configured per Modin settings.
     """
-    log_filename = f".modin/logs/job_{job_id}/{log_name}.log"
-    os.makedirs(os.path.dirname(log_filename), exist_ok=True)
+
```
anmyachev, line 120:
```suggestion
```

anmyachev (approved):
Since the cache is saved to the current working directory, I agree that this can get in the way and something needs to be done about it. The proposed approach suits me too.

vnlitvinov (dismissed):
LGTM!

modin/logging/config.py
rentruewang: With that being said, it's NBD and I'm fine either way.

rentruewang: I see. It's one of the checks. My bad.

anmyachev:
Thanks @rentruewang!

PR #6423, opened 2023-07-27
PERF-#6388: avoid masking in __getitem__ when the number of rows to be taken > 90%

### LINKED ISSUE #6388: `modin.array.__getitem__` works very inefficient when the key is modin.array and its size is > 50% of all rows

anmyachev:
## What do these changes do?

This change speeds up the workload I work on. I'll add a couple of synthetic examples a little later, but the pull request is ready for review.

Benchmark's result (best time): 5.4 sec (PR) vs 7.2 sec (master). Boost: ~25%.

<details>

<summary> Benchmark </summary>

```python
from time import time

import os

os.environ["MODIN_ENGINE"] = "Ray"
os.environ["MODIN_CPUS"] = "44"
os.environ["MODIN_BENCHMARK_MODE"] = "True"

import numpy as np
import modin.pandas as pd

df = pd.DataFrame(random_state.randint(0, 10**6, size=(5 * 10**6, 64)))

new_index = df.index.values[:int(len(df.index)*0.9) + 1]
random_state.shuffle(new_index)


for _ in range(10):
    start = time()
    new_df = df.iloc[new_index]

```
</details>

- [x] Resolves #6388 
- [x] tests added and passing

### DIFF
--- a/modin/core/dataframe/pandas/dataframe/dataframe.py
+++ b/modin/core/dataframe/pandas/dataframe/dataframe.py
@@ -991,7 +991,7 @@ class PandasDataframe(ClassLogger):
             return self.copy()
 
         if col_positions is None and row_positions is not None:
-            # Check if the optimization that first takes part of the data using the mask 
+            # Check if the optimization that first takes part of the data using the mask
             # operation so that later less data is concatenated into a whole column is useful.
             # In the case when only a small portion of the data is discarded, the overhead of the
             # engine (for putting data in and out of storage) can exceed the resulting speedup.
--- a/modin/core/dataframe/pandas/dataframe/dataframe.py
+++ b/modin/core/dataframe/pandas/dataframe/dataframe.py
@@ -991,7 +991,10 @@ class PandasDataframe(ClassLogger):
             return self.copy()
 
         if col_positions is None and row_positions is not None:
-            # fast path
+            # This way tries to eliminate the optimization that first takes part of the data using
+            # the mask operation, so that later less data is concated into a whole column.
+            # Note: in the case where a small part of the data is discarded, the overhead of the
+            # engines (for putting data in and out of storage) can exceed the resulting speedup.
             all_rows = None
             if self.has_materialized_index:
                 all_rows = len(self.index)
--- a/modin/core/dataframe/pandas/dataframe/dataframe.py
+++ b/modin/core/dataframe/pandas/dataframe/dataframe.py
@@ -991,17 +991,17 @@ class PandasDataframe(ClassLogger):
             return self.copy()
 
         if col_positions is None and row_positions is not None:
-            # This way tries to eliminate the optimization that first takes part of the data using
-            # the mask operation, so that later less data is concated into a whole column.
-            # Note: in the case where a small part of the data is discarded, the overhead of the
-            # engines (for putting data in and out of storage) can exceed the resulting speedup.
+            # Check if the optimization that first takes part of the data using the mask 
+            # operation so that later less data is concatenated into a whole column is useful.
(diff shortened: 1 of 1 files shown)


modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -990,6 +990,19 @@ def _take_2d_positional(
         if col_positions is None and row_positions is None:
             return self.copy()
 
+        if col_positions is None and row_positions is not None:
+            # fast path
```
vnlitvinov, line 994: an explanation of why it's a fast path and what's going on would be useful IMHO

anmyachev: @vnlitvinov comment added, synthetic benchmark result at the top.

modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -990,6 +990,22 @@ def _take_2d_positional(
...
+            # the mask operation, so that later less data is concated into a whole column.
+            # Note: in the case where a small part of the data is discarded, the overhead of the
+            # engines (for putting data in and out of storage) can exceed the resulting speedup.
+            all_rows = None
+            if self.has_materialized_index:
+                all_rows = len(self.index)
+            elif self._row_lengths_cache:
+                all_rows = sum(self._row_lengths_cache)
+            if all_rows:
+                if len(row_positions) > int(0.9 * all_rows):
```
vnlitvinov, line 1004:
```suggestion
                if len(row_positions) > 0.9 * all_rows:
```
in Python, there's no need to cast to `int` to compare 🙃 

Also I wonder if the threshold should be tuned - did you try different coefficients?

modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -990,6 +990,22 @@ def _take_2d_positional(
         if col_positions is None and row_positions is None:
             return self.copy()
 
+        if col_positions is None and row_positions is not None:
+            # This way tries to eliminate the optimization that first takes part of the data using
+            # the mask operation, so that later less data is concated into a whole column.
+            # Note: in the case where a small part of the data is discarded, the overhead of the
+            # engines (for putting data in and out of storage) can exceed the resulting speedup.
```
vnlitvinov, line 997:
```suggestion
            # Check if the optimization that first takes part of the data using the mask 
            # operation so that later less data is concatenated into a whole column is useful.
            # In the case when only a small portion of the data is discarded, the overhead of the
            # engine (for putting data in and out of storage) can exceed the resulting speedup.
```

vnlitvinov (dismissed):
LGTM overall, left a few optional suggestions.

vnlitvinov (approved):
LGTM!

anmyachev:
> Also I wonder if the threshold should be tuned - did you try different coefficients?

@vnlitvinov I believe that the coefficient can be further reduced, for this it is necessary to conduct more perf. experiments.

modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -990,6 +990,22 @@ def _take_2d_positional(
         if col_positions is None and row_positions is None:
             return self.copy()
 
+        if col_positions is None and row_positions is not None:
+            # Check if the optimization that first takes part of the data using the mask
```
anmyachev, line 994:
```suggestion
            # Check if the optimization that first takes part of the data using the mask
```

anmyachev:
Blocked by https://github.com/modin-project/modin/pull/6457

vnlitvinov (approved):
LGTM!

PR #6448, opened 2023-08-03
FIX-#6436: Support ~ in paths in IO functions correctly

### LINKED ISSUE #6436: BUG: Some Modin IO functions unnecessary default to pandas on ~path
### Modin version checks

### Reproducible Example

```python
import warnings
import pandas
import modin.pandas as pd
import modin.config as cfg
cfg.Engine.put('python') # to ease debugging

df1 = pandas.DataFrame({'a': [1,2], 'b': [3,4]})
df1.to_csv('~/simple-test-file.csv')

with warnings.catch_warnings(record=True) as warned:
    df2 = pd.read_csv('~/simple-test-file.csv')
df3 = pandas.read_csv('~/simple-test-file.csv')

assert (df2._to_pandas() == df3).all().all()
assert len(warned) == 0, warned[0].message
```

### Issue Description

Pandas supports transparent `~` expansion in IO functions, but Modin is not supporting this pattern everywhere.
In some places we're using `pandas.io.common.get_handle()` which will do this internally, but in others we'll just check for things like `os.path.exists(path)` or `os.path.isdir(path)`, and such checks fail for non-expanded paths, leading to defaulting to pandas for perfectly local files, which could degrade our IO performance significantly.

This also leads to biting some unaware users [like this](https://github.com/MrPowers/pydata-examples/blob/42e9f15b8180a343688ad434209dd881c9861e31/modin/modin_h2o_groupby_queries.ipynb):
image
So later they don't see as much speedup as we were claiming in our docs...

### Expected Behavior

Code example above should not be failing with assertion errors - we should not be defaulting to pandas here.

### Error Logs

<details>

```python-traceback

  File "b.py", line 15, in <module>
    assert len(warned) == 0, warned[0].message
AssertionError: `read_csv` is not currently supported by PandasOnPython, defaulting to pandas implementation.

```

</details>

### Installed Versions

<details>

```
INSTALLED VERSIONS
------------------
commit           : d6eb58911cc3ac6ec788f1d0d4c67b5c38bec297
python           : 3.10.10.final.0
python-bits      : 64
OS               : Windows
OS-release       : 10
Version          : 10.0.22621
machine          : AMD64
processor        : AMD64 Family 25 Model 80 Stepping 0, AuthenticAMD
byteorder        : little
LC_ALL           : None
LANG             : None
LOCALE           : Russian_Russia.1251

Modin dependencies
------------------
modin            : 0.23.0+22.gd6eb5891.dirty
ray              : 2.4.0
dask             : 2023.6.0
distributed      : 2023.6.0
hdk              : None

pandas dependencies
-------------------
pandas           : 2.0.2
numpy            : 1.25.0
pytz             : 2023.3
dateutil         : 2.8.2
setuptools       : 67.6.1
pip              : 23.1.2
Cython           : None
pytest           : 7.3.2
hypothesis       : None
sphinx           : 6.1.3
blosc            : None
feather          : 0.4.1
xlsxwriter       : None
lxml.etree       : 4.9.2
html5lib         : None
pymysql          : None
psycopg2         : 2.9.3
jinja2           : 3.1.2
IPython          : 8.12.0
pandas_datareader: None
bs4              : 4.12.2
bottleneck       : None
brotli           :
fastparquet      : 2022.12.0
fsspec           : 2023.6.0
gcsfs            : None
matplotlib       : 3.7.1
numba            : None
numexpr          : 2.7.3
odfpy            : None
openpyxl         : 3.1.2
pandas_gbq       : 0.15.0
pyarrow          : 12.0.0
pyreadstat       : None
pyxlsb           : None
s3fs             : 2023.6.0
scipy            : 1.10.1
snappy           : None
sqlalchemy       : 1.4.45
tables           : 3.8.0
tabulate         : None
xarray           : 2023.5.0
xlrd             : 2.0.1
zstandard        : None
tzdata           : 2023.3
qtpy             : 2.3.1
pyqt5            : None
```

</details>

vnlitvinov:
## What do these changes do?
Implemented a decorator which would expand user paths (i.g. `~/foo.txt` or `~user/bar.txt`) passed as given argument in the decorated function. This should stop unnecessary defaulting to pandas for some implementations (where we checked that file existed, which it doesn't until one expands the path).

- [x] Resolves #6436, #3180
- [ ] tests added and passing

Note: supersedes #4724.

### DIFF
--- a/modin/experimental/pandas/io.py
+++ b/modin/experimental/pandas/io.py
@@ -177,6 +177,8 @@ def _make_parser_func(sep: str, funcname: str) -> Callable:
     ----------
     sep : str
         The separator default to use for the parser.
+    funcname : str
+        The name of the generated parser function.
 
     Returns
     -------
--- a/modin/experimental/pandas/io.py
+++ b/modin/experimental/pandas/io.py
@@ -25,8 +25,11 @@ from pandas._typing import CompressionOptions, StorageOptions
 from . import DataFrame
 from modin.config import IsExperimental
 from modin.core.storage_formats import BaseQueryCompiler
+from modin.utils import expanduser_path_arg
+from modin.logging import enable_logging
 
 
+@enable_logging
 def read_sql(
     sql,
     con,
@@ -119,6 +122,8 @@ def read_sql(
     return (DataFrame(query_compiler=qc) for qc in result)
 
 
+@expanduser_path_arg("filepath_or_buffer")
+@enable_logging
 def read_custom_text(
     filepath_or_buffer,
     columns,
@@ -164,7 +169,7 @@ def read_custom_text(
 
 
 # CSV and table
-def _make_parser_func(sep: str) -> Callable:
+def _make_parser_func(sep: str, funcname: str) -> Callable:
     """
     Create a parser function from the given sep.
 
@@ -242,7 +247,8 @@ def _make_parser_func(sep: str) -> Callable:
         return _read(**kwargs)
 
     parser_func.__doc__ = _read.__doc__
-    return parser_func
+    parser_func.__name__ = funcname
+    return expanduser_path_arg()(enable_logging(parser_func))
 
 
 def _read(**kwargs) -> DataFrame:
@@ -296,9 +302,11 @@ def _read(**kwargs) -> DataFrame:
     return DataFrame(query_compiler=pd_obj)
 
 
-read_csv_glob = _make_parser_func(sep=",")
+read_csv_glob = _make_parser_func(sep=",", funcname="read_csv_glob")
 
 
+@expanduser_path_arg("filepath_or_buffer")
+@enable_logging
 def read_pickle_distributed(
     filepath_or_buffer,
     compression: Optional[str] = "infer",
@@ -345,6 +353,8 @@ def read_pickle_distributed(
     return DataFrame(query_compiler=FactoryDispatcher.read_pickle_distributed(**kwargs))
 
 
+@expanduser_path_arg("filepath_or_buffer")
+@enable_logging
 def to_pickle_distributed(
     self,
     filepath_or_buffer,
--- a/modin/experimental/pandas/io.py
+++ b/modin/experimental/pandas/io.py
@@ -26,10 +26,8 @@ from . import DataFrame
 from modin.config import IsExperimental
 from modin.core.storage_formats import BaseQueryCompiler
 from modin.utils import expanduser_path_arg
-from modin.logging import enable_logging
 
 
-@enable_logging
 def read_sql(
     sql,
     con,
@@ -123,7 +121,6 @@ def read_sql(
 
 
 @expanduser_path_arg("filepath_or_buffer")
-@enable_logging
 def read_custom_text(
     filepath_or_buffer,
     columns,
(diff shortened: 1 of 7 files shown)


anmyachev:
Should we close the issue https://github.com/modin-project/modin/issues/3180 then?

vnlitvinov:
> Should we close the issue #3180 then?

Uhh, thanks for finding that issue, I vaguely remembered there was one, but wasn't able to find it myself 🙃 
I should also take the tests from the PR that was aimed at closing that issue I guess...

modin/pandas/test/test_io.py
```diff
@@ -1729,6 +1752,18 @@ def test_read_parquet_s3_with_column_partitioning(self, engine):
         )
 
 
+# Leave this test apart from the test classes, which skip the default to pandas
+# warning check. We want to make sure we are NOT defaulting to pandas for a
+# path relative to user home.
+# TODO(https://github.com/modin-project/modin/issues/3655): Get rid of this
+# commment once we turn all default to pandas messages into errors.
+def test_read_parquet_relative_to_user_home(make_parquet_file):
```
anmyachev, line 1760: We should use `@pytest.mark.filterwarnings(default_to_pandas_ignore_string)` for HDK backend.

modin/experimental/pandas/io.py
```diff
@@ -25,8 +25,11 @@
 from . import DataFrame
 from modin.config import IsExperimental
 from modin.core.storage_formats import BaseQueryCompiler
+from modin.utils import expanduser_path_arg
+from modin.logging import enable_logging
 
 
+@enable_logging
```
anmyachev, line 32: We should `enable_logging` add separately, shouldn't?

modin/pandas/test/utils.py
```diff
@@ -977,7 +978,7 @@ def applyier(module, *args, **kwargs):
         return result
 
     def call_eval_general():
-        eval_general(
+        return eval_general(
```
anmyachev, line 981: Why?

Will need to update `eval_io` docstring.

modin/experimental/pandas/io.py
vnlitvinov: technically yes, it's a separate change... though I wanted to ensure all IO functions (both normal and experimental) were decorated identically

modin/pandas/test/utils.py
vnlitvinov: To enable comparison of reading results... though I've just thought of a different way to achieve the same thing, might as well try it, too.

modin/pandas/test/test_io.py
vnlitvinov: I've embedded the check to same place where we check for `Engine.get() == "Python"`

modin/experimental/pandas/io.py
vnlitvinov: but you're right anyway, so I've reverted this change; 1 PR should be about 1 logical change 🙃

modin/pandas/test/utils.py
vnlitvinov: I've changed the way I test for reading, so this change is no longer necessary; I've reverted that

anmyachev (approved):
LGTM!

modin/pandas/test/test_io.py
vnlitvinov: Hmm, that wasn't entirely correct, so I changed it again 🤣

anmyachev (approved):

PR #6458, opened 2023-08-05
FIX-#6459: support fastparquet>=2023.1.0

### LINKED ISSUE #6459: Unpin `fastparquet`
Motivation: https://github.com/modin-project/modin/pull/6442#issuecomment-1666379460

anmyachev:
## What do these changes do?

- [x] Resolves #6459 
- [x] tests added and passing

### DIFF
--- a/environment-dev.yml
+++ b/environment-dev.yml
@@ -37,7 +37,7 @@ dependencies:
   # pymssql==2.2.8 broken: https://github.com/modin-project/modin/issues/6429
   - pymssql>=2.1.5,!=2.2.8
   - psycopg2>=2.8.6
-  - fastparquet>=0.6.3,<2023.1.0
+  - fastparquet>=0.6.3
   - tqdm>=4.60.0
 
   # dependencies for making release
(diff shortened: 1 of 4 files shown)


environment-dev.yml
```diff
@@ -37,7 +37,7 @@ dependencies:
   # pymssql==2.2.8 broken: https://github.com/modin-project/modin/issues/6429
   - pymssql>=2.1.5,!=2.2.8
   - psycopg2>=2.8.6
-  - fastparquet>=0.6.3,<2023.1.0
+  - fastparquet>=0.6.3
```
vnlitvinov, line 40: I wonder if we should really bump the _lower_ bound instead, to something like `>2023.1.0`?.. as, if I understand correctly, earlier versions of `fastparquet` aren't properly compatible with pandas 2.0.

anmyachev: On the one hand, yes, on the other hand, pandas still supports fairly old versions. I guess there are rare compatibility issues, so they don't raise the minimum version. I think we should do the same.

vnlitvinov: Yeah, we should probably just mirror what pandas requires on `fastparquet`...

vnlitvinov (approved):
LGTM!

Issue #2533, opened 2020-12-14
File leaks during read_csv and read_fwf tests on Windows platform


amyskov:
### System information
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: Windows 10
- **Modin version**: 1a8cd0a3765df4cd9f76480e9862f522818333b2 + patch (listed below)
- **Python version**: 3.8.6
- **Code we can use to reproduce**:
patch:
```
diff --git a/modin/pandas/test/test_io.py b/modin/pandas/test/test_io.py
index 9342980..db14fc6 100644
--- a/modin/pandas/test/test_io.py
+++ b/modin/pandas/test/test_io.py
@@ -309,7 +309,7 @@ def make_csv_file():
             try:
                 os.remove(filename)
             except PermissionError:
-                pass
+                raise


 @pytest.fixture(scope="class")
@@ -357,7 +357,7 @@ def TestReadCSVFixture():
             try:
                 os.remove(filename)
             except PermissionError:
-                pass
+                raise


 def setup_json_file(row_size, force=False):
@@ -420,7 +420,7 @@ def teardown_excel_file():
         try:
             os.remove(TEST_EXCEL_FILENAME)
         except PermissionError:
-            pass
+            raise


 def setup_feather_file(row_size, force=False):
@@ -555,7 +555,7 @@ def teardown_fwf_file():
         try:
             os.remove(TEST_FWF_FILENAME)
         except PermissionError:
-            pass
+            raise


 @pytest.mark.usefixtures("TestReadCSVFixture")
@@ -2305,7 +2305,7 @@ def test_cleanup():
             try:
                 os.remove(f)
             except PermissionError:
-                pass
+                raise


 def test_from_arrow():
```
After patch applying run `pytest modin/pandas/test/test_io.py`

### Describe the problem

Looks like files are not properly closed as it was in https://github.com/modin-project/modin/pull/2514.

### Source code / logs

pytest logs:
```
========================================================================= short test summary info =========================================================================
FAILED modin/pandas/test/test_io.py::test_fwf_file_index_col - PermissionError: [WinError 32] The process cannot access the file because it is being used by another proc...
ERROR modin/pandas/test/test_io.py::TestReadCSV::test_read_csv_internal[None-True-True-True-,-python-False] - PermissionError: [WinError 32] The process cannot access th...
[...]

PR #6461, opened 2023-08-05
TEST-#6460: don't use 'repr' to force materialization

### LINKED ISSUE #6460: Do not use `repr` function to materialize the result of a function execution
`repr` function does not fully materialize Modin's dataframe or series now.

anmyachev:
## What do these changes do?

- [x] Resolves #6460 
- [x] tests added and passing

### DIFF
--- a/modin/core/dataframe/pandas/partitioning/partition_manager.py
+++ b/modin/core/dataframe/pandas/partitioning/partition_manager.py
@@ -879,16 +879,9 @@ class PandasDataframePartitionManager(ClassLogger, ABC):
             assert all(
                 [len(partition.list_of_blocks) == 1 for partition in partitions]
             ), "Implementation assumes that each partition contains a single block."
-            blocks_to_materialize = [
-                partition.list_of_blocks[0] for partition in partitions
-            ]
-            try:
-                result = cls._execution_wrapper.materialize(blocks_to_materialize)
-            except Exception:
-                # wait for the end of computations so that the processes free up the resources used
-                cls._execution_wrapper.wait(blocks_to_materialize)
-                raise
-            return result
+            return cls._execution_wrapper.materialize(
+                [partition.list_of_blocks[0] for partition in partitions]
+            )
         return [partition.get() for partition in partitions]
 
     @classmethod
--- a/modin/core/dataframe/pandas/partitioning/partition_manager.py
+++ b/modin/core/dataframe/pandas/partitioning/partition_manager.py
@@ -879,9 +879,16 @@ class PandasDataframePartitionManager(ClassLogger, ABC):
             assert all(
                 [len(partition.list_of_blocks) == 1 for partition in partitions]
             ), "Implementation assumes that each partition contains a single block."
-            return cls._execution_wrapper.materialize(
-                [partition.list_of_blocks[0] for partition in partitions]
-            )
+            blocks_to_materialize = [
+                partition.list_of_blocks[0] for partition in partitions
+            ]
+            try:
+                result = cls._execution_wrapper.materialize(blocks_to_materialize)
+            except Exception:
+                # wait for the end of computations so that the processes free up the resources used
+                cls._execution_wrapper.wait(blocks_to_materialize)
+                raise
+            return result
         return [partition.get() for partition in partitions]
 
     @classmethod
--- a/modin/experimental/pandas/test/test_io_exp.py
+++ b/modin/experimental/pandas/test/test_io_exp.py
@@ -30,6 +30,7 @@ from modin.pandas.test.utils import (
 )
 from modin.test.test_utils import warns_that_defaulting_to_pandas
 from modin.pandas.test.utils import parse_dates_values_by_id, time_parsing_csv_path
+from modin.utils import try_cast_to_pandas
 
 
 @pytest.mark.skipif(
@@ -159,8 +160,7 @@ class TestCsvGlob:
                 modin_df = pd.read_csv_glob(
                     time_parsing_csv_path, parse_dates=parse_dates
                 )
-                # Call _to_pandas on the modin df to force it to materialize.
-                _ = modin_df._to_pandas()
+                try_cast_to_pandas(modin_df)  # force materialization
             assert isinstance(
                 modin_exception.value, type(pandas_exception)
             ), "Got Modin Exception type {}, but pandas Exception type {} was expected".format(
--- a/modin/experimental/pandas/test/test_io_exp.py
+++ b/modin/experimental/pandas/test/test_io_exp.py
@@ -159,8 +159,8 @@ class TestCsvGlob:
                 modin_df = pd.read_csv_glob(
                     time_parsing_csv_path, parse_dates=parse_dates
                 )
-                # Call __repr__ on the modin df to force it to materialize.
-                repr(modin_df)
+                # Call _to_pandas on the modin df to force it to materialize.
+                _ = modin_df._to_pandas()
             assert isinstance(
                 modin_exception.value, type(pandas_exception)
             ), "Got Modin Exception type {}, but pandas Exception type {} was expected".format(
--- a/modin/pandas/test/test_series.py
+++ b/modin/pandas/test/test_series.py
@@ -26,7 +26,7 @@ import matplotlib
 import modin.pandas as pd
 from numpy.testing import assert_array_equal
 
-from modin.utils import get_current_execution
+from modin.utils import get_current_execution, try_cast_to_pandas
 from modin.test.test_utils import warns_that_defaulting_to_pandas
 
 from modin.utils import to_pandas
@@ -134,9 +134,7 @@ def inter_df_math_helper_one_side(
         pandas_result = pandas_attr(4)
     except Exception as err:
         with pytest.raises(type(err)):
-            res = modin_attr(4)
-            if hasattr(res, "_to_pandas"):
-                res._to_pandas()
+            try_cast_to_pandas(modin_attr(4))  # force materialization
     else:
         modin_result = modin_attr(4)
         df_equals(modin_result, pandas_result, **comparator_kwargs)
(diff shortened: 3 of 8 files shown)


modin/experimental/pandas/test/test_io_exp.py
```diff
@@ -159,8 +159,8 @@ def test_read_single_csv_with_parse_dates(self, parse_dates):
                 modin_df = pd.read_csv_glob(
                     time_parsing_csv_path, parse_dates=parse_dates
                 )
-                # Call __repr__ on the modin df to force it to materialize.
-                repr(modin_df)
+                # Call _to_pandas on the modin df to force it to materialize.
+                _ = modin_df._to_pandas()
```
vnlitvinov, line 163: Maybe we should use public `utils.to_pandas()` function instead? The protected (name starting with `_`) is called so for a reason, and is subject to (potential) change.

modin/pandas/test/test_series.py
```diff
@@ -135,7 +135,9 @@ def inter_df_math_helper_one_side(
         pandas_result = pandas_attr(4)
     except Exception as err:
         with pytest.raises(type(err)):
-            repr(modin_attr(4))  # repr to force materialization
+            res = modin_attr(4)
+            if hasattr(res, "_to_pandas"):
+                res._to_pandas()
```
vnlitvinov, line 140:
```python
res = utils.try_cast_to_pandas(modin_attr(4))
```
?

modin/pandas/test/test_series.py
```diff
@@ -144,7 +146,9 @@ def inter_df_math_helper_one_side(
         pandas_result = pandas_attr(4.0)
     except Exception as err:
         with pytest.raises(type(err)):
-            repr(modin_attr(4.0))  # repr to force materialization
+            res = modin_attr(4.0)
+            if hasattr(res, "_to_pandas"):
+                res._to_pandas()
```
vnlitvinov, line 151:
```python
res = utils.try_cast_to_pandas(modin_attr(4))
```
?

modin/pandas/test/test_series.py
```diff
@@ -166,35 +170,34 @@ def inter_df_math_helper_one_side(
...
-    except Exception as err:
-        with pytest.raises(type(err)):
-            repr(modin_attr(modin_series))  # repr to force materialization
-    else:
-        modin_result = modin_attr(modin_series)
-        df_equals(modin_result, pandas_result, **comparator_kwargs)
+    eval_general(
+        modin_series,
+        pandas_series,
+        lambda df: pandas_attr(df) if isinstance(pandas.Series) else modin_attr(df),
```
vnlitvinov, line 176:
```suggestion
        lambda df: (pandas_attr if isinstance(df, pandas.Series) else modin_attr)(df),
```
how does this pass any tests with this typo btw? 🤣 (missing argument to `isinstance()`)

modin/pandas/test/test_series.py
```diff
@@ -166,35 +170,34 @@ def inter_df_math_helper_one_side(
...
 
     list_test = random_state.randint(RAND_LOW, RAND_HIGH, size=(modin_series.shape[0]))
     try:
         pandas_result = pandas_attr(list_test)
     except Exception as err:
         with pytest.raises(type(err)):
-            repr(modin_attr(list_test))  # repr to force materialization
+            res = modin_attr(list_test)
+            if hasattr(res, "_to_pandas"):
+                res._to_pandas()
```
vnlitvinov, line 187:
```python
res = utils.try_cast_to_pandas(modin_attr(4))
```
?

modin/pandas/test/test_series.py
```diff
@@ -166,35 +170,34 @@ def inter_df_math_helper_one_side(
...
-        with pytest.raises(type(err)):
-            repr(modin_attr(series_test_modin))  # repr to force materialization
-    else:
-        modin_result = modin_attr(series_test_modin)
-        df_equals(modin_result, pandas_result, **comparator_kwargs)
+
+    eval_general(
+        series_test_modin,
+        series_test_pandas,
+        lambda df: pandas_attr(df) if isinstance(pandas.Series) else modin_attr(df),
```
vnlitvinov, line 198:
```suggestion
        lambda df: (pandas_attr if isinstance(df, pandas.Series) else modin_attr)(df),
```

modin/pandas/test/utils.py
```diff
@@ -889,8 +889,10 @@ def execute_callable(fn, inplace=False, md_kwargs={}, pd_kwargs={}):
             if check_exception_type is None:
                 return None
             with pytest.raises(Exception) as md_e:
-                # repr to force materialization
-                repr(fn(modin_df, **md_kwargs))
+                res = fn(modin_df, **md_kwargs)
+                if hasattr(res, "_to_pandas"):
+                    # force materialization
+                    _ = res._to_pandas()
```
vnlitvinov, line 895:
```suggestion
                utils.try_cast_to_pandas(fn(modin_df, **md_kwargs))  # force materialization
```

vnlitvinov (changes requested):
There seems to be at least two places with missing arguments... and I've left a few suggestions on using "more public" API instead of protected `._to_pandas()`

vnlitvinov:
Also @anmyachev I'm not entirely sure this fixes the problem unearthed in https://github.com/modin-project/modin/pull/6442 - I think [the analysis](https://github.com/modin-project/modin/pull/6442#issuecomment-1666379460) there rightfully assumes that "_at least on Ray, that error is propagated from the worker to the main process as soon as it happens, which means that another worker process may still be doing stuff in the background. The error "completes" the test, so we think we are ready for teardown_", and I don't think `._to_pandas()` would be any different here - i.e. that it would wait for _all_ tasks to finish before erroring out. I suppose we'd have to use `mgr.wait_partitions()` or something similar somehow instead.

modin/pandas/test/test_series.py
anmyachev: both pandas and modin threw the same exception and the test passed. Still, this behavior of `eval_general` function is very dangerous.

anmyachev: good catch!

anmyachev:
> Also @anmyachev I'm not entirely sure this fixes the problem unearthed in #6442 - I think [the analysis](https://github.com/modin-project/modin/pull/6442#issuecomment-1666379460) there rightfully assumes that "_at least on Ray, that error is propagated from the worker to the main process as soon as it happens, which means that another worker process may still be doing stuff in the background. The error "completes" the test, so we think we are ready for teardown_", and I don't think `._to_pandas()` would be any different here - i.e. that it would wait for _all_ tasks to finish before erroring out. I suppose we'd have to use `mgr.wait_partitions()` or something similar somehow instead.

Actually, in order to get an exception on the main process, it must be explicitly materialized through `ray.get` function.
Consider the example:
```python
import ray
import time
ray.init(num_cpus=4)

def example():
    raise ValueError("something isn't working")

rem_example = ray.remote(example)
future = rem_example.remote() # call it
time.sleep(10)  # to make sure exception is happened in remote `example` function
print("after sleep")  # this code works even if a exception is happened

ray.get(future)  # get the exception on the main process
print("after get")  # this code doesn't work, there was an exception in the previous line

```

vnlitvinov:
> The second is that if an error occurs in one of the partitions, then we do not wait for the materialization of other partitions to finish, where resources may still be occupied.

Yeah, that was the problem I talked about.

For the `repr()` building only bits of the dataframe - this is something I didn't consider, but you're right here. However I think that, if we're at squashing that "files not closed" bug, we should fix both problems in this PR...

anmyachev:
Need to check if the problem is the root cause for https://github.com/modin-project/modin/issues/2533.

modin/core/dataframe/pandas/partitioning/partition_manager.py
```diff
@@ -871,9 +871,16 @@ def get_objects_from_partitions(cls, partitions):
...
-            )
+            blocks_to_materialize = [
+                partition.list_of_blocks[0] for partition in partitions
+            ]
+            try:
+                result = cls._execution_wrapper.materialize(blocks_to_materialize)
+            except Exception:
+                # wait for the end of computations so that the processes free up the resources used
+                cls._execution_wrapper.wait(blocks_to_materialize)
+                raise
```
anmyachev, line 882: @YarShev this construction causes the unidist to hangs, although dask and ray work. Can you suggest what's wrong?

vnlitvinov:
> Need to check if the problem is the root cause for #2533.

For that we first need to see if that issue is still occurring... 😺

modin/core/dataframe/pandas/partitioning/partition_manager.py
vnlitvinov, line 882: I don't think this is exactly the right place to put this synchronization point... At least it should be waited conditionally based on some configuration variable, as I would think that in most scenarios the user doesn't want to wait for all computations when it's already known that an exception is happening.

Also, this is fixing a problem which isn't entirely related to the original issue, so maybe it's better to open a new separate PR with this change.

modin/core/dataframe/pandas/partitioning/partition_manager.py
```diff
@@ -871,9 +871,16 @@ def get_objects_from_partitions(cls, partitions):
...
+            return result
```
vnlitvinov, line 883: if that code is to stay, we can make it look nicer:
```suggestion
            try:
                return cls._execution_wrapper.materialize(blocks_to_materialize)
            except Exception:
                # wait for the end of computations so that the processes free up the resources used
                cls._execution_wrapper.wait(blocks_to_materialize)
                raise
```

vnlitvinov (commented):
Overall looks good, but CI is still not green, so cannot approve.

modin/core/dataframe/pandas/partitioning/partition_manager.py
anmyachev: > Also, this is fixing a problem which isn't entirely related to the original issue, so maybe it's better to open a new separate PR with this change.

I agree, let's continue in another PR.

vnlitvinov (commented):
Code at a glance looks good, but tests are failing.
@anmyachev please take a look at CI.

vnlitvinov (approved):
LGTM!

PR #6474, opened 2023-08-08
TEST-#6469: pin numexpr<2.8.5

### LINKED ISSUE #6469: numexpr 2.8.5 breaks CI
Pandas is not compatible with the release.

See https://github.com/pandas-dev/pandas/issues/54449

Broken CI: https://github.com/modin-project/modin/actions/runs/5799349333/job/15721083501

Example of the error:
```bash
def test_eval_and_query_with_local_and_global_var(method, data, local_var):
        modin_df, pandas_df = pd.DataFrame(data), pandas.DataFrame(data)
        op = "+" if method == "eval" else "<"
        for expr in (f"col1 {op} @local_var", f"col1 {op} @TEST_VAR"):
>           df_equals(getattr(modin_df, method)(expr), getattr(pandas_df, method)(expr))

modin/pandas/test/dataframe/test_udf.py:351: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
modin/logging/logger_decorator.py:128: in run_and_log
    return obj(*args, **kwargs)
modin/pandas/dataframe.py:818: in eval
    new_query_compiler = self._query_compiler.eval(expr, **kwargs)
modin/logging/logger_decorator.py:128: in run_and_log
    return obj(*args, **kwargs)
modin/core/storage_formats/pandas/query_compiler.py:[228](https://github.com/modin-project/modin/actions/runs/5799349333/job/15721083501#step:7:229)0: in eval
    pandas.DataFrame(columns=self.columns)
/usr/share/miniconda/envs/modin/lib/python3.8/site-packages/pandas/core/frame.py:4566: in eval
    return _eval(expr, inplace=inplace, **kwargs)
/usr/share/miniconda/envs/modin/lib/python3.8/site-packages/pandas/core/computation/eval.py:357: in eval
    ret = eng_inst.evaluate()
/usr/share/miniconda/envs/modin/lib/python3.8/site-packages/pandas/core/computation/engines.py:81: in evaluate
    res = self._evaluate()
/usr/share/miniconda/envs/modin/lib/python3.8/site-packages/pandas/core/computation/engines.py:121: in _evaluate
    return ne.evaluate(s, local_dict=scope)
/usr/share/miniconda/envs/modin/lib/python3.8/site-packages/numexpr/necompiler.py:943: in evaluate
    raise e
/usr/share/miniconda/envs/modin/lib/python3.8/site-packages/numexpr/necompiler.py:851: in validate
    _names_cache[expr_key] = getExprNames(ex, context)
/usr/share/miniconda/envs/modin/lib/python3.8/site-packages/numexpr/necompiler.py:714: in getExprNames
    ex = stringToExpression(text, {}, context)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

s = '(col1) + (__pd_eval_local_local_var)', types = {}
context = {'optimization': 'aggressive', 'truediv': False}

    def stringToExpression(s, types, context):
        """Given a string, convert it to a tree of ExpressionNode's.
        """
        # sanitize the string for obvious attack vectors that NumExpr cannot
        # parse into its homebrew AST. This is to protect the call to `eval` below.
        # We forbid `;`, `:`. `[` and `__`, and attribute access via '.'.
        # We cannot ban `.real` or `.imag` however...
        no_whitespace = re.sub(r'\s+', '', s)
        if _forbidden_re.search(no_whitespace) is not None:
>           raise ValueError(f'Expression {s} has forbidden control characters.')
E           ValueError: Expression (col1) + (__pd_eval_local_local_var) has forbidden control characters.
```

anmyachev:
## What do these changes do?

Fix CI on master branch.

- [x] Resolves #6469 
- [x] tests added and passing

### DIFF
--- a/environment-dev.yml
+++ b/environment-dev.yml
@@ -39,6 +39,8 @@ dependencies:
   - psycopg2>=2.8.6
   - fastparquet>=0.6.3
   - tqdm>=4.60.0
+  # pandas doesn't compatible with numexpr=2.8.5: https://github.com/modin-project/modin/issues/6469
+  - numexpr<2.8.5
 
   # dependencies for making release
   - pygithub>=v1.58.0
--- a/environment-dev.yml
+++ b/environment-dev.yml
@@ -39,7 +39,7 @@ dependencies:
   - psycopg2>=2.8.6
   - fastparquet>=0.6.3
   - tqdm>=4.60.0
-  # pandas doesn't compatible with numexpr=2.8.5: https://github.com/modin-project/modin/issues/6469
+  # pandas isn't compatible with numexpr=2.8.5: https://github.com/modin-project/modin/issues/6469
   - numexpr<2.8.5
 
   # dependencies for making release
--- a/requirements/env_hdk.yml
+++ b/requirements/env_hdk.yml
@@ -23,6 +23,8 @@ dependencies:
   # - xarray
   - pytables>=3.6.1
   - fastparquet>=0.6.3
+  # pandas doesn't compatible with numexpr=2.8.5: https://github.com/modin-project/modin/issues/6469
+  - numexpr<2.8.5
 
   # dependencies for making release
   - pygithub>=v1.58.0
--- a/requirements/env_hdk.yml
+++ b/requirements/env_hdk.yml
@@ -23,7 +23,7 @@ dependencies:
   # - xarray
   - pytables>=3.6.1
   - fastparquet>=0.6.3
-  # pandas doesn't compatible with numexpr=2.8.5: https://github.com/modin-project/modin/issues/6469
+  # pandas isn't compatible with numexpr=2.8.5: https://github.com/modin-project/modin/issues/6469
   - numexpr<2.8.5
 
   # dependencies for making release
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -31,7 +31,7 @@ connectorx>=0.2.6a4
 fastparquet>=0.6.3
 flask-cors
 tqdm>=4.60.0
-# pandas doesn't compatible with numexpr=2.8.5: https://github.com/modin-project/modin/issues/6469
+# pandas isn't compatible with numexpr=2.8.5: https://github.com/modin-project/modin/issues/6469
 numexpr<2.8.5
 # Latest modin-spreadsheet with widget fix
 git+https://github.com/modin-project/modin-spreadsheet.git@49ffd89f683f54c311867d602c55443fb11bf2a5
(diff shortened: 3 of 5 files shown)


environment-dev.yml
```diff
@@ -39,6 +39,8 @@ dependencies:
   - psycopg2>=2.8.6
   - fastparquet>=0.6.3
   - tqdm>=4.60.0
+  # pandas doesn't compatible with numexpr=2.8.5: https://github.com/modin-project/modin/issues/6469
```
vnlitvinov, line 42:
```suggestion
  # pandas isn't compatible with numexpr=2.8.5: https://github.com/modin-project/modin/issues/6469
```

requirements/env_hdk.yml
```diff
@@ -23,6 +23,8 @@ dependencies:
   # - xarray
   - pytables>=3.6.1
   - fastparquet>=0.6.3
+  # pandas doesn't compatible with numexpr=2.8.5: https://github.com/modin-project/modin/issues/6469
```
vnlitvinov, line 26:
```suggestion
  # pandas isn't compatible with numexpr=2.8.5: https://github.com/modin-project/modin/issues/6469
```

vnlitvinov (commented):
Changes LGTM, but let's fix the English 🙃

vnlitvinov (approved):
LGTM, thanks @anmyachev!

YarShev:
I think we should have the pin in setup.py too. If I installed modin from source, I would get an error like in the issue, right?

anmyachev:
> I think we should have the pin in setup.py too. If I installed modin from source, I would get an error like in the issue, right?

This is an optional dependency and should not be installed with Modin by default.

YarShev:
pydantic is also an optional dependency but we have a pin for it in setup.py.

anmyachev:
IIRC `pydantic` is installed with `ray[default]`. It is not optional for `modin[ray]`.

YarShev:
Anyway, if Modin doesn't work out of the box without the pin for numexpr, I think we should fix the things. Would you mind to open a PR so other maintaners take a look at the fix?

anmyachev:
> Anyway, if Modin doesn't work out of the box without the pin for numexpr, I think we should fix the things. Would you mind to open a PR so other maintaners take a look at the fix?

It's more quickly just ping @modin-project/modin-core here

PR #6579, opened 2023-09-18
FIX-#6578: ignore `FutureWarning`s in pandas kernel which used for defaulting to pandas

### LINKED ISSUE #6578: `align` func shows `FutureWarning`s in cases where they are not needed.

anmyachev:
## What do these changes do?

~Now there is only a basic implementation that defaults to pandas. We may not process parameters on the Modin side.~

- [x] Resolves #6578 
- [x] tests added and passing

### DIFF
--- a/modin/core/dataframe/algebra/default2pandas/default.py
+++ b/modin/core/dataframe/algebra/default2pandas/default.py
@@ -13,6 +13,8 @@
 
 """Module houses default functions builder class."""
 
+import warnings
+
 import pandas
 from pandas.core.dtypes.common import is_list_like
 
@@ -117,7 +119,9 @@ class DefaultMethod(Operator):
             # pandas default implementation doesn't know how to handle `dtypes` keyword argument
             kwargs.pop("dtypes", None)
             df = cls.frame_wrapper(df)
-            result = fn(df, *args, **kwargs)
+            with warnings.catch_warnings():
+                warnings.filterwarnings("ignore", category=FutureWarning)
+                result = fn(df, *args, **kwargs)
 
             if (
                 not isinstance(result, pandas.Series)
--- a/modin/pandas/base.py
+++ b/modin/pandas/base.py
@@ -749,6 +749,45 @@ class BasePandasDataset(ClassLogger):
         """
         Align two objects on their axes with the specified join method.
         """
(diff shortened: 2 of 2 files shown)


vnlitvinov (changes requested):
I don't understand why we are not giving these warnings. Higher-level API should not depend on lower-level implementation details to give warnings about these, as implementation might change without notice for higher-level parts.

On the contrary, it's lower level that ought to suppress the warnings that high-level had already given, if such warnings arise in the implementation.

IMHO, this change is not solving the original problem, but rather replacing it with another one 🙃

vnlitvinov:
> There is a high probability that we will filter out the warning that Modin does not show, that is, a situation will arise when the user does not receive a warning at all, which in my understanding is a bad decision.

Given Modin strives to be pandas drop-in replacement with whatever backend user picks, the _only_ place that should be raising warnings is the high-level API layer. These changes mean that on non-pandas backend the user will _not_ get the warnings you've removed here at all.

As for missing some warnings because Modin does not give them - this would be a bug in Modin, and relying on lower-level to give those for us is a workaround at best.

anmyachev:
> > There is a high probability that we will filter out the warning that Modin does not show, that is, a situation will arise when the user does not receive a warning at all, which in my understanding is a bad decision.
> 
> Given Modin strives to be pandas drop-in replacement with whatever backend user picks, the _only_ place that should be raising warnings is the high-level API layer. The changes in this PR mean that on non-pandas backend the user will _not_ get the warnings you've removed here at all.
> 
> As for missing some warnings because Modin does not give them - this would be a bug in Modin, and relying on lower-level to give those for us is a workaround at best. So the proper solution is to silence _all_ pandas warnings from the query compiler layer and explicitly raise them ourselves at API layer, IMO.

I agree, I also considered these changes as a workaround, given that we need to release it soon. However, I still changed my approach towards what you are talking about. Now all these `FutureWarning`s are filtered in the pandas kernel for operations that are defaulted by the pandas (it's enough for `align` op). I propose to correct all other places and other types of warnings separately.

anmyachev:
Blocked on #6595

anmyachev:
Will be fixed in #6595

PR #6652, opened 2023-10-16
FIX-#0000: follow up PR with fixes for #5634

anmyachev:
## What do these changes do?

https://github.com/modin-project/modin/pull/6650

- [ ] Resolves #? 
- [x] tests added and passing

### DIFF
--- a/modin/core/execution/unidist/implementations/pandas_on_unidist/partitioning/partition.py
+++ b/modin/core/execution/unidist/implementations/pandas_on_unidist/partitioning/partition.py
@@ -15,7 +15,6 @@
 
 import warnings
 
-import pandas
 import unidist
 
 from modin.core.dataframe.pandas.partitioning.partition import PandasDataframePartition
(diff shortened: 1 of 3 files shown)


anmyachev:
@YarShev

modin/core/execution/unidist/implementations/pandas_on_unidist/partitioning/partition.py
```diff
@@ -302,7 +301,7 @@ def ip(self, materialize=True):
             if len(self.call_queue):
                 self.drain_call_queue()
             else:
-                self._ip_cache = self.apply(lambda df: pandas.DataFrame([]))._ip_cache
+                self._ip_cache = self.apply(lambda df: None)._ip_cache
```
anmyachev, line 304: On a cluster, this may not be correct, since if all the processes of one node are busy, then the engine can start the calculation on another node, which will lead to incorrect receipt of the address.

One of the most likely solutions would be to return the entire dataframe and completely update the partition (all references to remote objects).

It may potentially be possible to use special API engines to obtain the address more efficiently. (for example, like in https://github.com/ray-project/ray/issues/13630)

@YarShev thanks for the explanation and solution proposal!

PR #6698, opened 2023-11-03
FIX-#6703: don't use `set_index_name(None)`

### LINKED ISSUE #6703: Executing `_query_compiler.set_index_name(None)` seems unnecessary.

anmyachev:
## What do these changes do?

This code was added quite a long time ago. I don’t know exactly which case this fixes, but all tests pass. I suggest deleting it, since it materializes the index (and it seems to be unnecessary).

- [x] Resolves #6703 
- [x] tests added and passing

### DIFF
--- a/modin/pandas/groupby.py
+++ b/modin/pandas/groupby.py
@@ -58,7 +58,6 @@ _DEFAULT_BEHAVIOUR = {
     "_axis",
     "_by",
     "_check_index",
-    "_check_index_name",
     "_columns",
     "_compute_index_grouped",
     "_default_to_pandas",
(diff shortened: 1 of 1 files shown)


dchigarev (approved):
Just for the history: the `set_index_name(None)` was needed in the past in order to avoid propagation of the `__reduced__` name into the index:
```python
>>> df = pd.DataFrame({"a": [2, 1, 3, 4], "b": [3, 4, 5, 6]})
>>> sr = pd.Series([1, 1, 2, 2])
>>> sr._query_compiler.columns
Index(['__reduced__'], dtype='object')
>>> df.groupby(sr).sum()
             a   b
__reduced__
1            3   7
2            7  11
```
The `_check_index_name()` logic was explicitly renaming index names to `None`. However, at some point in time, this logic was moved to the inside of groupby kernels [[1]](https://github.com/modin-project/modin/blob/d117e99f151284d8596fceb9c27078747c3fa3da/modin/core/dataframe/algebra/groupby.py#L280-L281) [[2]](https://github.com/modin-project/modin/blob/d117e99f151284d8596fceb9c27078747c3fa3da/modin/core/storage_formats/pandas/query_compiler.py#L4008-L4015), and it seems that we forgot to remove the logic from front-end.

**Q:** maybe this logic was kept there for HDK?
**A:** no, HDK only supports grouping on columns of the frame, so the case of grouping on an unnamed Series will be defaulted to pandas and so be handled correctly

PR #6699, opened 2023-11-03
PERF-#6702: don't materialize axes when calling `to_numpy`

### LINKED ISSUE #6702: Don't materialize axes when calling `to_numpy`

anmyachev:
## What do these changes do?

- [x] Resolves #6702 
- [x] tests added and passing

### DIFF
--- a/modin/core/dataframe/pandas/dataframe/dataframe.py
+++ b/modin/core/dataframe/pandas/dataframe/dataframe.py
@@ -4068,7 +4068,14 @@ class PandasDataframe(ClassLogger):
         -------
         np.ndarray
         """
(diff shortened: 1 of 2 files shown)


dchigarev (approved):
the obvious thing to mention is that we are losing the ability to catch errors when index/columns are not materialized, but I guess it's okay?

anmyachev:
> the obvious thing to mention is that we are losing the ability to catch errors when index/columns are not materialized, but I guess it's okay?

Not really, when the index is not materialized, we calculate it from internal partitions, in this case the check does not make sense. We do the same for `to_pandas`.

anmyachev:
@dchigarev let's merge it

PR #6725, opened 2023-11-07
PERF-#0000: Don't materialize index in `groupby` where `by` hashable

anmyachev:
## What do these changes do?

- [ ] Resolves #? 
- [ ] tests added and passing

### DIFF
--- a/modin/pandas/dataframe.py
+++ b/modin/pandas/dataframe.py
@@ -498,7 +498,12 @@ class DataFrame(BasePandasDataset):
         elif hashable(by) and not isinstance(by, (pandas.Grouper, FrozenList)):
             drop = by in self.columns
             idx_name = by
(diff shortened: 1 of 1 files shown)


modin/pandas/dataframe.py
```diff
@@ -497,7 +497,12 @@ def groupby(
         elif hashable(by) and not isinstance(by, (pandas.Grouper, FrozenList)):
             drop = by in self.columns
             idx_name = by
-            if by is not None and by in self._query_compiler.get_index_names(axis):
+            if (
+                by is not None
+                # `by` is already in `self.columns`
+                and not drop
+                and by in self._query_compiler.get_index_names(axis)
+            ):
```
anmyachev, line 505: @dchigarev what do you think about? As far as I know, `by` cannot be in both the columns and the index names at the same time; in this case, Pandas gives an error. So in this case efficiency is not needed.

PR #6764, opened 2023-11-23
FEAT-#0000: make pickling methods available when experimental mode is activated in runtime

anmyachev:
## What do these changes do?

- [ ] Resolves #? 
- [ ] tests added and passing


modin/experimental/pandas/__init__.py
```diff
@@ -48,6 +48,13 @@
     to_pickle_distributed,
 )
 
+
+class ExpDataFrame(DataFrame):  # noqa: F405
```
anmyachev, line 52: The new class is needed so that class modifications (for example: from `setattr` usage) do not affect the class that work in normal mode; this greatly simplifies the organization of interaction between the experimental context and the normal one.

anmyachev: Outdated

modin/utils.py
```diff
@@ -628,6 +630,43 @@ def execute(*objs: Iterable[Any], trigger_hdk_import: bool = False) -> None:
...
+    from modin.config import IsExperimental
+    from modin.core.execution.dispatching.factories.dispatcher import FactoryDispatcher
+
+    old_value = IsExperimental.get()
+    try:
+        IsExperimental.put(True)
+        FactoryDispatcher._update_factory_if_already_initialized()
+        yield
+    finally:
+        IsExperimental.put(old_value or False)
```
anmyachev, line 665: `old_value` can be `None`

modin/pandas/accessor.py
```diff
@@ -191,3 +195,72 @@ def __get__(self, obj, cls):
...
+    Namespace class for accessing experimental functionality.
+
+    Parameters
+    ----------
+    data : DataFrame or Series
+        Object to operate on.
+    """
+
+    def __init__(self, data):
+        if IsExperimental.get() is not True:
```
dchigarev, line 211: Can we get rid of this requirement? I mean, if a user already gone that far by accessing an underscore `df._exp` field, then he probably understands what he's doing and his intention to use experimental API is certain enough.

dchigarev: we could have added a decorator for each of the methods in `ExperimentalFunctions` that would enable/disable experimental mode in the fly

anmyachev: @dchigarev done, ready for review

modin/utils.py
```diff
@@ -628,6 +630,43 @@ def execute(*objs: Iterable[Any], trigger_hdk_import: bool = False) -> None:
...
+    >>>   df._exp.to_pickle_distributed("transactions_train_new*.pkl")
+    >>>   new_df = read_pickle_distributed("transactions_train_new*.pkl")
+    """
+    from modin.config import IsExperimental
+    from modin.core.execution.dispatching.factories.dispatcher import FactoryDispatcher
+
+    old_value = IsExperimental.get()
+    try:
+        IsExperimental.put(True)
+        FactoryDispatcher._update_factory_if_already_initialized()
```
dchigarev, line 662: can we hide the factory updating to the `IsExperimental.put()` method? We can then make this context-lib function generic, so it could work with other config variables that we tend to change often (for example, `ExperimentalGroupbyImpl`, `ExperimentalNumpyAPI`, `AsyncReading`, ...):
```python
with modin.utils.set_config_value(IsExperimental, value=True):
    ...
```

anmyachev: An interesting idea, it seems possible, but let's check it separately.

modin/experimental/pandas/__init__.py
```diff
@@ -45,11 +45,8 @@
     read_custom_text,
     read_pickle_distributed,
     read_sql,
-    to_pickle_distributed,
 )
 
-setattr(DataFrame, "to_pickle_distributed", to_pickle_distributed)  # noqa: F405
-
```
dchigarev, line 52: why? it seems like it breaks the old code:
```python
from modin.config import IsExperimental
IsExperimental.put(True)

import modin.experimental.pandas as pd
pd.DataFrame(...).to_pickle_distributed() # fails
```

dchigarev: I would expect the new behavior via `._exp` to be adding a new way of invoking experimental method, but not replacing the previous one

anmyachev: Yes, it breaks. Initially, this was done without taking into account the case when the experimental mode should work in conjunction with the regular one. Updating attributes during import makes context switching very difficult.

Since this is experimental code, I thought that such a breaking change would not be a big problem.

dchigarev: > Updating attributes during import makes context switching very difficult

why? shouldn't `enable_exp_mode` context do nothing if the experimental mode is already enabled?

anmyachev: > enable_exp_mode context do nothing if the experimental mode is already enabled?

Yes, but what if `import modin.experimental.pandas` is made in this context? The expected behavior would be that when exiting the context, no side effects remain.

dchigarev: Well, it still should be fine then. If I understand correctly, after we leave the context, the experimental methods of a DataFrame will still be accessible but not functioning, which seems like a pretty logical behavior
```python
with enable_exp_mode():
    import modin.experimental.pandas as pd
    pd.DataFrame([]).to_pickle_distributed() # works

pd.DataFrame([]).to_pickle_distributed() # raises: not in experimental mode
```

anmyachev: I brought back this function, but added a warning about future removal and an alternative to use. I hope this suits you.

YarShev:
Maybe we should initialize both a primary factory and a respective expetimental factory at once and expose experimental pandas API that is slightly different to the original one? Doing so we could avoid an extra accessor `_exp` and a new context manager `enable_exp_mode`, which seems to me not user-friendly a bit.

anmyachev:
> and expose experimental pandas API that is slightly different to the original one

Do you mean to provide access to experimental functions in the same namespace as regular functions, but which will only work when experimental mode is enabled?

> Maybe we should initialize both a primary factory and a respective expetimental factory at once

Moreover, we could simply combine them into one.

YarShev:
> Do you mean to provide access to experimental functions in the same namespace as regular functions, but which will only work when experimental mode is enabled?

Through `modin.experimental.pandas` and the experimental API shouldn't be depending on `MODIN_EXPERIMENTAL`.

> Moreover, we could simply combine them into one.

We could probably combine some of them into one, and, I think since we have the executions such as ExpetimentalPyarrowOnRay and ExperimentalHdkOnNative, we should provide them when MODIN_EXPERIMENTAL is enabled, i.e. as is now.

anmyachev:
> Through modin.experimental.pandas and the experimental API shouldn't be depending on MODIN_EXPERIMENTAL.

@YarShev I don’t really understand how in this case we will be able to provide access to experimental dataframe methods. Could you provide some more details?

YarShev:
> @YarShev I don’t really understand how in this case we will be able to provide access to experimental dataframe methods. Could you provide some more details?

Just like we provide access to the experimental xgboost API. We do not require users to set `MODIN_EXPERIMENTAL` for this. For experimental dataframe APIs, we would initialize both factories (main and experimental) and expose the API same way. `MODIN_EXPERIMENTAL` would only be responsible for enabling the executions we consider as experimental.

anmyachev:
> > @YarShev I don’t really understand how in this case we will be able to provide access to experimental dataframe methods. Could you provide some more details?
> 
> Just like we provide access to the experimental xgboost API. We do not require users to set `MODIN_EXPERIMENTAL` for this. For experimental dataframe APIs, we would initialize both factories (main and experimental) and expose the API same way. `MODIN_EXPERIMENTAL` would only be responsible for enabling the executions we consider as experimental.

Ok, I think I partially understand. This will help remove the new context manager that I added. However, I still don't understand how this will help get rid of the new `_exp` accessor.

YarShev:
If we provide experimental APIs that is slightly different to the original one (read_sql -> read_sql_exp, e.g.), I think `._exp` can be avoided.

dchigarev:
> @dchigarev your thoughts?

fine with me

YarShev:
> @YarShev I still wouldn’t want to confuse users with functions that are not available in Pandas. Besides, I think @devin-petersohn would be against it. However, what if we added not _exp, but modin attribute. This would clearly show the user that the functionality is missing in pandas.

Now I got your thoughts. I see two options. 1) provide `._exp` attribute or 2) make all experimental APIs as free functions accepting Modin object(s) if necessary. I am better to go with the first option.

modin/utils.py
```diff
@@ -695,7 +697,7 @@ def get_current_execution() -> str:
     str
         Returns <StorageFormat>On<Engine>-like string.
     """
-    return f"{'Experimental' if IsExperimental.get() else ''}{StorageFormat.get()}On{Engine.get()}"
+    return f"{StorageFormat.get()}On{Engine.get()}"
```
anmyachev, line 700: The factory name is now independent of the mode, since one factory should also provide entry points to experimental functionality so that they do not have to be switched when using functionality from different modes.

* For functions that do not exist in pandas, the entry points are separate namespaces (for methods: `DataFrame/Series.modin`, for free functions: `import modin.experimental.pandas`).
* For functions that exist in pandas and only have different implementations, switching implementations is done using environment variable `MODIN_EXPERIMENTAL` as before.

anmyachev: Experimental IO classes have been removed as unnecessary.

anmyachev: > We could probably combine some of them into one, and, I think since we have the executions such as ExpetimentalPyarrowOnRay and ExperimentalHdkOnNative, we should provide them when MODIN_EXPERIMENTAL is enabled, i.e. as is now.

@YarShev please note that for HDK the experimental mode is turned on automatically at the time of initialization.

Garra1980:
Should we document these changes?

modin/utils.py
YarShev: > The factory name is now independent of the mode, since one factory should also provide entry points to experimental functionality so that they do not have to be switched when using functionality from different modes.

What if we have experimental executions that are not initiatialized automatically as HDK or an external one?

> For functions that exist in pandas and only have different implementations, switching implementations is done using environment variable MODIN_EXPERIMENTAL as before.

> I think we should use MODIN_EXPERIMENTAL only for experimental executions and it should be decoupled from API. What if we just change function names that are present in pandas but only have different implementations?

> Experimental IO classes have been removed as unnecessary.

Could you elaborate this a bit why those are unnecesary?

anmyachev: > What if we have experimental executions (exp pyarrow on ray, external one) that are not initiatialized automatically like HDK?

If we want to make the factory work only if the experimental mode is enabled, then we can put a mode check in the method that is used to initialize the factory - `def prepare(cls)`.

anmyachev: > Could you elaborate this a bit why those are unnecesary?

All methods from experimental IO classes have been transferred to regular IO classes. And we moved them because these methods should be available in both modes, since now there are special namespaces through which the methods are **always** available. This is enough to understand the intent of the user who wants to use the experimental functionality.

modin/core/execution/dask/implementations/pandas_on_dask/io/io.py
```diff
@@ -31,6 +31,9 @@
     SQLDispatcher,
 )
 from modin.core.storage_formats.pandas.parsers import (
+    ExperimentalCustomTextParser,
+    ExperimentalPandasPickleParser,
```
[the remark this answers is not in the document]
anmyachev: Since there is now one factory for both modes, importing experimental functionality in a regular module will be a common thing.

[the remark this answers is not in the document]
YarShev: I mean should we move experimental parsers to modin.experimental.core.storage_formats.pandas.parsers?

modin/utils.py
anmyachev: > Why do you think we shouldn't hold experimental factories that initialize respective executions?

The factory only routes calls, this task is the same for both modes, so I think we don’t need another one factory that differs only in name, but does the same job.

> Me: I think we should use MODIN_EXPERIMENTAL only for experimental executions and it should be decoupled from API. What if we just change function names that are present in pandas but only have different implementations?

Let's look at `groupby` example, we have an experimental (although it is activated by another environment variable, but still) and a main implementation. Do I understand correctly that your proposal suggests making a new `groupby_exp` function? In this case, there will be two functions with the same signatures (from the user’s point of view), and I believe there will be duplication of code in our implementation (albeit a little likely). In addition, now it’s enough to simply run user's code in two modes and see which is better from a performance point of view. If we make your change, anyone will need to make changes to the sources every time, which is quite inconvenient.

modin/core/execution/dask/implementations/pandas_on_dask/io/io.py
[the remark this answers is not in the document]
anmyachev: Got it. I think with the current implementation it would be more consistent to move the code. However, the following question arises: why do we divide the code into experimental and non-experimental, only for IO functions? Should we do the same for all other cases of experimental code (groupby, asynchronous read, numpy module implementation and maybe some other cases)?

modin/core/io/sql/sql_dispatcher.py
```diff
@@ -176,3 +178,118 @@ def func(df):  # pragma: no cover
         cls.materialize(
             [part.list_of_blocks[0] for row in result._partitions for part in row]
         )
+
+    def __getattribute__(self, __name: str) -> Any:
```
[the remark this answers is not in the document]
anmyachev: Outdated code, automatically suggested by IDE.

modin/core/io/sql/sql_dispatcher.py
```diff
@@ -33,7 +34,14 @@ class SQLDispatcher(FileDispatcher):
     """Class handles utils for reading SQL queries or database tables."""
 
     @classmethod
-    def _read(cls, sql, con, index_col=None, **kwargs):
+    def _read(cls, *args, **kwargs):  # noqa: GL08
```
anmyachev, line 37: Placeholder.

modin/experimental/core/io/__init__.py
```diff
@@ -14,13 +14,11 @@
 """Experimental IO functions implementations."""
 
 from .pickle.pickle_dispatcher import ExperimentalPickleDispatcher
-from .sql.sql_dispatcher import ExperimentalSQLDispatcher
 from .text.csv_glob_dispatcher import ExperimentalCSVGlobDispatcher
 from .text.custom_text_dispatcher import ExperimentalCustomTextDispatcher
 
 __all__ = [
     "ExperimentalCSVGlobDispatcher",
-    "ExperimentalSQLDispatcher",
```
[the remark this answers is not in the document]
anmyachev: reverted

modin/core/io/sql/sql_dispatcher.py
```diff
@@ -176,3 +178,118 @@ def func(df):  # pragma: no cover
...
+            partition_ids.append(
+                [cls.frame_partition_cls(obj) for obj in partition_id[:-1]]
+            )
+            index_ids.append(partition_id[-1])
+        new_index = pandas.RangeIndex(sum(cls.materialize(index_ids)))
+        new_query_compiler = cls.query_compiler_cls(
+            cls.frame_cls(np.array(partition_ids), new_index, cols_names)
+        )
+        new_query_compiler._modin_frame.synchronize_labels(axis=0)
+        return new_query_compiler
```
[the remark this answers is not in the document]
anmyachev: reverted

modin/core/execution/dispatching/factories/factories.py
```diff
@@ -329,6 +329,33 @@ def _read_pickle(cls, **kwargs):
         method="read_sql",
     )
     def _read_sql(cls, **kwargs):
+        if IsExperimental.get():
+            supported_engines = ("Ray", "Unidist", "Dask")
+            if Engine.get() not in supported_engines:
```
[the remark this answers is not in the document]
anmyachev: I'm wondering if Pandas now optionally supports pyarrow as a backend, then perhaps we can completely remove this execution?

modin/core/execution/dispatching/factories/factories.py
```diff
@@ -569,37 +517,54 @@ def _to_pickle_distributed(cls, *args, **kwargs):
...
+        from modin.core.execution.dask.implementations.pandas_on_dask.io import (
+            PandasOnDaskIO,
+        )
+
+        cls.io_cls = PandasOnDaskIO
 
 
 @doc(_doc_factory_class, execution_name="experimental PyarrowOnRay")
-class ExperimentalPyarrowOnRayFactory(BaseFactory):  # pragma: no cover
+class PyarrowOnRayFactory(BaseFactory):  # pragma: no cover
```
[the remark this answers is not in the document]
anmyachev: reverted

modin/core/execution/dispatching/factories/factories.py
```diff
@@ -569,37 +517,54 @@ def _to_pickle_distributed(cls, *args, **kwargs):
         return cls.io_cls.to_pickle_distributed(*args, **kwargs)
 
 
-@doc(_doc_factory_class, execution_name="experimental PandasOnRay")
-class ExperimentalPandasOnRayFactory(ExperimentalBaseFactory, PandasOnRayFactory):
+@doc(_doc_factory_class, execution_name="cuDFOnRay")
+class CudfOnRayFactory(BaseFactory):
```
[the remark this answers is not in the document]
anmyachev: reverted

modin/core/execution/dask/implementations/pandas_on_dask/io/io.py
```diff
@@ -74,5 +83,18 @@ def __make_write(*classes, build_args=build_args):
...
+    read_pickle_distributed = __make_read(
+        ExperimentalPandasPickleParser, ExperimentalPickleDispatcher
+    )
+    to_pickle_distributed = __make_write(ExperimentalPickleDispatcher)
+    read_custom_text = __make_read(
+        ExperimentalCustomTextParser, ExperimentalCustomTextDispatcher
+    )
+    read_sql_distributed = __make_read(
+        ExperimentalSQLDispatcher, build_args={**build_args, "base_read": read_sql}
+    )
```
anmyachev, line 97: The experimental implementation of reading sql on the master is called a non-experimental implementation of the same function. Due to the fact that classes are built from several other classes at runtime, I have not found a more beautiful way to do the same thing, given the new class architecture (in the PR).

modin/experimental/pandas/io.py
```diff
@@ -115,7 +115,7 @@ def read_sql(
 
     assert IsExperimental.get(), "This only works in experimental mode"
 
-    result = FactoryDispatcher.read_sql(**kwargs)
+    result = FactoryDispatcher.read_sql_distributed(**kwargs)
```
anmyachev, line 118: The new name is for internal use only.

modin/core/execution/dispatching/factories/test/test_dispatcher.py
```diff
@@ -113,8 +113,8 @@ def test_factory_switch():
 
 def test_engine_wrong_factory():
     with pytest.raises(FactoryNotFoundError):
-        with _switch_value(StorageFormat, "Pyarrow"):
-            with _switch_value(Engine, "Dask"):
+        with _switch_value(Engine, "Dask"):
+            with _switch_value(StorageFormat, "Pyarrow"):
```
anmyachev, line 117: Due to the fact that the engine and the engine are not initialized at the same time, another error began to appear here, namely that `ExperimentalPyarrowOnRay` factory cannot be initialized because the experimental mode is not enabled.

modin/experimental/core/io/sql/sql_dispatcher.py
```diff
@@ -72,7 +72,7 @@ def _read(
             message = "Defaulting to Modin core implementation; \
                 'partition_column', 'lower_bound', 'upper_bound' must be different from None"
             warnings.warn(message)
-            return cls.base_io.read_sql(
```
anmyachev, line 75: In the current implementation, `base_io` is `BaseIO` instead of `PandasOnRayIO`.

modin/core/execution/dask/implementations/pandas_on_dask/io/io.py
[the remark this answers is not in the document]
anmyachev: Will do separately.

modin/experimental/core/io/sql/sql_dispatcher.py
anmyachev: TODO: Check for other experimental features to see if this change is necessary for them.

modin/core/execution/dispatching/factories/dispatcher.py
```diff
@@ -118,14 +118,13 @@ def get_factory(cls) -> factories.BaseFactory:
...
-    # FIXME: replace `_` parameter with `*args`
-    def _update_factory(cls, _):
+    def _update_factory(cls, *args):
         """
         Update and prepare factory with a new one specified via Modin config.
 
         Parameters
         ----------
-        _ : object
+        *args : iterable
```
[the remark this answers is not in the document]
anmyachev: agree, done

docs/usage_guide/advanced_usage/index.rst
```diff
@@ -31,7 +31,7 @@ Modin also supports these experimental APIs on top of pandas that are under acti
 - :py:func:`~modin.experimental.pandas.read_sql` -- add optional parameters for the database connection
 - :py:func:`~modin.experimental.pandas.read_custom_text` -- read custom text data from file
 - :py:func:`~modin.experimental.pandas.read_pickle_distributed`  -- read multiple files in a directory
-- :py:meth:`~modin.experimental.pandas.DataFrame.to_pickle_distributed` -- write to multiple files in a directory
+- :py:meth:`~modin.experimental.pandas.DataFrame.modin.to_pickle_distributed` -- write to multiple files in a directory
```
YarShev, line 34:
```suggestion
- :py:meth:`~modin.experimental.pandas.DataFrame._exp.to_pickle_distributed` -- write to multiple files in a directory
```
Check other related places.

modin/core/execution/dispatching/factories/dispatcher.py
```diff
@@ -118,40 +118,30 @@ def get_factory(cls) -> factories.BaseFactory:
...
             This parameters serves the compatibility purpose.
             Does not affect the result.
         """
         factory_name = get_current_execution() + "Factory"
+        experimental_factory_name = "Experimental" + factory_name
```
YarShev, line 132: Why do we need this additional var?

modin/core/execution/dispatching/factories/factories.py
```diff
@@ -427,58 +427,42 @@ def _to_parquet(cls, *args, **kwargs):
...
     @classmethod
-    @_inherit_docstrings(BaseFactory._read_sql)
-    def _read_sql(cls, **kwargs):
+    @doc(
+        _doc_io_method_raw_template,
+        source="SQL files",
+        params=_doc_io_method_kwargs_params,
+    )
+    def _read_sql_distributed(cls, **kwargs):
         supported_engines = ("Ray", "Unidist", "Dask")
```
YarShev, line 466: Should we check for executions here, but not engines? Remember PyarrowOnRay.

modin/core/execution/dispatching/factories/factories.py
```diff
@@ -569,33 +526,55 @@ def _to_pickle_distributed(cls, *args, **kwargs):
...
+    @doc(_doc_factory_prepare_method, io_module_name="``PandasOnUnidistIO``")
+    def prepare(cls):
+        from modin.core.execution.unidist.implementations.pandas_on_unidist.io import (
+            PandasOnUnidistIO,
+        )
+
+        cls.io_cls = PandasOnUnidistIO
+
+
+# EXPERIMENTAL EXECUTIONS
```
YarShev, line 577: This doesn't give any useful info. Either remove or expand explanation of experimental executions.

modin/experimental/pandas/__init__.py
```diff
@@ -48,6 +50,19 @@
     to_pickle_distributed,
 )
 
+old_to_pickle_distributed = to_pickle_distributed
+
+
+@functools.wraps(to_pickle_distributed)
+def to_pickle_distributed(*args, **kwargs):
+    warnings.warn(
+        "`DataFrame.to_pickle_distributed` will be removed after Modin 0.26.0 release. "
```
YarShev, line 59:
```suggestion
        "`DataFrame.to_pickle_distributed` is deprecated and will be removed in a future version. "
```

modin/experimental/pandas/__init__.py
```diff
@@ -48,6 +50,19 @@
...
+        + "Please use `DataFrame.modin.to_pickle_distributed` instead.",
```
YarShev, line 60:
```suggestion
        + "Please use `DataFrame._exp.to_pickle_distributed` instead.",
```

modin/experimental/pandas/io.py
```diff
@@ -316,7 +316,7 @@ def read_pickle_distributed(
 
     This experimental feature provides parallel reading from multiple pickle files which are
     defined by glob pattern. The files must contain parts of one dataframe, which can be
-    obtained, for example, by `to_pickle_distributed` function.
+    obtained, for example, by `DataFrame.modin.to_pickle_distributed` function.
```
YarShev, line 319:
```suggestion
    obtained, for example, by `DataFrame._exp.to_pickle_distributed` function.
```

modin/experimental/pandas/test/test_io_exp.py
```diff
@@ -260,7 +260,7 @@ def test_distributed_pickling(filename, compression):
         if filename_param == test_default_to_pickle_filename
         else contextlib.nullcontext()
     ):
-        df.to_pickle_distributed(filename, compression=compression)
+        df.modin.to_pickle_distributed(filename, compression=compression)
```
YarShev, line 263:
```suggestion
        df._exp.to_pickle_distributed(filename, compression=compression)
```

modin/pandas/accessor.py
```diff
@@ -191,3 +195,71 @@ def __get__(self, obj, cls):
...
+
+    Parameters
+    ----------
+    data : DataFrame or Series
+        Object to operate on.
+    """
+
+    def __init__(self, data):
+        warnings.warn(
+            "Thank you for using the Modin Experimental pandas API."
```
YarShev, line 212: Do we really need to emit this warning?

modin/pandas/dataframe.py
```diff
@@ -3169,3 +3169,6 @@ def __reduce__(self):
         return self._inflate_light, (self._query_compiler, pid)
 
     # Persistance support methods - END
+
+    # Namespace for experimental functions
+    modin = CachedAccessor("modin", ExperimentalFunctions)
```
YarShev, line 3174:
```suggestion
    _exp = CachedAccessor("_exp", ExperimentalFunctions)
```
I thought we agreed on _exp. I would prefer to use it instead of `modin` attribute, which is a bit vague to me.

modin/pandas/test/test_api.py
```diff
@@ -153,17 +153,21 @@ def test_dataframe_api_equality():
     modin_dir = [obj for obj in dir(pd.DataFrame) if obj[0] != "_"]
     pandas_dir = [obj for obj in dir(pandas.DataFrame) if obj[0] != "_"]
 
-    ignore = ["timetuple"]
+    ignore_in_pandas = ["timetuple"]
+    # modin - namespace for using experimental functionality
+    ignore_in_modin = ["modin"]
```
YarShev, line 158:
```suggestion
    ignore_in_modin = ["_exp"]
```

modin/core/execution/dispatching/factories/test/test_dispatcher.py
YarShev: Should we put an inline comment here?

modin/pandas/dataframe.py
anmyachev: It seemed to me that this was an open question, judging by what I wrote here https://github.com/modin-project/modin/pull/6764#issuecomment-1832028750.

Using private attributes can eliminate the desire to try, which is why I thought about naming the attribute with a public name, and to make it clear that the attribute is not implemented in pandas.

modin/core/execution/dispatching/factories/factories.py
anmyachev: Well `# EXPERIMENTAL FACTORIES` can be better here. 

I wanted to add this so as not to mix factories, giving a line for logical partitioning.

YarShev: Let's also put a comment on what the difference is between exp and non-exp factories.

modin/pandas/dataframe.py
YarShev: Makes sense, let's leave as is.

modin/experimental/pandas/__init__.py
anmyachev: Why don't you want to specify an explicit release?

modin/experimental/core/io/sql/sql_dispatcher.py
anmyachev: Not necessary

modin/core/execution/dispatching/factories/dispatcher.py
anmyachev: it is used in several places

modin/core/execution/dispatching/factories/factories.py
anmyachev: Fixed

modin/core/execution/dispatching/factories/factories.py
anmyachev: added

modin/pandas/accessor.py
anmyachev: Removed here as well as for importing `modin.experimental.pandas`. 
This is a debatable question, but I want to make it consistent. I also believe that the name of the module speaks for itself.

modin/experimental/pandas/__init__.py
YarShev: Our plans may change. Pandas also does not specify an exact release to deprecate functionality.

YarShev:
Left a few comments + there are some not addressed ones. Other than that, LGTM.

docs/usage_guide/advanced_usage/index.rst
YarShev, line 34: Why when we click on this method in the docs, we navigate to the method with the signature modin.to_pickle_distributed but not DataFrame.modin.to_pickle_distributed?

modin/core/execution/dispatching/factories/dispatcher.py
```diff
@@ -118,41 +118,33 @@ def get_factory(cls) -> factories.BaseFactory:
...
         try:
-            cls.__factory = getattr(factories, factory_name)
+            cls.__factory = getattr(factories, factory_name, None) or getattr(
+                factories, experimental_factory_name
```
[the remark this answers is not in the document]
anmyachev: To throw `AttributeError` exception, but only if there is no suitable option from two factories.

modin/core/execution/dispatching/factories/factories.py
```diff
@@ -529,12 +450,39 @@ def _read_csv_glob(cls, **kwargs):
...
 
+    @classmethod
+    @doc(
+        _doc_io_method_raw_template,
+        source="SQL files",
+        params=_doc_io_method_kwargs_params,
+    )
+    def _read_sql_distributed(cls, **kwargs):
+        current_execution = get_current_execution()
+        if current_execution not in supported_executions:
```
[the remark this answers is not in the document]
anmyachev: This requires adding the following attribute to all IO classes:
```python
    read_sql_distributed = __make_read(
        ExperimentalSQLDispatcher, build_args={**build_args, "base_read": read_sql}
    )
```
Looks like a complication.

modin/experimental/pandas/__init__.py
anmyachev: changed

docs/flow/modin/experimental/pandas.rst
```diff
@@ -13,4 +13,4 @@ Experimental API Reference
 .. autofunction:: read_csv_glob
 .. autofunction:: read_custom_text
 .. autofunction:: read_pickle_distributed
-.. automethod:: modin.experimental.pandas.DataFrame.to_pickle_distributed
+.. automethod:: modin.pandas.DataFrame.modin::to_pickle_distributed
```
anmyachev, line 16: `As a hint to autodoc extension, you can put a :: separator in between module name and object name to let autodoc know the correct module name if it is ambiguous.` from https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#directive-automethod

YarShev (approved):

YarShev:
@dchigarev, @Garra1980, any other comments?

dchigarev (approved):

PR #6788, opened 2023-11-30
FIX-#6781: Use `pandas.api.types.pandas_dtype` to convert to valid numpy and pandas only dtypes

### LINKED ISSUE #6781: TypeError: data type 'Int64' not understood
This problem will affect a fairly large part of the code, since we everywhere expect that the type exists in the numpy, but it does not.

However, the fix for this problem looks quite simple; we need to create a utility that will take into account the possible absence of a type in the numpy and determine it using pandas.

Modin: 76d741bec279305b041ba5689947438884893dad

Reproducer:
```python
import modin.pandas as pd
pd.DataFrame([1,2,3,4], dtype="Int64").sum()
```

Traceback:
```bash
  File "...\modin\logging\logger_decorator.py", line 129, in run_and_log
    return obj(*args, **kwargs)
  File "...\modin\pandas\dataframe.py", line 2074, in sum
    data._query_compiler.sum(
  File "...\modin\logging\logger_decorator.py", line 129, in run_and_log
    return obj(*args, **kwargs)
  File "...\modin\core\dataframe\algebra\tree_reduce.py", line 58, in caller
    query_compiler._modin_frame.tree_reduce(
  File "...\modin\logging\logger_decorator.py", line 129, in run_and_log
    return obj(*args, **kwargs)
  File "...\modin\core\dataframe\pandas\dataframe\utils.py", line 501, in run_f_on_minimally_updated_metadata
  File "...\modin\core\dataframe\pandas\dataframe\dataframe.py", line 2103, in tree_reduce
    return self._compute_tree_reduce_metadata(
  File "...\modin\logging\logger_decorator.py", line 129, in run_and_log
    return obj(*args, **kwargs)
  File "...\modin\core\dataframe\pandas\dataframe\dataframe.py", line 2014, in _compute_tree_reduce_metadata
    [np.dtype(dtypes)] * len(new_axes[1]), index=new_axes[1]
TypeError: data type 'Int64' not understood
```

JignyasAnand:
## What do these changes do?

- [x] Resolves #6781 
- [x] tests added and passing

### DIFF
--- a/modin/pandas/test/dataframe/test_reduce.py
+++ b/modin/pandas/test/dataframe/test_reduce.py
@@ -306,6 +306,22 @@ def test_sum(data, axis, skipna, is_transposed):
     df_equals(modin_result, pandas_result)
 
 
+@pytest.mark.parametrize("nptypes", [["int64"]])
+@pytest.mark.parametrize("pdtypes", [["Int64"]])
+def test_dtype_consistency(nptypes, pdtypes):
+    # test for issue #6781
+
+    # Test the resultant dtypes with numpy datatypes
+    for nptype in nptypes:
+        res_dtype = pd.DataFrame([1, 2, 3, 4], dtype=nptype).sum().dtype
+        assert res_dtype == np.dtype(nptype)
+
+    # Test the resultant dtypes with pandas datatypes
+    for pdtype in pdtypes:
+        res_dtype = pd.DataFrame([1, 2, 3, 4], dtype=pdtype).sum().dtype
+        assert res_dtype == pandas.api.types.pandas_dtype(pdtype)
+
+
 @pytest.mark.parametrize("fn", ["prod, sum"])
 @pytest.mark.parametrize(
     "numeric_only", bool_arg_values, ids=arg_keys("numeric_only", bool_arg_keys)
--- a/modin/pandas/test/dataframe/test_reduce.py
+++ b/modin/pandas/test/dataframe/test_reduce.py
@@ -18,7 +18,7 @@ import pytest
 from pandas._testing import assert_series_equal
 
 import modin.pandas as pd
-from modin.config import NPartitions, StorageFormat, Engine
+from modin.config import Engine, NPartitions, StorageFormat
 from modin.pandas.test.utils import (
     arg_keys,
     assert_dtypes_equal,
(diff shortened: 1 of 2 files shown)


anmyachev:
@JignyasAnand thanks for the contribution! Could you also add a test using code from the issue somewhere around?
https://github.com/modin-project/modin/blob/2fee14409bfda6661b8fcccbb33f3c93ae17204f/modin/pandas/test/dataframe/test_reduce.py#L287

JignyasAnand:
Hi @anmyachev . I have added the tests. Can you take a look at them?

modin/pandas/test/dataframe/test_reduce.py
```diff
@@ -306,6 +306,22 @@ def test_sum(data, axis, skipna, is_transposed):
...
+
+    # Test the resultant dtypes with numpy datatypes
+    for nptype in nptypes:
+        res_dtype = pd.DataFrame([1, 2, 3, 4], dtype=nptype).sum().dtype
+        assert res_dtype == np.dtype(nptype)
+
+    # Test the resultant dtypes with pandas datatypes
+    for pdtype in pdtypes:
+        res_dtype = pd.DataFrame([1, 2, 3, 4], dtype=pdtype).sum().dtype
+        assert res_dtype == pandas.api.types.pandas_dtype(pdtype)
```
anmyachev, line 322: Let's use one logic for all dtypes as it done in `dataframe.py`.

```suggestion
@pytest.mark.parametrize("dtype", ["int64", "Int64"])
def test_dtype_consistency(dtype):
    # test for issue #6781
    res_dtype = pd.DataFrame([1, 2, 3, 4], dtype=dtype).sum().dtype
    assert res_dtype == pandas.api.types.pandas_dtype(pdtype)
```

Also, your test fails for the HDK engine, but for now it's ok, let's just skip this case. For this you can use the following code `@pytest.mark.skipif(Engine.get() == "Native", reason="Fails on HDK")`.

**Note:** `Engine` can be imported from `modin.config`.

JignyasAnand: Thanks for the suggestion. I have modified the code

anmyachev:
Thanks @JignyasAnand!

PR #6836, opened 2023-12-21
FEAT-#6835: Do not put binary functions to the Ray storage multiple times.

### LINKED ISSUE #6835: FEAT: Do not put binary functions to the Ray storage multiple times.
Currently, the binary operations are wrapped into lambdas which are put into the Ray storage on each operation.

AndreyPavlenko:
Added functions cache to RayWrapper.
Removed lambda wrapper for the binary functions.

## What do these changes do?

- [x] Resolves #6835 
- [ ] tests added and passing

### DIFF
--- a/modin/core/execution/ray/common/engine_wrapper.py
+++ b/modin/core/execution/ray/common/engine_wrapper.py
@@ -18,6 +18,7 @@ To be used as a piece of building a Ray-based engine.
 """
 
 import asyncio
+from types import FunctionType
 
 import ray
 from ray.util.client.common import ClientObjectRef
@@ -47,6 +48,8 @@ def _deploy_ray_func(func, *args, **kwargs):  # pragma: no cover
 
 class RayWrapper:
     """Mixin that provides means of running functions remotely and getting local results."""
+
+    _func_cache = {}
 
     @classmethod
     def deploy(cls, func, f_args=None, f_kwargs=None, num_returns=1):
@@ -127,6 +130,14 @@ class RayWrapper:
         ray.ObjectID
             Ray object identifier to get the value by.
         """
+        if isinstance(data, FunctionType):
+            qname = data.__qualname__
+            if "<locals>" not in qname and "<lambda>" not in qname:
+                ref = cls._func_cache.get(data, None)
+                if ref is None:
+                    ref = ray.put(data)
+                    cls._func_cache[data] = ref
+                return ref
         return ray.put(data, **kwargs)
 
     @classmethod
--- a/modin/core/execution/ray/common/engine_wrapper.py
+++ b/modin/core/execution/ray/common/engine_wrapper.py
@@ -18,7 +18,7 @@ To be used as a piece of building a Ray-based engine.
 """
 
 import asyncio
-from os import environ
+import os
 from types import FunctionType
 
 import ray
@@ -143,7 +143,7 @@ class RayWrapper:
                         cls._func_cache[data] = ref
                     else:
                         msg = "To many functions in the RayWrapper cache!"
-                        assert "MODIN_GITHUB_CI" not in environ, msg
+                        assert "MODIN_GITHUB_CI" not in os.environ, msg
                         ErrorMessage.warn(msg)
                 return ref
         return ray.put(data, **kwargs)
--- a/modin/core/execution/ray/common/engine_wrapper.py
+++ b/modin/core/execution/ray/common/engine_wrapper.py
@@ -18,10 +18,13 @@ To be used as a piece of building a Ray-based engine.
 """
 
 import asyncio
+from os import environ
 from types import FunctionType
 
 import ray
 from ray.util.client.common import ClientObjectRef
+
+from modin.error_message import ErrorMessage
 
 
 @ray.remote
@@ -135,8 +138,13 @@ class RayWrapper:
             if "<locals>" not in qname and "<lambda>" not in qname:
                 ref = cls._func_cache.get(data, None)
                 if ref is None:
-                    ref = ray.put(data)
-                    cls._func_cache[data] = ref
+                    if len(cls._func_cache) < 1024:
+                        ref = ray.put(data)
+                        cls._func_cache[data] = ref
+                    else:
+                        msg = "To many functions in the RayWrapper cache!"
+                        assert "MODIN_GITHUB_CI" not in environ, msg
+                        ErrorMessage.warn(msg)
                 return ref
         return ray.put(data, **kwargs)
 
--- a/modin/core/dataframe/algebra/binary.py
+++ b/modin/core/dataframe/algebra/binary.py
@@ -415,7 +415,9 @@ class Binary(Operator):
                     ):
                         shape_hint = "column"
                     new_modin_frame = query_compiler._modin_frame.map(
-                        lambda df: func(df, other, *args, **kwargs),
+                        func,
+                        func_args=(other, *args),
+                        func_kwargs=kwargs,
                         dtypes=dtypes,
                     )
                 return query_compiler.__constructor__(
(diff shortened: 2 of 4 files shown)


anmyachev:
@AndreyPavlenko what are the advantages of this approach?

AndreyPavlenko:
> what are the advantages of this approach?

The functions are put to the Ray storage only once, the reference is cached and then reused.
The current implementation creates a lambda on each operation and puts this, short living, lambda to the storage.
Another advantage is an opportunity for further optimizations. When passing a real function, but not lambda, to the map() function, it allows the map() function to choose the execution mode(lazy or not, ex: #6839) and, may be, apply some optimizations/precomputations. For example, a sequence of arithmetic functions with scalar arguments, could be precomputed and replaced with a single one.

modin/core/execution/ray/common/engine_wrapper.py
```diff
@@ -108,6 +111,14 @@ def put(cls, data, **kwargs):
         ray.ObjectID
             Ray object identifier to get the value by.
         """
+        if isinstance(data, FunctionType):
```
anmyachev, line 114: Will this work?
```suggestion
        if not isinstance(data, types.LambdaType):
```

modin/core/execution/ray/common/engine_wrapper.py
```diff
@@ -108,6 +111,14 @@ def put(cls, data, **kwargs):
...
+            qname = data.__qualname__
+            if "<locals>" not in qname and "<lambda>" not in qname:
```
anmyachev, line 116: Why check `<locals>` as well?

modin/core/execution/ray/common/engine_wrapper.py
AndreyPavlenko: Unfortunately, it doesn't work. Look at the source code of the types module:
```python
LambdaType = type(lambda: None)         # Same as FunctionType
```

modin/core/execution/ray/common/engine_wrapper.py
AndreyPavlenko: To filter out nested functions. I didn't find a better approach than using __qualname__.

anmyachev: > To filter out nested functions.

Why?

AndreyPavlenko: Because of a new function instance on each call. Most probably, the cached functions, will never be reused, it's just a memory leak.

modin/core/execution/ray/common/engine_wrapper.py
```diff
@@ -108,6 +111,14 @@ def put(cls, data, **kwargs):
...
+                ref = cls._func_cache.get(data, None)
```
anmyachev, line 117: Could there be a key collision?

modin/core/execution/ray/common/engine_wrapper.py
```diff
@@ -47,6 +48,8 @@ def _deploy_ray_func(func, *args, **kwargs):  # pragma: no cover
 class RayWrapper:
     """Mixin that provides means of running functions remotely and getting local results."""
 
+    _func_cache = {}
```
anmyachev, line 51: Please add a type hint here.

Have you estimated the cache size that might result?

modin/core/execution/ray/common/engine_wrapper.py
AndreyPavlenko: Do you mean a hash table collision? It's possible, like for any other keys, but unlikely. All the functions have a different hash code:
```python
>>> hash(pandas.DataFrame.add)
8744465041078
>>> hash(pandas.DataFrame.mul)
8744465041118
```
Even, in case of collisions, I believe it's not critical, because the hash table is quite small.

modin/core/execution/ray/common/engine_wrapper.py
AndreyPavlenko: No, but my assumption is that it must not be greater than pandas.DataFrame.* + a limited set of internal modin functions. If it's not true, than something is wrong and should be fixed. My initial approach was to use weak references, but it could be error-prone and results in performance issues, when short living objects are added to the cache and never reused. It's easier to detect such issues, when using strong references.
I think, it would be reasonable to add a size assert here, but we need to roughly estimate the maximum cache size.

anmyachev: > I think, it would be reasonable to add a size assert here, but we need to roughly estimate the maximum cache size.

It’s not easy to make an estimate by bytes; it would be nice to have at least the maximum number of keys.

By the way, doesn't Ray itself cache anything?

Does it make sense to do caching for other engines now? (Dask and Unidist)

AndreyPavlenko: Perhaps, it does cache, but i'm not sure, because ray.put() always returns a new object reference:

```python
>>> id(ray.put(pandas.DataFrame.add))
140361903189728
>>> id(ray.put(pandas.DataFrame.add))
140361903189584
>>> id(ray.put(pandas.DataFrame.add))
140360098017280
```

AndreyPavlenko: We could either add something like:
```python
assert len(_func_cache) < 64
```
or use lru cache here, but I'm not sure if it makes sense.

anmyachev: > We could either add something like:
assert len(_func_cache) < 64

Maybe we’ll make 128 functions and it won’t crash with an error if the 129th function appears, but we’ll delete the very first of the remaining functions and add a new one instead?

AndreyPavlenko: We can just use lru cache with max 128 items, but it hides the problem when the function filter does not work properly and we add/remove short-living functions to the cache. In case of explicit  size check, if we hit the assertion error, then either the limit should be increased or the filter is broken and should be fixed.

anmyachev: > In case of explicit size check, if we hit the assertion error, then either the limit should be increased or the filter is broken and should be fixed.

In this case, it is the users who will most likely encounter errors, not us. It turns out that something that worked before may break due to caching; this is an unacceptable option.

AndreyPavlenko: That's why we must either properly estimate the limit, or set a large limit, for example, 1024. In case of an invalid filtering, we will definitely hit this limit in unit tests.
Another option - use warning instead of assert.

anmyachev: 1024 limit and using warnings LGTM.

> Maybe we’ll make 128 functions and it won’t crash with an error if the 129th function appears, but we’ll delete the very first of the remaining functions and add a new one instead?

>> but it hides the problem when the function filter does not work properly and we add/remove short-living functions to the cache. 

We can introduce a separate counter that will count how many functions have been cached.

`pytest` seems to have a function that is called at the end of the testing session, we can check how many functions were cached during the tests, however, we do not run all tests with one `pytest` call, so such a check is not sufficient. There is a feeling that such things can mainly be caught using complex integration tests, which we do not have.

AndreyPavlenko: Added the size check.

modin/core/execution/ray/common/engine_wrapper.py
```diff
@@ -18,9 +18,13 @@
 """
 
 import asyncio
+from os import environ
```
anmyachev, line 21: Let's use the following import (and further like `os.environ`)
```suggestion
import os
```

anmyachev:
@AndreyPavlenko please also rebase on master branch to fix CI failures.

anmyachev (approved):
LGTM!

PR #6839, opened 2023-12-21
FEAT-#6838: Prefer lazy execution for binary operations with scalar.

### LINKED ISSUE #6838: Prefer lazy execution for binary operations.

AndreyPavlenko:
**Depends on #6836.**

## What do these changes do?

Multiple binary operations are faster to execute lazy, in a single batch. Here is an example:

```python
import ray
ray.init()

import time
import modin.pandas as pd

df = pd.Series(range(100000))
start = time.time()
for i in range(100):
    df = df + i
execute(df)
print(f"Time: {time.time() - start}")
```
Time: 2.4991016387939453
Time on master: 5.286777019500732
- [x] Resolves #6838 
- [ ] tests added and passing

### DIFF
--- a/modin/core/dataframe/algebra/binary.py
+++ b/modin/core/dataframe/algebra/binary.py
@@ -419,6 +419,7 @@ class Binary(Operator):
                         func_args=(other, *args),
                         func_kwargs=kwargs,
                         dtypes=dtypes,
+                        lazy=True,
                     )
                 return query_compiler.__constructor__(
                     new_modin_frame, shape_hint=shape_hint
--- a/modin/core/dataframe/pandas/dataframe/dataframe.py
+++ b/modin/core/dataframe/pandas/dataframe/dataframe.py
@@ -2153,7 +2153,7 @@ class PandasDataframe(ClassLogger):
         func_kwargs : dict, optional
             Keyword arguments for the 'func' callable.
         lazy : bool, default: False
-            Prefer lazy execution.
+            Whether to prefer lazy execution or not.
 
         Returns
         -------
--- a/modin/core/dataframe/pandas/dataframe/dataframe.py
+++ b/modin/core/dataframe/pandas/dataframe/dataframe.py
@@ -2132,6 +2132,7 @@ class PandasDataframe(ClassLogger):
         new_columns: Optional[pandas.Index] = None,
         func_args=None,
         func_kwargs=None,
+        lazy=False,
     ) -> "PandasDataframe":
         """
         Perform a function that maps across the entire dataset.
@@ -2151,15 +2152,20 @@ class PandasDataframe(ClassLogger):
             Positional arguments for the 'func' callable.
         func_kwargs : dict, optional
             Keyword arguments for the 'func' callable.
+        lazy : bool, default: False
+            Prefer lazy execution.
 
         Returns
         -------
         PandasDataframe
             A new dataframe.
         """
-        new_partitions = self._partition_mgr_cls.map_partitions(
-            self._partitions, func, func_args, func_kwargs
+        map_fn = (
+            self._partition_mgr_cls.lazy_map_partitions
+            if lazy
+            else self._partition_mgr_cls.map_partitions
         )
+        new_partitions = map_fn(self._partitions, func, func_args, func_kwargs)
         if new_columns is not None and self.has_materialized_columns:
             assert len(new_columns) == len(
                 self.columns


YarShev:
@AndreyPavlenko, #6836 is merged. Can we continue this?

modin/core/dataframe/algebra/binary.py
```diff
@@ -270,6 +270,8 @@ def try_compute_new_dtypes(
 class Binary(Operator):
     """Builder class for Binary operator."""
 
+    _binary_ops = set()
```
YarShev, line 273: Let's add a comment on what this is needed for.

modin/core/dataframe/algebra/binary.py
```diff
@@ -424,4 +426,5 @@ def caller(
                     new_modin_frame, shape_hint=shape_hint
                 )
 
+        cls._binary_ops.add(func)
```
YarShev, line 429: Why should we put a function for all binary cases? I guess this PR only addresses map case, doesn't it?

modin/core/execution/ray/common/engine_wrapper.py
```diff
@@ -53,6 +54,7 @@ class RayWrapper:
     """Mixin that provides means of running functions remotely and getting local results."""
 
     _func_cache = {}
+    _lazy_op_refs = set()
```
YarShev, line 57: Let's add comments on what these are needed for.

modin/core/execution/ray/implementations/pandas_on_ray/partitioning/partition.py
```diff
@@ -94,6 +94,9 @@ def apply(self, func, *args, **kwargs):
         It does not matter if `func` is callable or an ``ray.ObjectRef``. Ray will
         handle it correctly either way. The keyword arguments are sent as a dictionary.
         """
+        if self.execution_wrapper.prefer_lazy(func):
+            return self.add_to_apply_calls(func, *args, **kwargs)
```
YarShev, line 98: It seems to me we should move this to the partition manager layer.

YarShev: Also, I think we should discuss on a meeting the introduced lazy execution in general because now we are adding more cases when Modin is lazy even if it is not enabled.

modin/core/dataframe/algebra/binary.py
AndreyPavlenko: Done.

modin/core/dataframe/algebra/binary.py
AndreyPavlenko: I think, it makes sense to use the lazy execution for all the binary functions.

modin/core/execution/ray/common/engine_wrapper.py
AndreyPavlenko: Done.

modin/core/execution/ray/implementations/pandas_on_ray/partitioning/partition.py
AndreyPavlenko: > It seems to me we should move this to the partition manager layer.

It seems reasonable, but it will require more changes, because the apply() function is called by multiple functions in PandasDataframePartitionManager.

> Also, I think we should discuss on a meeting the introduced lazy execution in general because now we are adding more cases when Modin is lazy even if it is not enabled.

I think, the MODIN_LAZY_EXECUTION variable could be extended to have 3 values: AUTO, ON, OFF. AUTO is by default and we choose which mode to use.

modin/core/dataframe/algebra/binary.py
YarShev: Then I think we would need to look at perf measurements for all cases of binary operations (aligned frames, unaligned frames, etc.), which may take time to investigate this and do not promise positive results. I would propose to leave this PR only for the map case of binary operation.

modin/core/execution/ray/implementations/pandas_on_ray/partitioning/partition.py
YarShev: > > It seems to me we should move this to the partition manager layer.
> 
> It seems reasonable, but it will require more changes, because the apply() function is called by multiple functions in PandasDataframePartitionManager.
> 
> > Also, I think we should discuss on a meeting the introduced lazy execution in general because now we are adding more cases when Modin is lazy even if it is not enabled.
> 
> I think, the MODIN_LAZY_EXECUTION variable could be extended to have 3 values: AUTO, ON, OFF. AUTO is by default and we choose which mode to use.

I see. Maybe we could do simpler just by calling `lazy_map_partitions` in the binary operator.



> > It seems to me we should move this to the partition manager layer.
> 
> It seems reasonable, but it will require more changes, because the apply() function is called by multiple functions in PandasDataframePartitionManager.
> 
> > Also, I think we should discuss on a meeting the introduced lazy execution in general because now we are adding more cases when Modin is lazy even if it is not enabled.
> 
> I think, the MODIN_LAZY_EXECUTION variable could be extended to have 3 values: AUTO, ON, OFF. AUTO is by default and we choose which mode to use.

I agree. Let's address this in a separate PR.

modin/core/dataframe/algebra/binary.py
AndreyPavlenko: Done.

YarShev (approved):

PR #6880, opened 2024-01-25
FIX-#6879: Convert the right DF to single partition before broadcasting in query_compiler.merge

### LINKED ISSUE #6879: The query_compiler.merge reconstructs the Right dataframe for every partition of Left Dataframe
The query_compiler.merge reconstructs the Right dataframe from its partitions for every partition of Left Dataframe, The concat operation results in higher memory consumption when the size of right dataframe is large.

A possible option is to combine the right Dataframe partitions to a single partition dataframe by calling a remote function. This single partiotion dataframe is then passed to each partition of left dataframe thus avoiding the reconstruction in every worker while doing merge.

arunjose696:
## What do these changes do?

- [x] Resolves #6879  
- [ ] tests added and passing

### DIFF
--- a/modin/core/dataframe/pandas/dataframe/dataframe.py
+++ b/modin/core/dataframe/pandas/dataframe/dataframe.py
@@ -2771,27 +2771,22 @@ class PandasDataframe(ClassLogger):
         PandasDataframe
             A single partition PandasDataframe.
         """
-        partitions = self._partition_mgr_cls.to_single_partition(self._partitions)
-        new_row_lengths = (
-            [sum(self._row_lengths_cache)]
-            if self._row_lengths_cache is not None
-            else None
-        )
-        new_column_widths = (
-            [sum(self._column_widths_cache)]
-            if self._column_widths_cache is not None
-            else None
-        )
-        index = self.copy_index_cache()
-        columns = self.copy_columns_cache()
-        dtypes = self.copy_dtypes_cache()
+        partitions = self._partition_mgr_cls.combine(self._partitions)
         result = self.__constructor__(
             partitions,
-            index=index,
-            columns=columns,
-            dtypes=dtypes,
-            row_lengths=new_row_lengths,
-            column_widths=new_column_widths,
+            index=self.copy_index_cache(),
+            columns=self.copy_columns_cache(),
+            row_lengths=(
+                [sum(self._row_lengths_cache)]
+                if self._row_lengths_cache is not None
+                else None
+            ),
+            column_widths=(
+                [sum(self._column_widths_cache)]
+                if self._column_widths_cache is not None
+                else None
+            ),
+            dtypes=self.copy_dtypes_cache(),
         )
         result.synchronize_labels()
         return result
--- a/modin/core/dataframe/pandas/dataframe/dataframe.py
+++ b/modin/core/dataframe/pandas/dataframe/dataframe.py
@@ -2764,7 +2764,7 @@ class PandasDataframe(ClassLogger):
 
     def to_pandas_in_remote_function(self) -> "PandasDataframe":
         """
-        Create a PandasFrame with single partition from the partitions of the current dataframe.
+        Create a single partition PandasDataframe from the partitions of the current dataframe.
 
         Returns
         -------
--- a/modin/core/dataframe/pandas/dataframe/dataframe.py
+++ b/modin/core/dataframe/pandas/dataframe/dataframe.py
@@ -2771,10 +2771,7 @@ class PandasDataframe(ClassLogger):
         PandasDataframe
             A single partition PandasDataframe.
         """
-        single_partition = [
-            self._partition_mgr_cls.to_single_partition(self._partitions)
-        ]
-        partitions = np.array(single_partition).reshape(1, -1)
+        partitions = self._partition_mgr_cls.to_single_partition(self._partitions)
         new_row_lengths = (
             [sum(self._row_lengths_cache)]
             if self._row_lengths_cache is not None
--- a/modin/core/dataframe/pandas/dataframe/dataframe.py
+++ b/modin/core/dataframe/pandas/dataframe/dataframe.py
@@ -2762,25 +2762,27 @@ class PandasDataframe(ClassLogger):
             partitions, new_index, new_columns, row_lengths, column_widths
         )
 
-    def force_materialization(self) -> "PandasDataframe":
+    def to_pandas_in_remote_function(self) -> "PandasDataframe":
         """
-        Materialize axis partitions into a single partition.
-
-        Applies the identity function first across rows, and thereafter across columns
-        to get a df single  partition.
+        Create a PandasFrame with single partition from the partitions of the current dataframe.
 
         Returns
         -------
         PandasDataframe
             An PandasDataframe containing only a single partition.
         """
-        row_partitions = self._partition_mgr_cls.row_partitions(self._partitions)
-        col_partition = self._partition_mgr_cls.column_partitions(
-            np.asarray([row_partitions]).T
+        partition_with_to_pandas_data = [
+            self._partition_mgr_cls.convert_partitions_to_pandas_in_remote_func(
+                self._partitions
+            )
+        ]
+        partitions = np.array(partition_with_to_pandas_data).reshape(1, -1)
+
+        result = self.__constructor__(
+            partitions, self.index, columns=self.columns, dtypes=self.dtypes
         )
-        new_frame = np.array([col_partition[0].apply(lambda df: df, num_splits=1)])
-        single_partition_df = self.__constructor__(new_frame)
-        return single_partition_df
+        result.synchronize_labels()
+        return result
 
     @lazy_metadata_decorator(apply_axis="both")
     def apply_full_axis(
--- a/modin/core/dataframe/pandas/dataframe/dataframe.py
+++ b/modin/core/dataframe/pandas/dataframe/dataframe.py
@@ -2772,14 +2772,26 @@ class PandasDataframe(ClassLogger):
             An PandasDataframe containing only a single partition.
         """
         partition_with_to_pandas_data = [
-            self._partition_mgr_cls.convert_partitions_to_pandas_in_remote_func(
-                self._partitions
-            )
+            self._partition_mgr_cls.to_single_partition(self._partitions)
         ]
         partitions = np.array(partition_with_to_pandas_data).reshape(1, -1)
-
+        new_row_lengths = (
+            [sum(self._row_lengths_cache)]
+            if self._row_lengths_cache is not None
+            else None
+        )
+        new_column_widths = (
+            [sum(self._column_widths_cache)]
+            if self._column_widths_cache is not None
+            else None
+        )
         result = self.__constructor__(
-            partitions, self.index, columns=self.columns, dtypes=self.dtypes
+            partitions,
+            self.index,
+            columns=self.columns,
+            dtypes=self.dtypes,
+            row_lengths=new_row_lengths,
+            column_widths=new_column_widths,
         )
         result.synchronize_labels()
         return result
--- a/modin/core/dataframe/pandas/dataframe/dataframe.py
+++ b/modin/core/dataframe/pandas/dataframe/dataframe.py
@@ -2774,22 +2774,12 @@ class PandasDataframe(ClassLogger):
         PandasDataframe
             An PandasDataframe containing only a single partition.
         """
-        single_col_aggregated_df = self.apply_full_axis(
-            axis=0,
-            func=lambda x: x,
-            keep_partitioning=False,
-            num_splits=1,
-            sync_labels=True,
-            pass_axis_lengths_to_partitions=False,
+        row_partitions = self._partition_mgr_cls.row_partitions(self._partitions)
+        col_partition = self._partition_mgr_cls.column_partitions(
+            np.asarray([row_partitions]).T
         )
-        single_partition_df = single_col_aggregated_df.apply_full_axis(
-            axis=1,
-            func=lambda x: x,
-            keep_partitioning=False,
-            num_splits=1,
-            sync_labels=True,
-            pass_axis_lengths_to_partitions=False,
-        )
+        new_frame = np.array([col_partition[0].apply(lambda df: df, num_splits=1)])
+        single_partition_df = self.__constructor__(new_frame)
         return single_partition_df
 
     @lazy_metadata_decorator(apply_axis="both")
--- a/modin/core/dataframe/pandas/dataframe/dataframe.py
+++ b/modin/core/dataframe/pandas/dataframe/dataframe.py
@@ -2762,19 +2762,19 @@ class PandasDataframe(ClassLogger):
             partitions, new_index, new_columns, row_lengths, column_widths
         )
 
-    def to_pandas_in_remote_function(self) -> "PandasDataframe":
+    def combine(self) -> "PandasDataframe":
         """
         Create a single partition PandasDataframe from the partitions of the current dataframe.
 
         Returns
         -------
         PandasDataframe
-            An PandasDataframe containing only a single partition.
+            A single partition PandasDataframe.
         """
-        partition_with_to_pandas_data = [
+        single_partition = [
             self._partition_mgr_cls.to_single_partition(self._partitions)
         ]
-        partitions = np.array(partition_with_to_pandas_data).reshape(1, -1)
+        partitions = np.array(single_partition).reshape(1, -1)
         new_row_lengths = (
             [sum(self._row_lengths_cache)]
             if self._row_lengths_cache is not None
@@ -2785,11 +2785,14 @@ class PandasDataframe(ClassLogger):
             if self._column_widths_cache is not None
             else None
         )
+        index = self.copy_index_cache()
+        columns = self.copy_columns_cache()
+        dtypes = self.copy_dtypes_cache()
         result = self.__constructor__(
             partitions,
-            self.index,
-            columns=self.columns,
-            dtypes=self.dtypes,
+            index=index,
+            columns=columns,
+            dtypes=dtypes,
             row_lengths=new_row_lengths,
             column_widths=new_column_widths,
         )
--- a/modin/core/dataframe/pandas/dataframe/dataframe.py
+++ b/modin/core/dataframe/pandas/dataframe/dataframe.py
@@ -2762,6 +2762,36 @@ class PandasDataframe(ClassLogger):
             partitions, new_index, new_columns, row_lengths, column_widths
         )
 
+    def force_materialization(self) -> "PandasDataframe":
+        """
+        Materialize axis partitions into a single partition.
+
+        Applies the identity function first across rows, and thereafter across columns
+        to get a df single  partition.
+
+        Returns
+        -------
+        PandasDataframe
+            An PandasDataframe containing only a single partition.
+        """
+        single_col_aggregated_df = self.apply_full_axis(
+            axis=0,
+            func=lambda x: x,
+            keep_partitioning=False,
+            num_splits=1,
+            sync_labels=True,
+            pass_axis_lengths_to_partitions=False,
+        )
+        single_partition_df = single_col_aggregated_df.apply_full_axis(
+            axis=1,
+            func=lambda x: x,
+            keep_partitioning=False,
+            num_splits=1,
+            sync_labels=True,
+            pass_axis_lengths_to_partitions=False,
+        )
+        return single_partition_df
+
     @lazy_metadata_decorator(apply_axis="both")
     def apply_full_axis(
         self,
--- a/modin/core/dataframe/pandas/partitioning/partition_manager.py
+++ b/modin/core/dataframe/pandas/partitioning/partition_manager.py
@@ -1159,11 +1159,68 @@ class PandasDataframePartitionManager(ClassLogger, ABC):
         PandasDataframePartition
             A PandasDataframePartition object which holds the data of all partitions.
         """
+        cls.finalize(partitions)
 
-        def call_to_pandas(data, partitions):
-            return cls.to_pandas(partitions)
+        def convert_to_pandas(data, partition_shape, *partition_data):
+            if all(
+                isinstance(obj, (pandas.DataFrame, pandas.Series))
+                for obj in partition_data
+            ):
+                height, width, *_ = tuple(partition_shape) + (0,)
+                # restore 2d array
+                objs = iter(partition_data)
+                partition_data = [
+                    [next(objs) for _ in range(width)] for __ in range(height)
+                ]
+            else:
+                # Partitions do not always contain pandas objects, for example, hdk uses pyarrow tables.
+                # This implementation comes from the fact that calling `partition.get`
+                # function is not always equivalent to `partition.to_pandas`.
+                partition_data = [
+                    [obj.to_pandas() for obj in part] for part in partitions
+                ]
+            if all(
+                isinstance(part, pandas.Series)
+                for row in partition_data
+                for part in row
+            ):
+                axis = 0
+            elif all(
+                isinstance(part, pandas.DataFrame)
+                for row in partition_data
+                for part in row
+            ):
+                axis = 1
+            else:
+                ErrorMessage.catch_bugs_and_request_email(True)
 
-        return partitions.flat[0].apply(call_to_pandas, partitions=partitions)
+            def is_part_empty(part):
+                return part.empty and (
+                    not isinstance(part, pandas.DataFrame) or (len(part.columns) == 0)
+                )
+
+            df_rows = [
+                pandas.concat([part for part in row], axis=axis, copy=False)
+                for row in partition_data
+                if not all(is_part_empty(part) for part in row)
+            ]
+            if len(df_rows) == 0:
+                return pandas.DataFrame()
+            else:
+                return concatenate(df_rows)
+
+        preprocessed_func = cls.preprocess_func(convert_to_pandas)
+        partition_shape = partitions.shape
+        partitions_flattened = partitions.flatten()
+        for idx, part in enumerate(partitions_flattened):
+            if hasattr(part, "force_materialization"):
+                partitions_flattened[idx] = part.force_materialization()
+        partition_data = [
+            partition.list_of_blocks[0] for partition in partitions_flattened
+        ]
+        return partitions.flat[0].apply(
+            preprocessed_func, partition_shape, *partition_data
+        )
 
     @classmethod
     @wait_computations_if_benchmark_mode
--- a/modin/core/dataframe/pandas/partitioning/partition_manager.py
+++ b/modin/core/dataframe/pandas/partitioning/partition_manager.py
@@ -1144,10 +1144,7 @@ class PandasDataframePartitionManager(ClassLogger, ABC):
     @classmethod
     def to_single_partition(cls, partitions):
         """
-        Call to_pandas for the given NumPy array of partitions.
-
-        For a NumPy array of partitions create a single partition
-        that has the data for all partitions.
+        Convert a NumPy 2D array of partitions to a NumPy 2D array of a single partition.
 
         Parameters
         ----------
@@ -1156,12 +1153,12 @@ class PandasDataframePartitionManager(ClassLogger, ABC):
 
         Returns
         -------
-        PandasDataframePartition
-            A PandasDataframePartition object which holds the data of all partitions.
+        np.ndarray
+                    A NumPy 2D array of a single partition.
         """
 
         def to_pandas_remote(data, partition_shape, *partition_data):
-        """Copy of ``cls.to_pandas()`` method adapted for a remote function."""
+            """Copy of ``cls.to_pandas()`` method adapted for a remote function."""
             if all(
                 isinstance(obj, (pandas.DataFrame, pandas.Series))
                 for obj in partition_data
@@ -1218,9 +1215,14 @@ class PandasDataframePartitionManager(ClassLogger, ABC):
         partition_refs = [
             partition.list_of_blocks[0] for partition in partitions_flattened
         ]
-        return partitions.flat[0].apply(
-            preprocessed_func, partition_shape, *partition_refs
-        )
+        result = np.array(
+            [
+                partitions.flat[0].apply(
+                    preprocessed_func, partition_shape, *partition_refs
+                )
+            ]
+        ).reshape(1, -1)
+        return result
 
     @classmethod
     @wait_computations_if_benchmark_mode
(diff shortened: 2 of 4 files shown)


arunjose696:
Few issues I have with the approach
1)peak memory consumption in the worker that converts dataframe to single partition would be still high and same as the previous approach.
2)As we are broadcasting the  right df as modin DF this is slightly heavier in memory compared to converting the dataframe to pandas. When comparing peak memory consumption of workers for the below snipet the observation was.
to_pandas<current<master

```
modin_df =  pd.DataFrame(np.random.randint(0,100,size=(1000, 1000)),)
modin_df2 = pd.DataFrame(np.random.randint(0,100,size=(10000, 1000)))
modin_result = pd.merge(modin_df, modin_df2, how="left")
```

anmyachev:
> Few issues I have with the approach 1)peak memory consumption in the worker that converts dataframe to single partition would be still high and same as the previous approach. 2)As we are broadcasting the right df as modin DF this is slightly heavier in memory compared to converting the dataframe to pandas. When comparing peak memory consumption of workers for the below snippet the observation was. **Approach with converting right.to_pandas<current<master**
> 
> ```
> modin_df =  pd.DataFrame(np.random.randint(0,100,size=(1000, 1000)),)
> modin_df2 = pd.DataFrame(np.random.randint(0,100,size=(10000, 1000)))
> modin_result = pd.merge(modin_df, modin_df2, how="left")
> ```
> 
> Can I have some suggestions on how to improve this?

Might it be possible to call `right.to_pandas` in the worker process instead of the main one? (with some changes)

arunjose696:
> Might it be possible to call `right.to_pandas` in the worker process instead of the main one? (with some changes)

To call right.to_pandas on workers we would need to still send the right modin dataframe to worker.  Wouldnt it still increase the memory consumption same as that in this case, why would it be better compared to the current approach?

YarShev:
@arunjose696, I think @anmyachev means calling `to_pandas` in a single worker process to get a single partitioned Modin DataFrame out of all partitions of the right Modin DataFrame. It seems we could do the following but I am not sure if this works. I guess some changes would be required for the implementation to work.

```python
def force_materialization(self) -> "PandasDataframe":
    row_partitions = self._partition_mgr_cls.row_partitions(self._partitions)
    col_partition = self._partition_mgr_cls.column_partitions(row_partitions)
    new_frame = np.array([col_partition[0].apply(lambda df: df, num_splits=1)])
    return new_frame
```

arunjose696:
> 
> ```python
> def force_materialization(self) -> "PandasDataframe":
>     row_partitions = self._partition_mgr_cls.row_partitions(self._partitions)
>     col_partition = self._partition_mgr_cls.column_partitions(row_partitions)
>     new_frame = np.array([col_partition[0].apply(lambda df: df, num_splits=1)])
>     return new_frame
> ```

I tried this approach. by making a small change, it works. However the  the memory consumption is increasing for several workers during the force_materialization even though only one remote worker is utilized. I tried checking the memory consumption of workers with below script in which I log the memory consumption in workers before and after the force materialization call. The memory consumption in multiple workers have gone up.

<details>
<summary>force_materialization.py</summary>

```
import modin.pandas as pd
import numpy as np
import time
import sys
import os
import re
import platform
import warnings

import psutil

_VM_PEAK_PATTERN = r"VmHWM:\s+(\d+)"

def get_max_memory_usage(proc=psutil.Process()):
    """Reads maximum memory usage in MB from process history. Returns 0 on non-linux systems
    or if the process is not alive."""
    max_mem = 0
    try:
        with open(f"/proc/{proc.pid}/status", "r") as stat:
            for match in re.finditer(_VM_PEAK_PATTERN, stat.read()):
                max_mem = float(match.group(1))
                # MB conversion
                max_mem = int(max_mem / 1024)
                break
    except FileNotFoundError:
        if platform.system() == "Linux":
            warnings.warn(f"Couldn't open `/proc/{proc.pid}/status` file. Is the process alive?")
        else:
            warnings.warn("Couldn't get the max memory usage on a non-Linux platform.")
        return 0
    max_mem_used=max_mem + sum(get_max_memory_usage(c) for c in proc.children())
    print(f"for proceess with name {proc.name()} and  { len(proc.children())} children the mem used is {max_mem}")
    return max_mem_used
modin_df =  pd.DataFrame(np.random.randint(0,100,size=(1000, 1000)),)

execute(modin_df)
print("before force_materialization")
print(f"total memory consumed ={get_max_memory_usage()}")
xf=modin_df._query_compiler._modin_frame.force_materialization()

print("/n/nafter force_materialization")
print(f"total memory consumed ={get_max_memory_usage()}")
```
</details>

anmyachev:
> ```python
> def force_materialization(self) -> "PandasDataframe":
>     row_partitions = self._partition_mgr_cls.row_partitions(self._partitions)
>     col_partition = self._partition_mgr_cls.column_partitions(row_partitions)
>     new_frame = np.array([col_partition[0].apply(lambda df: df, num_splits=1)])
>     return new_frame
> ```
> I tried this approach. by making a small change, it works. However the the memory consumption is increasing for several workers during the force_materialization even though only one remote worker is utilized. I tried checking the memory consumption of workers with below script in which I log the memory consumption in workers before and after the force materialization call. The memory consumption in multiple workers have gone up.

As far as I remember, with this approach it is possible that intermediate partitions are created using method `force_materialization`, which may explain the increased memory consumption.

https://github.com/modin-project/modin/blob/fe19363d58dfd50834971ee56e5038a822374594/modin/core/dataframe/pandas/partitioning/axis_partition.py#L131-L139

I would like to consider the possibility of creating a pandas dataframe in a worker process, without creating intermediate objects. The closest implementation to what I think is the best solution here is the following code:

https://github.com/modin-project/modin/blob/fe19363d58dfd50834971ee56e5038a822374594/modin/core/dataframe/pandas/partitioning/partition_manager.py#L762

modin/core/dataframe/pandas/partitioning/partition_manager.py
```diff
@@ -1148,6 +1148,30 @@ def _apply_func_to_list_of_partitions(cls, func, partitions, **kwargs):
...
+        partitions : np.ndarray
+            The partitions which have to be converted to a single partition.
+
+        Returns
+        -------
+        PandasDataframePartition
+            A PandasDataframePartition object which holds the data of all partitions.
+        """
+
+        def call_to_pandas(data, partitions):
```
anmyachev, line 1170: Maybe it can help to fix tests on Ray in CI:
```suggestion
        cls.finalize(partitions)
        def call_to_pandas(data, partitions):
```

modin/core/dataframe/pandas/partitioning/partition_manager.py
```diff
@@ -1148,6 +1148,84 @@ def _apply_func_to_list_of_partitions(cls, func, partitions, **kwargs):
...
+            The partitions which have to be converted to a single partition.
+
+        Returns
+        -------
+        PandasDataframePartition
+            A PandasDataframePartition object which holds the data of all partitions.
+        """
+        cls.finalize(partitions)
+
+        def convert_to_pandas(data, partition_shape, *partition_data):
```
YarShev, line 1171:
```suggestion
        def to_pandas_remote(data, partition_shape, *partition_data):
            """Copy of ``cls.to_pandas()`` but adapted for a remote call."""
```

modin/core/dataframe/pandas/partitioning/partition_manager.py
```diff
@@ -1148,6 +1148,84 @@ def _apply_func_to_list_of_partitions(cls, func, partitions, **kwargs):
...
+        ----------
+        partitions : np.ndarray
+            The partitions which have to be converted to a single partition.
+
+        Returns
+        -------
+        PandasDataframePartition
+            A PandasDataframePartition object which holds the data of all partitions.
+        """
+        cls.finalize(partitions)
```
YarShev, line 1169: Do we really need this? See my question about necessacity of the call `list_of_blocks`. Do we really need both (they do similar stuff)? Do we need only one call (either finalize or list_of_blocks)?

modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -2703,6 +2703,28 @@ def explode(self, axis: Union[int, Axis], func: Callable) -> "PandasDataframe":
             partitions, new_index, new_columns, row_lengths, column_widths
         )
 
+    def to_pandas_in_remote_function(self) -> "PandasDataframe":
```
YarShev, line 2706:
```suggestion
    def combine(self) -> "PandasDataframe":
```
or `to_single_partition`

modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -2703,6 +2703,28 @@ def explode(self, axis: Union[int, Axis], func: Callable) -> "PandasDataframe":
...
+        """
+        Create a PandasFrame with single partition from the partitions of the current dataframe.
```
YarShev, line 2708:
```suggestion
        Create a single partition PandasDataframe from the partitions of the current dataframe.
```

modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -2703,6 +2703,28 @@ def explode(self, axis: Union[int, Axis], func: Callable) -> "PandasDataframe":
...
+        """
+        partition_with_to_pandas_data = [
+            self._partition_mgr_cls.convert_partitions_to_pandas_in_remote_func(
+                self._partitions
+            )
+        ]
+        partitions = np.array(partition_with_to_pandas_data).reshape(1, -1)
+
+        result = self.__constructor__(
+            partitions, self.index, columns=self.columns, dtypes=self.dtypes
```
YarShev, line 2723: What about row_lengths and column_widths?

modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -2703,6 +2703,28 @@ def explode(self, axis: Union[int, Axis], func: Callable) -> "PandasDataframe":
...
+        )
+        result.synchronize_labels()
```
YarShev, line 2725: Do we really need this?

modin/core/dataframe/pandas/partitioning/partition_manager.py
```diff
@@ -1148,6 +1148,84 @@ def _apply_func_to_list_of_partitions(cls, func, partitions, **kwargs):
         preprocessed_func = cls.preprocess_func(func)
         return [obj.apply(preprocessed_func, **kwargs) for obj in partitions]
 
+    @classmethod
+    def convert_partitions_to_pandas_in_remote_func(cls, partitions):
```
YarShev, line 1152:
```suggestion
    def combine(cls, partitions):
```
or to_single_partition

modin/core/dataframe/pandas/dataframe/dataframe.py
arunjose696: As the partitions are different for the new dataframe the row_lengths and col_widths would not be same as self

modin/core/dataframe/pandas/dataframe/dataframe.py
arunjose696: Yes, the index could be used as "on" parameter in merge if, the indexes are not synchronized when the result dataframe broadcasted it would not have the index information in the remote function causing an error.

modin/core/dataframe/pandas/partitioning/partition_manager.py
```diff
@@ -1148,6 +1148,84 @@ def _apply_func_to_list_of_partitions(cls, func, partitions, **kwargs):
...
+                objs = iter(partition_data)
+                partition_data = [
+                    [next(objs) for _ in range(width)] for __ in range(height)
+                ]
+            else:
+                # Partitions do not always contain pandas objects, for example, hdk uses pyarrow tables.
+                # This implementation comes from the fact that calling `partition.get`
+                # function is not always equivalent to `partition.to_pandas`.
+                partition_data = [
+                    [obj.to_pandas() for obj in part] for part in partitions
```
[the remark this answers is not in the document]
arunjose696: This was a mistake, will fix it

modin/core/dataframe/pandas/dataframe/dataframe.py
YarShev: We could sum them up if they are available.

```python
new_row_lengths = [sum(self._row_lengths_cache)]
new_column_widths = [sum(self._colum_widths_cache)]
```

modin/core/dataframe/pandas/dataframe/dataframe.py
YarShev: I see, thanks!

modin/core/dataframe/pandas/partitioning/partition_manager.py
arunjose696: Got it,  as draining call queue is done is list_of_blocks, I will remove this.

modin/core/dataframe/pandas/partitioning/partition_manager.py
```diff
@@ -1148,6 +1148,84 @@ def _apply_func_to_list_of_partitions(cls, func, partitions, **kwargs):
...
+            if len(df_rows) == 0:
+                return pandas.DataFrame()
+            else:
+                return concatenate(df_rows)
+
+        preprocessed_func = cls.preprocess_func(convert_to_pandas)
+        partition_data = [
+            block
+            for partition in partitions.flatten()
+            for block in partition.list_of_blocks
```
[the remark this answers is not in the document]
arunjose696: This is clear to me, intent here is to pass the object references/futures itself as materialization would be expensive. The idea is to get the data of partitions in the remote functions by using object references.

 If partitions are passed as in `partitions.flat[0].apply(fun, partitions) ` when trying to access data of a partition in remote function as in  `partitions[0].list_of_blocks` or `partitions[0]._data` we get the object reference in the remote function and not the data in object reference, The object reference would still need to be converted to actual data calling the engine_wrapper.materialize(), which is why the data is extracted with list_of_blocks and is passed to remote function.

modin/core/dataframe/pandas/dataframe/dataframe.py
arunjose696: This should help

modin/core/dataframe/pandas/partitioning/partition_manager.py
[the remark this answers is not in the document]
YarShev: I see, makes sense.

modin/core/dataframe/pandas/dataframe/dataframe.py
YarShev: ping

modin/core/dataframe/pandas/partitioning/partition_manager.py
```diff
@@ -1141,6 +1141,86 @@ def _apply_func_to_list_of_partitions(cls, func, partitions, **kwargs):
         preprocessed_func = cls.preprocess_func(func)
         return [obj.apply(preprocessed_func, **kwargs) for obj in partitions]
 
+    @classmethod
+    def to_single_partition(cls, partitions):
+        """
+        Call to_pandas for the given NumPy array of partitions.
+
+        For a NumPy array of partitions create a single partition
+        that has the data for all partitions.
```
YarShev, line 1150:
```suggestion
        Convert a NumPy 2D array of partitions to a NumPy 2D array of a single partition.
```

modin/core/dataframe/pandas/partitioning/partition_manager.py
```diff
@@ -1141,6 +1141,86 @@ def _apply_func_to_list_of_partitions(cls, func, partitions, **kwargs):
...
+
+        Parameters
+        ----------
+        partitions : np.ndarray
+            The partitions which have to be converted to a single partition.
+
+        Returns
+        -------
+        PandasDataframePartition
+            A PandasDataframePartition object which holds the data of all partitions.
```
YarShev, line 1160:
```suggestion
        np.ndarray
            A NumPy 2D array of a single partition.
```

modin/core/dataframe/pandas/partitioning/partition_manager.py
```diff
@@ -1141,6 +1141,86 @@ def _apply_func_to_list_of_partitions(cls, func, partitions, **kwargs):
...
+        """
+
+        def to_pandas_remote(data, partition_shape, *partition_data):
```
YarShev, line 1163:
```suggestion
        def to_pandas_remote(data, partition_shape, *partition_data)
        """Copy of cls.to_pandas() method adapted for a remote function."""
```

modin/core/dataframe/pandas/partitioning/partition_manager.py
```diff
@@ -1141,6 +1141,86 @@ def _apply_func_to_list_of_partitions(cls, func, partitions, **kwargs):
...
+            else:
+                return concatenate(df_rows)
+
+        preprocessed_func = cls.preprocess_func(to_pandas_remote)
+        partition_shape = partitions.shape
+        partitions_flattened = partitions.flatten()
+        for idx, part in enumerate(partitions_flattened):
+            if hasattr(part, "force_materialization"):
+                partitions_flattened[idx] = part.force_materialization()
+        partition_data = [
```
YarShev, line 1217:
```suggestion
        partition_refs = [
```

modin/core/dataframe/pandas/partitioning/partition_manager.py
arunjose696: This function returns the output of partition.apply() which would just be one PandasDataframePartition this would be converted to numpy array in dataframe class.

YarShev: I see, but we need to change this to return a numpy array of partitions. The partition manager's methods should return a numpy array of new partitions so the dataframe could take it in the constrctor. We should not wrap partitions in a numpy array at the dataframe layer.

arunjose696: okay

modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -2762,6 +2762,40 @@ def explode(self, axis: Union[int, Axis], func: Callable) -> "PandasDataframe":
...
+    def combine(self) -> "PandasDataframe":
+        """
+        Create a single partition PandasDataframe from the partitions of the current dataframe.
+
+        Returns
+        -------
+        PandasDataframe
+            A single partition PandasDataframe.
+        """
+        partitions = self._partition_mgr_cls.to_single_partition(self._partitions)
```
YarShev, line 2774:
```suggestion
        partitions = self._partition_mgr_cls.combine(self._partitions)
```
Let's align the naming then.

modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -2762,6 +2762,40 @@ def explode(self, axis: Union[int, Axis], func: Callable) -> "PandasDataframe":
...
+        columns = self.copy_columns_cache()
+        dtypes = self.copy_dtypes_cache()
+        result = self.__constructor__(
+            partitions,
+            index=index,
+            columns=columns,
+            dtypes=dtypes,
+            row_lengths=new_row_lengths,
+            column_widths=new_column_widths,
+        )
```
YarShev, line 2795:
```suggestion
        result = self.__constructor__(
            partitions,
            index=self.copy_index_cache(),
            columns=self.copy_columns_cache(),
            row_lengths=(
                [sum(self._row_lengths_cache)]
                if self._row_lengths_cache is not None
                else None
            ),
            column_widths=(
                [sum(self._column_widths_cache)]
                if self._column_widths_cache is not None
                else None
            ),
            dtypes=self.copy_dtypes_cache(),
        )
```

modin/core/dataframe/pandas/partitioning/partition_manager.py
```diff
@@ -1141,6 +1141,89 @@ def _apply_func_to_list_of_partitions(cls, func, partitions, **kwargs):
         preprocessed_func = cls.preprocess_func(func)
         return [obj.apply(preprocessed_func, **kwargs) for obj in partitions]
 
+    @classmethod
+    def to_single_partition(cls, partitions):
```
YarShev, line 1145:
```suggestion
    def combine(cls, partitions):
```

modin/core/dataframe/pandas/partitioning/partition_manager.py
```diff
@@ -1141,6 +1141,89 @@ def _apply_func_to_list_of_partitions(cls, func, partitions, **kwargs):
...
+
+        Parameters
+        ----------
+        partitions : np.ndarray
+            The partitions which have to be converted to a single partition.
+
+        Returns
+        -------
+        np.ndarray
+                    A NumPy 2D array of a single partition.
```
YarShev, line 1157:
```suggestion
        np.ndarray
            A NumPy 2D array of a single partition.
```

modin/core/storage_formats/pandas/query_compiler.py
```diff
@@ -614,13 +615,12 @@ def map_func(
                         .set_index(list(left_renamer.values()))
                     )
                     new_dtypes = ModinDtypes.concat([left_dtypes, right_dtypes])
-
```
anmyachev, line 617: Please revert

modin/core/dataframe/pandas/partitioning/partition_manager.py
```diff
@@ -1141,6 +1141,89 @@ def _apply_func_to_list_of_partitions(cls, func, partitions, **kwargs):
...
+
+        preprocessed_func = cls.preprocess_func(to_pandas_remote)
+        partition_shape = partitions.shape
+        partitions_flattened = partitions.flatten()
+        for idx, part in enumerate(partitions_flattened):
+            if hasattr(part, "force_materialization"):
+                partitions_flattened[idx] = part.force_materialization()
+        partition_refs = [
+            partition.list_of_blocks[0] for partition in partitions_flattened
+        ]
```
anmyachev, line 1217: Why not use `finalize` for this?

modin/core/dataframe/pandas/partitioning/partition_manager.py
```diff
@@ -1141,6 +1141,89 @@ def _apply_func_to_list_of_partitions(cls, func, partitions, **kwargs):
...
+        partitions : np.ndarray
+            The partitions which have to be converted to a single partition.
+
+        Returns
+        -------
+        np.ndarray
+            A NumPy 2D array of a single partition.
+        """
+
+        def to_pandas_remote(data, partition_shape, *partition_data):
```
anmyachev, line 1160: Can we reduce code duplication for that?

modin/core/dataframe/pandas/partitioning/partition_manager.py
```diff
@@ -1141,6 +1141,89 @@ def _apply_func_to_list_of_partitions(cls, func, partitions, **kwargs):
...
+            partition.list_of_blocks[0] for partition in partitions_flattened
+        ]
+        result = np.array(
+            [
+                partitions.flat[0].apply(
+                    preprocessed_func, partition_shape, *partition_refs
+                )
+            ]
+        ).reshape(1, -1)
+        return result
```
anmyachev, line 1225: For better readability
```suggestion
        combined_partition = partitions.flat[0].apply(
            preprocessed_func, partition_shape, *partition_refs
        )
        return np.array([combined_partition]).reshape(1, -1)
```

modin/core/dataframe/pandas/partitioning/partition_manager.py
arunjose696: Removed finalize because when calling partition.list_of_blocks the drain_call_queue() is called in  list_of_blocks function call.

anmyachev: > Removed finalize because when calling partition.list_of_blocks the drain_call_queue() is called in list_of_blocks function call.

Ok.

`force_materialization` is also called in `def list_of_blocks` (through `list_of_block_partitions`). Can we remove `force_materialization` call here?

arunjose696: In the case when axis partitions are composed of other axis partitions as [here ](https://github.com/arunjose696/modin/blob/master/modin/core/dataframe/pandas/partitioning/axis_partition.py#L119), the partitions would not be force_materialized, and would have multiple partitions as the current implementation of converting to_pandas   requires partitions to have only single partition(https://github.com/modin-project/modin/pull/6880/files#diff-ff5fbf152c79b8591d2ff791f04cda1a8db8e1f082271c98373d6b7f863fd11cR1216) this would not work. 

Also I could see a comment about removing force_materialization() might increase the complexity 
https://github.com/modin-project/modin/pull/4776#discussion_r1032628169

anmyachev: got it, thanks! ok for now.

modin/core/dataframe/pandas/utils.py
```diff
@@ -65,3 +67,54 @@ def concatenate(dfs):
...
+
+
+def create_dataframe_from_partition_data(partition_data, partition_shape):
+    """
+    Convert partition data of multiple dataframes to a single dataframe.
+
+    Parameters
+    ----------
+    partition_data : list
+        List of pandas DataFrames/Object references holding pandas DataFrames.
```
anmyachev, line 79: ?
```suggestion
        List of pandas DataFrames/Object references holding pandas DataFrames.
    partition_shape : int or tuple
```

modin/core/dataframe/pandas/utils.py
```diff
@@ -65,3 +67,56 @@ def concatenate(dfs):
...
+
+    df_rows = [
+        pandas.concat([part for part in row], axis=axis, copy=False)
+        for row in partition_data
+        if not all(is_part_empty(part) for part in row)
+    ]
+    if len(df_rows) == 0:
+        return pandas.DataFrame()
+    else:
+        return concatenate(df_rows)
```
YarShev, line 122: Let's add a parameter named `copy`, which we would pass in pandas.concat. We could avoid a data copy in a remote function because Ray itself will store a copy of data.

anmyachev (approved):
LGTM!

modin/core/dataframe/pandas/partitioning/partition_manager.py
arunjose696: Done

anmyachev: Thanks!

modin/core/dataframe/pandas/partitioning/partition_manager.py
```diff
@@ -1141,6 +1098,42 @@ def _apply_func_to_list_of_partitions(cls, func, partitions, **kwargs):
...
+        partitions : np.ndarray
+            The partitions which have to be converted to a single partition.
+
+        Returns
+        -------
+        np.ndarray
+            A NumPy 2D array of a single partition.
+        """
+
+        def to_pandas_remote(data, partition_shape, *partition_data):
```
YarShev, line 1117: `data` parameter is not used anywhere. Should be removed.

modin/core/dataframe/pandas/utils.py
```diff
@@ -28,6 +30,8 @@ def concatenate(dfs):
     ----------
     dfs : list
         List of pandas DataFrames to concatenate.
+    make_copy : bool, default: True
```
YarShev, line 33:
```suggestion
    copy : bool, default: True
```

modin/core/dataframe/pandas/utils.py
```diff
@@ -60,8 +64,66 @@ def concatenate(dfs):
...
-    if len(dfs) == 1:
+    if len(dfs) == 1 and make_copy:
         # concat doesn't make a copy if len(dfs) == 1,
         # so do it explicitly
         return dfs[0].copy()
-    return pandas.concat(dfs, copy=True)
+    return pandas.concat(dfs, copy=make_copy)
+
+
+def create_dataframe_from_partition_data(
```
YarShev, line 74:
```suggestion
def create_pandas_df_from_partitions(
```

modin/core/dataframe/pandas/partitioning/partition_manager.py
arunjose696: data parameter  is by default present in any function passed to partition.apply() this would hold the data of partition, I am quite not sure if it would be possible to remove this.

modin/core/dataframe/pandas/utils.py
arunjose696: Done.

modin/core/dataframe/pandas/partitioning/partition_manager.py
YarShev: Ah, I see. Then we should probably pass partition_refs[1:] in this function and combine it with data argument to avoid an extra deserialization of a pandas dataframe.

modin/core/dataframe/pandas/partitioning/partition_manager.py
```diff
@@ -1141,6 +1098,42 @@ def _apply_func_to_list_of_partitions(cls, func, partitions, **kwargs):
...
+            )
+
+        preprocessed_func = cls.preprocess_func(to_pandas_remote)
+        partition_shape = partitions.shape
+        partitions_flattened = partitions.flatten()
+        for idx, part in enumerate(partitions_flattened):
+            if hasattr(part, "force_materialization"):
+                partitions_flattened[idx] = part.force_materialization()
+        partition_refs = [
+            partition.list_of_blocks[0] for partition in partitions_flattened
```
YarShev, line 1130:
```suggestion
            partition.list_of_blocks[0] for partition in partitions_flattened[1:]
```

modin/core/dataframe/pandas/partitioning/partition_manager.py
```diff
@@ -1141,6 +1098,42 @@ def _apply_func_to_list_of_partitions(cls, func, partitions, **kwargs):
...
+        Returns
+        -------
+        np.ndarray
+            A NumPy 2D array of a single partition.
+        """
+
+        def to_pandas_remote(data, partition_shape, *partition_data):
+            """Copy of ``cls.to_pandas()`` method adapted for a remote function."""
+            return create_pandas_df_from_partitions(
+                partition_data, partition_shape, called_from_remote=True
```
YarShev, line 1120:
```suggestion
                partition_data.insert(0, data), partition_shape, called_from_remote=True
```

modin/core/dataframe/pandas/partitioning/partition_manager.py
YarShev, line 1117:
```suggestion
        def to_pandas_remote(df, partition_shape, *dfs):
```

YarShev (approved):

YarShev:
@anmyachev, any comments?

anmyachev (approved):

modin/core/dataframe/pandas/partitioning/partition_manager.py
arunjose696: This is smart. Thanks

PR #6915, opened 2024-02-06
FEAT-#6914: Add a config for setting a number of threads per Dask worker

### LINKED ISSUE #6914: Number of threads per Dask worker
We should add a configuration value for setting a number of threads per Dask worker.

YarShev:
## What do these changes do?

- [x] Resolves #6914  
- [x] tests passing

### DIFF
--- a/modin/core/execution/dask/common/utils.py
+++ b/modin/core/execution/dask/common/utils.py
@@ -19,10 +19,10 @@ from modin.config import (
     CIAWSAccessKeyID,
     CIAWSSecretAccessKey,
     CpuCount,
+    DaskThreadsPerWorker,
     GithubCI,
     Memory,
     NPartitions,
-    DaskThreadsPerWorker,
 )
 from modin.core.execution.utils import set_env
 from modin.error_message import ErrorMessage
--- a/modin/core/execution/dask/common/utils.py
+++ b/modin/core/execution/dask/common/utils.py
@@ -22,6 +22,7 @@ from modin.config import (
     GithubCI,
     Memory,
     NPartitions,
+    DaskThreadsPerWorker,
 )
 from modin.core.execution.utils import set_env
 from modin.error_message import ErrorMessage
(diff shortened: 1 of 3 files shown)


dchigarev (dismissed):

modin/core/execution/dask/common/utils.py
```diff
@@ -54,12 +55,17 @@ def _disable_warnings():
...
+        threads_per_worker = DaskThreadsPerWorker.get()
         memory_limit = Memory.get()
         worker_memory_limit = memory_limit // num_cpus if memory_limit else "auto"
 
         # when the client is initialized, environment variables are inherited
         with set_env(PYTHONWARNINGS="ignore::FutureWarning"):
-            client = Client(n_workers=num_cpus, memory_limit=worker_memory_limit)
+            client = Client(
+                n_workers=num_cpus,
+                threads_per_worker=threads_per_worker,
```
anmyachev, line 66: What is the default value for this parameter?

YarShev: There is no a default value for this parameter in Dask. Dask computes it based on some logic in depend on available CPUs and n_workers to spawn. I think it makes sense that every distributed engine in Modin by default relies on multiprocessing rather than multithreading+multiprocessing.

anmyachev: I saw the following definition:
```python
        if n_workers and threads_per_worker is None:
            # Overcommit threads per worker, rather than undercommit
            threads_per_worker = max(1, int(math.ceil(CPU_COUNT / n_workers)))
```

This doesn't seem so bad. Well, I'm not against this change, but don't we need to split this pull request into 2: adding a new configuration option and changing the default parameter value for Dask?

YarShev: It doesn't make sense to add a config and don't use it anywhere. The change in this PR explicitly says that the default value is 1 and it is used.

anmyachev: > It doesn't make sense to add a config and don't use it anywhere. The change in this PR explicitly says that the default value is 1 and it is used.

If you set the default value to `None`, then everything will be as before, but the user will have the opportunity to specify their own values for this variable. Although I don’t insist, it seems to me that this way we will more clearly show that we are changing Dask’s behavior for the users.

YarShev: Makes sense

anmyachev (approved):
LGTM!

PR #6941, opened 2024-02-16
FIX-#6935: Fix Merge failed when right operand is an empty dataframe

### LINKED ISSUE #6935: Merge failed when right operand is an empty dataframe
Modin 0.27.0

```python
import modin.pandas as pd
import numpy as np
df = pd.DataFrame(np.random.rand(100, 100))
df2 = df.iloc[:0]
pd.merge(df, df2) <- failed
```

Traceback:
```python
  File "modin\logging\logger_decorator.py", line 129, in run_and_log
    return obj(*args, **kwargs)
  File "modin\pandas\general.py", line 97, in merge
    return left.merge(
  File "modin\logging\logger_decorator.py", line 129, in run_and_log
    return obj(*args, **kwargs)
  File "modin\pandas\dataframe.py", line 1305, in merge
    query_compiler=self._query_compiler.merge(
  File "modin\logging\logger_decorator.py", line 129, in run_and_log
    return obj(*args, **kwargs)
  File "modin\core\storage_formats\pandas\query_compiler.py", line 523, in merge
    right_to_broadcast = right._modin_frame.combine()
  File "modin\logging\logger_decorator.py", line 129, in run_and_log
    return obj(*args, **kwargs)
  File "modin\core\dataframe\pandas\dataframe\dataframe.py", line 2774, in combine
    partitions = self._partition_mgr_cls.combine(self._partitions)
  File "modin\logging\logger_decorator.py", line 129, in run_and_log
    return obj(*args, **kwargs)
  File "modin\core\dataframe\pandas\partitioning\partition_manager.py", line 1132, in combine
    combined_partition = partitions.flat[0].apply(
IndexError: index 0 is out of bounds for axis 0 with size 0
```

@arunjose696 could you take a look?

arunjose696:
## What do these changes do?
Fixing corner case when partitions are empty for merge.

- [ ] Resolves #6935 ? 
- [ ] tests added and passing

### DIFF
--- a/modin/core/dataframe/pandas/dataframe/dataframe.py
+++ b/modin/core/dataframe/pandas/dataframe/dataframe.py
@@ -3354,10 +3354,9 @@ class PandasDataframe(ClassLogger):
             if df._partitions.size > 0:
                 return df._partitions
             else:
-                empty_partition = self._partition_mgr_cls.create_partition_from_data(
+                return np.array([[self._partition_mgr_cls._partition_class.put(
                     pandas.DataFrame(index=df.index, columns=df.columns)
-                )
-                return empty_partition
+                ]])
 
         if other is not None:
             if not isinstance(other, list):
--- a/modin/core/dataframe/pandas/dataframe/dataframe.py
+++ b/modin/core/dataframe/pandas/dataframe/dataframe.py
@@ -3348,10 +3348,21 @@ class PandasDataframe(ClassLogger):
         PandasDataframe
             New Modin DataFrame.
         """
+
+        def get_partitions(df):
+            """Deal with the corner case if the "other" dataframe has no partitions."""
+            if df._partitions.size > 0:
+                return df._partitions
+            else:
+                empty_partition = self._partition_mgr_cls.create_partition_from_data(
+                    pandas.DataFrame(index=df.index, columns=df.columns)
+                )
+                return empty_partition
+
         if other is not None:
             if not isinstance(other, list):
                 other = [other]
-            other = [o._partitions for o in other] if len(other) else None
+            other = [get_partitions(o) for o in other] if len(other) else None
 
         if apply_indices is not None:
             numeric_indices = self.get_axis(axis ^ 1).get_indexer_for(apply_indices)
--- a/modin/core/dataframe/pandas/dataframe/dataframe.py
+++ b/modin/core/dataframe/pandas/dataframe/dataframe.py
@@ -3354,9 +3354,15 @@ class PandasDataframe(ClassLogger):
             if df._partitions.size > 0:
                 return df._partitions
             else:
-                return np.array([[self._partition_mgr_cls._partition_class.put(
-                    pandas.DataFrame(index=df.index, columns=df.columns)
-                ]])
+                return np.array(
+                    [
+                        [
+                            self._partition_mgr_cls._partition_class.put(
+                                pandas.DataFrame(index=df.index, columns=df.columns)
+                            )
+                        ]
+                    ]
+                )
 
         if other is not None:
             if not isinstance(other, list):
--- a/modin/core/dataframe/pandas/dataframe/dataframe.py
+++ b/modin/core/dataframe/pandas/dataframe/dataframe.py
@@ -3203,6 +3203,25 @@ class PandasDataframe(ClassLogger):
             passed_len += len(internal)
         return result_dict
 
+    def _extract_partitions(self):
+        """
+        Extract partitions if partitions are present.
+
+        If partitions are empty return a dummy partition with empty data but
+        index and columns of current dataframe.
+
+        Returns
+        -------
+        np.ndarray
+            NumPy array with extracted partitions.
+        """
+        if self._partitions.size > 0:
+            return self._partitions
+        else:
+            return self._partition_mgr_cls.create_partition_from_metadata(
+                index=self.index, columns=self.columns
+            )
+
     @lazy_metadata_decorator(apply_axis="both")
     def broadcast_apply_select_indices(
         self,
@@ -3348,26 +3367,10 @@ class PandasDataframe(ClassLogger):
         PandasDataframe
             New Modin DataFrame.
         """
-
-        def get_partitions(df):
-            """Deal with the corner case if the "other" dataframe has no partitions."""
-            if df._partitions.size > 0:
-                return df._partitions
-            else:
-                return np.array(
-                    [
-                        [
-                            self._partition_mgr_cls._partition_class.put(
-                                pandas.DataFrame(index=df.index, columns=df.columns)
-                            )
-                        ]
-                    ]
-                )
-
         if other is not None:
             if not isinstance(other, list):
                 other = [other]
-            other = [get_partitions(o) for o in other] if len(other) else None
+            other = [o._extract_partitions() for o in other] if len(other) else None
 
         if apply_indices is not None:
             numeric_indices = self.get_axis(axis ^ 1).get_indexer_for(apply_indices)
--- a/modin/core/dataframe/pandas/partitioning/partition_manager.py
+++ b/modin/core/dataframe/pandas/partitioning/partition_manager.py
@@ -176,6 +176,23 @@ class PandasDataframePartitionManager(ClassLogger, ABC):
     # END Abstract Methods
 
     @classmethod
+    def create_partition_from_data(cls, data):
+        """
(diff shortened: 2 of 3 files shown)


modin/core/dataframe/pandas/partitioning/partition_manager.py
```diff
@@ -1120,6 +1120,9 @@ def to_pandas_remote(df, partition_shape, *dfs):
                 (df,) + dfs, partition_shape, called_from_remote=True
             )
 
+        if partitions.size == 0:
```
anmyachev, line 1123: @arunjose696 please add also a test

arunjose696: Wrote the test, the fix is slightly complicated as the broadcasted right dataframes being empty could not be reconstructed as the partitions would be empty (and a simple reconstruction of empty datatframe does not works as  the column information is not available).

 I assume I would need to modify broadcast_apply_full_axis logic as other modin operations eg. df.compare(df2) which use broadcast_apply_full_axis would also fail if the second dataframe is empty.

anmyachev: In the case of empty partitions, could we also send data about the indexes that we store at the dataframe level?

arunjose696: I have tried an approach with creating a new partition (with indexes and columns from dataframe level)in broadcast_apply_full axis if the dataframe does not have partitions, this seems to work for the test case.

modin/core/dataframe/pandas/partitioning/partition.py
```diff
@@ -427,6 +427,23 @@ def empty(cls):
         """
         return cls.put(pandas.DataFrame(), 0, 0)
 
+    @classmethod
+    def partition_from_data(cls, df):
```
anmyachev, line 431: It's pretty clear to use the old `put` method I think.

arunjose696: Added this for verbosity , will remove.

anmyachev:
@arunjose696 please rebase

modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -3348,10 +3348,21 @@ def broadcast_apply_full_axis(
...
         """
+
+        def get_partitions(df):
+            """Deal with the corner case if the "other" dataframe has no partitions."""
+            if df._partitions.size > 0:
+                return df._partitions
+            else:
+                empty_partition = self._partition_mgr_cls.create_partition_from_data(
+                    pandas.DataFrame(index=df.index, columns=df.columns)
+                )
```
anmyachev, line 3359: Maybe we should make this as a part of `combine` function?

modin/pandas/test/dataframe/test_join_sort.py
```diff
@@ -383,6 +383,20 @@ def test_merge(test_data, test_data2):
...
+    pandas_df = pandas.DataFrame(data)
+    pandas_df2 = pandas_df.iloc[:0]
+    modin_df = pd.DataFrame(data)
+    modin_df_2 = modin_df.iloc[:0]
+    modin_result = pd.merge(
+        modin_df,
+        modin_df_2,
+    )
+    pandas_result = pandas.merge(pandas_df, pandas_df2)
+    df_equals(modin_result, pandas_result)
```
YarShev, line 397: Let's use `eval_general`.

modin/core/dataframe/pandas/dataframe/dataframe.py
YarShev, line 3359:
```suggestion
                return np.array([[self._partition_mgr_cls._partition_class.put(
                    pandas.DataFrame(index=df.index, columns=df.columns)
                ]])
```

modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -3348,10 +3348,21 @@ def broadcast_apply_full_axis(
...
+                return empty_partition
```
YarShev, line 3360:
```suggestion
```

modin/core/dataframe/pandas/dataframe/dataframe.py
YarShev: I think we can leave out create_partition_from_data. What do you think?

cc @arunjose696, @anmyachev

anmyachev: Do you mean leave it as is? Then this will only work for `broadcast_apply_full ` function; if we want to use this approach with other functions, we will have to duplicate this logic.

arunjose696: I think it would be better to leave the `create_partition_from_data` function as is. The issue is indeed  in `broadcast_apply_full_axis` as we dont send any data to remote function for empty partition dataframe, so I assume the issue would be better addressed in this function itself. 

The `combine` function is called only in merge()/join() so even if we use the logic there to create a empty partition(with index and columns), the `broadcast_apply_full_axis` function would fail for other operations eg `df.compare(df2)`, 

To use this approach in other functions, the alternative I think would be to make [get_partitions()](https://github.com/arunjose696/modin/blob/mergefix/modin/core/dataframe/pandas/dataframe/dataframe.py#L3352) a method of dataframe(and let it handle the case if partitions are empty) class would this be fine ?

modin/pandas/test/dataframe/test_join_sort.py
arunjose696: Done

modin/core/dataframe/pandas/dataframe/dataframe.py
arunjose696: committed this, for now but I think the readability would be better with using `empty_partition` or `empty_data_partition` as a variable and returning the variable. Is  it a better practice to not create a variable and return directly as this would be slightly confusing?

modin/core/dataframe/pandas/partitioning/partition_manager.py
```diff
@@ -1120,6 +1137,9 @@ def to_pandas_remote(df, partition_shape, *dfs):
                 (df,) + dfs, partition_shape, called_from_remote=True
             )
 
+        if partitions.size <= 1:
```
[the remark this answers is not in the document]
arunjose696: Done.

modin/core/dataframe/pandas/dataframe/dataframe.py
arunjose696: Done.

modin/core/dataframe/pandas/partitioning/partition_manager.py
```diff
@@ -175,6 +175,23 @@ def preprocess_func(cls, map_func):
 
     # END Abstract Methods
 
+    @classmethod
+    def create_partition_from_data(cls, data):
```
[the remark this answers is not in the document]
arunjose696: Can I keep this function as is or else we would have to use ` self._partition_mgr_cls._partition_class.put` which would be calling a private function from a private attribute?

anmyachev (approved):
LGTM! Thanks @arunjose696

modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -3203,6 +3203,25 @@ def _prepare_frame_to_broadcast(self, axis, indices, broadcast_all):
...
+        Returns
+        -------
+        np.ndarray
+            NumPy array with extracted partitions.
+        """
+        if self._partitions.size > 0:
+            return self._partitions
+        else:
+            return self._partition_mgr_cls.create_partition_from_data(
+                pandas.DataFrame(index=self.index, columns=self.columns)
```
YarShev, line 3222: Maybe we should also pass in dtypes?

modin/core/dataframe/pandas/partitioning/partition_manager.py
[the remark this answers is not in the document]
YarShev: Maybe rename this method to create_partition_from_metadata?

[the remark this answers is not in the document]
arunjose696: Have renamed and changed arguments to take in a metadata dict as the name would suggest it accepts metadata.

YarShev (approved):

PR #6955, opened 2024-02-21
FEAT-#7139: Use ray-core instead of ray-default

### LINKED ISSUE #7139: Simplify Modin on Ray installation
Details in https://github.com/modin-project/modin/pull/6955#issue-2147677569

anmyachev:
## What do these changes do?

These changes are aimed at reducing the number of dependencies and thereby simplifying the installation of Modin with the Ray engine in the user environment. For example, we might not install dependencies for the Ray dashboard. ~To do this, we could use either package `ray[client]`, but it appears only from version `Ray>=2.6.0` (there is no desire to increase the minimally supported version of Ray so much for now), or switch to using the basic package itself (`ray`), but this will not work for fairly new versions of Ray, so how they don't include `grpcio` dependency.~

~As a transitional option, we can use the basic package with an explicit indication of `grpcio` package . This option will work, in this pull request, I tried it with different versions of Ray. However, there is a chance that `grpcio` package of some version will be installed, with which Ray has problems. Ray can explicitly exclude any versions in such cases (**note:** the exceptions are not made for the basic package, but for `ray[client]` package, so they cannot be reused on our side), but using the proposed approach, we most likely will not know about this until the problem arises for users.~

~If we decide to switch to such a transitional option, then I believe we should also consider switching to `ray[client]` package for `Modin 1.0` release.~

For example, dependencies for `ray[default]==2.9.3`:
```bash
        "default": [
            # If adding dependencies necessary to launch the dashboard api server,
            # please add it to dashboard/optional_deps.py as well.
            "aiohttp >= 3.7",
            "aiohttp_cors",
            "colorful",
            "py-spy >= 0.2.0",
            "requests",
            "gpustat >= 1.0.0",  # for windows
            "grpcio >= 1.32.0; python_version < '3.10'",  # noqa:E501
            "grpcio >= 1.42.0; python_version >= '3.10'",  # noqa:E501
            "opencensus",
            "pydantic!=2.0.*,!=2.1.*,!=2.2.*,!=2.3.*,!=2.4.*,<3",
            "prometheus_client >= 0.7.1",
            "smart_open",
            "virtualenv >=20.0.24, !=20.21.1",  # For pip runtime env.
        ],
```

- [x] Resolves #7139  
- [x] tests ~added and~ passing


docs/requirements-doc.txt
```diff
@@ -13,7 +13,7 @@ recommonmark
 sphinx<6.0.0
 sphinx-click
 # ray==2.5.0 broken: https://github.com/conda-forge/ray-packages-feedstock/issues/100
-ray[default]>=1.13.0,!=2.5.0
+ray[client]>=1.13.0,!=2.5.0
```
anmyachev, line 16: `client` target appeared in Ray 2.6.0

anmyachev:
@YarShev @sfc-gh-dpetersohn @dchigarev thoughts?

docs/requirements-doc.txt
```diff
@@ -13,7 +13,8 @@ recommonmark
 sphinx<6.0.0
 sphinx-click
 # ray==2.5.0 broken: https://github.com/conda-forge/ray-packages-feedstock/issues/100
-ray[default]>=2.1.0,!=2.5.0
+ray>=2.1.0,!=2.5.0
```
YarShev, line 16:
```suggestion
ray-core>=2.1.0,!=2.5.0
```
?

requirements-dev.txt
```diff
@@ -7,8 +7,9 @@ psutil>=5.8.0
 
 ## optional dependencies
 # ray==2.5.0 broken: https://github.com/conda-forge/ray-packages-feedstock/issues/100
-ray[default]>=2.1.0,!=2.5.0
+ray>=2.1.0,!=2.5.0
```
YarShev, line 10:
```suggestion
ray-core>=2.1.0,!=2.5.0
```
?

setup.py
```diff
@@ -7,7 +7,7 @@
 
 dask_deps = ["dask>=2.22.0", "distributed>=2.22.0"]
 # ray==2.5.0 broken: https://github.com/conda-forge/ray-packages-feedstock/issues/100
-ray_deps = ["ray[default]>=2.1.0,!=2.5.0", "pyarrow>=7.0.0"]
+ray_deps = ["ray>=2.1.0,!=2.5.0", "pyarrow>=7.0.0", "grpcio"]
```
YarShev, line 10:
```suggestion
ray_deps = ["ray-core>=2.1.0,!=2.5.0", "pyarrow>=7.0.0", "grpcio"]
```

YarShev:
> switch to using the basic package itself (ray), but this will not work for fairly new versions of Ray, so how they don't include grpcio dependency.

Why did they exclude grpcio? If we install ray-core, it won't work out of the box, right? Doesn't this look like an issue for Ray?

> As a transitional option, we can use the basic package with an explicit indication of grpcio package . This option will work, in this pull request, I tried it with different versions of Ray. However, there is a chance that grpcio package of some version will be installed, with which Ray has problems. Ray can explicitly exclude any versions in such cases (note: the exceptions are not made for the basic package, but for ray[client] package, so they cannot be reused on our side), but using the proposed approach, we most likely will not know about this until the problem arises for users.

If Ray limits grpcio to install specific versions, it seems there shouln't be a problem when installing `pip install grpcio ray-core`. The pip resolver should handle a case with incompatible versions of grpcio for Ray.

> If we decide to switch to such a transitional option, then I believe we should also consider switching to ray[client] package for Modin 1.0 release.

Why?

docs/requirements-doc.txt
anmyachev: Analogue of conda package `ray-core` for pip - just `ray`.

requirements-dev.txt
anmyachev: Analogue of conda package `ray-core` for pip - just `ray`.

setup.py
anmyachev: Analogue of conda package `ray-core` for pip - just `ray`.

anmyachev:
> Why did they exclude grpcio? If we install ray-core, it won't work out of the box, right? Doesn't this look like an issue for Ray?

I don’t know for sure, but this is a conscious decision to transfer this package to another target: `ray[client]`.

> If Ray limits grpcio to install specific versions, it seems there shouldn't be a problem when installing pip install grpcio ray-core. The pip resolver should handle a case with incompatible versions of grpcio for Ray.

The default target does not contain version pins for `grpcio` dependency (only `ray[client]` or `ray[default]`).

>> If we decide to switch to such a transitional option, then I believe we should also consider switching to ray[client] package for Modin 1.0 release.

> Why?

I consider this a more reliable solution, since this is not our direct dependency, but a transitive one from Ray.

YarShev (dismissed):

setup.py
YarShev: ping

anmyachev:
Hi @rkooo567! I would like to be more confident that we are going in the right direction towards reducing Ray dependencies installed by default. Do I understand correctly that in a situation where core functionality and a cluster launcher are needed, it will be enough to install `ray[client]` instead of `ray[default]`? (since the dashboard from 'default' target is not needed)

rkooo567:
I think there's no package that only includes cluster launcher. ray[client] is not for the cluster launcher (it is for ray client mode https://docs.ray.io/en/master/cluster/running-applications/job-submission/ray-client.html, which is not recommended to use)

anmyachev:
> I think there's no package that only includes cluster launcher. ray[client] is not for the cluster launcher (it is for ray client mode https://docs.ray.io/en/master/cluster/running-applications/job-submission/ray-client.html, which is not recommended to use)

@rkooo567 thanks! Now it’s more clear how to deal with this.

anmyachev:
It looks like we can safely remove `ClientObjectRef` from Modin sources. It was introduced in https://github.com/modin-project/modin/pull/2851. Ray fixed the problem a long time ago and defined the type like this: `class ClientObjectRef(raylet.ObjectRef):`

README.md
```diff
@@ -76,6 +76,9 @@ Otherwise, installation of `modin[mpi]` may fail. Refer to
 [Installing with pip](https://unidist.readthedocs.io/en/latest/installation.html#installing-with-pip)
 section of the unidist documentation for more details about installation.
 
+**Note:** Since Modin 0.30.0 we use a reduced set of Ray dependencies: ray-core instead of ray-default.
+This means that the dashboard and cluster launcher are no longer installed by default.
```
YarShev, line 80:
```suggestion
**Note:** Since Modin 0.30.0 we use a reduced set of Ray dependencies: `ray[core]` instead of `ray[default]`.
This means that the dashboard and cluster launcher are no longer installed by default.
If you need those, consider installing `ray[default]` along with `modin[ray]`.
```
Should we also add this note for intallation with conda?

docs/getting_started/installation.rst
```diff
@@ -40,6 +40,9 @@ it is required to have a working MPI implementation installed beforehand.
 Otherwise, installation of ``modin[mpi]`` may fail. Refer to
 `Installing with pip`_ section of the unidist documentation for more details about installation.
 
+**Note:** Since Modin 0.30.0 we use a reduced set of Ray dependencies: ray-core instead of ray-default.
```
YarShev, line 43: same

modin/core/execution/ray/implementations/pandas_on_ray/partitioning/partition.py
```diff
@@ -328,7 +325,7 @@ def ip(self, materialize=True):
         return ip
 
     @property
-    def _data(self) -> Union[ray.ObjectRef, "ClientObjectRef"]:  # noqa: GL08
+    def _data(self) -> Union[ray.ObjectRef]:  # noqa: GL08
```
YarShev, line 328:
```suggestion
    def _data(self) -> ray.ObjectRef:  # noqa: GL08
```

docs/getting_started/installation.rst
```diff
@@ -40,6 +40,9 @@ it is required to have a working MPI implementation installed beforehand.
...
+This means that the dashboard and cluster launcher are no longer installed by default.
```
anmyachev, line 44:
```suggestion
**Note:** Since Modin 0.30.0 we use a reduced set of Ray dependencies: `ray` instead of `ray[default]`.
This means that the dashboard and cluster launcher are no longer installed by default.
If you need those, consider installing `ray[default]` (through pip) along with `modin[ray]` or `ray-default` (through conda).
```

setup.py
anmyachev: I don't think so, this knowledge is easily obtained from the ray installation page (if needed).

README.md
```diff
@@ -76,6 +76,10 @@ Otherwise, installation of `modin[mpi]` may fail. Refer to
 [Installing with pip](https://unidist.readthedocs.io/en/latest/installation.html#installing-with-pip)
 section of the unidist documentation for more details about installation.
 
+**Note:** Since Modin 0.30.0 we use a reduced set of Ray dependencies: `ray` instead of `ray[default]`.
+This means that the dashboard and cluster launcher are no longer installed by default.
+If you need those, consider installing `ray[default]` (through pip) along with `modin[ray]` or `ray-default` (through conda).
```
YarShev, line 81:
```suggestion
If you need those, consider installing `ray[default]` (through pip) or `ray-default` (through conda) along with `modin[ray]`.
```
The first suggestion above.
The second is maybe we should also add this note to installation with conda below like.

```
**Note:** Since Modin 0.30.0 we use a reduced set of Ray dependencies: `ray-core` instead of `ray-default`.
This means that the dashboard and cluster launcher are no longer installed by default.
If you need those, consider installing `ray-default` along with `modin-ray`.
```

YarShev: I would go with the second one.

anmyachev: added

docs/getting_started/installation.rst
```diff
@@ -40,6 +40,10 @@ it is required to have a working MPI implementation installed beforehand.
 Otherwise, installation of ``modin[mpi]`` may fail. Refer to
 `Installing with pip`_ section of the unidist documentation for more details about installation.
 
+**Note:** Since Modin 0.30.0 we use a reduced set of Ray dependencies: `ray` instead of `ray[default]`.
```
YarShev, line 43:
```suggestion
**Note:** Since Modin 0.30.0 we use a reduced set of Ray dependencies: ``ray`` instead of ``ray[default]``.
```

docs/getting_started/installation.rst
```diff
@@ -40,6 +40,10 @@ it is required to have a working MPI implementation installed beforehand.
...
+This means that the dashboard and cluster launcher are no longer installed by default.
+If you need those, consider installing `ray[default]` along with `modin[ray]`.
```
YarShev, line 45:
```suggestion
If you need those, consider installing ``ray[default]`` along with ``modin[ray]``.
```

docs/getting_started/installation.rst
```diff
@@ -129,6 +133,10 @@ it is possible to install modin with chosen engine(s) alongside. Current options
 | modin-all                       | Dask, Ray, Unidist, HDK   |          Linux              |
 +---------------------------------+---------------------------+-----------------------------+
 
+**Note:** Since Modin 0.30.0 we use a reduced set of Ray dependencies: `ray-core` instead of `ray-default`.
```
YarShev, line 136:
```suggestion
**Note:** Since Modin 0.30.0 we use a reduced set of Ray dependencies: ``ray-core`` instead of ``ray-default``.
```

docs/getting_started/installation.rst
```diff
@@ -129,6 +133,10 @@ it is possible to install modin with chosen engine(s) alongside. Current options
...
+This means that the dashboard and cluster launcher are no longer installed by default.
+If you need those, consider installing `ray-default` along with `modin-ray`.
```
YarShev, line 138:
```suggestion
If you need those, consider installing ``ray-default`` along with ``modin-ray``.
```

YarShev (approved):

PR #6972, opened 2024-02-27
FEAT-#6929: Implement Series.case_when in a distributed way

### LINKED ISSUE #6929: Implement Series.case_when in a distributed way

AndreyPavlenko:
## What do these changes do?

- [x] Resolves #6929 
- [x] tests added and passing

### DIFF
--- a/modin/core/dataframe/pandas/dataframe/dataframe.py
+++ b/modin/core/dataframe/pandas/dataframe/dataframe.py
@@ -4688,10 +4688,10 @@ class PandasDataframe(ClassLogger, modin_layer="CORE-DATAFRAME"):
             def case_when(df, caselist):  # pragma: no cover
                 caselist = [
                     tuple(
-                        d.iloc[:, 0] if isinstance(d, pandas.DataFrame) else d
-                        for d in c
+                        data.iloc[:, 0] if isinstance(data, pandas.DataFrame) else data
+                        for data in case_tuple
                     )
-                    for c in caselist
+                    for case_tuple in caselist
                 ]
                 series = df.iloc[:, 0]
                 return pandas.DataFrame({series.name: series.case_when(caselist)})
--- a/modin/core/dataframe/pandas/dataframe/dataframe.py
+++ b/modin/core/dataframe/pandas/dataframe/dataframe.py
@@ -21,8 +21,6 @@ import datetime
 import re
 from typing import TYPE_CHECKING, Callable, Dict, Hashable, List, Optional, Union
 
-import numpy as np
-import pandas
 from pandas._libs.lib import no_default
 from pandas.api.types import is_object_dtype
 from pandas.core.dtypes.common import is_dtype_equal, is_list_like, is_numeric_dtype
@@ -30,7 +28,7 @@ from pandas.core.indexes.api import Index, RangeIndex
 
 from modin.config import Engine, IsRayCluster, NPartitions
 from modin.core.dataframe.base.dataframe.dataframe import ModinDataframe
-from modin.core.dataframe.base.dataframe.utils import Axis, JoinType
+from modin.core.dataframe.base.dataframe.utils import Axis, JoinType, is_trivial_index
 from modin.core.dataframe.pandas.dataframe.utils import (
     ShuffleSortFunctions,
     add_missing_categories_to_groupby,
@@ -54,6 +52,9 @@ if TYPE_CHECKING:
         ProtocolDataframe,
     )
     from pandas._typing import npt
+
+import numpy as np
+import pandas
 
 from modin.logging import ClassLogger
 from modin.pandas.indexing import is_range_like
@@ -4668,3 +4669,166 @@ class PandasDataframe(ClassLogger, modin_layer="CORE-DATAFRAME"):
         ErrorMessage.default_to_pandas(message="`from_dataframe`")
         pandas_df = from_dataframe_to_pandas(df)
         return cls.from_pandas(pandas_df)
+
+    def case_when(self, caselist):
+        """
+        Replace values where the conditions are True.
+
+        Parameters
+        ----------
+        caselist : list of tuples
+
+        Returns
+        -------
+        PandasDataframe
+        """
+        cls = type(self)
+        wrapper_put = self._partition_mgr_cls._execution_wrapper.put
+        if (remote_fn := getattr(cls, "_CASE_WHEN_FN", None)) is None:
+
+            def case_when(df, caselist):  # pragma: no cover
+                caselist = [
+                    tuple(
+                        d.iloc[:, 0] if isinstance(d, pandas.DataFrame) else d
+                        for d in c
+                    )
+                    for c in caselist
+                ]
+                series = df.iloc[:, 0]
+                return pandas.DataFrame({series.name: series.case_when(caselist)})
+
+            cls._CASE_WHEN_FN = remote_fn = wrapper_put(case_when)
+
+        parts_len = len(self._partitions)
+        parts = None
+        lengths = None
+        is_trivial_idx = None
+
+        def copartition(df, fill_value):
+            nonlocal is_trivial_idx
+            copartition = False
+            if is_trivial_idx is None:
+                is_trivial_idx = is_trivial_index(self.index)
+            if (
+                is_trivial_idx != is_trivial_index(df.index)
+                or (not is_trivial_idx and any(self.index != df.index))
+                or (parts_len < len(df._partitions))
+            ):
+                copartition = True
+            else:
+                nonlocal parts, lengths
+                if parts is None:
+                    parts = [p[0] for p in self._partitions]
+                    lengths = self._get_lengths(parts, Axis.ROW_WISE)
+
+                df_parts = [p[0] for p in df._partitions]
+                df_part_lengths = df._get_lengths(df_parts, Axis.ROW_WISE)
+                if any(lengths[i] != df_part_lengths[i] for i in range(len(df_parts))):
+                    copartition = True
+            if copartition:
+                _, list_of_right_parts, joined_index, row_lengths = self._copartition(
+                    Axis.ROW_WISE.value,
+                    df,
+                    how="left",
+                    sort=False,
+                    fill_value=fill_value,
+                )
+                df = self.__constructor__(
+                    list_of_right_parts[0],
+                    joined_index,
+                    df.columns,
+                    row_lengths,
+                    df.column_widths,
+                )
+            return df
+
+        # For Dask the callebles are wrapped for each partition in the map_data() function.
+        # If the same callable is wrapped only once for all partitions, CancelledError is raised.
+        wrap_callable = Engine.get() != "Dask"
+        use_map = wrap_callable
+        new_caselist = []
+        for condition, replacement in caselist:
+            if callable(condition):
+                if wrap_callable:
+                    condition = wrapper_put(condition)
+            else:
+                use_map = False
+                if isinstance(condition, cls):
+                    condition = copartition(condition, True)
+            if callable(replacement):
+                if wrap_callable:
+                    replacement = wrapper_put(replacement)
+            elif use_map and is_list_like(replacement):
+                use_map = False
+                if isinstance(replacement, cls):
+                    replacement = copartition(replacement, None)
+            new_caselist.append((condition, replacement))
+
+        # If all the conditions are callable and the replacements are either
+        # callable or scalar, use map().
+        if use_map:
+            return self.map(func=remote_fn, func_args=[new_caselist], lazy=True)
+
+        def map_data(
+            part_offset,
+            part_len,
+            data,
+            data_offset,
+            fill_value,
+        ):
+            if isinstance(data, cls):
+                if part_offset < len(data._partitions):
+                    return data._partitions[part_offset][0]._data
+                else:
+                    return [fill_value] * part_len
+
+            if isinstance(data, pandas.Series):
+                nonlocal is_trivial_idx
+                if is_trivial_idx is None:
+                    is_trivial_idx = is_trivial_index(self.index)
+                if is_trivial_idx and is_trivial_index(data.index):
+                    data = data[data_offset : data_offset + part_len]
+                    diff = part_len - len(data)
+                    if diff > 0:
+                        data = pandas.concat((data, pandas.Series([fill_value] * diff)))
+                    return data
+                return data.reindex(
+                    self.index[data_offset : data_offset + part_len],
+                    fill_value=fill_value,
+                )
+
+            if not wrap_callable and callable(data):
+                return wrapper_put(data)
+
+            return (
+                data[data_offset : data_offset + part_len]
+                if is_list_like(data)
+                else data
+            )
+
+        if parts is None:
+            parts = [p[0] for p in self._partitions]
+            lengths = self._get_lengths(parts, Axis.ROW_WISE)
+        new_parts = []
+        data_offset = 0
+        for i in range(0, parts_len):
+            part = parts[i]
+            part_len = lengths[i]
+            cases = [
+                tuple(
+                    map_data(i, part_len, d[0], data_offset, d[1])
+                    for d in zip(c, (True, None))
+                )
+                for c in new_caselist
+            ]
+            new_parts.append(
+                part.add_to_apply_calls(
+                    remote_fn,
+                    cases,
+                    length=part_len,
+                    width=1,
+                )
+            )
+            data_offset += part_len
+        new_parts = np.array([[p] for p in new_parts])
+        return self.__constructor__(new_parts, columns=self.columns, index=self.index)
--- a/modin/core/dataframe/pandas/dataframe/dataframe.py
+++ b/modin/core/dataframe/pandas/dataframe/dataframe.py
@@ -4691,145 +4691,141 @@ class PandasDataframe(ClassLogger, modin_layer="CORE-DATAFRAME"):
             or (remote_fn := getattr(cls, "_CASE_WHEN_FN", None)) is None
         ):
 
-            def case_when(df, caselist):  # pragma: no cover
+            def case_when(df, name, caselist):  # pragma: no cover
                 caselist = [
                     tuple(
-                        data.iloc[:, 0] if isinstance(data, pandas.DataFrame) else data
+                        data.squeeze(1) if isinstance(data, pandas.DataFrame) else data
                         for data in case_tuple
                     )
                     for case_tuple in caselist
                 ]
-                series = df.iloc[:, 0]
-                return pandas.DataFrame({series.name: series.case_when(caselist)})
+                return pandas.DataFrame({name: df.squeeze(1).case_when(caselist)})
 
             if single_wrap:
                 cls._CASE_WHEN_FN = remote_fn = wrapper_put(case_when)
             else:
                 remote_fn = case_when
 
-        parts_len = len(self._partitions)
-        parts = None
-        lengths = None
+        name = self.columns[0]
+        use_map = single_wrap
         is_trivial_idx = None
-
-        def copartition(df, fill_value):
-            nonlocal is_trivial_idx
-            copartition = False
-            if is_trivial_idx is None:
-                is_trivial_idx = is_trivial_index(self.index)
-            if (
-                is_trivial_idx != is_trivial_index(df.index)
-                or (not is_trivial_idx and any(self.index != df.index))
-                or (parts_len < len(df._partitions))
+        # Lists of modin frames: first for conditions, second for replacements
+        modin_lists = [[], []]
+        # Fill values for conditions and replacements respectively
+        fill_values = [True, None]
+        new_caselist = []
+        for case_tuple in caselist:
+            new_case = []
+            for data, modin_list, fill_value in zip(
+                case_tuple, modin_lists, fill_values
             ):
-                copartition = True
-            else:
-                nonlocal parts, lengths
-                if parts is None:
-                    parts = [p[0] for p in self._partitions]
-                    lengths = self._get_lengths(parts, Axis.ROW_WISE)
+                if isinstance(data, cls):
+                    modin_list.append(data)
+                elif callable(data):
+                    if single_wrap:
+                        data = wrapper_put(data)
+                elif isinstance(data, pandas.Series):
+                    use_map = False
+                    if is_trivial_idx is None:
+                        self_idx = self.index
+                        length = len(self_idx)
+                        is_trivial_idx = is_trivial_index(self_idx)
+                    if is_trivial_idx and is_trivial_index(data.index):
+                        data = data[:length]
+                        diff = length - len(data)
+                        if diff > 0:
+                            data = pandas.concat(
+                                [data, pandas.Series([fill_value] * diff)]
+                            )
+                    else:
+                        data = data.reindex(self_idx, fill_value=fill_value)
+                elif use_map and is_list_like(data):
+                    use_map = False
+                new_case.append(data)
+            new_caselist.append(tuple(new_case))
 
-                df_parts = [p[0] for p in df._partitions]
-                df_part_lengths = df._get_lengths(df_parts, Axis.ROW_WISE)
-                if any(lengths[i] != df_part_lengths[i] for i in range(len(df_parts))):
-                    copartition = True
-            if copartition:
+        if modin_lists[0] or modin_lists[1]:
+            # Copartition modin frames
+            use_map = False
+            columns = self.columns
+            column_widths = [1]
+            for modin_list, fill_value in zip(modin_lists, fill_values):
                 _, list_of_right_parts, joined_index, row_lengths = self._copartition(
                     Axis.ROW_WISE.value,
-                    df,
+                    modin_list,
                     how="left",
                     sort=False,
                     fill_value=fill_value,
                 )
-                df = self.__constructor__(
-                    list_of_right_parts[0],
-                    joined_index,
-                    df.columns,
-                    row_lengths,
-                    df.column_widths,
+                modin_list.clear()
+                modin_list.extend(
+                    self.__constructor__(
+                        part,
+                        joined_index,
+                        columns,
+                        row_lengths,
+                        column_widths,
+                    )
+                    for part in list_of_right_parts
                 )
-            return df
 
-        use_map = single_wrap
-        new_caselist = []
-        for condition, replacement in caselist:
-            if callable(condition):
-                if single_wrap:
-                    condition = wrapper_put(condition)
-            else:
-                use_map = False
-                if isinstance(condition, cls):
-                    condition = copartition(condition, True)
-            if callable(replacement):
-                if single_wrap:
-                    replacement = wrapper_put(replacement)
-            elif use_map and is_list_like(replacement):
-                use_map = False
-                if isinstance(replacement, cls):
-                    replacement = copartition(replacement, None)
-            new_caselist.append((condition, replacement))
+            # Replace modin frames with copartitioned
+            caselist = new_caselist
+            new_caselist = []
+            for i in range(2):
+                modin_lists[i] = iter(modin_lists[i])
+            for case_tuple in caselist:
+                new_case = tuple(
+                    next(modin_list) if isinstance(data, cls) else data
+                    for data, modin_list in zip(case_tuple, modin_lists)
+                )
+                new_caselist.append(new_case)
 
         # If all the conditions are callable and the replacements are either
         # callable or scalar, use map().
         if use_map:
-            return self.map(func=remote_fn, func_args=[new_caselist], lazy=True)
+            return self.map(func=remote_fn, func_args=[name, new_caselist], lazy=True)
 
+        # Get the chunk of data corresponding the the specified partition
         def map_data(
-            part_offset,
+            part_idx,
             part_len,
             data,
             data_offset,
             fill_value,
         ):
             if isinstance(data, cls):
-                if part_offset < len(data._partitions):
-                    return data._partitions[part_offset][0]._data
-                else:
-                    return [fill_value] * part_len
-
+                return data._partitions[part_idx][0]._data
             if isinstance(data, pandas.Series):
-                nonlocal is_trivial_idx
-                if is_trivial_idx is None:
-                    is_trivial_idx = is_trivial_index(self.index)
-                if is_trivial_idx and is_trivial_index(data.index):
-                    data = data[data_offset : data_offset + part_len]
-                    diff = part_len - len(data)
-                    if diff > 0:
-                        data = pandas.concat((data, pandas.Series([fill_value] * diff)))
-                    return data
-                return data.reindex(
-                    self.index[data_offset : data_offset + part_len],
-                    fill_value=fill_value,
-                )
-
+                return data[data_offset : data_offset + part_len]
             # As mentioned above, this is required for Dask
             if not single_wrap and callable(data):
                 return wrapper_put(data)
-
             return (
                 data[data_offset : data_offset + part_len]
                 if is_list_like(data)
                 else data
             )
 
-        if parts is None:
-            parts = [p[0] for p in self._partitions]
-            lengths = self._get_lengths(parts, Axis.ROW_WISE)
+        parts = [p[0] for p in self._partitions]
+        lengths = self._get_lengths(parts, Axis.ROW_WISE)
         new_parts = []
         data_offset = 0
-        for i in range(0, parts_len):
-            part = parts[i]
-            part_len = lengths[i]
+
+        # Split the data and apply the remote function to each partition
+        # with the corresponding chunk of data
+        for i, part, part_len in zip(range(0, len(parts)), parts, lengths):
             cases = [
                 tuple(
-                    map_data(i, part_len, d[0], data_offset, d[1])
-                    for d in zip(c, (True, None))
+                    map_data(i, part_len, data, data_offset, fill_value)
+                    for data, fill_value in zip(c, (True, None))
                 )
                 for c in new_caselist
             ]
             new_parts.append(
                 part.add_to_apply_calls(
                     remote_fn if single_wrap else wrapper_put(remote_fn),
+                    name,
                     cases,
                     length=part_len,
                     width=1,
--- a/modin/core/dataframe/pandas/dataframe/dataframe.py
+++ b/modin/core/dataframe/pandas/dataframe/dataframe.py
@@ -21,6 +21,8 @@ import datetime
 import re
 from typing import TYPE_CHECKING, Callable, Dict, Hashable, List, Optional, Union
 
+import numpy as np
+import pandas
 from pandas._libs.lib import no_default
 from pandas.api.types import is_object_dtype
 from pandas.core.dtypes.common import is_dtype_equal, is_list_like, is_numeric_dtype
@@ -52,9 +54,6 @@ if TYPE_CHECKING:
         ProtocolDataframe,
     )
     from pandas._typing import npt
-
-import numpy as np
-import pandas
 
 from modin.logging import ClassLogger
 from modin.pandas.indexing import is_range_like
@@ -4742,7 +4741,7 @@ class PandasDataframe(ClassLogger, modin_layer="CORE-DATAFRAME"):
                 )
             return df
 
-        # For Dask the callebles are wrapped for each partition in the map_data() function.
+        # For Dask the callables are wrapped for each partition in the map_data() function.
         # If the same callable is wrapped only once for all partitions, CancelledError is raised.
         wrap_callable = Engine.get() != "Dask"
         use_map = wrap_callable
@@ -4797,6 +4796,7 @@ class PandasDataframe(ClassLogger, modin_layer="CORE-DATAFRAME"):
                     fill_value=fill_value,
                 )
 
+            # As mentioned above, this is required for Dask
             if not wrap_callable and callable(data):
                 return wrapper_put(data)
 
@@ -4831,4 +4831,10 @@ class PandasDataframe(ClassLogger, modin_layer="CORE-DATAFRAME"):
             )
             data_offset += part_len
         new_parts = np.array([[p] for p in new_parts])
-        return self.__constructor__(new_parts, columns=self.columns, index=self.index)
+        return self.__constructor__(
+            new_parts,
+            columns=self.columns,
+            index=self.index,
+            row_lengths=lengths,
+            column_widths=[1],
+        )
--- a/modin/core/dataframe/pandas/dataframe/dataframe.py
+++ b/modin/core/dataframe/pandas/dataframe/dataframe.py
@@ -4681,9 +4681,15 @@ class PandasDataframe(ClassLogger, modin_layer="CORE-DATAFRAME"):
         -------
         PandasDataframe
         """
+        # For Dask the callables must wrapped for each partition, otherwise
+        # the execution could fail with CancelledError.
+        single_wrap = Engine.get() != "Dask"
         cls = type(self)
         wrapper_put = self._partition_mgr_cls._execution_wrapper.put
-        if (remote_fn := getattr(cls, "_CASE_WHEN_FN", None)) is None:
+        if (
+            not single_wrap
+            or (remote_fn := getattr(cls, "_CASE_WHEN_FN", None)) is None
+        ):
 
             def case_when(df, caselist):  # pragma: no cover
                 caselist = [
@@ -4696,7 +4702,10 @@ class PandasDataframe(ClassLogger, modin_layer="CORE-DATAFRAME"):
                 series = df.iloc[:, 0]
                 return pandas.DataFrame({series.name: series.case_when(caselist)})
 
-            cls._CASE_WHEN_FN = remote_fn = wrapper_put(case_when)
+            if single_wrap:
+                cls._CASE_WHEN_FN = remote_fn = wrapper_put(case_when)
+            else:
+                remote_fn = case_when
 
         parts_len = len(self._partitions)
         parts = None
@@ -4741,21 +4750,18 @@ class PandasDataframe(ClassLogger, modin_layer="CORE-DATAFRAME"):
                 )
             return df
 
-        # For Dask the callables are wrapped for each partition in the map_data() function.
-        # If the same callable is wrapped only once for all partitions, CancelledError is raised.
-        wrap_callable = Engine.get() != "Dask"
-        use_map = wrap_callable
+        use_map = single_wrap
         new_caselist = []
         for condition, replacement in caselist:
             if callable(condition):
-                if wrap_callable:
+                if single_wrap:
                     condition = wrapper_put(condition)
             else:
                 use_map = False
                 if isinstance(condition, cls):
                     condition = copartition(condition, True)
             if callable(replacement):
-                if wrap_callable:
+                if single_wrap:
                     replacement = wrapper_put(replacement)
             elif use_map and is_list_like(replacement):
                 use_map = False
@@ -4797,7 +4803,7 @@ class PandasDataframe(ClassLogger, modin_layer="CORE-DATAFRAME"):
                 )
 
             # As mentioned above, this is required for Dask
-            if not wrap_callable and callable(data):
+            if not single_wrap and callable(data):
                 return wrapper_put(data)
 
             return (
@@ -4823,7 +4829,7 @@ class PandasDataframe(ClassLogger, modin_layer="CORE-DATAFRAME"):
             ]
             new_parts.append(
                 part.add_to_apply_calls(
-                    remote_fn,
+                    remote_fn if single_wrap else wrapper_put(remote_fn),
                     cases,
                     length=part_len,
                     width=1,
--- a/modin/core/dataframe/pandas/dataframe/dataframe.py
+++ b/modin/core/dataframe/pandas/dataframe/dataframe.py
@@ -4673,6 +4673,9 @@ class PandasDataframe(ClassLogger, modin_layer="CORE-DATAFRAME"):
         """
         Replace values where the conditions are True.
 
+        This is Series.case_when() implementation and, thus, it's designed to work
+        only with single-column DataFrames.
+
         Parameters
         ----------
         caselist : list of tuples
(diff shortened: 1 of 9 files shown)


modin/core/storage_formats/pandas/query_compiler.py
```diff
@@ -4683,3 +4688,150 @@
...
+        ):
+            if isinstance(data, type(frame)):
+                if part_offset < len(data._partitions):
+                    return data._partitions[part_offset][0]._data
+                else:
+                    return [fill_value] * part_len
+
+            if isinstance(data, pandas.Series):
+                if trivial[0] is None:
+                    trivial[0] = is_trivial_index(frame.index)
```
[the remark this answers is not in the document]
YarShev: Can we rewrite this code to fix the alert?

modin/core/storage_formats/pandas/query_compiler.py
```diff
@@ -43,7 +43,8 @@
 from pandas.core.indexing import check_bool_indexer
 from pandas.errors import DataError, MergeError
 
-from modin.config import CpuCount, RangePartitioningGroupby
+import modin.pandas as pd
```
YarShev, line 46: Why do we need to import this module here? I guess we would need to filter arguments at the API layer and pass QueryCompilers to this layer.

modin/core/storage_formats/pandas/query_compiler.py
```diff
@@ -4683,3 +4688,150 @@ def compare(self, other, **kwargs):
                 other._modin_frame,
             )
         )
+
+    def case_when(self, caselist):
+        if Engine.get() in ("Unidist", "Dask"):
```
YarShev, line 4693: Does Dask also fail?

modin/core/storage_formats/pandas/query_compiler.py
```diff
@@ -4683,3 +4688,150 @@ def compare(self, other, **kwargs):
...
+        parts = [p[0] for p in frame._partitions]
+        parts_len = len(parts)
+        lengths = frame._get_lengths(parts, Axis.ROW_WISE)
+        new_caselist = []
+        is_trivial_idx = None
+        for c in caselist:
+            c = list(c)
+            for i in range(2):
+                data = c[i]
+                if not isinstance(data, pd.Series):
```
YarShev, line 4716: We shouldn't operate on the high-layer objects in the lower layers? See a comment above.

modin/core/storage_formats/pandas/query_compiler.py
```diff
@@ -4683,3 +4688,150 @@ def compare(self, other, **kwargs):
...
+            )
+            data_offset += part_len
+        new_parts = np.array([[p] for p in new_parts])
+        frame = frame.__constructor__(
+            new_parts, columns=frame.columns, index=frame.index
+        )
+        return self.__constructor__(frame)
+
+
+def _case_when(df, caselist):  # pragma: no cover
```
YarShev, line 4819: Usually we declare such wrappers inside concrete methods.

cc @dchigarev

modin/core/storage_formats/pandas/query_compiler.py
```diff
@@ -4683,3 +4688,150 @@ def compare(self, other, **kwargs):
...
+            part_len = lengths[i]
+            cases = [
+                tuple(
+                    map_data(i, part_len, d[0], data_offset, d[1])
+                    for d in zip(c, (True, None))
+                )
+                for c in new_caselist
+            ]
+            new_parts.append(
+                part.add_to_apply_calls(
```
YarShev, line 4804: QueryCompilers should not know about low level details. Why can't we use the Modin DataFrame functional?

modin/core/storage_formats/pandas/query_compiler.py
AndreyPavlenko: The arguments are not filtered - https://github.com/modin-project/modin/blob/9adaf3396cd88b1cc1ab25a72d4373d874cc6069/modin/pandas/series.py#L1018

modin/core/storage_formats/pandas/query_compiler.py
AndreyPavlenko: Yes.

modin/core/storage_formats/pandas/query_compiler.py
AndreyPavlenko: There is a function cache in RayWrapper.put(). Nested functions are not cached.

modin/core/storage_formats/pandas/query_compiler.py
AndreyPavlenko: This implementation rely on the partition internals. Should I move it to the DataFrame class?

modin/core/storage_formats/pandas/query_compiler.py
AndreyPavlenko: Should I add the filter?

modin/core/storage_formats/pandas/query_compiler.py
anmyachev: What error occurs? (For Dask)

dchigarev: I once had a problem where some tests were failing/hanging for Unidist and Dask specifically (not for Ray), when I accidentally passed a distributed object to a kernel, for example passing a low-level PandasDataframe, or an object that contains a PandasDataframe in it (LazyCategorical, ModinDtypes, ModinIndex). Could you double-check that it's not the case here?

I also had a similar problem when declaring kernel functions in a global score (like `_case_when()` in you RP), could you try moving the declaration inside the qc method and see if it helps?

modin/core/storage_formats/pandas/query_compiler.py
dchigarev: I'm just wondering, whether we can implement `case_when` using [`PandasDataframe.n_ary_op()`](https://github.com/modin-project/modin/blob/cee9b9819803e029c31961b4cf2c913b9f00b962/modin/core/dataframe/pandas/dataframe/dataframe.py#L3665)? It seems like it does the same thing: copartition + a send kernel with broadcasting other partitions

modin/core/storage_formats/pandas/query_compiler.py
YarShev: Yes, the API layer should filter arguments to pass those in the QC.

modin/core/storage_formats/pandas/query_compiler.py
```diff
@@ -4683,3 +4688,150 @@ def compare(self, other, **kwargs):
...
+            return super().case_when(caselist)
+
+        frame = self._modin_frame
+        # If all the conditions are callable and the replacements are either
+        # callable or scalar, use map().
+        if all(
+            callable(c[0]) and (callable(c[1]) or not is_list_like(c[1]))
+            for c in caselist
+        ):
+            frame = frame.map(func=_case_when, func_args=[caselist], lazy=True)
```
YarShev, line 4704: Why do we want this to be lazy? Is it more beneficial?

modin/core/storage_formats/pandas/query_compiler.py
AndreyPavlenko: [PandasDataframe.n_ary_op()](https://github.com/modin-project/modin/blob/cee9b9819803e029c31961b4cf2c913b9f00b962/modin/core/dataframe/pandas/dataframe/dataframe.py#L3665) operates on partitions, but the caselist may contain not only data frames but also callables and scalars. Can it handle this case?

YarShev: I don't think it can handle this case in the state it is now. We should either add some logic to this operator or add a new operator, which could handle this case.

AndreyPavlenko: Would it  be reasonable to move the implementation of case_when() to PandasOnRayDataframe? Currently it works only for the Ray backend.

YarShev: Ok to me.

YarShev: In future we could try to pull some general stuff out of it and put in a new core operator, which could be more generic but not only for case_when.

anmyachev: > Would it be reasonable to move the implementation of case_when() to PandasOnRayDataframe? Currently it works only for the Ray backend.

Here is a suggestion on how to fix the code for other engines: https://github.com/modin-project/modin/pull/6972#discussion_r1508734112. If we don’t have strong reasons, then it’s better to implement it for all (main) engines at once.

AndreyPavlenko: Another reason of moving to the data frame class is that the current implementation uses the data frame internals.

modin/core/storage_formats/pandas/query_compiler.py
AndreyPavlenko: Fixed for Unidist and Dask. It was required to do `execution_wrapper.put()` for it each callable argument. Moreover, for Dask it's required for each partition, otherwise CancelledError is raised.

modin/core/storage_formats/pandas/query_compiler.py
AndreyPavlenko: Added the filtering, but when defaulting to pandas the arguments are not properly converted to Series. Should be fixed in a separate PR.

modin/core/dataframe/pandas/partitioning/partition_manager.py
```diff
@@ -657,7 +657,11 @@ def lazy_map_partitions(
         NumPy array
             An array of partitions
         """
-        preprocessed_map_func = cls.preprocess_func(map_func)
+        preprocessed_map_func = (
+            map_func
+            if cls._execution_wrapper.is_future(map_func)
```
YarShev, line 662: When do we pass map_func as a future in this method?

modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -55,6 +53,9 @@
     )
     from pandas._typing import npt
 
+import numpy as np
+import pandas
```
YarShev, line 57: These should be moved above the modin imports.

modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -4629,3 +4637,166 @@ def from_dataframe(cls, df: "ProtocolDataframe") -> "PandasDataframe":
...
+        -------
+        PandasDataframe
+        """
+        cls = type(self)
+        wrapper_put = self._partition_mgr_cls._execution_wrapper.put
+        if (remote_fn := getattr(cls, "_CASE_WHEN_FN", None)) is None:
+
+            def case_when(df, caselist):  # pragma: no cover
+                caselist = [
+                    tuple(
```
YarShev, line 4659: Why do we have to wrap this in a tuple?

modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -4629,3 +4637,166 @@ def from_dataframe(cls, df: "ProtocolDataframe") -> "PandasDataframe":
...
+            elif use_map and is_list_like(replacement):
+                use_map = False
+                if isinstance(replacement, cls):
+                    replacement = copartition(replacement, None)
+            new_caselist.append((condition, replacement))
+
+        # If all the conditions are callable and the replacements are either
+        # callable or scalar, use map().
+        if use_map:
+            return self.map(func=remote_fn, func_args=[new_caselist], lazy=True)
```
YarShev, line 4738: Why is it beneficial to use lazy=True? Are there any measurements?

modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -4629,3 +4637,166 @@ def from_dataframe(cls, df: "ProtocolDataframe") -> "PandasDataframe":
...
+                df = self.__constructor__(
+                    list_of_right_parts[0],
+                    joined_index,
+                    df.columns,
+                    row_lengths,
+                    df.column_widths,
+                )
+            return df
+
+        # For Dask the callebles are wrapped for each partition in the map_data() function.
```
YarShev, line 4713:
```suggestion
        # For Dask the callables are wrapped for each partition in the map_data() function.
```

modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -4629,3 +4637,166 @@ def from_dataframe(cls, df: "ProtocolDataframe") -> "PandasDataframe":
...
+            lengths = self._get_lengths(parts, Axis.ROW_WISE)
+        new_parts = []
+        data_offset = 0
+        for i in range(0, parts_len):
+            part = parts[i]
+            part_len = lengths[i]
+            cases = [
+                tuple(
+                    map_data(i, part_len, d[0], data_offset, d[1])
+                    for d in zip(c, (True, None))
```
YarShev, line 4788: What is the purpose of None?

modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -4629,3 +4637,166 @@ def from_dataframe(cls, df: "ProtocolDataframe") -> "PandasDataframe":
...
+                    diff = part_len - len(data)
+                    if diff > 0:
+                        data = pandas.concat((data, pandas.Series([fill_value] * diff)))
+                    return data
+                return data.reindex(
+                    self.index[data_offset : data_offset + part_len],
+                    fill_value=fill_value,
+                )
+
+            if not wrap_callable and callable(data):
```
YarShev, line 4768: Let's put a comment why we put data every time.

modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -4629,3 +4637,166 @@ def from_dataframe(cls, df: "ProtocolDataframe") -> "PandasDataframe":
...
+                part.add_to_apply_calls(
+                    remote_fn,
+                    cases,
+                    length=part_len,
+                    width=1,
+                )
+            )
+            data_offset += part_len
+        new_parts = np.array([[p] for p in new_parts])
+        return self.__constructor__(new_parts, columns=self.columns, index=self.index)
```
YarShev, line 4802: Can we know in advance row_lengths and column_widths in this case?

modin/core/storage_formats/pandas/query_compiler.py
YarShev: Do you want to first fix this case in a separate PR and only after that complete this PR?

modin/core/storage_formats/pandas/query_compiler.py
AndreyPavlenko: There is no difference in case of a single call, but in case of multiple calls or if another lazy operation is performed on the result, it's faster because it allows to perform a batch of operations on each partition in parallel.

modin/core/dataframe/pandas/dataframe/dataframe.py
AndreyPavlenko: According to the doc, caselist is a list of tuples.

modin/core/dataframe/pandas/dataframe/dataframe.py
AndreyPavlenko: Answered here - https://github.com/modin-project/modin/pull/6972/#discussion_r1514478027

modin/core/dataframe/pandas/dataframe/dataframe.py
AndreyPavlenko: These are the fill values - True fro condition and None for replacement.

modin/core/dataframe/pandas/dataframe/dataframe.py
AndreyPavlenko: Added.

modin/core/dataframe/pandas/partitioning/partition_manager.py
AndreyPavlenko: Here https://github.com/modin-project/modin/pull/6972/files#diff-718e1f98ecbed40356292a477fc27bafb81e62c796491c81e398d5b397d6d69eR4668 the remote function is put to the wrapper. After tat, an object reference is passed to this function.

modin/core/storage_formats/pandas/query_compiler.py
AndreyPavlenko: I tried to fix in this PR by adding flag to the Series qc and then squeeze when converting to Pandas, but got a lot of test failures. Seems, the fix is a bit more complicated. In this PR I've added a workaround by converting the arguments, when defaulting to pandas, explicitly. We could live it as is and then remove then a proper conversion mechanism is implemented.

modin/pandas/test/test_series.py
```diff
@@ -4518,18 +4520,63 @@ def test_struct_general():
...
+    nparts = NPartitions.get()
+    part_size = MinPartitionSize.get()
+    new_nparts = max(1, min(math.ceil(len(base) / part_size), part_size)) + 1
+    NPartitions.put(new_nparts)
+    MinPartitionSize.put(math.ceil(len(base) / new_nparts))
+    base_repart = pd.Series(base)
+    NPartitions.put(nparts)
+    MinPartitionSize.put(part_size)
+    for df in (pd.Series(base), base_repart):
+        df_equals(pd_result, df.case_when(caselist))
```
YarShev, line 4570: Can we use eval_general instead of df_equals?

modin/core/storage_formats/base/query_compiler.py
```diff
@@ -6706,6 +6706,14 @@ def case_when(self, caselist):  # noqa: PR01, RT01, D200
         """
         Replace values where the conditions are True.
         """
+        qc_type = type(self)
+        caselist = [
+            tuple(
+                d.to_pandas().squeeze(axis=1) if isinstance(d, qc_type) else d
+                for d in c
```
YarShev, line 6713: Let's aling the namings for c and d everywhere.

YarShev (commented):
Overall, LGTM! Left some minor comments.

@dchigarev, any comments?

modin/pandas/test/test_series.py
AndreyPavlenko: I don't think it makes sense because the pandas result is calculated only once here and all the modin results are compared with it.

YarShev (dismissed):

modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -4629,3 +4636,179 @@ def from_dataframe(cls, df: "ProtocolDataframe") -> "PandasDataframe":
...
+        use_map = single_wrap
+        new_caselist = []
+        for condition, replacement in caselist:
+            if callable(condition):
+                if single_wrap:
+                    condition = wrapper_put(condition)
+            else:
+                use_map = False
+                if isinstance(condition, cls):
+                    condition = copartition(condition, True)
```
dchigarev, line 4730: btw, `._copartition()` support copartitioning along several dataframes at the same time, would it be more efficient to first filter out all modin dataframes and then call `._copartition()` only once?

modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -4629,3 +4636,179 @@ def from_dataframe(cls, df: "ProtocolDataframe") -> "PandasDataframe":
...
+            if callable(replacement):
+                if single_wrap:
+                    replacement = wrapper_put(replacement)
+            elif use_map and is_list_like(replacement):
+                use_map = False
+                if isinstance(replacement, cls):
+                    replacement = copartition(replacement, None)
+            new_caselist.append((condition, replacement))
```
dchigarev, line 4738: to avoid code duplication
```suggestion
        for conditions in caselist:
            new_conds = []
            for fill_value, cond in zip([True, None], conditions):
                if callable(cond):
                    if single_wrap:
                        cond = wrapper_put(cond)
                else:
                    use_map = False
                    if isinstance(cond, cls):
                        cond = copartition(cond, fill_value)
                new_conds.append(cond)
            new_caselist.append(tuple(new_conds))
```

modin/core/dataframe/pandas/partitioning/partition_manager.py
anmyachev: We should update docstring.

modin/core/storage_formats/pandas/query_compiler.py
```diff
@@ -4529,3 +4529,16 @@ def compare(self, other, **kwargs):
...
+            qc_type = type(self)
+            caselist = [
+                tuple(
+                    data._modin_frame if isinstance(data, qc_type) else data
+                    for data in case_tuple
+                )
+                for case_tuple in caselist
+            ]
+            return self.__constructor__(impl(caselist))
+        return super().case_when(caselist)
```
anmyachev, line 4544: Looks like an unreachable branch.

modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -4629,3 +4636,179 @@ def from_dataframe(cls, df: "ProtocolDataframe") -> "PandasDataframe":
...
+            lengths = self._get_lengths(parts, Axis.ROW_WISE)
+        new_parts = []
+        data_offset = 0
+        for i in range(0, parts_len):
+            part = parts[i]
+            part_len = lengths[i]
+            cases = [
+                tuple(
+                    map_data(i, part_len, d[0], data_offset, d[1])
+                    for d in zip(c, (True, None))
```
dchigarev, line 4794:
```suggestion
                    map_data(i, part_len, data, data_offset, fill_value)
                    for data, fill_value in zip(c, (True, None))
```

modin/pandas/test/test_series.py
```diff
@@ -4518,18 +4520,67 @@ def test_struct_general():
...
+    "caselist",
+    _case_when_caselists(),
+)
+def test_case_when(base, caselist):
+    pd_result = base.case_when(caselist)
+    nparts = NPartitions.get()
+    part_size = MinPartitionSize.get()
+    new_nparts = max(1, min(math.ceil(len(base) / part_size), part_size)) + 1
+    NPartitions.put(new_nparts)
+    MinPartitionSize.put(math.ceil(len(base) / new_nparts))
```
anmyachev, line 4565: Let's use `set_min_partition_size` and `set_num_partitions` fixtures.

modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -4629,3 +4636,179 @@ def from_dataframe(cls, df: "ProtocolDataframe") -> "PandasDataframe":
...
+            part_len,
+            data,
+            data_offset,
+            fill_value,
+        ):
+            if isinstance(data, cls):
+                if part_offset < len(data._partitions):
+                    return data._partitions[part_offset][0]._data
+                else:
+                    return [fill_value] * part_len
```
dchigarev, line 4756: Does this ever happen? I though `._copartition()` should have aligned partitioning

modin/pandas/test/test_series.py
anmyachev: However it's better to choose https://github.com/modin-project/modin/pull/6972/files#r1515897730

modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -4629,3 +4636,179 @@ def from_dataframe(cls, df: "ProtocolDataframe") -> "PandasDataframe":
...
+                nonlocal is_trivial_idx
+                if is_trivial_idx is None:
+                    is_trivial_idx = is_trivial_index(self.index)
+                if is_trivial_idx and is_trivial_index(data.index):
+                    data = data[data_offset : data_offset + part_len]
+                    diff = part_len - len(data)
+                    if diff > 0:
+                        data = pandas.concat((data, pandas.Series([fill_value] * diff)))
+                    return data
+                return data.reindex(
```
dchigarev, line 4768: can we perform the reindexing only once at the copartitioning stage, so here we would only need to slice the data?

modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -4629,3 +4636,179 @@ def from_dataframe(cls, df: "ProtocolDataframe") -> "PandasDataframe":
...
+            data_offset,
+            fill_value,
+        ):
+            if isinstance(data, cls):
+                if part_offset < len(data._partitions):
+                    return data._partitions[part_offset][0]._data
+                else:
+                    return [fill_value] * part_len
+
+            if isinstance(data, pandas.Series):
```
dchigarev, line 4758: do we need support for non-distributed pandas objects?

modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -4629,3 +4636,179 @@ def from_dataframe(cls, df: "ProtocolDataframe") -> "PandasDataframe":
...
+                        data.iloc[:, 0] if isinstance(data, pandas.DataFrame) else data
+                        for data in case_tuple
+                    )
+                    for case_tuple in caselist
+                ]
+                series = df.iloc[:, 0]
+                return pandas.DataFrame({series.name: series.case_when(caselist)})
+
+            if single_wrap:
+                cls._CASE_WHEN_FN = remote_fn = wrapper_put(case_when)
```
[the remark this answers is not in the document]
AndreyPavlenko: There is no such cache for other backends.

modin/core/storage_formats/base/query_compiler.py
```diff
@@ -6706,6 +6706,14 @@ def case_when(self, caselist):  # noqa: PR01, RT01, D200
         """
         Replace values where the conditions are True.
         """
+        qc_type = type(self)
+        caselist = [
+            tuple(
+                data.to_pandas().squeeze(axis=1) if isinstance(data, qc_type) else data
```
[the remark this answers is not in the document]
AndreyPavlenko: We need a proper mechanism for qc to Series conversion in to_pandas().

modin/pandas/test/test_series.py
```diff
@@ -4518,18 +4520,67 @@ def test_struct_general():
...
+    base_repart = pd.Series(base)
+    NPartitions.put(nparts)
+    MinPartitionSize.put(part_size)
```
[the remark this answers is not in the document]
AndreyPavlenko: It doesn't work for Series:

```python traceback
modin/core/dataframe/pandas/partitioning/partition_manager.py", line 1043, in <lambda>
E                           index_func = lambda df: df.axes[axis]  # noqa: E731
```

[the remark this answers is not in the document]
dchigarev: ?
```python
construct_modin_df_by_scheme(base.to_frame(), ...).squeeze(axis=1)
```

modin/core/dataframe/pandas/dataframe/dataframe.py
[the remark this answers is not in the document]
dchigarev: maybe it's worth introducing a common mechanism in a separate PR then instead of manually writing it for every method?

modin/core/storage_formats/base/query_compiler.py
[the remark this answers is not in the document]
anmyachev: Could you create an issue for this and add it into this code?

modin/core/dataframe/pandas/dataframe/dataframe.py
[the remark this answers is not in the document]
dchigarev: > but you mentioned that there are problems with global functions in Unidist. Will that work?

not sure about that

modin/core/storage_formats/base/query_compiler.py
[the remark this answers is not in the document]
AndreyPavlenko: https://github.com/modin-project/modin/issues/7041

modin/pandas/test/test_series.py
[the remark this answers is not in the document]
AndreyPavlenko: It works, but Unidist and HDK tests fail with NotImplementedError - https://github.com/modin-project/modin/actions/runs/8194577509/job/22410966951?pr=6972. On Ray it works, but the test runs ~20% longer.

modin/pandas/test/test_series.py
```diff
@@ -4510,18 +4512,70 @@ def test_struct_general():
...
+        pandas.Series(range(40)),
+        pandas.Series([0, 7, 8, 9] * 10, name="c", index=range(0, 80, 2)),
+    ],
+)
+@pytest.mark.parametrize(
+    "caselist",
+    _case_when_caselists(),
+)
+def test_case_when(base, caselist):
+    pd_result = base.case_when(caselist)
```
YarShev, line 4552:
```suggestion
    pandas_result = base.case_when(caselist)
```
It is easy to misunderstand pd with modin/pandas. Let's be explicit.

modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -4629,3 +4636,175 @@ def from_dataframe(cls, df: "ProtocolDataframe") -> "PandasDataframe":
...
+
+        Parameters
+        ----------
+        caselist : list of tuples
+
+        Returns
+        -------
+        PandasDataframe
+        """
+        # For Dask the callables must wrapped for each partition, otherwise
```
YarShev, line 4652:
```suggestion
        # For Dask the callables must be wrapped for each partition, otherwise
```

modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -4629,3 +4636,175 @@ def from_dataframe(cls, df: "ProtocolDataframe") -> "PandasDataframe":
...
+
+            def case_when(df, name, caselist):  # pragma: no cover
+                caselist = [
+                    tuple(
+                        data.squeeze(1) if isinstance(data, pandas.DataFrame) else data
+                        for data in case_tuple
+                    )
+                    for case_tuple in caselist
+                ]
+                return pandas.DataFrame({name: df.squeeze(1).case_when(caselist)})
```
YarShev, line 4670:
```suggestion
                return pandas.DataFrame({name: df.squeeze(axis=1).case_when(caselist)})
```

modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -4629,3 +4636,175 @@ def from_dataframe(cls, df: "ProtocolDataframe") -> "PandasDataframe":
...
+            for modin_list, fill_value in zip(modin_lists, fill_values):
+                _, list_of_right_parts, joined_index, row_lengths = self._copartition(
+                    Axis.ROW_WISE.value,
+                    modin_list,
+                    how="left",
+                    sort=False,
+                    fill_value=fill_value,
+                )
+                modin_list.clear()
+                modin_list.extend(
```
YarShev, line 4729: Is it okay if we modify the list here? Will `modin_lists` be affected?

AndreyPavlenko: Yes, this list is used to collect modin frames and then replace them with copartitioned.

YarShev: I mean won't this be a problem for `modin_lists` further or it is done intentionally?

modin/pandas/test/test_series.py
YarShev: ping

modin/core/dataframe/pandas/dataframe/dataframe.py
YarShev: ping

modin/core/dataframe/pandas/dataframe/dataframe.py
YarShev: ping

modin/core/dataframe/pandas/dataframe/dataframe.py
AndreyPavlenko: Intentionally. This is a temporary list for collecting the modin frames.

modin/core/dataframe/pandas/dataframe/dataframe.py
AndreyPavlenko: I think, all the possible cases are covered here.

YarShev (approved):

modin/pandas/test/test_series.py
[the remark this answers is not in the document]
anmyachev: It should be quite easy to add `construct_modin_series_by_scheme` (by analogue). Then there will be no slowdown and the code will be more reliable (there is no way to change variables and not return the previous values).

modin/core/dataframe/pandas/dataframe/dataframe.py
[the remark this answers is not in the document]
anmyachev: Maybe implement `_func_cache` only for Dask for now to fix the issue?

modin/pandas/test/test_series.py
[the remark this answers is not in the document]
AndreyPavlenko: It should be moved to some utility class, because when it's used from test_internals, NotImplementedError is raised for Unidist and HDK.
I don't know why, but the test runs slower. Perhaps, because the constructed from partitions frame does not have the lengths info.

modin/core/dataframe/pandas/dataframe/dataframe.py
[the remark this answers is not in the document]
AndreyPavlenko: Seems, for Dask it does not work. If the function is cached, the tests randomly fail with CancelledError.

modin/pandas/test/test_series.py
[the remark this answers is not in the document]
anmyachev: I see. Let's use fixtures: https://github.com/modin-project/modin/blob/b1501d8dd838d36842d7eeccd07a2fad9073daf1/modin/conftest.py#L462 and https://github.com/modin-project/modin/blob/b1501d8dd838d36842d7eeccd07a2fad9073daf1/modin/conftest.py#L486 in this case.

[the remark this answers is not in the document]
AndreyPavlenko: But, in this case, all the frames will be identically partitioned. The idea is to test cases when the base and the arguments have different partitioning.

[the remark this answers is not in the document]
dchigarev: `NotImplementedError` is raised because the utility function being imported from `modin.test.storage_formats.pandas.test_internals.py` that doesn't support testing nor with Unidist nor with HDK:
https://github.com/modin-project/modin/blob/b1501d8dd838d36842d7eeccd07a2fad9073daf1/modin/test/storage_formats/pandas/test_internals.py#L86-L89
The utility function should work just fine with Unidist if imported from any other place (for HDK we could just skip the test, or make the `construct_by_scheme_function` simply return `pd.DataFrame(data)` since HDK doesn't support modin's partitioning)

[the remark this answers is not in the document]
AndreyPavlenko: Added minor optimizations to construct_modin_df_by_scheme(). Now it works for Unidist and for HDK it's not used.

modin/pandas/test/test_series.py
```diff
@@ -85,6 +86,11 @@
     test_string_list_data_values,
 )
 
+if StorageFormat.get() != "Hdk":
+    from ...test.storage_formats.pandas.test_internals import (
```
YarShev, line 90: Can we use the absolute path instead of the relative one?

modin/core/storage_formats/pandas/query_compiler.py
```diff
@@ -4529,3 +4529,14 @@ def compare(self, other, **kwargs):
                 other._modin_frame,
             )
         )
+
+    def case_when(self, caselist):
+        qc_type = BaseQueryCompiler
```
YarShev, line 4534: Why was this changed? Why doesn't `qc_type = type(self)` work?

modin/core/storage_formats/base/query_compiler.py
```diff
@@ -6706,6 +6706,15 @@ def case_when(self, caselist):  # noqa: PR01, RT01, D200
         """
         Replace values where the conditions are True.
         """
+        # A workaround for https://github.com/modin-project/modin/issues/7041
+        qc_type = BaseQueryCompiler
```
YarShev, line 6710: same

YarShev (approved):

modin/core/dataframe/pandas/dataframe/dataframe.py
anmyachev: We could handle this at a higher level.

AndreyPavlenko:
> I don't understand how this differs from how you cache in this pull request. Could you provide more details?

The function reference is saved in the class field. In case of Dask it's not cached.

modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -4629,3 +4636,179 @@ def from_dataframe(cls, df: "ProtocolDataframe") -> "PandasDataframe":
...
+            new_case = []
+            for data, modin_list, fill_value in zip(
+                case_tuple, modin_lists, fill_values
+            ):
+                if isinstance(data, cls):
+                    modin_list.append(data)
+                elif callable(data):
+                    if single_wrap:
+                        data = wrapper_put(data)
+                elif isinstance(data, pandas.Series):
```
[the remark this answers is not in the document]
AndreyPavlenko: I'm not sure if it's efficient. In that case we will need to put all the objects into the remote storage, replace them with references and then pass the references to the remote functions. It may result in additional serialization, mere over, these objects are required for a single call only and will not be used anymore.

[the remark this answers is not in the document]
dchigarev: > Wouldn't it be easier to convert to Modin objects at the top level?

I think Anatoly meant, wouldn't it be easier for us? :)

I don't think that we support non-distributed pandas object in any other modin's method. Most of the time we have an explicit check at the high level [verifying that we the function received a <i>modin</i> dataframe](https://github.com/modin-project/modin/blob/7c835a2761ede41d402c1febd29826c1d0b9512f/modin/pandas/general.py#L236-L239).

It's okay for this method to support pandas objects (since the time was already spent on it), but in the future, we shouldn't spend time on supporting such cases.

dchigarev:
> > I don't understand how this differs from how you cache in this pull request. Could you provide more details?
> 
> The function reference is saved in the class field. In case of Dask it's not cached.

I still don't like the idea of a method being responsible to cache itself. We should either have a common mechanism that does this automatically or not to have this logic at all. Does the caching gives any significant improvement? Can we remove it?

modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -4629,3 +4636,179 @@ def from_dataframe(cls, df: "ProtocolDataframe") -> "PandasDataframe":
...
+                    if is_trivial_idx is None:
+                        self_idx = self.index
+                        length = len(self_idx)
+                        is_trivial_idx = is_trivial_index(self_idx)
+                    if is_trivial_idx and is_trivial_index(data.index):
+                        data = data[:length]
+                        diff = length - len(data)
+                        if diff > 0:
+                            data = pandas.concat(
+                                [data, pandas.Series([fill_value] * diff)]
```
dchigarev, line 4710:
```suggestion
                                [data, pandas.Series([fill_value] * diff)], ignore_index=True
```
we probably need to pass `ignore_index=True` here, otherwise the output index would be incorrect:
```python
>>> data
0    1
1    2
2    3
3    4
dtype: int64
>>> pandas.concat([data, pandas.Series([0] * 4)])
0    1
1    2
2    3
3    4
0    0
1    0
2    0
3    0
dtype: int64
```

modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -4629,3 +4636,179 @@ def from_dataframe(cls, df: "ProtocolDataframe") -> "PandasDataframe":
...
+            if not single_wrap and callable(data):
+                return wrapper_put(data)
+            return (
+                data[data_offset : data_offset + part_len]
+                if is_list_like(data)
+                else data
+            )
+
+        parts = [p[0] for p in self._partitions]
+        lengths = self._get_lengths(parts, Axis.ROW_WISE)
```
dchigarev, line 4783: Why not like this? There could be already computed cache for them
```suggestion
        lengths = self.row_lengths
```

modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -4629,3 +4636,179 @@ def from_dataframe(cls, df: "ProtocolDataframe") -> "PandasDataframe":
...
+
+        Parameters
+        ----------
+        caselist : list of tuples
+
+        Returns
+        -------
+        PandasDataframe
+        """
+        # For Dask the callables must be wrapped for each partition, otherwise
```
dchigarev, line 4652: let's add this assertion at the beginning, and also mentioned in the doc-string that this method only works for a 1-column dataframe
```suggestion
        assert self._partitions.shape[1] == 1
        # For Dask the callables must be wrapped for each partition, otherwise
```

modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -4629,3 +4636,179 @@ def from_dataframe(cls, df: "ProtocolDataframe") -> "PandasDataframe":
...
+            )
+
+        parts = [p[0] for p in self._partitions]
+        lengths = self._get_lengths(parts, Axis.ROW_WISE)
+        new_parts = []
+        data_offset = 0
+
+        # Split the data and apply the remote function to each partition
+        # with the corresponding chunk of data
+        for i, part, part_len in zip(range(0, len(parts)), parts, lengths):
```
dchigarev, line 4789:
```suggestion
        for i, part, part_len in zip(range(len(parts)), parts, lengths):
```

modin/pandas/test/test_series.py
```diff
@@ -4510,18 +4516,75 @@ def test_struct_general():
...
+    "caselist",
+    _case_when_caselists(),
+)
+def test_case_when(base, caselist):
+    pandas_result = base.case_when(caselist)
+    modin_bases = [pd.Series(base)]
+
+    # 'base' and serieses from 'caselist' must have equal lengths, however in this test we want
+    # to verify that 'case_when' works correctly even if partitioning of 'base' and 'caselist' isn't equal
+    if StorageFormat.get() != "Hdk":  # HDK always uses a single partition.
```
dchigarev, line 4561: we should better fix [this problem](https://github.com/modin-project/modin/pull/6972/files#r1521307926) at the test level
```suggestion
    if StorageFormat.get() != "Hdk" and StorageFormat.get() != "BaseOnPython":  # HDK and BaseOnPython always use a single partition.
```

modin/core/storage_formats/pandas/query_compiler.py
dchigarev: this should be [fixed in tests](https://github.com/modin-project/modin/pull/6972/files#r1531803079) then, not in QC

AndreyPavlenko:
> I still don't like the idea of a method being responsible to cache itself. We should either have a common mechanism that does this automatically or not to have this logic at all. Does the caching gives any significant improvement? Can we remove it?

I've not benchmarked, but it should be more efficient, because, if not cached, the same function is serialized multiple times for each partition on each call.

dchigarev:
> but it should be more efficient

There's no doubt that it's more efficient, but my point is that, in reality, nobody runs `.case_when()` thousand times in a row. Even if one does, they would get ~100-200ms saving because of the caching. This may worth it if implemented properly (like automatical caching in `RayWraper._funcs_cache`), but I personally can't justify the current approach with a method manually caching itself in a dataframe's attribute.

> Seems, for Dask it does not work. If the function is cached, the tests randomly fail with CancelledError.

This should actually be considered as a huge red flag, pointing out that there could be other unexpected side-effects of such implementation of caching with other engines.

> if not cached, the same function is serialized multiple times for each partition

Let's call `PartitionMaganer.preprocess_func()` in the beginning of `PandasDataframe.case_when()`, this is a common way of how other dataframe's method 'cache' a kernel so each partition would reuse it.

YarShev:
@AndreyPavlenko, what else is required to be done in this PR? Please also resolve conflicts.

modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -4629,3 +4636,179 @@ def from_dataframe(cls, df: "ProtocolDataframe") -> "PandasDataframe":
...
+        ----------
+        caselist : list of tuples
+
+        Returns
+        -------
+        PandasDataframe
+        """
+        # For Dask the callables must be wrapped for each partition, otherwise
+        # the execution could fail with CancelledError.
+        single_wrap = Engine.get() != "Dask"
```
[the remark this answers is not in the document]
AndreyPavlenko: https://github.com/modin-project/modin/issues/7148

YarShev (approved):

modin/core/storage_formats/pandas/query_compiler.py
```diff
@@ -4500,3 +4500,14 @@ def compare(self, other, **kwargs):
...
+    def case_when(self, caselist):
+        qc_type = type(self)
+        caselist = [
+            tuple(
+                data._modin_frame if isinstance(data, qc_type) else data
+                for data in case_tuple
+            )
+            for case_tuple in caselist
+        ]
+        return self.__constructor__(self._modin_frame.case_when(caselist))
```
anmyachev, line 4513: @YarShev ?
```suggestion
        return self.__constructor__(self._modin_frame.case_when(caselist), shape_hint=self._shape_hint)
```

anmyachev (dismissed):

YarShev (approved):

anmyachev (approved):

modin/core/dataframe/pandas/dataframe/dataframe.py
```diff
@@ -4629,3 +4636,179 @@ def from_dataframe(cls, df: "ProtocolDataframe") -> "PandasDataframe":
         ErrorMessage.default_to_pandas(message="`from_dataframe`")
         pandas_df = from_dataframe_to_pandas(df)
         return cls.from_pandas(pandas_df)
+
+    def case_when(self, caselist):
```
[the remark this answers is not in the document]
AndreyPavlenko: Now it's a bit shorter, I'm not sure if it's still required to split it into multiple functions.

PR #7022, opened 2024-03-07
FEAT-#7021: Implement to/from_dask functions

### LINKED ISSUE #7021: Implement to/from_dask_dataframe functions

Retribution98:
## What do these changes do?

- [x] Resolves #7021 
- [x] tests added and passing

### DIFF
--- a/docs/ecosystem.rst
+++ b/docs/ecosystem.rst
@@ -57,17 +57,17 @@ where Ray Dataset can be used and what libraries it powers.
 
     ray_dataset = to_ray_dataset(modin_df)
 
-to_dask_dataframe
------------------
+to_dask
+-------
 
 You can refer to `Dask DataFrame`_ page to get more details on
 where Dask DataFrame can be used and what libraries it powers.
 
 .. code-block:: python
 
-    from modin.pandas.io import to_dask_dataframe
+    from modin.pandas.io import to_dask
 
-    dask_df = to_dask_dataframe(modin_df)
+    dask_df = to_dask(modin_df)
 
 .. _pandas ecosystem: https://pandas.pydata.org/community/ecosystem.html
 .. _NumPy ecosystem: https://numpy.org
--- a/docs/ecosystem.rst
+++ b/docs/ecosystem.rst
@@ -57,7 +57,20 @@ where Ray Dataset can be used and what libraries it powers.
 
     ray_dataset = to_ray_dataset(modin_df)
 
+to_dask_dataframe
+-----------------
+
+You can refer to `Dask DataFrame`_ page to get more details on
+where Dask DataFrame can be used and what libraries it powers.
+
+.. code-block:: python
+
+    from modin.pandas.io import to_dask_dataframe
+
+    dask_df = to_dask_dataframe(modin_df)
+
 .. _pandas ecosystem: https://pandas.pydata.org/community/ecosystem.html
 .. _NumPy ecosystem: https://numpy.org
 .. _Ray Data: https://docs.ray.io/en/latest/data/data.html
+.. _Dask DataFrame: https://docs.dask.org/en/stable/dataframe.html
 
--- a/modin/core/execution/dask/implementations/pandas_on_dask/io/io.py
+++ b/modin/core/execution/dask/implementations/pandas_on_dask/io/io.py
@@ -13,7 +13,6 @@
 
 """Module houses class that implements ``BaseIO`` using Dask as an execution engine."""
 
-import dask.dataframe as dd
 from distributed.client import default_client
 
 from modin.core.execution.dask.common import DaskWrapper
@@ -138,7 +137,7 @@ class PandasOnDaskIO(BaseIO):
     del __make_write  # to not pollute class namespace
 
     @classmethod
-    def from_dask_dataframe(cls, dask_obj):
+    def from_dask(cls, dask_obj):
         """
         Create a Modin `query_compiler` from a Dask Dataframe.
 
@@ -171,6 +170,8 @@ class PandasOnDaskIO(BaseIO):
         dask.dataframe.DataFrame or dask.dataframe.Series
             Converted object with type depending on input.
         """
+        from dask.dataframe import from_delayed
+
         partitions = unwrap_partitions(modin_obj, axis=0)
 
         # partiotions must be converted to pandas Series
@@ -185,4 +186,4 @@ class PandasOnDaskIO(BaseIO):
 
             partitions = [client.submit(df_to_series, part) for part in partitions]
 
-        return dd.from_delayed(partitions)
+        return from_delayed(partitions)
--- a/modin/core/execution/dask/implementations/pandas_on_dask/io/io.py
+++ b/modin/core/execution/dask/implementations/pandas_on_dask/io/io.py
@@ -61,7 +61,7 @@ from modin.experimental.core.storage_formats.pandas.parsers import (
     ExperimentalPandasXmlParser,
 )
 from modin.pandas.series import Series
-from modin.utils import MODIN_UNNAMED_SERIES_LABEL
+from modin.utils import MODIN_UNNAMED_SERIES_LABEL, import_optional_dependency
 
 
 class PandasOnDaskIO(BaseIO):
@@ -171,6 +171,9 @@ class PandasOnDaskIO(BaseIO):
         dask.dataframe.DataFrame or dask.dataframe.Series
             Converted object with type depending on input.
         """
+        import_optional_dependency(
+            "dask-expr", "dask-expr is required to create a Dask DataFrame."
+        )
         from dask.dataframe import from_delayed
 
         partitions = unwrap_partitions(modin_obj, axis=0)
--- a/modin/core/execution/dask/implementations/pandas_on_dask/io/io.py
+++ b/modin/core/execution/dask/implementations/pandas_on_dask/io/io.py
@@ -139,21 +139,21 @@ class PandasOnDaskIO(BaseIO):
     @classmethod
     def from_dask(cls, dask_obj):
         """
-        Create a Modin `query_compiler` from a Dask Dataframe.
+        Create a Modin `query_compiler` from a Dask DataFrame.
 
         Parameters
         ----------
         dask_obj : dask.dataframe.DataFrame
-            The Dask Dataframe to convert from.
+            The Dask DataFrame to convert from.
 
         Returns
         -------
         BaseQueryCompiler
-            QueryCompiler containing data from the Dask Dataframe.
+            QueryCompiler containing data from the Dask DataFrame.
         """
         client = default_client()
         dask_fututures = client.compute(dask_obj.to_delayed())
-        modin_df = from_partitions(dask_fututures, axis=0, columns=[])._query_compiler
+        modin_df = from_partitions(dask_fututures, axis=0)._query_compiler
         return modin_df
 
     @classmethod
--- a/modin/core/execution/dask/implementations/pandas_on_dask/io/io.py
+++ b/modin/core/execution/dask/implementations/pandas_on_dask/io/io.py
@@ -13,6 +13,9 @@
 
 """Module houses class that implements ``BaseIO`` using Dask as an execution engine."""
 
+import dask.dataframe as dd
+from distributed.client import default_client
+
 from modin.core.execution.dask.common import DaskWrapper
 from modin.core.execution.dask.implementations.pandas_on_dask.dataframe import (
     PandasOnDaskDataframe,
@@ -40,6 +43,10 @@ from modin.core.storage_formats.pandas.parsers import (
     PandasSQLParser,
 )
 from modin.core.storage_formats.pandas.query_compiler import PandasQueryCompiler
+from modin.distributed.dataframe.pandas.partitions import (
+    from_partitions,
+    unwrap_partitions,
+)
 from modin.experimental.core.io import (
     ExperimentalCSVGlobDispatcher,
     ExperimentalCustomTextDispatcher,
@@ -54,6 +61,8 @@ from modin.experimental.core.storage_formats.pandas.parsers import (
     ExperimentalPandasPickleParser,
     ExperimentalPandasXmlParser,
 )
+from modin.pandas.series import Series
+from modin.utils import MODIN_UNNAMED_SERIES_LABEL
 
 
 class PandasOnDaskIO(BaseIO):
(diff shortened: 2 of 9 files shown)


modin/core/io/io.py
```diff
@@ -140,6 +140,30 @@ def from_ray_dataset(cls, ray_obj):
             "Modin Dataframe can only be converted to a Ray Dataset if Modin uses a Ray engine."
         )
 
+    @classmethod
+    def from_dask_dataframe(cls, dask_obj):
```
anmyachev, line 144: Let's add a test that catch this exceptions (for example, when ray is using).

modin/core/execution/dask/implementations/pandas_on_dask/io/io.py
```diff
@@ -13,6 +13,9 @@
 
 """Module houses class that implements ``BaseIO`` using Dask as an execution engine."""
 
+import dask.dataframe as dd
+from distributed.client import default_client
```
anmyachev, line 17: Let's use `DaskWrapper` to interact with Dask API.

modin/pandas/test/test_io.py
```diff
@@ -90,6 +90,10 @@
 except ImportError:
     EXCEPTIONS = ()
 
+try:
+    import dask.dataframe as dd
+except ImportError:
+    pass
```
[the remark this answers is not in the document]
YarShev: I guess this is required for executions like hdk and unidist, which do not have dask installed in the env.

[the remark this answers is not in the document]
anmyachev: What for? There is a `pass` on `ImportError`.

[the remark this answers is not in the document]
YarShev: I see. Since we skipif tests for non Dask engines, we could remove this then.

modin/pandas/accessor.py
```diff
@@ -236,6 +236,21 @@ def to_ray_dataset(self):
         """
         return to_ray_dataset(self._data)
 
+    def to_dask_dataframe(self):
```
YarShev, line 239: It seems we should have both to_dask_dataframe and to_dask_series for a Modin df and series, respectively.

YarShev: Since there is a direct mapping dask series <-> modin series and dask df <-> modin df, I think it makes sense to just leave to_dask.

Retribution98: Dask is a flexible parallel computing library for analytics. If you want to process tabular data by parallelizing pandas, you need to use Dask DataFrame. Since you previously required the function to be called "to_ray_dataset", this function must be called "to_dask_dataframe"!

modin/pandas/test/test_io.py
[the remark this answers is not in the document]
Retribution98: I agree, we can only import it for `test_from_dask_dataframe`. If we implement other tests using Dask Dataframe, we will move it back to the top of the file.

modin/core/execution/dask/implementations/pandas_on_dask/io/io.py
Retribution98: Are you sure? As I see it, we use the public API without wrapping it in Wrapper, for both dask and ray.
Also `DaskWrapper` has annotation: "The class responsible for execution of remote operations.", but creating a Dask Dataframe is not valid for it.

Considering your wishes, I can suggest changing `import dask.dataframe as dd` to `from dask.dataframe import from_delayed`. What do you think about it?

anmyachev: > Are you sure? As I see it, we use the public API without wrapping it in Wrapper, for both dask and ray.
Also DaskWrapper has annotation: "The class responsible for execution of remote operations.", but creating a Dask Dataframe is not valid for it.

Yes, I'm sure. It's okay to change the description because that's what these classes were designed for. All places where the API is used directly need to be updated to work through a wrapper.

modin/pandas/accessor.py
anmyachev: > Dask is a flexible parallel computing library for analytics. If you want to process tabular data by parallelizing pandas, you need to use Dask DataFrame. Since you previously required the function to be called "to_ray_dataset", this function must be called "to_dask_dataframe"!

Ok for me, but we also have to add `to_dask_series` in that case.

modin/core/execution/dask/implementations/pandas_on_dask/io/io.py
Retribution98: Ok, I see reasons for this, so I just want to clarify. In this case, we just create the same method in `DaskWrapper` that proxies it to DaskDatafame, like this:
```python
@classmethod
def from_delayed(cls, partitions):
    dd.from_delayed(partitions)
```
Is correct?

Do I need to create an issue to refactor another public API calls?

modin/pandas/accessor.py
YarShev: Guys, I am now thinking that we should have named `to_ray_dataset` as just `to_ray`. Here I think we should name the function just `to_dask` and convert a Modin df to a dask df and a Modin series to a dask series. Just came to my mind that we overcomplicate things. `to_ray_dataset`, `to_ray_something_else`, `to_dask_df`, `to_dask_series`, `to_dask_array`, etc. We are adding a lot of stuff like that we have to support. Having just `to_ray` and `to_dask` seem sufficient since conversion to other ray objects or dask objects should be implemeneted by those libraries and they support that. E.g, dask_s.to_frame(), dask_df.to_dask_array(), etc. What do you think?

Retribution98: You moved these function to Modin API and it is inherited by both Dataframe and Series. How do you suggest to split them?
Also we have the different implementation for `to_pandas` function. In that case we have the union way to cast our DataFrame/Series to pandas DataFrame/Series respectively.

Retribution98: > Guys, I am now thinking that we should have named `to_ray_dataset` as just `to_ray`. Here I think we should name the function just `to_dask` and convert a Modin df to a dask df and a Modin series to a dask series. Just came to my mind that we overcomplicate things. `to_ray_dataset`, `to_ray_something_else`, `to_dask_df`, `to_dask_series`, `to_dask_array`, etc. We are adding a lot of stuff like that we have to support. Having just `to_ray` and `to_dask` seem sufficient since conversion to other ray objects or dask objects should be implemeneted by those libraries and they support that. E.g, dask_s.to_frame(), dask_df.to_dask_array(), etc. What do you think?

Yes, this is exactly what I wanted to explain before. I think this is a more suitable option. But we have already released the `to_ray_dataset` function. So I think now we will have more difficulties with renaming it (mark it as deprecated and remove later), if we choose this option.

anmyachev: > You moved these function to Modin API and it is inherited by both Dataframe and Series. How do you suggest to split them? Also we have the different implementation for `to_pandas` function. In that case we have the union way to cast our DataFrame/Series to pandas DataFrame/Series respectively.

I don't see a good way to do that, that's why I suggest to implement only `to_dask`.

anmyachev: BTW, I look at Ray dataset as a component of the framework, and not as a specific type. The problem that arose here is that the function name uses a specific type (DataFrame or Series). I still don't mind calling to_dask_[name of the component in Dask].

YarShev: @anmyachev, I would prefer to rename `to_ray_dataset` to `to_ray` by analogy with `to_dask`, `to_pandas`, etc. and deprecate `to_ray_dataset`. Why do you think we shouldn't rename `to_ray_dataset`?

YarShev: > I look at Ray dataset as a component of the framework, and not as a specific type

We can choose any type of the data format of a library and convert a Modin object to it. All other conversions to a third party library object should be (is) implemented by that library. No need for us to support that stuff.

Retribution98: > BTW, I look at Ray dataset as a component of the framework, and not as a specific type. The problem that arose here is that the function name uses a specific type (DataFrame or Series). I still don't mind calling to_dask_[name of the component in Dask].

The fact is that "dataset" is not component of the Ray, but "dataframe" is a component of the Dask.
In code it seems like this:
```python
import ray.data as rd
ray_df : rd.Dataset = rd.from_pandas(...)

import dask.dataframe as dd
dask_df : dd.DataFrame = dd.from_delayed(...)
```

anmyachev: I assume you see a one-to-one correspondence between Python modules and library components?

anmyachev: @YarShev @Retribution98 If you're on the same side on this issue, feel free to continue with this decision.

modin/core/execution/dask/implementations/pandas_on_dask/io/io.py
anmyachev: > Do I need to create an issue to refactor another public API calls?

It would be great.

> In this case, we just create the same method in DaskWrapper that proxies it to DaskDatafame, like this:

We discussed it offline, this can also be done separately.

modin/pandas/test/test_io.py
```diff
@@ -3295,3 +3294,56 @@ def test_from_ray_dataset():
...
+)
+@pytest.mark.filterwarnings(default_to_pandas_ignore_string)
+def test_series_to_dask():
+    index = pandas.DatetimeIndex(
+        pandas.date_range("2000", freq="h", periods=len(TEST_DATA["col1"]))
+    )
+
+    pandas_df = pandas.DataFrame(TEST_DATA, index=index)
+    pandas_s = pandas_df.iloc[0]
+    modin_s = pd.Series(pandas_s)
```
YarShev, line 3327: We can use create_test_series as for df test (create_test_dfs).

modin/pandas/test/test_io.py
```diff
@@ -3295,3 +3294,56 @@ def test_from_ray_dataset():
...
+    dask_series = modin_s.modin.to_dask()
+    pandas_s.equals(dask_series.compute())
+
+
+@pytest.mark.skipif(
+    condition=Engine.get() != "Dask",
+    reason="Dask Dataframe can only be converted to a Modin Dataframe if Modin uses a Dask engine.",
+)
+@pytest.mark.filterwarnings(default_to_pandas_ignore_string)
+def test_from_dask_dataframe():
```
YarShev, line 3338: We decided to leave to_dask and from_dask only, don't we?

modin/pandas/test/test_io.py
Retribution98: Thanks, I updated this test

modin/pandas/test/test_io.py
Retribution98: Sure, the both functions changed

docs/ecosystem.rst
```diff
@@ -57,7 +57,20 @@ where Ray Dataset can be used and what libraries it powers.
 
     ray_dataset = to_ray_dataset(modin_df)
 
+to_dask_dataframe
+-----------------
```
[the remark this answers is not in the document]
Retribution98: Thanks, it has already fixed.

docs/ecosystem.rst
```diff
@@ -57,7 +57,20 @@ where Ray Dataset can be used and what libraries it powers.
...
+
+You can refer to `Dask DataFrame`_ page to get more details on
+where Dask DataFrame can be used and what libraries it powers.
+
+.. code-block:: python
+
+    from modin.pandas.io import to_dask_dataframe
+
+    dask_df = to_dask_dataframe(modin_df)
```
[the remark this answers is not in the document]
Retribution98: Thanks, it has already fixed.

modin/core/execution/dask/implementations/pandas_on_dask/io/io.py
```diff
@@ -127,3 +135,55 @@ def __make_write(*classes, build_args=build_args):
...
+
+    @classmethod
+    def from_dask(cls, dask_obj):
+        """
+        Create a Modin `query_compiler` from a Dask Dataframe.
+
+        Parameters
+        ----------
+        dask_obj : dask.dataframe.DataFrame
+            The Dask Dataframet to convert from.
```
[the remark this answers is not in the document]
Retribution98: Thanks, it has already fixed.

modin/pandas/test/test_io.py
```diff
@@ -90,7 +91,6 @@
 except ImportError:
     EXCEPTIONS = ()
 
-
```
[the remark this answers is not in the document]
Retribution98: Done

modin/core/execution/dask/implementations/pandas_on_dask/io/io.py
```diff
@@ -127,3 +135,55 @@ def __make_write(*classes, build_args=build_args):
...
+
+        Returns
+        -------
+        BaseQueryCompiler
+            QueryCompiler containing data from the Dask Dataframe.
+        """
+        client = default_client()
+        dask_fututures = client.compute(dask_obj.to_delayed())
+        return from_partitions(dask_fututures, axis=0)._query_compiler
```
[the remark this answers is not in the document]
YarShev: Okay, I see. Let's leave as is for now to be in line with `from_pandas` behavior.

modin/pandas/io.py
```diff
@@ -1059,6 +1059,29 @@ def from_ray_dataset(ray_obj):
     )
 
 
+def from_dask(dask_obj):
+    """
+    Convert a Dask DataFrame into Modin DataFrame.
```
YarShev, line 1064:
```suggestion
    Convert a Dask DataFrame to a Modin DataFrame.
```

modin/pandas/test/test_io.py
```diff
@@ -3295,3 +3296,50 @@ def test_from_ray_dataset():
...
+def test_series_to_dask():
+    modin_s, pandas_s = create_test_series(TEST_DATA["col1"])
+
+    dask_series = modin_s.modin.to_dask()
+    df_equals(dask_series.compute(), pandas_s)
+
+
+@pytest.mark.skipif(
+    condition=Engine.get() != "Dask",
+    reason="Dask Dataframe can only be converted to a Modin Dataframe if Modin uses a Dask engine.",
```
YarShev, line 3331:
```suggestion
    reason="Dask DataFrame can only be converted to a Modin DataFrame if Modin uses a Dask engine.",
```
Fix spelling everywhere.

YarShev (commented):
Please also take a loot at CI, which is failing.

modin/core/execution/dask/implementations/pandas_on_dask/io/io.py
```diff
@@ -127,3 +135,56 @@ def __make_write(*classes, build_args=build_args):
...
+
+        Returns
+        -------
+        BaseQueryCompiler
+            QueryCompiler containing data from the Dask Dataframe.
+        """
+        client = default_client()
+        dask_fututures = client.compute(dask_obj.to_delayed())
+        modin_df = from_partitions(dask_fututures, axis=0, columns=[])._query_compiler
+        return modin_df
```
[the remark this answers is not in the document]
Retribution98: This change was pushed by mistake and it is cause of CI failing. It will revert soon.

modin/pandas/test/test_io.py
Retribution98: Okay, I has done it, but I think it's a waste of time, because this kind of writing goes through all the code and successfully applied in other PRs.

modin/pandas/io.py
Retribution98: Done

modin/core/execution/dask/implementations/pandas_on_dask/io/io.py
```diff
@@ -171,9 +171,6 @@ def to_dask(cls, modin_obj):
         dask.dataframe.DataFrame or dask.dataframe.Series
             Converted object with type depending on input.
         """
-        import_optional_dependency(
-            "dask-expr", "dask-expr is required to create a Dask DataFrame."
-        )
```
Retribution98: The default message from dask.dataframe is more user friendly, so I suggest not using this method.

Retribution98: image

Retribution98: I also think that we should not check the version for dusk_expr. Yes, it has problems in 1.0.1 version, but this is a problem in Dusk, not our compatibility with it. In addition, this library releases new versions very quickly and, in my opinion, users are unlikely to encounter this problem in the future.

YarShev (approved):

PR #7042, opened 2024-03-07
FIX-#6227: Make sure `Series.unique()` with pyarrow dtype returns `ArrowExtensionArray`

### LINKED ISSUE #6227: BUG: Series.unique with pyarrow dtype
### Modin version checks

### Reproducible Example

```python
ser = pd.Series([1, 0, pd.NA], dtype="uint8[pyarrow]")

>>> ser.unique()  # <- modin master
array([1, 0, <NA>], dtype=object)

>>> ser.unique()  # <- pandas main
<ArrowExtensionArray>
[1, 0, <NA>]
Length: 3, dtype: uint8[pyarrow]
```

### Issue Description

NA

### Expected Behavior

NA

### Error Logs

<details>

```python-traceback

```

</details>

### Installed Versions

<details>

</details>

anmyachev:
## What do these changes do?

- [x] Resolves #6227 
- [ ] tests added and passing


modin/core/storage_formats/pandas/query_compiler.py
```diff
@@ -1936,7 +1936,10 @@ def str_split(self, pat=None, n=-1, expand=False, regex=None):
     def unique(self, keep="first", ignore_index=True, subset=None):
```
anmyachev, line 1936: @dchigarev when [combining functions](https://github.com/modin-project/modin/pull/7091) `drop_duplicates` and `unique` into one, did you take into account that they return different data types? It seems to me that because of this, using `qc.unique` in its current form is inconvenient. From my side, the best solution seems to be to split this function into several. What do you think?

modin/core/storage_formats/pandas/query_compiler.py
```diff
@@ -1936,7 +1936,10 @@ def str_split(self, pat=None, n=-1, expand=False, regex=None):
...
         # kernels with 'pandas.Series.unique()' work faster
         can_use_unique_kernel = (
-            subset is None and ignore_index and len(self.columns) == 1 and keep
+            subset is None
+            and ignore_index
+            and len(self.columns) == 1
+            and keep is not False
```
anmyachev, line 1942: @dchigarev If we do not make this change, then the string is stored in the variable (for example, `"first"`). Did you do this on purpose?

modin/core/storage_formats/pandas/query_compiler.py
```diff
@@ -1956,13 +1959,10 @@ def unique(self, keep="first", ignore_index=True, subset=None):
...
+            return self.__constructor__(new_modin_frame, shape_hint=self._shape_hint)
         else:
-            new_modin_frame = self._modin_frame.apply_full_axis(
-                0,
-                lambda x: x.squeeze(axis=1).unique(),
-                new_columns=self.columns,
-            )
-        return self.__constructor__(new_modin_frame, shape_hint=self._shape_hint)
+            res = self.to_pandas().squeeze(axis=1)
+            return res.unique()
```
anmyachev, line 1965: @dchigarev I measured the new approach on small data (using your script from https://github.com/modin-project/modin/pull/7091). It seems to work faster than the previous one. Do you have the ability to measure on big sizes? (maybe you still have your env)
image

modin/core/storage_formats/pandas/query_compiler.py
dchigarev: nope, it's a bug, thanks for the fix!

modin/core/storage_formats/pandas/query_compiler.py
dchigarev: > when https://github.com/modin-project/modin/pull/7091 drop_duplicates and unique into one, did you take into account that they return different data types

Initially, I haven't seen a problem with it, since the result of both `.unique()` and `.drop_duplicates()` was wrapped into a pandas dataframe anyway inside of partition class. I didn't want to add another unique-like qc method in order not to bloat QCs api (which is already quite big and sometimes redundant).

An ideal solution I see is to continue wrapping the results with a pandas dataframe inside of partitions and then make the front-end handle the conversion to the desired output type (for example, `.drop_duplicates()` returns a `DataFrame(result_qc)` and `.unique()` returns `result_qc.to_values()`). Arrow extension arrays seem to round-trip to pandas with no problems:
```python
>>> arr = pd.array([1, 1, None], dtype="int64[pyarrow]")
>>> type(arr)
<class 'pandas.core.arrays.arrow.array.ArrowExtensionArray'>
>>> arr is pd.DataFrame(arr).squeeze().values
True
```

modin/core/storage_formats/pandas/query_compiler.py
dchigarev: I think it will always work faster, since the old full-column approach is basically '.to_pandas() + splitting + putting splits to ray storage', and in your approach it's only .to_pandas().

anmyachev: > I think it will always work faster, since the old full-column approach is basically '.to_pandas() + splitting + putting splits to ray storage', and in your approach it's only .to_pandas().

Yes, but I compare with range-partitioning implementation.

dchigarev: Running the same scripts on linux gives different results (range-partitioning is still faster in a lot of cases):

<details><summary>perf results for linux</summary>

MODIN_CPUS=8
image

MODIN_CPUS=16
image

</details>

I've been able to reproduce your results though when tried to run the same on windows:

<details><summary>perf results for windows</summary>

image

</details>

I don't think that the main reasons for the difference is OS here. I think that it's a hardware difference, since for windows runs I've used my own laptop instead of a dedicated machine.

anmyachev: @dchigarev thank you very much for running the benchmark!

> I don't think that the main reasons for the difference is OS here. I think that it's a hardware difference, since for windows runs I've used my own laptop instead of a dedicated machine.

Interesting detail.

modin/tests/pandas/test_series.py
```diff
@@ -3726,6 +3731,20 @@ def test_unique(data):
...
+
+def test_unique_pyarrow_dtype():
+    # See #6227 for details
+    modin_series, pandas_series = create_test_series(
+        [1, 0, pd.NA], dtype="uint8[pyarrow]"
+    )
+    modin_result = modin_series.unique()
+    pandas_result = pandas_series.unique()
+
+    df_equals(modin_result, pandas_result)
```
YarShev, line 3745: Why don't we use eval_general?

modin/pandas/series.py
```diff
@@ -2107,9 +2107,13 @@ def unique(self) -> ArrayLike:  # noqa: RT01, D200
...
         """
-        return self.__constructor__(
-            query_compiler=self._query_compiler.unique()
-        ).to_numpy()
+        # `values` can't be used here because it performs unnecessary conversion,
+        # after which the result type does not match the pandas
+        return (
+            self.__constructor__(query_compiler=self._query_compiler.unique())
+            .modin.to_pandas()
+            ._values
```
YarShev, line 2115: Is there a public method that returns the same?

modin/core/storage_formats/pandas/query_compiler.py
```diff
@@ -2039,6 +2046,8 @@ def unique(self, keep="first", ignore_index=True, subset=None):
                 preserve_columns=True,
             )
         else:
+            # return self.to_pandas().squeeze(axis=1).unique() works faster
+            # but returns pandas type instead of query compiler
```
YarShev, line 2050: Do we want to perform this in a separate issue? I guess so.

modin/pandas/series.py
anmyachev: I searched but didn't find it (`values` and `array` are not suitable).

modin/tests/pandas/test_series.py
anmyachev: rewritten

modin/core/storage_formats/pandas/query_compiler.py
anmyachev: https://github.com/modin-project/modin/issues/7182

YarShev (approved):

PR #7067, opened 2024-03-12
TEST-#7066: Explicitly check for exceptions in `test_io.py`

### LINKED ISSUE #7066: Explicitly check for exceptions in `test_io.py`

anmyachev:
## What do these changes do?

- [x] Resolves #7066 
- [x] tests added and passing

### DIFF
--- a/modin/pandas/test/test_io.py
+++ b/modin/pandas/test/test_io.py
@@ -46,7 +46,6 @@ from modin.config import (
     TestReadFromPostgres,
     TestReadFromSqlServer,
 )
-from modin.core.storage_formats.pandas.utils import split_result_of_axis_func_pandas
 from modin.db_conn import ModinDatabaseConnection, UnsupportedDatabaseException
 from modin.pandas.io import from_arrow, from_ray_dataset, to_pandas
 from modin.test.test_utils import warns_that_defaulting_to_pandas
@@ -61,7 +60,6 @@ from .utils import (
     eval_io_from_str,
     generate_dataframe,
     get_unique_filename,
-    io_ops_bad_exc,
     json_long_bytes,
     json_long_string,
     json_short_bytes,
@@ -288,6 +286,9 @@ class TestCsv:
             filepath_or_buffer=unique_filename,
             delimiter=delimiter,
             sep=sep,
+            raising_exceptions=ValueError(
+                "Specified a sep and a delimiter; you can only specify one."
+            ),
         )
 
     @pytest.mark.parametrize(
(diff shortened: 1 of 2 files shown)


modin/pandas/test/test_io.py
```diff
@@ -537,12 +549,15 @@ def test_read_csv_datetime(
...
-            # In this case raised TypeError: <lambda>() takes 1 positional argument but 2 were given
-            raising_exceptions = list(io_ops_bad_exc)
-            raising_exceptions.remove(TypeError)
+        raising_exceptions = None
+
+        if "format-Ymd" in request.node.callspec.id and (
+            "parse_dates3" in request.node.callspec.id
+            or "parse_dates4" in request.node.callspec.id
+        ):
+            raising_exceptions = False
```
YarShev, line 560: Should we put a similar FIXME here?

modin/pandas/test/utils.py
```diff
@@ -905,6 +911,12 @@ def execute_callable(fn, inplace=False, md_kwargs={}, pd_kwargs={}):
                         md_e, tuple(raising_exceptions)
                     ), f"not acceptable exception type: {md_e}"
                 elif raising_exceptions and type(raising_exceptions) is not type:
+                    if Engine.get() == "Ray":
+                        from ray.exceptions import RayTaskError
+
+                        # unwrap ray exceptions from remote worker
+                        if isinstance(md_e, RayTaskError):
```
YarShev, line 918: Do we have similar problems for Dask and Unidist?

anmyachev: It seems not, otherwise they would have to break these tests too.

modin/pandas/test/test_io.py
anmyachev: I specified an exception

modin/pandas/test/test_io.py
```diff
@@ -380,9 +383,14 @@ def test_read_csv_parsing_1(
                 ).columns
             }
 
+        raising_exceptions = None
+        if engine == "c" and skipfooter != 0:
```
YarShev, line 387: In all other cases there are no exceptions, right? or we just don't check raising exceptions in those cases?

anmyachev: > In all other cases there are no exceptions, right?

Yes and no. There really are no exceptions in these situations, but this is not being checked now, but it will be after https://github.com/modin-project/modin/pull/6954

YarShev (approved):

PR #7074, opened 2024-03-12
TEST-#7073: Explicitly check for exceptions in `test_default.py`

### LINKED ISSUE #7073: Explicitly check for exceptions in `test_default.py`

anmyachev:
## What do these changes do?

- [x] Resolves #7073 
- [x] tests added and passing

### DIFF
--- a/modin/pandas/test/utils.py
+++ b/modin/pandas/test/utils.py
@@ -56,7 +56,7 @@ from modin.utils import try_cast_to_pandas
 random_state = np.random.RandomState(seed=42)
 
 DATASET_SIZE_DICT = {
-    "Small": (2**2, 2**3),
+    "Small": (2**6, 2**6),
     "Normal": (2**6, 2**8),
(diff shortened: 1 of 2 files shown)


YarShev:
BaseOnPython failed

modin/pandas/test/utils.py
```diff
@@ -50,7 +56,7 @@
 random_state = np.random.RandomState(seed=42)
 
 DATASET_SIZE_DICT = {
-    "Small": (2**2, 2**3),
+    "Small": (2**6, 2**6),
```
anmyachev, line 59: Right now it only affects this `BaseOnPython` execution. Due to the different number of columns, it is not very convenient to write tests, since this has to be taken into account. For now, it’s enough for me that the number of columns is no less than the number that is in normal mode.

Another option is to run all executions on the same data sizes, that is, use the normal dataset size.

YarShev (approved):

PR #7079, opened 2024-03-13
FEAT-#4527: Add Modin logging to AxisPartition and Block…

### LINKED ISSUE #4527: Add Modin logging to engine layer

arunjose696:
…Partition classes

## What do these changes do?

In #4527 it was mentioned to add modin logging to different layers, Among these logging support has already been added to Partition Manager, The current PR adds logging to Block and Axis partitions. 

For logging in remote function it would be recommended to use external tools specific to backends for example [Ray timeline](https://docs.ray.io/en/latest/ray-core/api/doc/ray.timeline.html) for ray backend , UNIDIST_MPI_LOG for unidist etc.

- [ ] Resolves #4527? 
- [ ] tests added and passing


modin/core/dataframe/base/partitioning/axis_partition.py
```diff
@@ -16,8 +16,10 @@
 from abc import ABC, abstractmethod
 from typing import Any, Callable, Iterable, Optional, Tuple, Union
 
+from modin.logging import ClassLogger
 
-class BaseDataframeAxisPartition(ABC):  # pragma: no cover
+
+class BaseDataframeAxisPartition(ABC, ClassLogger):  # pragma: no cover
```
YarShev, line 22: Since we use the methods of this class inside remote functions, how does this addition of ClassLogger here affect logging for remote functions?

modin/core/execution/ray/implementations/pandas_on_ray/partitioning/partition.py
```diff
@@ -93,6 +93,7 @@ def __init__(
             )
         )
 
+    @disable_logging
```
YarShev, line 96: Why should we disable logging for this method?

arunjose696: Logging requires modules like sys, which might already have been cleaned up by the time __del__ is invoked(As we couldnt be sure when __del__ would be invoked).  This  leads to ModuleNotFoundError. So disabling logging of __del__

modin/core/dataframe/base/partitioning/axis_partition.py
arunjose696: As I understand BaseDataframeAxisPartition's methods  would call the remote functions, I could not see the other way around where methods of BaseDataframeAxisPartition are called in remote functions, I could see data of the partitions are send to remote functions and operations are performed on them, Is there anyplace where partitions would be directly passed to remote functions?

When running remote functions from the methods of BaseDataframeAxisPartition the class logger would work normally as it would log BaseDataframeAxisPartition methods and leave out logging of remote functions.

arunjose696: Got it Thanks

When we enable the logger with `LogMode.enable() ` in modin code, the logger gets enabled only on the main process the logger remains disabled in workers so the remote functions execute the function without logging so info about the function wouldn't be logged from remote function.

YarShev (approved):

